PR #37, opened 2020-12-02
config-argsparser applet

xwu4lab:
This is a config and argsparser applet modified from YACS


config/modules/featureExtractor.yaml
```diff
@@ -0,0 +1,13 @@
+# @package _global_
+FeatureExtractor:
+  _target_: gtsfm_cfgnode.FeatureExtractor
```
johnwlambert, line 3: Thanks for the PR, @xwu4lab, this is cool.

Could we have a single hierarchal yaml file instead of 4 different ones? i think would be more readable

xwu4lab: According to your comment, I also provide an example called gtsfm_cfgnode_from_single_file.py to make it work.

johnwlambert:
Decided to stick with the current design. May revisit hierarchical config or strongly typed in the future

PR #80, opened 2021-01-31
Fixing case of no ground truth pose

ayushbaid:
In the scene optimizer, we try to read the pose graph to generate visualizations. This crashed the program when there was no ground truth pose available. In this PR, I fix it by simply returning a `None` instead of the list of poses.

### DIFF
--- a/gtsfm/loader/loader_base.py
+++ b/gtsfm/loader/loader_base.py
@@ -101,13 +101,16 @@ class LoaderBase(metaclass=abc.ABCMeta):
 
         return [dask.delayed(self.get_camera_intrinsics)(x) for x in range(N)]
 
-    def create_computation_graph_for_poses(self) -> List[Delayed]:
+    def create_computation_graph_for_poses(self) -> Optional[List[Delayed]]:
         """Creates the computation graph for camera poses.
 
         Returns:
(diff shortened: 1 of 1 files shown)


gtsfm/loader/loader_base.py
```diff
@@ -101,14 +101,17 @@ def create_computation_graph_for_intrinsics(self) -> List[Delayed]:
...
-    def create_computation_graph_for_poses(self) -> List[Delayed]:
+    def create_computation_graph_for_poses(self) -> Optional[List[Delayed]]:
         """Creates the computation graph for camera poses.
 
         Returns:
             list of delayed tasks for camera poses.
         """
         N = self.__len__()
 
+        if self.get_camera_pose(0) is None:
```
johnwlambert, line 112: looks good, maybe you could add one inline comment mentioning that we assume if no ground truth pose at frame 0, none for the whole sequence?

johnwlambert (approved):

johnwlambert:
Feel free to merge

gtsfm/loader/loader_base.py
ayushbaid: Done. Thanks!

PR #82, opened 2021-02-02
Add view frustum functionality

johnwlambert:
Plot 3d frustums in 3d
Screen Shot 2021-02-02 at 3 39 58 PM

### DIFF
--- a/gtsfm/common/view_frustum.py
+++ b/gtsfm/common/view_frustum.py
@@ -1,5 +1,4 @@
-
-from typing import Tuple
+from typing import List, Tuple
 
 import numpy as np
 from gtsam import Point3, Pose3
@@ -10,7 +9,13 @@ DEFAULT_FRUSTUM_RAY_LEN = 0.3  # meters, arbitrary
 class ViewFrustum:
     """ Generates edges of a 5-face mesh for drawing pinhole camera in 3d"""
 
-    def __init__(self, fx: float, img_w: int, img_h: int, frustum_ray_len: int = DEFAULT_FRUSTUM_RAY_LEN) -> None:
+    def __init__(
+        self,
+        fx: float,
+        img_w: int,
+        img_h: int,
+        frustum_ray_len: int = DEFAULT_FRUSTUM_RAY_LEN,
+    ) -> None:
         """
         Args:
             fx: focal length in x-direction, assuming square pixels (fx == fy)
@@ -24,17 +29,19 @@ class ViewFrustum:
         self.img_h_ = img_h
         self.frustum_ray_len_ = frustum_ray_len
 
-    def get_frustum_vertices(self) -> Tuple[np.ndarray,np.ndarray,np.ndarray,np.ndarray,np.ndarray]:
-        """Obtain 3d positions of all 5 frustum vertices
+    def get_frustum_vertices_camfr(
+        self,
+    ) -> Tuple[np.ndarray, np.ndarray, np.ndarray, np.ndarray, np.ndarray]:
+        """Obtain 3d positions of all 5 frustum vertices in the camera frame
 
           (x,y,z)               (x,y,z)                (x,y,z)              (x,y,z)
               \\=================//                      \\                   //
                \\               //                        \\ 1-------------2 //
         (-w/2,-h/2,fx)       (w/2,-h/2,fx)                 \\| IMAGE PLANE |//
                  1-------------2                             |             |/
-                 |\\         //| IMAGE PLANE  (-w/2, h/2,fx) 3-------------4 (w/2, h/2,fx)
+                 |\\         //| IMAGE PLANE  (-w/2, h/2,fx) 4-------------3 (w/2, h/2,fx)
                  | \\       // | IMAGE PLANE                  \\         //
-                 3--\\-----//--4                               \\       //
+                 4--\\-----//--3                               \\       //
                      \\   //                                    \\     //
                       \\ //                                      \\   //
                         O PINHOLE                                 \\ //
@@ -45,14 +52,12 @@ class ViewFrustum:
                 [self.img_w_ // 2, self.img_h_ // 2],  # optical center
                 [0, 0],  # top-left
                 [self.img_w_ - 1, 0],  # top-right
-                [0, self.img_h_ - 1],  # bottom-left
                 [self.img_w_ - 1, self.img_h_ - 1],  # bottom-right
+                [0, self.img_h_ - 1],  # bottom-left
             ]
         )
 
-        ray_dirs = compute_pixel_ray_directions_vectorized(
-            uv, self.fx_, self.img_w_, self.img_h_
-        )
+        ray_dirs = compute_pixel_ray_directions_vectorized(uv, self.fx_, self.img_w_, self.img_h_)
         v0 = ray_dirs[0] * 0
         v1 = ray_dirs[1] * self.frustum_ray_len_
         v2 = ray_dirs[2] * self.frustum_ray_len_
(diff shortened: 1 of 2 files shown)


gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+        e4 = np.array([v4, v2])
+
+        # optical center to bottom-left (e5)
+        e5 = np.array([v0, v3])
+
+        # optical center to bottom-right (e6)
+        e6 = np.array([v0, v4])
+
+        # bottom-left to bottom-right (e7)
+        e7 = np.array([v3, v4])
```
ayushbaid, line 100: we can remove the whitespaces?

```
e0 = np.array([v1, v2])  # top-left to top-right
e1 = np.array([v0, v1])  # optical center to top-left
e2 = np.array([v0, v2])  # optical center to top-right
e3 = np.array([v3, v1])  # bottom-left to top-left
e4 = np.array([v4, v2])  # bottom-right to top-right
e5 = np.array([v0, v3])  # optical center to bottom-left
e6 = np.array([v0, v4])  # optical center to bottom-right
e7 = np.array([v3, v4])  # bottom-left to bottom-right
```

tests/common/test_view_frustum.py
```diff
@@ -0,0 +1,77 @@
+
```
ayushbaid, line 1: This is not technically a unit test, right?

ayushbaid (changes requested):
Looks good in general. I have a few suggestions regarding code quality

tests/common/test_view_frustum.py
johnwlambert: Thanks for the review Ayush. I think it still qualifies as a unit test "smokescreen" because if the code crashes then the test fails

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+    y_center_offs = v - py
+
+    x_center_offs = x_center_offs.reshape(num_rays, 1)
+    y_center_offs = y_center_offs.reshape(num_rays, 1)
+    fx_tiled = np.tile(fx, (num_rays)).reshape(num_rays, 1)
+
+    ray_dirs = np.hstack([x_center_offs, y_center_offs, fx_tiled])
+
+    # elementwise multiplication of scalars requires last dim to match
+    ray_dirs = (ray_dirs.T / np.linalg.norm(ray_dirs, axis=1)).T
```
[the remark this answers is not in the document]
johnwlambert: nice, i'll make that change

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+        img_h: image height (in pixels)
+        """
+        self.fx = fx
+        self.img_w = img_w
+        self.img_h = img_h
+
+    def get_frustum_vertices(self) -> Tuple[np.ndarray,np.ndarray,np.ndarray,np.ndarray,np.ndarray]:
+        """Obtain 3d positions of all 5 frustum vertices
+
+          (x,y,z)               (x,y,z)                (x,y,z)              (x,y,z)
```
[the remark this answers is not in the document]
johnwlambert: couldn't get the rays to be clear through the bottom and top vertices otherwise

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+    """ Generates edges of a 5-face mesh for drawing pinhole camera in 3d"""
+
+    def __init__(self, fx: float, img_w: int, img_h: int) -> None:
+        """
+        Args:
+        fx: focal length in x-direction, assuming square pixels (fx == fy)
+        img_w: image width (in pixels)
+        img_h: image height (in pixels)
+        """
+        self.fx = fx
```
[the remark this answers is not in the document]
johnwlambert: good point, will do

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
+
+from typing import Tuple
+
+import numpy as np
+from gtsam import Point3, Pose3
+
+FRUSTUM_RAY_LEN = 0.3  # meters, arbitrary
```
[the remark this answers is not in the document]
johnwlambert: good point, will make that change

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+
+    # offset from center
+    x_center_offs = u - px
+    y_center_offs = v - py
+
+    x_center_offs = x_center_offs.reshape(num_rays, 1)
+    y_center_offs = y_center_offs.reshape(num_rays, 1)
+    fx_tiled = np.tile(fx, (num_rays)).reshape(num_rays, 1)
+
+    ray_dirs = np.hstack([x_center_offs, y_center_offs, fx_tiled])
```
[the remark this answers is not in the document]
johnwlambert: that's cleaner, thanks

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+            v_end = edge_camfr[1]
+
+            edge_worldfr = [wTc.transformFrom(Point3(v_start)), wTc.transformFrom(Point3(v_end))]
+            edges_worldfr += [edge_worldfr]
+
+        return edges_worldfr
+
+
+def compute_pixel_ray_directions_vectorized(
+    uv: np.ndarray, fx: float, img_w: int, img_h: int
```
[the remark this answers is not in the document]
johnwlambert: kind of prefer to have it functional (and in case we need it somewhere else later down the road independently of a specific frustum object)

gtsfm/common/view_frustum.py
johnwlambert: good point, will do

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+                [self.img_w - 1, 0],  # top-right
+                [0, self.img_h - 1],  # bottom-left
+                [self.img_w - 1, self.img_h - 1],  # bottom-right
+            ]
+        )
+
+        ray_dirs = compute_pixel_ray_directions_vectorized(
+            uv, self.fx, self.img_w, self.img_h
+        )
+        v0 = ray_dirs[0] * 0
```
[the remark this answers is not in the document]
johnwlambert: just was trying to show v0 is 0 meters away from optical center

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,165 @@
...
+
+FRUSTUM_RAY_LEN = 0.3  # meters, arbitrary
+
+
+class ViewFrustum:
+    """ Generates edges of a 5-face mesh for drawing pinhole camera in 3d"""
+
+    def __init__(self, fx: float, img_w: int, img_h: int) -> None:
+        """
+        Args:
```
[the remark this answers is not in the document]
johnwlambert: I use Sublime :-)

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,146 @@
...
+        """
+        uv = np.array(
+            [
+                [self.img_w_ // 2, self.img_h_ // 2],  # optical center
+                [0, 0],  # top-left
+                [self.img_w_ - 1, 0],  # top-right
+                [0, self.img_h_ - 1],  # bottom-left
+                [self.img_w_ - 1, self.img_h_ - 1],  # bottom-right
+            ]
+        )
```
akshay-krishnan, line 51: Doesnt this disagree with the image above?

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,146 @@
...
+        uv: array of shape (N,2) with (u,v) coordinates
+        fx: focal length in x-direction, assuming square pixels (fx == fy)
+        img_w: image width (in pixels)
+        img_h: image height (in pixels)
+
+    Returns:
+        ray_dirs: Array of shape (N,3) with ray directions in camera frame
+    """
+    assert uv.shape[1] == 2
+    assert uv.ndim == 2
```
akshay-krishnan, line 129: I think ndim must be asserted before checking shape. What if uv has ndim less than 2?

gtsfm/common/view_frustum.py
akshay-krishnan: I understood after reading `compute_pixel_ray_directions_vectorized` how it works. But someone reading this method alone will likely be confused.

akshay-krishnan: Maybe that function can be modified to accept offsets?

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,146 @@
...
+        e3 = np.array([v3, v1]) # bottom-left to top-left (e3)
+        e4 = np.array([v4, v2]) # bottom-right to top-right (e4)
+        e5 = np.array([v0, v3]) # optical center to bottom-left (e5)
+        e6 = np.array([v0, v4]) # optical center to bottom-right (e6)
+        e7 = np.array([v3, v4]) # bottom-left to bottom-right (e7)
+
+        edges_camfr = np.stack([e0, e1, e2, e3, e4, e5, e6, e7])
+        return edges_camfr
+
+    def get_mesh_edges_worldframe(self, wTc: Pose3) -> np.ndarray:
```
akshay-krishnan, line 93: Are both worldframe and cameraframe useful? If only worldframe will be used, the code can be more simplified.

akshay-krishnan (changes requested):

gtsfm/common/view_frustum.py
johnwlambert: Thanks for the review, Akshay. `cameraframe` is likely not as useful, but I think it helps keep the code more modular

johnwlambert: that way both functions are <10 lines

gtsfm/common/view_frustum.py
johnwlambert: good point, i'll fix that

gtsfm/common/view_frustum.py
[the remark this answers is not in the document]
ayushbaid: Oh ok. So it is infact the origin.

ayushbaid (approved):

gtsfm/common/view_frustum.py
akshay-krishnan: Computationally, there are only 5 points that need to be transformed from camera to world frame, but we look over the 8 edges (each with 2 points), so 16 points and transform them. Points are repeated many times. So you could directly call get_frustum_vertices() in get_mesh_edges_worldframe(), transform the points to world frame first, and then make edges out of them.
Also, the code to form edges from vertices can be moved to a different helper and be used in both get_mesh_edges_camframe and get_mesh_edges_worldframe.

gtsfm/common/view_frustum.py
```diff
@@ -0,0 +1,146 @@
...
+        v0, v1, v2, v3, v4 = self.get_frustum_vertices()
+
+        e0 = np.array([v1, v2]) # top-left to top-right (e0)
+        e1 = np.array([v0, v1]) # optical center to top-left (e1)
+        e2 = np.array([v0, v2]) # optical center to top-right (e2)
+        e3 = np.array([v3, v1]) # bottom-left to top-left (e3)
+        e4 = np.array([v4, v2]) # bottom-right to top-right (e4)
+        e5 = np.array([v0, v3]) # optical center to bottom-left (e5)
+        e6 = np.array([v0, v4]) # optical center to bottom-right (e6)
+        e7 = np.array([v3, v4]) # bottom-left to bottom-right (e7)
```
akshay-krishnan, line 88: I think using a for loop to connect the origin to the 4 points might be cleaner. 
And another for loop could be used to connect v1 - v2, v2 - v3, v3 - v4, v4- v1. 
Also think it might be better to move this to a different helper and use it in both get_mesh_edges_camframe and get_mesh_edges_worldframe.

adisingh50 (approved):

gtsfm/common/view_frustum.py
johnwlambert: should be fixed now, thanks

PR #84, opened 2021-02-03
Logging metrics for front-end output

ayushbaid:
In this PR, we log the relative rotation and translation error from the `TwoViewEstimator` class. In a follow-up PR, we will persist these metrics to the disk.

Sample logs:
```
DEBUG:root:[Two View Estimator] Relative rotation error 0.040799
DEBUG:root:[Two View Estimator] Relative unit-translation error 0.151869
DEBUG:root:[Two View Estimator] Relative rotation error 0.008509
DEBUG:root:[Two View Estimator] Relative unit-translation error 0.133831
DEBUG:root:[Two View Estimator] Relative rotation error 0.064076
DEBUG:root:[Two View Estimator] Relative unit-translation error 0.165390
DEBUG:root:[Two View Estimator] Relative rotation error 0.017157

....

DEBUG:root:[Two view optimizer] [Summary] Rotation success: 66/66
DEBUG:root:[Two view optimizer] [Summary] Translation success: 43/66
DEBUG:root:[Two view optimizer] [Summary] Pose success: 43/66

```


gtsfm/scene_optimizer.py
```diff
@@ -56,8 +55,21 @@
 from gtsfm.frontend.verifier.verifier_base import VerifierBase
 from gtsfm.loader.folder_loader import FolderLoader
 
-# configure loggers to avoid DEBUG level stdout messages
-logging.basicConfig(stream=sys.stdout, level=logging.INFO)
+
+def get_logger():
```
johnwlambert, line 59: Thanks for the PR, Ayush. I think we define this `get_logger()` function in multiple places, so it could probably go in `gtsfm/utils/logger_utils.py` or something

gtsfm/scene_optimizer.py
```diff
@@ -87,6 +99,42 @@ def __init__(self, matcher: MatcherBase, verifier: VerifierBase):
...
         self.verifier = verifier
 
+    def __compute_metrics(
+        self,
+        i2Ri1_computed: Optional[Rot3],
+        i2Ui1_computed: Optional[Rot3],
+        i2Ri1_expected: Rot3,
+        i2Ui1_expected: Unit3,
+    ) -> Tuple[float, float]:
+        """[summary]
```
johnwlambert, line 109: missing docstring summary here

gtsfm/scene_optimizer.py
```diff
@@ -96,7 +144,8 @@ def create_computation_graph(
         camera_intrinsics_i1_graph: Delayed,
         camera_intrinsics_i2_graph: Delayed,
         exact_intrinsics: bool = True,
-    ) -> Tuple[Delayed, Delayed, Delayed]:
+        i2Ti1_expected_graph: Optional[Delayed] = None,
```
johnwlambert, line 147: i might add 1-sentence to docstring explaining that `i2Ti1_expected_graph` is the ground truth

gtsfm/scene_optimizer.py
```diff
@@ -87,6 +99,42 @@ def __init__(self, matcher: MatcherBase, verifier: VerifierBase):
...
+
+        R_error = comp_utils.compute_relative_rotation_angle(
+            i2Ri1_computed, i2Ri1_expected
+        )
+
+        U_error = comp_utils.compute_relative_unit_translation_angle(
+            i2Ui1_computed, i2Ui1_expected
+        )
+
+        logging.debug(
```
johnwlambert, line 129: might get too verbose to print for every pair, maybe median and average are more useful?

johnwlambert: (but still dump all to JSON)

gtsfm/scene_optimizer.py
```diff
@@ -119,7 +168,26 @@ def create_computation_graph(
             exact_intrinsics,
         )
 
-        return i2Ri1_graph, i2Ui1_graph, v_corr_idxs_graph
+        # if we have the expected data, evaluate the computed relative pose
+        if i2Ti1_expected_graph is not None:
+            error_graphs = dask.delayed(self.__compute_metrics)(
+                i2Ri1_graph,
+                i2Ui1_graph,
+                dask.delayed(lambda x: x.rotation())(i2Ti1_expected_graph),
```
johnwlambert, line 176: is lambda necessary here?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -46,30 +46,21 @@ def align_poses(input_list: List[Pose3], ref_list: List[Pose3]) -> List[Pose3]:
...
-    w2Ti0_ = ref_list[0]
-    # origin transform -- map the origin of the input list to the reference list
-    w2Tw1 = w2Ti0_.compose(i0Tw1)
-    
-    # origin shifted list
-    input_shifted_list = [w2Tw1.compose(w1Ti) for w1Ti in input_list]
-
-    # get distances w.r.t origin for both the list to compute the scale
-    w2Ti0 = input_shifted_list[0] # set this as origin
+    # match the scales first
```
johnwlambert, line 49: what's the change we're making here?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -192,3 +194,53 @@ def compare_global_poses(
...
+        U_2: the second unit-translation.
+
+    Returns:
+        the angle between the two unit-vectors.
+    """
+    if U_1 is None or U_2 is None:
+        return np.pi
+
+    dot_product = np.dot(U_1.point3(), U_2.point3())
+
```
johnwlambert, line 243: extra whitespace probably not needed between L242 and L244

tests/utils/test_geometry_comparisons.py
```diff
@@ -26,42 +26,78 @@ def test_compare_poses_with_uniform_scaled_translations(self):
...
-        self.assertFalse(geometry_comparisons.compare_global_poses(
-            pose_list, pose_list_))
+        self.assertFalse(
+            geometry_comparisons.compare_global_poses(pose_list, pose_list_)
+        )
+
+    def test_compute_relative_rotation_angle(self):
+        """Tests the relative angle between two rotations."""
+
+        R_1 = POSE_LIST[0].rotation()
```
johnwlambert, line 76: probably simpler to make this unit test self-defined and just define R_1 and R_2 ourselves here, to avoid looking up https://github.com/borglab/gtsam/blob/develop/python/gtsam/examples/SFMdata.py

gtsfm/scene_optimizer.py
ayushbaid: Median and average won't help here. I think maybe bin the errors and dump the histogram?

I want to decide on the format/tech used to dump and follow up in a second pair.

gtsfm/scene_optimizer.py
ayushbaid: how will I do it otherwise?

gtsfm/utils/geometry_comparisons.py
ayushbaid: I was first matching the first pose, and then aligning the scale (by using the median value). This creates and error. 

Example: we have 5 poses on the circle. The circle center is at the origin. Now a scaled version of this scene is the one where we have doubled the radius of the circle.

Earlier, the align_pose function first matched the 0th pose (i.e. set the same rotation and translation) and then scaled the scene. That logic will not work. So now I first match the scale and then align the 0th pose.

I had a unit test for this, but it did not run earlier because I messed up the unit test name to create two tests with the same name.

gtsfm/scene_optimizer.py
```diff
@@ -96,8 +133,30 @@ def create_computation_graph(
...
+            i2Ti1_expected_graph (optional): ground truth relative pose, used
+                                             for evaluation if available.
+                                             Defaults to None.
+
+        Returns:
+            computed relative rotation wrapped as Delayed.
+            computed relative translation direction wrapped as Delayed.
+            indices of verified correspondences wrapped as Delayed.
+            error in relative rotation wrapped as Delayed
+            error in relative translation direction wrapped as Delayed.
```
akshay-krishnan, line 158: It might be better to have a different accessor (or property in the python world) for the errors. We should try to and keep the API for TwoViewEstimator as close to its functionality as possible. Now someone who does not want the errors will need dummy variables in their code.

gtsfm/scene_optimizer.py
```diff
@@ -87,6 +87,43 @@ def __init__(self, matcher: MatcherBase, verifier: VerifierBase):
         self.matcher = matcher
         self.verifier = verifier
 
+    def __compute_metrics(
```
akshay-krishnan, line 90: Can we move this entirely to a different util file? 
This file will probably be the most frequently read in this repository, so I hope you agree that we should keep it very high level. 
I really like how this file is organized with very high-level classes, but it is already getting too big.

gtsfm/scene_optimizer.py
```diff
@@ -381,11 +471,15 @@ def create_computation_graph(
...
                 use_intrinsics_in_verification,
+                gt_relative_pose,
             )
             i2Ri1_graph_dict[(i1, i2)] = i2Ri1
             i2Ui1_graph_dict[(i1, i2)] = i2Ui1
             v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
 
+            auxiliary_graph_list.append(rot_error)
+            auxiliary_graph_list.append(unit_tran_error)
+
```
akshay-krishnan, line 482: If the API change I suggested is made, we can access these and append them to the graph only if a condition is met (I think thats if gt_relative_pose is not None).

gtsfm/scene_optimizer.py
```diff
@@ -119,7 +178,26 @@ def create_computation_graph(
...
-        return i2Ri1_graph, i2Ui1_graph, v_corr_idxs_graph
+        # if we have the expected data, evaluate the computed relative pose
+        if i2Ti1_expected_graph is not None:
+            error_graphs = dask.delayed(self.__compute_metrics)(
+                i2Ri1_graph,
+                i2Ui1_graph,
+                dask.delayed(lambda x: x.rotation())(i2Ti1_expected_graph),
+                dask.delayed(lambda x: Unit3(x.translation()))(
+                    i2Ti1_expected_graph
+                ),
```
akshay-krishnan, line 189: I think it might be cleaner to move the logic to separate translation and rotation to the compute_metrics function. Also, doesnt the translation have to be normalized?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -192,3 +194,51 @@ def compare_global_poses(
...
+    if R_1 is None or R_2 is None:
+        return np.pi
+
+    relative_rot = R_1.between(R_2)
+
+    origin = Rot3()
+
+    relative_rot_axisangle = origin.Logmap(relative_rot)
+    relative_rot_angle = np.linalg.norm(relative_rot_axisangle) % (2 * np.pi)
```
akshay-krishnan, line 222: I thought rotations in gtsam had an axis-angle method, did you check that?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -192,3 +194,51 @@ def compare_global_poses(
...
+        U_1: the first unit-translation.
+        U_2: the second unit-translation.
+
+    Returns:
+        the angle between the two unit-vectors.
+    """
+    if U_1 is None or U_2 is None:
+        return np.pi
+
+    dot_product = np.dot(U_1.point3(), U_2.point3())
```
akshay-krishnan, line 242: Doesnt Unit3 have a dot() method?

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: I feel like either of them should work since the scale is with respect to the first pose and not the world frame.

akshay-krishnan (changes requested):

gtsfm/utils/geometry_comparisons.py
ayushbaid: Nope. I use the median of scale of all (i, 0) pairs, i.e. translation magnitude ratio for all poses w.r.t. 0th pose.

I'll replace it with whatever we have in Zhang18iros_evaluation. But for now I'll leave this as is.

gtsfm/scene_optimizer.py
ayushbaid: Nice point, but I'll need your suggestions on how to organize it.

We need the logging to be in this file, so that leaves us with just 2 lines of code (calling the two error functions in `comp_utils`). I can remove this function, and copy those two lines to the `create_computation_graph` function?

I thought having a separate function just allows cleaner and smaller functions, even if they call each other.

gtsfm/scene_optimizer.py
johnwlambert: I agree it is getting a bit too big -- maybe in a follow-up PR we should put each of TwoViewEstimator, MultiViewOptimizer into their own files

gtsfm/scene_optimizer.py
ayushbaid: Thanks. I casted it to Unit3 for normalization.

gtsfm/utils/geometry_comparisons.py
ayushbaid: Thanks. Much cleaner!

gtsfm/utils/geometry_comparisons.py
ayushbaid: Oh it does. It is not exposed in python yet. I'll maybe leave this as a TODO.

gtsfm/scene_optimizer.py
ayushbaid: If they go in their separate files, do we still move the functions to other files as Akshay recommends?

gtsfm/utils/geometry_comparisons.py
johnwlambert: This is a nice fix @ayushbaid . I'll add another test case that now passes with the fixed code. Can we put the geometry comparison fix in its own PR though to make the PRs simpler? Then we can accept it immediately while we discuss this other stuff

tests/utils/test_geometry_comparisons.py
```diff
@@ -26,42 +27,106 @@ def test_compare_poses_with_uniform_scaled_translations(self):
...
+
+    wTv0 = Pose3(Rot3(R0), Point3(np.array([4, 3, 0])))
+    wTv1 = Pose3(Rot3(R90), Point3(np.array([8, 3, 0])))
+    wTv2 = Pose3(Rot3(R180), Point3(np.array([8, 7, 0])))
+    wTv3 = Pose3(Rot3(R270), Point3(np.array([4, 7, 0])))
+
+    wTi_list_ = [wTv0, wTv1, wTv2, wTv3]
+
+    pose_graphs_equal = geometry_comparisons.compare_global_poses(wTi_list, wTi_list_)
+    assert self.assertTrue(pose_graphs_equal)
```
johnwlambert, line 128: IMG_2371

gtsfm/utils/geometry_comparisons.py
ayushbaid: Yeah the test case was already there. I mistakenly named it w/ name of another unit test and hence it was running earlier. It will run now.

gtsfm/scene_optimizer.py
ayushbaid: @johnwlambert

johnwlambert:
Looks like CI is failing: 
```
======================================================================
ERROR: test_sift_twoway_degensac (frontend.test_frontend_argoverse.TestFrontend)
Check DoG + SIFT + 2-way Matcher + DEGENSAC-8pt frontend.
----------------------------------------------------------------------
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 105, in test_sift_twoway_degensac
    self.__compare_frontend_result_error(
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 124, in __compare_frontend_result_error
    ) = self.__get_frontend_computation_graph(
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 70, in __get_frontend_computation_graph
    (i2Ri1, i2Ui1, _,) = two_view_estimator.create_computation_graph(
ValueError: too many values to unpack (expected 3)

======================================================================
ERROR: test_sift_twoway_ransac (frontend.test_frontend_argoverse.TestFrontend)
Check DoG + SIFT + 2-way Matcher + RANSAC-5pt frontend.
----------------------------------------------------------------------
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 91, in test_sift_twoway_ransac
    self.__compare_frontend_result_error(
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 124, in __compare_frontend_result_error
    ) = self.__get_frontend_computation_graph(
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 70, in __get_frontend_computation_graph
    (i2Ri1, i2Ui1, _,) = two_view_estimator.create_computation_graph(
ValueError: too many values to unpack (expected 3)
```

gtsfm/scene_optimizer.py
akshay-krishnan: @ayushbaid up to you. I thought it might be cleaner, but it might be okay if they are in separate files, as long as we only have two function calls for metrics.

gtsfm/scene_optimizer.py
akshay-krishnan: Okay, you could also consider doing that in compute_metrics. What are your thoughts on that?

gtsfm/scene_optimizer.py
akshay-krishnan: That is not what I was talking about in this comment. It's good to have the _compute_metrics method. This comment is about not returning the errors in `create_computation_graph`. You should still compute them here, but store them as member variables and return them in another get_metrics() method.

gtsfm/scene_optimizer.py
akshay-krishnan: Basically, passing ground truth Pose3 to _compute_metrics and allow it to separate them and normalize. Keeps code here cleaner.

gtsfm/two_view_estimator.py
```diff
@@ -86,4 +94,51 @@ def create_computation_graph(
...
+        return (
+            i2Ri1_graph,
+            i2Ui1_graph,
+            v_corr_idxs_graph,
+            error_graphs[0],
+            error_graphs[1],
+        )
+
+
+def compute_metrics(
```
johnwlambert, line 114: Thanks for the updates @ayushbaid . maybe this could be `compute_two_view_error_metrics()` is a  better name?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -194,3 +194,49 @@ def compare_global_poses(
...
+
+    Returns:
+        the angle between two rotations.
+    """
+
+    if R_1 is None or R_2 is None:
+        return np.pi
+
+    relative_rot = R_1.between(R_2)
+
```
johnwlambert, line 218: nit: prefer to remove this extra whitespace between lines

gtsfm/utils/geometry_comparisons.py
```diff
@@ -194,3 +194,49 @@ def compare_global_poses(
...
+    Args:
+        R_1: the first rotation.
+        R_2: the second rotation.
+
+    Returns:
+        the angle between two rotations.
+    """
+
+    if R_1 is None or R_2 is None:
+        return np.pi
```
johnwlambert, line 215: should this be a module-level private global default?

`MAX_ROTATION_ERROR = np.pi`?

gtsfm/two_view_estimator.py
ayushbaid: Its already in the `two_view_estimator.py` file, so does the new name makes sense? What about `compute_relative_pose_metrics()`?

gtsfm/scene_optimizer.py
```diff
@@ -336,6 +363,50 @@ def write_sfmdata_to_disk(sfm_data: SfmData, save_fpath: str) -> None:
...
+    success_count_unit3 = np.sum(unit3_errors < angular_err_threshold)
+    success_count_pose = np.sum(pose_errors < angular_err_threshold)
+
+    logging.debug(
+        "[Two view optimizer] [Summary] Rotation success: %d/%d",
+        success_count_rot3,
+        num_entries,
+    )
+
+    logging.debug(
```
johnwlambert, line 397: I would prefer `logger.info` instead of `logging.debug`

gtsfm/two_view_estimator.py
johnwlambert: sounds good

gtsfm/two_view_estimator.py
```diff
@@ -86,4 +94,51 @@ def create_computation_graph(
...
+
+    R_error = comp_utils.compute_relative_rotation_angle(
+        i2Ri1_computed, i2Ti1_expected.rotation()
+    )
+
+    U_error = comp_utils.compute_relative_unit_translation_angle(
+        i2Ui1_computed, Unit3(i2Ti1_expected.translation())
+    )
+
+    logging.debug("[Two View Estimator] Relative rotation error %f", R_error)
```
johnwlambert, line 139: prefer to pull get the default logger here with the logger utils. Is it printing out currently as expected when you run it?

gtsfm/scene_optimizer.py
ayushbaid: It doesn't work with `logger.info`. I might be doing something wrong here

ayushbaid: As discussed, leaving it for later

gtsfm/two_view_estimator.py
ayushbaid: leaving this for now, as discussed

johnwlambert:
It looks like CI is still failing @ayushbaid 

```
----------------------------------------------------------------------
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 91, in test_sift_twoway_ransac
    self.__compare_frontend_result_error(
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 124, in __compare_frontend_result_error
    ) = self.__get_frontend_computation_graph(
  File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 70, in __get_frontend_computation_graph
    (i2Ri1, i2Ui1, _,) = two_view_estimator.create_computation_graph(
ValueError: too many values to unpack (expected 3)
```

gtsfm/scene_optimizer.py
ayushbaid: kicking for later

gtsfm/scene_optimizer.py
ayushbaid: I have to return it in the `create_computation_graph` function because it has to be made part of the main computation in the scene_optimizer class. The user anyway doesn't have to interact/change these files at all so I personally feel having a single function is the best option

gtsfm/scene_optimizer.py
ayushbaid: We have moved to individual files and removed class functions to general functions in the same file.

gtsfm/scene_optimizer.py
ayushbaid: Thanks. implemented this

ayushbaid:
> It looks like CI is still failing @ayushbaid
> 
> ```
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 91, in test_sift_twoway_ransac
>     self.__compare_frontend_result_error(
>   File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 124, in __compare_frontend_result_error
>     ) = self.__get_frontend_computation_graph(
>   File "/home/runner/work/gtsfm/gtsfm/tests/frontend/test_frontend_argoverse.py", line 70, in __get_frontend_computation_graph
>     (i2Ri1, i2Ui1, _,) = two_view_estimator.create_computation_graph(
> ValueError: too many values to unpack (expected 3)
> ```

Fixed it!

gtsfm/scene_optimizer.py
johnwlambert: Not sure if we can store side effects in member variables with Dask since everything seems functional in the computational graph

johnwlambert:
Looks like we have some conflicts we need to resolve now vs. latest master

johnwlambert (approved):

PR #89, opened 2021-02-08
Alternate version of hydra config

johnwlambert:
(Draft) version of GTSFM config using FAIR's hydra, but without structured config

Pros:
- Simpler

Cons:
- No type validation


xwu4lab:
I think it may be not quite correct when we say it is no type checking for hydra. Actually, hydra can achieve type checking through schema, and we have already made it in the latest update of gtsfm_config module.

gtsfm/config.yaml
```diff
@@ -0,0 +1,37 @@
+SceneOptimizer:
+  _target_: gtsfm.scene_optimizer.SceneOptimizer
```
ayushbaid, line 2: do we need all these targets?

setup.py
```diff
@@ -41,5 +41,7 @@
     keywords="computer-vision",
     packages=find_packages(),
     python_requires=">= 3.6",
-    install_requires=[],
+    install_requires=[
```
ayushbaid, line 44: can we move it to the conda environment file?

johnwlambert: @ayushbaid  Only hydra >=1.10 supports recursive object instantiation, and I didnt see 1.1.0 in conda (only available via pip install --pre)

gtsfm/config.yaml
johnwlambert: @ayushbaid we need them to be able to instantiate instantly without manually calling the constructors

johnwlambert:
> I think it may be not quite correct when we say it is no type checking for hydra. Actually, hydra can achieve type checking through schema, and we have already made it in the latest update of gtsfm_config module.

@xwu4lab you're right that Hydra does have type checking, and I saw that you were using it in the structured config approach. My approach here though is not using type checking -- I was just playing around with an approach that didn't involve the structured configs

gtsfm/run_scene_optimizer.py
```diff
@@ -16,28 +17,32 @@
...
 
-    sfm_result_graph = scene_optimizer.create_computation_graph(
-        len(loader),
-        loader.get_valid_pairs(),
-        loader.create_computation_graph_for_images(),
-        loader.create_computation_graph_for_intrinsics(),
-        use_intrinsics_in_verification=True,
-        gt_pose_graph=loader.create_computation_graph_for_poses(),
-    )
+        loader = FolderLoader(os.path.join(TEST_ROOT, "data", "set1_lund_door"), image_extension="JPG")
```
ayushbaid, line 28: should we have this from config too?

tests/test_scene_optimizer.py
```diff
@@ -87,60 +62,52 @@ def test_find_largest_connected_component(self):
         (
             computed_relative_rotations,
             computed_relative_unit_translations,
-        ) = select_largest_connected_component(
-            input_relative_rotations, input_relative_unit_translations
-        )
+        ) = select_largest_connected_component(input_relative_rotations, input_relative_unit_translations)
```
ayushbaid, line 65: Are you using 120 line width?

tests/test_scene_optimizer.py
```diff
@@ -87,60 +62,52 @@ def test_find_largest_connected_component(self):
...
 
-        # generate the dask computation graph
-        sfm_result_graph = self.obj.create_computation_graph(
-            len(self.loader),
-            self.loader.get_valid_pairs(),
-            self.loader.create_computation_graph_for_images(),
-            self.loader.create_computation_graph_for_intrinsics(),
-            use_intrinsics_in_verification=use_intrinsics_in_verification,
-        )
+        with initialize_config_module(config_module="gtsfm.configs"):
```
ayushbaid, line 82: why are we doing this in a test function and not the setup?

johnwlambert: using a scope like this was the idiomatic way proposed in the [Hydra unit test examples](https://github.com/facebookresearch/hydra/tree/master/examples/advanced/hydra_app_example/tests#L29), so I followed it. But we couldn't have the scope if we set it up in `setUp(self)`.

tests/test_scene_optimizer.py
johnwlambert: yup

gtsfm/run_scene_optimizer.py
johnwlambert: I think we had talked about how the `SceneGenerator` would provide these paths to the `SceneOptimizer`, so I think they would go in a SceneGenerator config, not the SceneOptimizer config

tests/test_scene_optimizer.py
johnwlambert: I generally am thinking these days 120 length makes for more readability in the single column format in the github.com UI.

In 2-column review, 80 chars can be a bit easier, but widening the browser just for review works as a fix for me.

gtsfm/run_scene_optimizer.py
```diff
@@ -0,0 +1,49 @@
+import os
```
ayushbaid, line 1: Should we move this file to `gtsfm/runner/run_scene_optimizer.py`? In that case, I can created runners for different modules like front-end only and stuff?

johnwlambert: i like that idea, i'll move it, one sec

johnwlambert: done

gtsfm/config.yaml
ayushbaid: got it. Thanks!

setup.py
ayushbaid: Can we still move it to the yml file? (like we did for degensac)

johnwlambert: sure, forgot that was possible

ayushbaid (approved):

PR #95, opened 2021-02-11
Adding metric to check correctness of correspondences

ayushbaid:
This PR logs the number and ratio of correct correspondences for each image pair, and also the summary count of number of pairs with 100% correct correspondences.

### DIFF
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -339,7 +339,7 @@ def aggregate_frontend_metrics(
     Args:
         rot3_errors: angular errors in rotations, measured in degrees.
         unit3_errors: angular errors in unit-translations, measures in degrees.
-        corr_correctness: correctness of correspondences, as (num_correct, ratio_of_correct).
+        corr_correctness: correctness of correspondences, as (num_correct, inlier_ratio).
         angular_err_threshold: threshold to classify the error as success, in degrees.
     """
     num_entries = len(rot3_errors)
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -358,7 +358,7 @@ def aggregate_frontend_metrics(
     success_count_unit3 = np.sum(unit3_errors < angular_err_threshold_deg)
     success_count_pose = np.sum(pose_errors < angular_err_threshold_deg)
 
-    # count entries with all correct correspondences
+    # count entries with inlier ratio == 1.
     all_correct = np.count_nonzero(np.array(corr_correctness)[:, 1] == 1.0)
 
     logger.debug(
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -128,7 +128,7 @@ class SceneOptimizer:
                 v_corr_idxs,
                 rot_error,
                 unit_tran_error,
-                corr_correctness_count,
+                correspondence_stats,
             ) = self.two_view_estimator.create_computation_graph(
                 keypoints_graph_list[i1],
                 keypoints_graph_list[i2],
@@ -147,8 +147,8 @@ class SceneOptimizer:
                 frontend_metrics_dict[(i1, i2)] = (
                     rot_error,
                     unit_tran_error,
-                    corr_correctness_count[0],
-                    corr_correctness_count[1],
+                    correspondence_stats[0],
+                    correspondence_stats[1],
                 )
 
             if self._save_viz:
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -339,7 +339,7 @@ def aggregate_frontend_metrics(
     Args:
         rot3_errors: angular errors in rotations, measured in degrees.
         unit3_errors: angular errors in unit-translations, measures in degrees.
-        corr_correctness: correctness of correspondences.
+        corr_correctness: correctness of correspondences, as (num_correct, ratio_of_correct).
         angular_err_threshold: threshold to classify the error as success, in degrees.
     """
     num_entries = len(rot3_errors)
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -107,6 +107,7 @@ class SceneOptimizer:
 
         frontend_rot3_errors = []
         frontend_unit3_errors = []
+        frontend_corr_correctness = []
 
         for (i1, i2) in image_pair_indices:
             if gt_pose_graph is not None:
@@ -114,7 +115,14 @@ class SceneOptimizer:
             else:
                 gt_relative_pose = None
 
-            (i2Ri1, i2Ui1, v_corr_idxs, rot_error, unit_tran_error,) = self.two_view_estimator.create_computation_graph(
+            (
+                i2Ri1,
+                i2Ui1,
+                v_corr_idxs,
+                rot_error,
+                unit_tran_error,
+                corr_correctness_count,
+            ) = self.two_view_estimator.create_computation_graph(
                 keypoints_graph_list[i1],
                 keypoints_graph_list[i2],
                 descriptors_graph_list[i1],
@@ -131,6 +139,7 @@ class SceneOptimizer:
             if gt_pose_graph is not None:
                 frontend_rot3_errors.append(rot_error)
                 frontend_unit3_errors.append(unit_tran_error)
+                frontend_corr_correctness.append(corr_correctness_count)
 
             if self._save_viz:
                 os.makedirs("plots/correspondences", exist_ok=True)
@@ -151,6 +160,7 @@ class SceneOptimizer:
                 dask.delayed(aggregate_frontend_metrics)(
                     frontend_rot3_errors,
                     frontend_unit3_errors,
+                    frontend_corr_correctness,
                     self._pose_angular_error_thresh,
                 )
             )
@@ -321,6 +331,7 @@ def write_sfmdata_to_disk(sfm_data: SfmData, save_fpath: str) -> None:
 def aggregate_frontend_metrics(
     rot3_errors: List[Optional[float]],
     unit3_errors: List[Optional[float]],
+    corr_correctness: List[Tuple[int, float]],
     angular_err_threshold_deg: float,
 ) -> None:
     """Aggregate the front-end metrics to log summary statistics.
@@ -328,6 +339,7 @@ def aggregate_frontend_metrics(
     Args:
         rot3_errors: angular errors in rotations.
         unit3_errors: angular errors in unit-translations.
+        corr_correctness: correctness of correspondences.
         angular_err_threshold: threshold to classify the error as success.
     """
     angular_err_threshold_rad = np.deg2rad(angular_err_threshold_deg)
@@ -347,6 +359,9 @@ def aggregate_frontend_metrics(
     success_count_unit3 = np.sum(unit3_errors < angular_err_threshold_rad)
     success_count_pose = np.sum(pose_errors < angular_err_threshold_rad)
 
+    # count entries with all correct correspondences
+    all_correct = [1 for x in corr_correctness if x[1] == 1.0]
+
     logger.debug(
         "[Two view optimizer] [Summary] Rotation success: %d/%d/%d",
         success_count_rot3,
@@ -365,5 +380,11 @@ def aggregate_frontend_metrics(
         "[Two view optimizer] [Summary] Pose success: %d/%d/%d",
         success_count_pose,
         num_valid_entries,
+        num_entries,
+    )
+
+    logger.debug(
+        "[Two view optimizer] [Summary] All correct correspondences: %d/%d",
+        len(all_correct),
         num_entries,
     )
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -4,7 +4,7 @@ Authors: Ayush Baid, John Lambert
 """
 import logging
 import os
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import dask
 import gtsam
@@ -18,7 +18,6 @@ from dask.delayed import Delayed
 from gtsam import (
     Pose3,
     SfmData,
-    Unit3,
 )
 
 import gtsfm.utils.geometry_comparisons as comp_utils
@@ -26,22 +25,26 @@ import gtsfm.utils.io as io_utils
 import gtsfm.utils.logger as logger_utils
 import gtsfm.utils.serialization  # import needed to register serialization fns
 import gtsfm.utils.viz as viz_utils
-from gtsfm.averaging.rotation.rotation_averaging_base import (
-    RotationAveragingBase,
-)
-from gtsfm.averaging.translation.translation_averaging_base import (
-    TranslationAveragingBase,
-)
 from gtsfm.common.image import Image
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.feature_extractor import FeatureExtractor
-from gtsfm.frontend.detector_descriptor.detector_descriptor_base import (
-    DetectorDescriptorBase,
-)
-from gtsfm.frontend.matcher.matcher_base import MatcherBase
-from gtsfm.frontend.verifier.verifier_base import VerifierBase
 from gtsfm.multi_view_optimizer import MultiViewOptimizer
 from gtsfm.two_view_estimator import TwoViewEstimator
+
+# paths for storage
+PLOT_PATH = "plots"
+PLOT_CORRESPONDENCE_PATH = os.path.join(PLOT_PATH, "correspondences")
+METRICS_PATH = "result_metrics"
+RESULTS_PATH = "results"
+
+"""
+data type for frontend metrics on a pair of images, containing:
+1. rotation angular error
+2. translation angular error
+3. number of correct correspondences
+4. inlier ratio
+"""
+FRONTEND_METRICS_FOR_PAIR = Tuple[Optional[float], Optional[float], int, float]
 
 logger = logger_utils.get_logger()
 
@@ -74,6 +77,12 @@ class SceneOptimizer:
         self._save_bal_files = save_bal_files
         self._pose_angular_error_thresh = pose_angular_error_thresh
 
+        # make directories for persisting data
+        os.makedirs(PLOT_PATH, exist_ok=True)
+        os.makedirs(PLOT_CORRESPONDENCE_PATH, exist_ok=True)
+        os.makedirs(METRICS_PATH, exist_ok=True)
+        os.makedirs(RESULTS_PATH, exist_ok=True)
+
     def create_computation_graph(
         self,
         num_images: int,
@@ -105,9 +114,7 @@ class SceneOptimizer:
         i2Ui1_graph_dict = {}
         v_corr_idxs_graph_dict = {}
 
-        frontend_rot3_errors = []
-        frontend_unit3_errors = []
-        frontend_corr_correctness = []
+        frontend_metrics_dict = {}
 
         for (i1, i2) in image_pair_indices:
             if gt_pose_graph is not None:
@@ -137,12 +144,14 @@ class SceneOptimizer:
             v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
 
             if gt_pose_graph is not None:
-                frontend_rot3_errors.append(rot_error)
-                frontend_unit3_errors.append(unit_tran_error)
-                frontend_corr_correctness.append(corr_correctness_count)
+                frontend_metrics_dict[(i1, i2)] = (
+                    rot_error,
+                    unit_tran_error,
+                    corr_correctness_count[0],
+                    corr_correctness_count[1],
+                )
 
             if self._save_viz:
-                os.makedirs("plots/correspondences", exist_ok=True)
                 auxiliary_graph_list.append(
                     dask.delayed(visualize_twoview_correspondences)(
                         image_graph[i1],
@@ -150,17 +159,17 @@ class SceneOptimizer:
                         keypoints_graph_list[i1],
                         keypoints_graph_list[i2],
                         v_corr_idxs,
-                        "plots/correspondences/{}_{}.jpg".format(i1, i2),
+                        os.path.join(PLOT_CORRESPONDENCE_PATH, "{}_{}.jpg".format(i1, i2)),
                     )
                 )
 
-        # aggregate metrics for frontend
+        # persist all front-end metrics and its summary
         if gt_pose_graph is not None:
+            auxiliary_graph_list.append(dask.delayed(persist_frontend_metrics_full)(frontend_metrics_dict))
+
             auxiliary_graph_list.append(
                 dask.delayed(aggregate_frontend_metrics)(
-                    frontend_rot3_errors,
-                    frontend_unit3_errors,
-                    frontend_corr_correctness,
+                    frontend_metrics_dict,
                     self._pose_angular_error_thresh,
                 )
             )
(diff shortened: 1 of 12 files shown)


johnwlambert:
Thanks for the PR @ayushbaid . Can you merge master in again?

gtsfm/scene_optimizer.py
```diff
@@ -347,6 +359,9 @@ def aggregate_frontend_metrics(
     success_count_unit3 = np.sum(unit3_errors < angular_err_threshold_rad)
     success_count_pose = np.sum(pose_errors < angular_err_threshold_rad)
 
+    # count entries with all correct correspondences
+    all_correct = [1 for x in corr_correctness if x[1] == 1.0]
```
johnwlambert, line 363: is `np.sum()` faster?

gtsfm/utils/features.py
```diff
@@ -47,3 +47,65 @@ def normalize_coordinates(coordinates: np.ndarray, intrinsics: Cal3Bundler) -> n
     """
 
     return np.vstack([intrinsics.calibrate(x[:2].reshape(2, 1)) for x in coordinates])
+
```
johnwlambert, line 50: is features.py the right name for this file?

johnwlambert: keypoint_utils.py?

gtsfm/scene_optimizer.py
ayushbaid: I have to convert it to a numpy array anyway?

gtsfm/utils/features.py
ayushbaid: Good point. But should I do it in this review?

gtsfm/utils/features.py
```diff
@@ -47,3 +47,65 @@ def normalize_coordinates(coordinates: np.ndarray, intrinsics: Cal3Bundler) -> n
...
+    Args:
+        points: non-homogenous 2D points, of shape Nx2.
+        lines: lines, of shape Nx3.
+
+    Returns:
+        Point-line distance for each row, of shape N.
+    """
+    line_norms = np.linalg.norm(lines[:, :2], axis=1)
+
+    return np.abs(np.einsum("ij,ij->i", convert_to_homogenous_coordinates(points), lines)) / line_norms
```
johnwlambert, line 111: i personally dont find einsum to be too readable, so I'd prefer not to use it

gtsfm/two_view_estimator.py
```diff
@@ -23,6 +26,8 @@
 pil_logger = logging.getLogger("PIL")
 pil_logger.setLevel(logging.INFO)
 
+ESSENTIAL_MATRIX_EPIPOLAR_DISTANCE_THRESHOLD = 0.1  # TODO: move to config?
```
johnwlambert, line 29: i prefer in config

gtsfm/scene_optimizer.py
johnwlambert: maybe if the numpy array can use a memcpy from the list array that is faster than a for-loop in the list comprehension

tests/utils/test_feature_utils.py
```diff
@@ -21,3 +21,110 @@ def test_normalize_coordinates(self):
...
+        )
+        lines = np.array(
+            [
+                [4.0, -3.0, -4.0],
+                [0.0, -0.5, 1.0],
+                [-3.0, -4.0, -5.0],
+            ]
+        )
+        expected = np.array([3.0, 1.0, 4.6])
+
```
johnwlambert, line 123: i'd prefer to remove the whitespace between these 3 lines

tests/utils/test_feature_utils.py
```diff
@@ -21,3 +21,110 @@ def test_normalize_coordinates(self):
...
+        points = np.array(
+            [
+                [-2.0, 1.0],
+                [5.0, 1.0],
+                [2.0, 3.0],
+            ]
+        )
+        lines = np.array(
+            [
+                [4.0, -3.0, -4.0],
```
johnwlambert, line 117: could you add an inline commend stating these are `a,b,c` coefficients?

gtsfm/utils/features.py
```diff
@@ -47,3 +47,65 @@ def normalize_coordinates(coordinates: np.ndarray, intrinsics: Cal3Bundler) -> n
...
+    return epipolar_lines
+
+
+def compute_point_line_distances(points: np.ndarray, lines: np.ndarray) -> np.ndarray:
+    """Computes the distance of a point from a line in 2D. The function processed multiple inputs independently in a
+    vectorized fashion.
+
+    Args:
+        points: non-homogenous 2D points, of shape Nx2.
+        lines: lines, of shape Nx3.
```
johnwlambert, line 104: i presume 3-tuple consists of a, b, c, where ax + by = c? or ax + by + c = 0? -- can we add that to the docstring

tests/utils/test_feature_utils.py
```diff
@@ -21,3 +21,110 @@ def test_normalize_coordinates(self):
...
+                [3.5, 20.0],
+            ]
+        )  # 2d points in homogenous coordinates
+        essential_mat = EssentialMatrix(Rot3.RzRyRx(0, np.deg2rad(45), 0), Unit3(np.array([-5, 2, 0])))
+
+        computed = feature_utils.convert_to_epipolar_lines(points, essential_mat)
+
+        expected = np.array(
+            [
+                [-2.36351579, -5.90878948, 1.75364193],
```
johnwlambert, line 78: where do these numbers come from? just doing the paper on math, or from a real example?

johnwlambert:
@ayushbaid could you make a plot, showing an image pair, and color keypoints away from line more and more red, in green-to-red colormap?

For example:
```python
from colour import Color

NUM_RANGE_BINS = 50

colors_arr = np.array(
      [[color_obj.rgb] for color_obj in Color("red").range_to(Color("green"), NUM_RANGE_BINS)]
).squeeze()
np.fliplr(colors_arr)

pt_ranges = np.linalg.norm(uv_cam[:, :3], axis=1)
rgb_bins = np.round(pt_ranges).astype(np.int32)
# account for moving past 100 meters, loop around again
rgb_bins = rgb_bins % NUM_RANGE_BINS
uv_colors = (255 * colors_arr[rgb_bins]).astype(np.int32)

img = draw_point_cloud_in_img_cv2(img, uv, np.fliplr(uv_colors))
```

tests/utils/test_feature_utils.py
ayushbaid: I did math on paper upto two decimal places, checked it, and copied over the computed values for more precision

gtsfm/utils/features.py
ayushbaid: Here I want to do row-wise dot product. If I do not use einsum, I have to use `np.sum(np.multiply(a, b), axis=1)`. I thought this alternative is not elegant.

gtsfm/two_view_estimator.py
ayushbaid: This is technically an evaluation related param though.

johnwlambert: evaluation is something to configure, though, since there's no one correct way to evaluate

gtsfm/utils/features.py
johnwlambert: can we write it with a matrix multiplication?

johnwlambert: bmm?

ayushbaid: Does numpy has bmm?

gtsfm/two_view_estimator.py
ayushbaid: Makes sense. Although I'd like to somehow have two different configs (or sections) for building the pipeline and evaluating the pipeline.

johnwlambert: agreed, let's leave that as a to-do

gtsfm/utils/features.py
johnwlambert: can you open an issue for it to be addressed in the future?

ayushbaid:
> @ayushbaid could you make a plot, showing an image pair, and color keypoints away from line more and more red, in green-to-red colormap?
> 
> For example:
> 
> ```python
> from colour import Color
> 
> NUM_RANGE_BINS = 50
> 
> colors_arr = np.array(
>       [[color_obj.rgb] for color_obj in Color("red").range_to(Color("green"), NUM_RANGE_BINS)]
> ).squeeze()
> np.fliplr(colors_arr)
> 
> pt_ranges = np.linalg.norm(uv_cam[:, :3], axis=1)
> rgb_bins = np.round(pt_ranges).astype(np.int32)
> # account for moving past 100 meters, loop around again
> rgb_bins = rgb_bins % NUM_RANGE_BINS
> uv_colors = (255 * colors_arr[rgb_bins]).astype(np.int32)
> 
> img = draw_point_cloud_in_img_cv2(img, uv, np.fliplr(uv_colors))
> 
> 
> 
> def draw_point_cloud_in_img_cv2(img: np.ndarray, xy: np.ndarray, colors: np.ndarray, radius: int = 5) -> np.ndarray:
>     """Plot a point cloud in an image by drawing small circles centered at (x,y) locations.
>     Note these are not (u,v) but rather (v,u) coordinate pairs.
>     Args:
>         img: Array of shape (M, N, 3), representing an image with channel order BGR, per the OpenCV convention
>         xy: Array of shape (K, 2) representing the center coordinates of each circle
>         colors: Array of shape (K, 3), with BGR values in [0, 255] representing the fill color for each circle
>         radius: radius of all circles
>     Returns:
>         img: Array of shape (M, N, 3), with all circles plotted
>     """
>     for i, (x, y) in enumerate(xy):
>         rgb = colors[i]
>         rgb = tuple([int(intensity) for intensity in rgb])
>         img = cv2.circle(img, (x, y), radius, tuple(rgb), -1)
>     return img
> ```

Finally, I have the image!

Correspondence distance w.r.t epipolar lines

gtsfm/utils/features.py
johnwlambert: matmul can do something similar. But i personally think `np.sum(np.multiply(a, b), axis=1)` is clear and good

ayushbaid: resolved

gtsfm/utils/features.py
ayushbaid: done

gtsfm/scene_optimizer.py
```diff
@@ -321,14 +331,16 @@ def write_sfmdata_to_disk(sfm_data: SfmData, save_fpath: str) -> None:
...
 ) -> None:
     """Aggregate the front-end metrics to log summary statistics.
 
     Args:
-        rot3_errors: angular errors in rotations, measured in degrees
-        unit3_errors: angular errors in unit-translations, measured in degrees
-        angular_err_threshold_deg: threshold to classify the error as success, in degrees
+        rot3_errors: angular errors in rotations, measured in degrees.
+        unit3_errors: angular errors in unit-translations, measures in degrees.
+        corr_correctness: correctness of correspondences.
```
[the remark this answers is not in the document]
ayushbaid: ah yes. They will represent `(num_correct_correspondence, ratio_of_correct_correspondences)`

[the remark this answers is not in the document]
johnwlambert: judging from another file, it seems like 
```
	        Number of correct correspondences.
        Ratio of correspondences which are correct.
```
but I would not have guessed that from this docstring.

Is a NamedTuple more appropriate?
```python
class TwoViewCorrectnessResult(NamedTuple):
      num_correct_corr: float
      inlier_fraction: float
```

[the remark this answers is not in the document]
ayushbaid: Named tuple will create issues for dask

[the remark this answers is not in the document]
johnwlambert: we use it in data association without trouble though right?

tests/utils/test_verification_utils.py
```diff
@@ -20,13 +21,28 @@ def test_recover_relative_pose_from_essential_matrix(self):
...
-            Cal3Bundler(),
+            i2Ei1.matrix(), corr_i1.coordinates, corr_i2.coordinates, Cal3Bundler(), Cal3Bundler()
         )
 
         # compare the recovered R and U with the ground truth
         self.assertTrue(i2Ri1.equals(i2Ei1.rotation(), 1e-3))
         self.assertTrue(i2Ui1.equals(i2Ei1.direction(), 1e-3))
+
+    def test_compute_epipolar_distances(self):
+        """Test for epipolar distance computation using 2 sets of points and the essential matrix."""
```
johnwlambert, line 32: where does this example come from? maybe we could add some more intuition behind it

gtsfm/scene_optimizer.py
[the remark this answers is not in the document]
ayushbaid: Yes we do. But I remember I had to change one of keypoints or images to a normal class because dask was unable to handle it. I guess it depends on whether the variable has to be passed around. In our case, it will be returned from a function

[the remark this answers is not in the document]
ayushbaid: Yes we do. But I remember I had to change one of keypoints or images to a normal class because dask was unable to handle it. I guess it depends on whether the variable has to be passed around. In our case, it will be returned from a function

tests/utils/test_feature_utils.py
```diff
@@ -21,3 +21,109 @@ def test_normalize_coordinates(self):
...
+        points = np.array(
+            [
+                [-2.0, 1.0],
+                [5.0, 1.0],
+                [2.0, 3.0],
+            ]
+        )
+        lines = np.array(  # coefficients (a, b, c) for the line ax + by + cz = 0
+            [
+                [4.0, -3.0, -4.0],
```
johnwlambert, line 117: could we make this test case simpler? like lines with normal (1,0) or (1,2) or (0,1)?

(4,-3) is a bit hard already to do in one's head

gtsfm/utils/verification.py
```diff
@@ -71,3 +56,32 @@ def fundamental_to_essential_matrix(
...
+        normalized_coords_i2: corr. normalized coordinates in image i2, of shape Nx2.
+        i2Ei1: essential matrix between two images
+
+    Returns:
+        Symmetric epipolar distances for each row of the input, of shape N.
+    """
+    # construct the essential matrix in the opposite directin
+    i2Ti1 = Pose3(i2Ei1.rotation(), i2Ei1.direction().point3())
+    i1Ti2 = i2Ti1.inverse()
+    i1Ei2 = EssentialMatrix(i1Ti2.rotation(), Unit3(i1Ti2.translation()))
```
johnwlambert, line 77: is this equivalent to `i1Ei2.T`?

gtsfm/scene_optimizer.py
```diff
@@ -346,6 +358,9 @@ def aggregate_frontend_metrics(
     success_count_unit3 = np.sum(unit3_errors < angular_err_threshold_deg)
     success_count_pose = np.sum(pose_errors < angular_err_threshold_deg)
 
+    # count entries with all correct correspondences
```
johnwlambert, line 361: i might just add to the inline comment that we are checking which rows have inlier ratio == 1

gtsfm/two_view_estimator.py
```diff
@@ -85,19 +90,74 @@ def create_computation_graph(
...
+    number_correct = metric_utils.count_correct_correspondences(
+        keypoints_i1.extract_indices(corr_idxs_i1i2[:, 0]),
+        keypoints_i2.extract_indices(corr_idxs_i1i2[:, 1]),
+        intrinsics_i1,
+        intrinsics_i2,
+        i2Ti1,
+        epipolar_distance_threshold,
+    )
+
+    # logger.debug(
```
johnwlambert, line 152: did we mean to comment this one out?

gtsfm/two_view_estimator.py
```diff
@@ -115,10 +175,8 @@ def compute_relative_pose_metrics(
         Unit translation error, in degrees
     """
     R_error_deg = comp_utils.compute_relative_rotation_angle(i2Ri1_computed, i2Ti1_expected.rotation())
-
-    U_error_deg = comp_utils.compute_relative_unit_translation_angle(i2Ui1_computed, Unit3(i2Ti1_expected.translation()))
-
-    logger.debug("[Two View Estimator] Relative rotation error %f deg.", R_error_deg)
```
johnwlambert, line 121: i think these could be useful to keep printing these in some sort of debug mode. or at least i'd like to dump them to disk instead of remove them

johnwlambert: i prefer more verbosity now while we figure out all the issues

gtsfm/utils/features.py
```diff
@@ -45,5 +45,66 @@ def normalize_coordinates(coordinates: np.ndarray, intrinsics: Cal3Bundler) -> n
...
+        i2Ei1: essential matrix.
+
+    Returns:
+        Corr. epipolar lines in i2, of shape Nx3
+    """
+    if normalized_coordinates_i1 is None or normalized_coordinates_i1.size == 0:
+        return normalized_coordinates_i1
+
+    epipolar_lines = convert_to_homogenous_coordinates(normalized_coordinates_i1) @ i2Ei1.matrix().T
+
```
johnwlambert, line 93: nit: prefer to remove this extra whitespace here

gtsfm/utils/features.py
```diff
@@ -45,5 +45,66 @@ def normalize_coordinates(coordinates: np.ndarray, intrinsics: Cal3Bundler) -> n
...
+    return np.hstack(
+        (
+            non_homogenous_coordinates,
+            np.ones((non_homogenous_coordinates.shape[0], 1)),
+        )
+    )
+
+
+def convert_to_epipolar_lines(normalized_coordinates_i1: np.ndarray, i2Ei1: EssentialMatrix) -> np.array:
+    """Convert coordinates to epipolar lines.
```
johnwlambert, line 80: maybe inline math is helpful here in the comment? like 

```python
points live on the line if
    x2^T @ E @ x1 = 0
where E is i2Ei1
```

tests/utils/test_verification_utils.py
ayushbaid: I just wrote some random points. No intuition behind them.

tests/utils/test_feature_utils.py
ayushbaid: I deliberately choose combination of 3 and 4 because the L2 norm is 5. This is important because we divide by the L2 norm of `[a, b]`.

gtsfm/utils/verification.py
ayushbaid: Yes it is. 

If there was a transpose function, we could use it directly. But the `EssentialMatrix` object internally stores a `Rot3` and `Unit3` so there is no direct way to transpose. I can move this exact logic to GTSAM if you think thats better.

gtsfm/scene_optimizer.py
ayushbaid: done.

gtsfm/two_view_estimator.py
ayushbaid: should be removed, as we are not logging them anymore.

gtsfm/two_view_estimator.py
ayushbaid: We will dump them as JSON anyways. Will just keep them off stdout. What do you prefer?

gtsfm/utils/features.py
ayushbaid: done

gtsfm/utils/features.py
ayushbaid: better?

johnwlambert: yes, thanks

gtsfm/two_view_estimator.py
johnwlambert: are they getting dumped to JSON now? if not, do you mind opening an issue for us to do that ASAP?

gtsfm/utils/verification.py
johnwlambert: fine as it is, no worries

tests/utils/test_feature_utils.py
```diff
@@ -21,3 +21,109 @@ def test_normalize_coordinates(self):
...
+        """Test for 2D point-line distance computation."""
+
+        points = np.array(
+            [
+                [-2.0, 1.0],
+                [5.0, 1.0],
+                [2.0, 3.0],
+            ]
+        )
+        lines = np.array(  # coefficients (a, b, c) for the line ax + by + cz = 0
```
johnwlambert, line 115: there is no z here right? just ax + by + c = 0?

tests/utils/test_feature_utils.py
johnwlambert: i still think there is an easier example

johnwlambert: for example, (a,b,c) = (1,0,-1) and (x,y) = (-2,1). distance should be 3

johnwlambert: another simple one is (a,b,c) = (-1,-1,2) and (x,y) = (2,2) should be distance = sqrt(2)

johnwlambert: can we just use those two simpler examples instead?

johnwlambert (approved):
Approved, pending a few minor comments i left

tests/utils/test_feature_utils.py
ayushbaid: yes thats right. Thanks!

gtsfm/two_view_estimator.py
ayushbaid: I can make it part of this PR?

tests/utils/test_feature_utils.py
ayushbaid: Changed. Does it look better?

gtsfm/two_view_estimator.py
johnwlambert: sure, whichever you prefer

gtsfm/scene_optimizer.py
```diff
@@ -367,6 +382,12 @@ def aggregate_frontend_metrics(
         num_entries,
     )
 
+    logger.debug(
+        "[Two view optimizer] [Summary] All correct correspondences: %d/%d",
```
johnwlambert, line 386: Maybe a clearer log message is: `Number of image pairs with 100% inlier ratio:`

gtsfm/utils/features.py
```diff
@@ -45,5 +45,67 @@ def normalize_coordinates(coordinates: np.ndarray, intrinsics: Cal3Bundler) -> n
...
+
+    Args:
+        normalized_coordinates_i1: normalized coordinates in i1, of shape Nx2.
+        i2Ei1: essential matrix.
+
+    Returns:
+        Corr. epipolar lines in i2, of shape Nx3
+    """
+    if normalized_coordinates_i1 is None or normalized_coordinates_i1.size == 0:
+        return normalized_coordinates_i1
```
johnwlambert, line 92: are we handling this failure mode correctly? why return normalized coords instead of epipolar lines? should we return None?

tests/utils/test_feature_utils.py
johnwlambert: yes, thanks

gtsfm/utils/features.py
ayushbaid: I wanted to return an empty array if the input was empty, and None if the input was None. I thought this will be easier to do instead of two if conditions

gtsfm/scene_optimizer.py
ayushbaid: oh yeah. this is way better :P

gtsfm/utils/features.py
johnwlambert: aren't we returning an incorrect quantity though?

ayushbaid: I did not get this.

johnwlambert: @ayushbaid I just meant that this function is supposed to return epipolar_lines.

But if we `return normalized_coordinates_i1`, we are not returning what this function is supposed to return. So we can't trust its output in that case, and we'll get inaccurate distances down the road.

I would prefer we return something like None and catch it, so we can abort for this image pair.

ayushbaid: got it. Makes sense and I'll implement this. I guess returning None is cleaner so I'll make that change.

gtsfm/utils/verification.py
```diff
@@ -42,32 +42,62 @@ def recover_relative_pose_from_essential_matrix(
...
+
+    # get lines in i2 and i1
+    epipolar_lines_i2 = feature_utils.convert_to_epipolar_lines(normalized_coords_i1, i2Ei1)
```
johnwlambert, line 96: @ayushbaid we need to check if these are None right? if so, let's return NaN?

gtsfm/scene_optimizer.py
```diff
@@ -18,31 +18,34 @@
...
+RESULTS_PATH = "results"
+
+"""
+data type for frontend metrics on a pair of images, containing:
+1. rotation angular error
+2. translation angular error
+3. number of correct correspondences
+4. inlier ratio
+"""
+FRONTEND_METRICS_FOR_PAIR = Tuple[Optional[float], Optional[float], int, float]
```
johnwlambert, line 47: why are only the first two optional @ayushbaid ? not all 4?

ayushbaid: When the verification fails, the number of correct correspondences is zero and inlier ratio is 0 too. Hence I don't need optional for them.

gtsfm/utils/verification.py
ayushbaid: NaN is more appropriate for elements in the array, right? How can I return a NaN in place of a numpy array?

johnwlambert: forgot we were returning an array of distances.

i guess "None" then?

ayushbaid: Yes None is better imo.

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,45 @@
...
+        return 0
+
+    normalized_coords_i1 = feature_utils.normalize_coordinates(keypoints_i1.coordinates, intrinsics_i1)
+    normalized_coords_i2 = feature_utils.normalize_coordinates(keypoints_i2.coordinates, intrinsics_i2)
+    i2Ei1 = EssentialMatrix(i2Ti1.rotation(), Unit3(i2Ti1.translation()))
+
+    epipolar_distances = verification_utils.compute_epipolar_distances(
+        normalized_coords_i1, normalized_coords_i2, i2Ei1
+    )
+    return np.count_nonzero(epipolar_distances < epipolar_dist_threshold)
```
johnwlambert, line 45: if we return None for `epipolar_distances` here, do we need to catch it?

gtsfm/scene_optimizer.py
```diff
@@ -105,16 +114,22 @@ def create_computation_graph(
...
                 gt_relative_pose = None
 
-            (i2Ri1, i2Ui1, v_corr_idxs, rot_error, unit_tran_error,) = self.two_view_estimator.create_computation_graph(
+            (
+                i2Ri1,
+                i2Ui1,
+                v_corr_idxs,
+                rot_error,
+                unit_tran_error,
+                corr_correctness_count,
```
johnwlambert, line 131: not sure if `corr_correctness_count,` is the right variable name -- maybe `correspondence_stats`? or `pair_corr_stats`?

ayushbaid: I'd go with `correspondence_stats`.

gtsfm/utils/metrics.py
ayushbaid: If I add this check: `if len(keypoints_i1) == 0 or len(keypoints_i2) == 0`, we wont run into Nones

johnwlambert (approved):

PR #94, opened 2021-02-11
Argoverse image loader

johnwlambert:
Provides data loader for Argoverse imagery and ground truth

PR #96, opened 2021-02-15
Adding metrics for rotation and translation averaging. 

akshay-krishnan:
The following 3 metrics have been added: 
- Rotation angle error (comparison to ground truth). 
- Translation distance error (comparison to ground truth). 
- Translation angle with direction measurements (comparison to measurements). 

The metrics are currently logged to a JSON file. 
Example log [here](https://drive.google.com/file/d/1C6ZQRwjWNTf3p4PKGWcu6y-6jhzj5z2m/view?usp=sharing)


akshay-krishnan:
I understand we wanted to limit to one metric per module, but as mentioned in the PR description, the translation distance error is a comparison to Ground truth and translation angle is a comparison to measurements. The former should be preferred if we have datasets with good metric ground truth camera poses.

akshay-krishnan:
@ayushbaid or @johnwlambert is there a formatting standard or setting that is shared for this project? I am using PEP8 on pyyapf which is based on [YAPF](https://github.com/google/yapf) and it looks like our formatting settings differ.

johnwlambert:
Thanks for the PR, @akshay-krishnan. We've been using Python black (https://github.com/psf/black).

Would you mind pulling the latest master into your branch?

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+            translations.append(pose.translation())
+        return rotations, translations
+
+    wRi_aligned_list, wti_aligned_list = get_rotations_translations_from_poses(
+        wTi_aligned_list)
+    gt_wRi_list, gt_wti_list = get_rotations_translations_from_poses(
+        gt_wTi_list)
+
+    metrics = {}
+    metrics['rotation_averaging_angle'] = compute_rotation_angle_metrics(
```
johnwlambert, line 161: looks good. Thanks for the PR Akshay.

Could we save this as "'rotation_averaging_angle_deg" to make it super clear to the reader what the units are?

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+        for pose in poses:
+            if pose is None:
+                rotations.append(None)
+                translations.append(None)
+                continue
+            rotations.append(pose.rotation())
+            translations.append(pose.translation())
+        return rotations, translations
+
+    wRi_aligned_list, wti_aligned_list = get_rotations_translations_from_poses(
```
johnwlambert, line 155: can we reformat this file with python black?

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+from gtsam import (Point3, Rot3, Pose3, Unit3)
+
+import gtsfm.utils.geometry_comparisons as comp_utils
+from gtsfm.common.sfm_result import SfmResult
+
+# A StatsDict is a dict from string to optional floats or their lists.
+StatsDict = Dict[str, Union[Optional[float], List[Optional[float]]]]
+
+
+def get_errors_statistics(errors: List) -> StatsDict:
```
johnwlambert, line 20: would this be `List[float]` in type hint?

gtsfm/scene_optimizer.py
```diff
@@ -168,8 +159,22 @@ def create_computation_graph(
             i2Ui1_graph_dict,
             v_corr_idxs_graph_dict,
             camera_intrinsics_graph,
+            gt_pose_graph
         )
 
+        # aggregate metrics
+        if gt_pose_graph is not None:
```
johnwlambert, line 166: i would prefer to keep `aggregate_frontend_metrics` before the multi-view step, so front-end metrics get dumped by Dask, even if the back-end crashes.

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+    Args:
+        errors: List of errors for a metric.
+
+    Returns: 
+        A dict with keys min_error, max_error, median_error, 
+        and errors_list mapping to the respective stats.
+    """
+
+    metrics = {}
+    valid_errors = [error for error in errors if error is not None]
```
johnwlambert, line 32: is numpy without a list comprehension faster?
```python
is_valid_bool = ~np.np.isnan(errors.astype(np.float32)
valid_errors = np.array(errors)[is_valid_bool]
```

gtsfm/utils/geometry_comparisons.py
```diff
@@ -205,3 +205,30 @@ def compute_relative_unit_translation_angle(U_1: Optional[Unit3], U_2: Optional[
     angle_rad = np.arccos(dot_product)
     angle_deg = np.rad2deg(angle_rad)
     return angle_deg
+
+
+def compute_translation_to_direction_angle(
```
johnwlambert, line 210: looks good, can we add a unit test for this one?

gtsfm/utils/metrics.py
johnwlambert: can we add  unit test for this fn?

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+        A statistics dict of the metrics errors in degrees. 
+    """
+    errors = []
+    for (wRi, gt_wRi) in zip(wRi_list, gt_wRi_list):
+        angle = comp_utils.compute_relative_rotation_angle(wRi, gt_wRi)
+        errors.append(angle)
+    return get_errors_statistics(errors)
+
+
+def compute_translation_distance_metrics(
```
johnwlambert, line 62: unit test here would be nice too

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,197 @@
...
+        gt_wTi_list)
+
+    metrics = {}
+    metrics['rotation_averaging_angle'] = compute_rotation_angle_metrics(
+        wRi_aligned_list, gt_wRi_list)
+    metrics[
+        'translation_averaging_distance'] = compute_translation_distance_metrics(
+            wti_aligned_list, gt_wti_list)
+    metrics[
+        'translation_to_direction_angle'] = compute_translation_angle_metrics(
```
johnwlambert, line 167: i would add that this is in degrees also:

`translation_to_direction_angle_deg`

gtsfm/utils/metrics.py
akshay-krishnan: Apparently, it is not, I just tried it out. List: 2.19e-05 np: 9.08e-05 (in seconds). Probably it is the two iterations that it needs if done the np way?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -205,3 +205,50 @@ def compute_relative_unit_translation_angle(U_1: Optional[Unit3], U_2: Optional[
...
+    i2Ui1: Optional[Unit3],
+    wTi2: Optional[Point3],
```
johnwlambert, line 212: Thanks for the updates, @akshay-krishnan . Should wTi2 be Optional[Pose3] in args?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -205,3 +205,50 @@ def compute_relative_unit_translation_angle(U_1: Optional[Unit3], U_2: Optional[
...
+    wTi1: Optional[Pose3]
+) -> Optional[float]:
+    """Compute the angle between a translation direction and point.
```
johnwlambert, line 215: is this docstring accurate? no point is passed in, right? just 2 poses and a translation dir?

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: Good catch! updated.

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: Updated it, hope its better now.

gtsfm/utils/metrics.py
akshay-krishnan: Do we really need unit tests for this file? All the logic has been moved to geometry_comparisons utils and is unit tested. Here we are using the same logic on a container and logging to JSON.

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,195 @@
...
+    metrics["translation_averaging_distance"] = compute_translation_distance_metrics(
+        wti_aligned_list, gt_wti_list
+    )
+    metrics["translation_to_direction_angle_deg"] = compute_translation_angle_metrics(
+        i2Ui1_dict, wTi_aligned_list
+    )
+    return metrics
+
+
+def save_averaging_metrics(
```
ayushbaid, line 172: This function should not be in the file. Ideally we would want only computations here.

gtsfm/multi_view_optimizer.py
```diff
@@ -80,8 +83,15 @@ def create_computation_graph(
 
         ba_result_graph = self.ba_optimizer.create_computation_graph(ba_input_graph)
 
+        if gt_poses_graph is not None:
+            self.metrics_graph = dask.delayed(metrics.save_averaging_metrics)(
+                i2Ui1_graph, wRi_graph, wti_graph, gt_poses_graph, 'metrics')
+
         return ba_input_graph, ba_result_graph
 
+    def get_metrics_computation_graph(self) -> Delayed:
```
ayushbaid, line 92: docstring

gtsfm/multi_view_optimizer.py
```diff
@@ -80,8 +83,15 @@ def create_computation_graph(
 
         ba_result_graph = self.ba_optimizer.create_computation_graph(ba_input_graph)
 
+        if gt_poses_graph is not None:
+            self.metrics_graph = dask.delayed(metrics.save_averaging_metrics)(
```
ayushbaid, line 87: This is a clean design, but I dont like the fact the function `create_computation_graph` changes the attributes of the class. As we sticking with a functional paradigm, I'd like no side-effects

gtsfm/utils/geometry_comparisons.py
```diff
@@ -205,3 +205,50 @@ def compute_relative_unit_translation_angle(U_1: Optional[Unit3], U_2: Optional[
     angle_rad = np.arccos(dot_product)
     angle_deg = np.rad2deg(angle_rad)
     return angle_deg
+
+
+def compute_translation_to_direction_angle(
+    i2Ui1: Optional[Unit3],
+    wTi2: Optional[Pose3],
+    wTi1: Optional[Pose3]
+) -> Optional[float]:
```
ayushbaid, line 214: I have a function which does similar stuff in the front-end. I compute angles between two `Unit3`s. I think that is a cleaner API instead of having a `Unit3` and two `Pose3`s.

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,195 @@
...
+    metrics = {}
+    valid_errors = [error for error in errors if error is not None]
+    metrics["median_error"] = np.median(valid_errors)
+    metrics["min_error"] = np.min(valid_errors)
+    metrics["max_error"] = np.max(valid_errors)
+    metrics["errors_list"] = errors
+    return metrics
+
+
+def compute_rotation_angle_metrics(
```
ayushbaid, line 40: We have a similar function in `geometry_comparisons`. Can you check if we need an additional function.

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,195 @@
...
+from gtsam import Point3, Rot3, Pose3, Unit3
+
+import gtsfm.utils.geometry_comparisons as comp_utils
+from gtsfm.common.sfm_result import SfmResult
+
+# A StatsDict is a dict from string to optional floats or their lists.
+StatsDict = Dict[str, Union[Optional[float], List[Optional[float]]]]
+
+
+def get_errors_statistics(errors: List[Optional[float]]) -> StatsDict:
```
ayushbaid, line 20: I'd rename it to `compute_error_summary` or `compute_errors_statistics` as it is not a getter.

ayushbaid (commented):
Some comments about the functions and overall design.

tests/utils/test_geometry_comparisons.py
```diff
@@ -25,33 +25,46 @@ def test_compare_poses_with_uniform_scaled_translations(self):
         """Check pose comparison with all translations in input #2 scaled by
         the same scalar factor."""
         scale_factor = 1.2
-        pose_list_ = [Pose3(x.rotation(), x.translation() * scale_factor) for x in POSE_LIST]
+        pose_list_ = [
```
johnwlambert, line 28: i forgot to mention we are using `black myfile.py -l 120` to allow for a 120-line limit, instead of the default 88, since we thought it made things a bit more readable and less choppy

gtsfm/utils/metrics.py
johnwlambert: i think i agree, perhaps better in `multi_view_optimizer.py`?

gtsfm/multi_view_optimizer.py
```diff
@@ -80,8 +83,15 @@ def create_computation_graph(
...
+                i2Ui1_graph, wRi_graph, wti_graph, gt_poses_graph, 'metrics')
```
johnwlambert, line 88: we recently merged another branch that saves to "result_metrics", can we use the same one here?

https://github.com/borglab/gtsfm/pull/94/files#diff-e715d8778a0c656f504e46468a51441480965c02932c4b2256e30a2239335e4eR388

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,195 @@
...
+    Args:
+        wti_list: List of estimated camera translations.
+        gt_wti_list: List of ground truth camera translations.
+
+    Returns:
+        A statistics dict of the metrics errors in degrees.
+    """
+    errors = []
+    for (wti, gt_wti) in zip(wti_list, gt_wti_list):
+        errors.append(comp_utils.compute_points_distance_l2(wti, gt_wti))
```
johnwlambert, line 78: is there a gauge ambiguity here that we need to account for?

gtsfm/utils/metrics.py
akshay-krishnan: But we have no computations here, maybe I should name the file something different? All the functions here simply iterate over multiple cameras and call other functions. I'm happy to move it though. I don't think it would belong in `multi_view_optimizer` either. Do we have a different utils file for IO?

akshay-krishnan: Looks like we have `io.py`

gtsfm/utils/metrics.py
akshay-krishnan: `compute_errors_statistics` sounds good.

gtsfm/utils/metrics.py
akshay-krishnan: There is a function that compares two lists of rotations. 
There is another function that returns the angle between two individual rotations (which I call in this function). I could do all the looping within `compute_averaging_metrics`, but I think the code is more modular this way.

akshay-krishnan: Let me know if you have other ideas.

tests/utils/test_geometry_comparisons.py
akshay-krishnan: Sure, I can change that. But I should mention that the 80 character limit is actually based on a lot of discussion on readability and is widely accepted. 
For more details: https://www.python.org/dev/peps/pep-0008/#maximum-line-length
https://google.github.io/styleguide/pyguide.html#32-line-length

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: Indeed, and I call that function here. I've been seeing this as a utility file, and `compute_translation_to_direction_angle` is a high-level utility function, I could move the logic of finding the unit translation between two translations to within the function that calls this, but would that not make the code less modular?

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+    if len(wRi_list) != len(wti_list) or len(wRi_list) != len(gt_wTi_list):
+        raise ValueError("Lengths of wRi_list, wti_list and gt_wTi_list should be the same.")
+
+    wTi_list = []
+    for (wRi, wti) in zip(wRi_list, wti_list):
+        wTi_list.append(Pose3(wRi, wti))
+    wTi_aligned_list = comp_utils.align_poses(wTi_list, gt_wTi_list)
+
+    def get_rotations_translations_from_poses(poses):
```
johnwlambert, line 134: Thanks for all the updates, Akshay.
An inline type hint would be nice here

gtsfm/utils/metrics.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Args:
+        i2Ui1_dict: List of translation direction measurements.
+        wTi_list: List of estimated camera poses.
+
+    Returns:
+        A statistics dict of the metrics errors in degrees.
+    """
+    angles = []
+    for (i1, i2) in i2Ui1_dict:
```
johnwlambert, line 93: i might prefer to explicitly say `i2Ui1_dict.keys()` to emphasize its not a list

gtsfm/utils/metrics.py
johnwlambert: cool, thanks for checking

tests/utils/test_geometry_comparisons.py
```diff
@@ -104,6 +104,44 @@ def test_compare_global_poses_scaled_squares(self):
...
+        wTi1_estimated = Pose3(Rot3(), Point3(2, 0, 0))
+        self.assertEqual(
+            geometry_comparisons.compute_translation_to_direction_angle(i2Ui1_measured, wTi2_estimated, wTi1_estimated),
+            0.0,
+        )
+
+    def test_compute_translation_to_direction_angle_is_nonzero(self):
+        i2Ui1_measured = Unit3(Point3(0, 1, 0))
+        wRi2 = Rot3.RzRyRx(-np.deg2rad(90), 0, 0)  # x-axis points to -y in world frame
+        wRi1 = Rot3.RzRyRx(np.deg2rad(30), np.deg2rad(60), np.deg2rad(90))  # irrelevant
```
johnwlambert, line 119: wRi1 defined but not used?

tests/utils/test_geometry_comparisons.py
```diff
@@ -104,6 +104,44 @@ def test_compare_global_poses_scaled_squares(self):
...
+        wTi2_estimated = Pose3(wRi2, Point3(0, 0, 0))
+        wTi1_estimated = Pose3(Rot3(), Point3(0, -1, 0))  # along -y axis
+        # estimated direction along x-axis and measured along y-axis in i2 frame.
+        self.assertTrue(
+            geometry_comparisons.compute_translation_to_direction_angle(i2Ui1_measured, wTi2_estimated, wTi1_estimated),
+            90.0,
```
johnwlambert, line 125: should it be 180 degrees, not 90 degrees @akshay-krishnan ? Since i2Ui1_measured is (0,1,0)

but poses suggest (0,-1,0)

gtsfm/utils/metrics.py
johnwlambert: i'm a bit ambivalent, either way

gtsfm/utils/geometry_comparisons.py
johnwlambert: yes, thanks

gtsfm/utils/metrics.py
johnwlambert: i dont think there is any ideal place to put it. so i think `metrics.py` is ok.

`io.py` is for library fns, and i think this is too specific for `io.py`

gtsfm/utils/geometry_comparisons.py
johnwlambert: IMHO it is ok to have this as a separate more modular function

gtsfm/utils/metrics.py
johnwlambert: i think it's ok

johnwlambert:
Looks great @akshay-krishnan . I left 3 minor comments, then I will approve

johnwlambert:
@akshay-krishnan can you merge master into your branch again? After latest PR merge

gtsfm/utils/metrics.py
akshay-krishnan: Style guides generally prefer using default iterators when iterating over lists and dicts. 
https://google.github.io/styleguide/pyguide.html#28-default-iterators-and-operators
The only disadvantage is as you said, it is less explicit that it is a dict, but I think the name of the variable and the type hint should make that clear. Let me know if this is really a concern and I can change it.

tests/utils/test_geometry_comparisons.py
akshay-krishnan: `i2Ui1_measured` is in i2 frame, and since this frame is rotated by -90 wrt world frame, (0, 1, 0) is actually (1, 0, 0) in the world frame. The angle between (0, -1, 0) and (1, 0, 0) in world frame is 90 degrees.

gtsfm/utils/metrics.py
johnwlambert: thanks

tests/utils/test_geometry_comparisons.py
johnwlambert: resolved

gtsfm/multi_view_optimizer.py
```diff
@@ -80,7 +83,16 @@ def create_computation_graph(
...
+        if gt_poses_graph is None:
+            return ba_input_graph, ba_result_graph, None
+
+        metrics_graph = dask.delayed(metrics.compute_averaging_metrics)(
+            i2Ui1_graph, wRi_graph, wti_graph, gt_poses_graph
+        )
+        saved_metrics_graph = dask.delayed(io.save_json_file)(
+            "result_metrics/multiview_optimizer_metrics.json", metrics_graph
+        )
+        return ba_input_graph, ba_result_graph, saved_metrics_graph
```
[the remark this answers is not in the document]
akshay-krishnan: I dont understand, add what to the auxiliary list? I don't see an auxiliary list in this file. I referred to the two_view_estimator and it seems to be returning the pose_error_graphs.

tests/utils/test_geometry_comparisons.py
akshay-krishnan: I think roll, pitch, yaw makes it more confusing. I always end up wondering which one corresponds to rotation around X, Y and Z. But I do agree the RzRyRx API is odd. 
wRi1 is irrelevant, and it seems like too much effort for something irrelevant. I can make that an identity rotation and have the rx change for wRi2. 
I don't think I understand your suggestion on yaw, not roll. Let me make this change, and could we please speak in terms of x, y, z? :)

akshay-krishnan:
@johnwlambert do we know why the CI is failing at `master`?

gtsfm/multi_view_optimizer.py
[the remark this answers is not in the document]
ayushbaid: Oh my bad. You are right. This stuff will happen in `scene_optimizer.py`

ayushbaid:
> @johnwlambert do we know why the CI is failing at `master`?

There are two expected failures: one in `SceneOptimizer` and one in `1DSFM`. Both fail because our computed poses do not match the expected results.

johnwlambert (approved):

PR #101, opened 2021-02-19
Adding sample poses for unit tests

ayushbaid:
Adding some sample pose configurations, to be used for unit tests.

The configurations are:
1. 4 poses in a circle, each pose having 2 edges to its neighbor
2. 4 poses in a circle, each pose having 3 edges to its neighbor
3. 3 poses in forward motion with large translations
4. 3 poses in forward motion with small translations
5. 3 poses in panorama


tests/data/sample_poses.py
```diff
@@ -0,0 +1,90 @@
...
+"""4 poses in the circle of radius 5m, all looking at the center of the circle.
+
+For relative poses, each pose has just 2 edges, connecting to the immediate neighbors.
+"""
+CIRCLE_TWO_EDGES_GLOBAL_POSES = [
+    Pose3(Rot3.RzRyRx(0, 0, 0), np.array([0, 0, 0])),
+    Pose3(Rot3.RzRyRx(np.deg2rad(90), 0, 0), np.array([0, 5, 5])),
+    Pose3(Rot3.RzRyRx(np.deg2rad(180), 0, 0), np.array([0, 0, 10])),
+    Pose3(Rot3.RzRyRx(np.deg2rad(270), 0, 0), np.array([0, -5, 5])),
+]
```
johnwlambert, line 36: why are these poses in the yz plane? Maybe simpler in xy plane, and rotate about z? @ayushbaid

tests/data/sample_poses.py
```diff
@@ -0,0 +1,90 @@
...
+    Pose3(Rot3(), np.array([0, 0, 0])),
+    Pose3(Rot3(), np.array([0, 0, 1e-3])),
+    Pose3(Rot3(), np.array([0, 0, 5e-3])),
+]
+
+FORWARD_MOTION_SMALL_TRANSLATIONS_RELATIVE_POSES = generate_relative_from_global(
+    FORWARD_MOTION_SMALL_TRANSLATIONS_GLOBAL_POSES, [(0, 1), (0, 2), (1, 2)]
+)
+
+"""3 poses in a panorama (i.e. no translations but large relative rotations)
```
johnwlambert, line 80: i might just add in docstring that 3 different pitch values are specified at the same location

tests/data/sample_poses.py
ayushbaid: I am somehow unable to get it set properly. I was initially trying that.

tests/data/sample_poses.py
ayushbaid: for panorama? I think `no translations` mean the same thing?

johnwlambert: Yeah, i think just adding "3 different pitch values" would make even more obvious though

tests/data/sample_poses.py
johnwlambert: What was the issue you were seeing? And which frame were you aiming to define the poses in?

tests/data/sample_poses.py
```diff
@@ -0,0 +1,90 @@
...
+    wTi_list: List[Pose3], pose_pairs: List[Tuple[int, int]]
+) -> Dict[Tuple[int, int], Pose3]:
+    """Generate relative poses from global poses.
+
+    Args:
+        wTi_list: global poses.
+        pose_pairs: pairs (i1, i2) to construct relative poses for.
+
+    Returns:
+        Relative pose i2Ti1 for all requested pairs, as a dictionary.
```
akshay-krishnan, line 22: Please change to: Dict from (i1, i2) to i2Ti1. That makes it clear that the keys are in order (i1, i2).

tests/data/sample_poses.py
```diff
@@ -0,0 +1,90 @@
...
+
+CIRCLE_ALL_EDGES_RELATIVE_POSES = generate_relative_from_global(
+    CIRCLE_TWO_EDGES_GLOBAL_POSES, [(0, 1), (0, 2), (0, 3), (1, 2), (1, 3), (2, 3)]
+)
+
+"""3 poses simulating a forward motion with large translations, with no relative rotation.
+
+For relative poses, we have a fully connected graph.
+"""
+FORWARD_MOTION_LARGE_TRANSLATIONS_GLOBAL_POSES = [
```
akshay-krishnan, line 56: I think the concept of motion can be avoided here. If we are using CIRCLE_TWO_EDGES_RELATIVE_POSES, we could have LINE_LARGE_EDGES_RELATIVE_POSES instead of FORWARD_MOTION_LARGE_TRANSLATIONS_GLOBAL_POSES, to have some naming convention.

akshay-krishnan (changes requested):
Another suggestion: 
To change all zero translations to some non-zero (but constant) translations, and all zero rotations to some non-zero but constant rotations. That could be useful for testing the alignment as well.

tests/data/sample_poses.py
ayushbaid: I was trying to draw a circle on the x-y plane. And all the cameras point to the center of the circle.

tests/data/sample_poses.py
```diff
@@ -0,0 +1,90 @@
...
+"""
+import copy
+from typing import Dict, List, Tuple
+
+import numpy as np
+from gtsam import Pose3, Rot3
+
+
+def generate_relative_from_global(
+    wTi_list: List[Pose3], pose_pairs: List[Tuple[int, int]]
```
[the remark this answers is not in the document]
ayushbaid: Yup. Thanks

tests/data/sample_poses.py
ayushbaid: Thanks

ayushbaid:
> Another suggestion:
> To change all zero translations to some non-zero (but constant) translations, and all zero rotations to some non-zero but constant rotations. That could be useful for testing the alignment as well.

I did not get what you mean by testing the alignment. 

Also, I still think having examples with the exact same translations/rotations will be helpful.

tests/data/sample_poses.py
ayushbaid: Good point. I'll update.

akshay-krishnan:
> I did not get what you mean by testing the alignment.
I assume these are ground truth poses to be used for testing? We have been aligning the estimated poses and the ground truth poses before comparing them. If the ground truth poses have one camera at the origin, or if the constant rotations are all identity, we cannot be sure whether the alignment code is working correctly. Because the estimate will also be at origin translation and identity rotation without alignment. 

> Also, I still think having examples with the exact same translations/rotations will be helpful.
Totally agree, I just think the exact same translations/rotations should be non-zero.

akshay-krishnan:
Just curious, how are we planning on writing tests for these? I think it would be helpful to have a function that returns these as a list, or a dict (with the current names as keys?), so we can run the tests on the collection.

ayushbaid:
> Just curious, how are we planning on writing tests for these? I think it would be helpful to have a function that returns these as a list, or a dict (with the current names as keys?), so we can run the tests on the collection.

I was thinking of creating multiple functions, each with one of this pose.

ayushbaid:
> > I did not get what you mean by testing the alignment.
> 
> I assume these are ground truth poses to be used for testing? We have been aligning the estimated poses and the ground truth poses before comparing them. If the ground truth poses have one camera at the origin, or if the constant rotations are all identity, we cannot be sure whether the alignment code is working correctly. Because the estimate will also be at origin translation and identity rotation without alignment.
> 
> > Also, I still think having examples with the exact same translations/rotations will be helpful.
> 
> Totally agree, I just think the exact same translations/rotations should be non-zero.

Fixed it. Does it look better?

akshay-krishnan:
> I was thinking of creating multiple functions, each with one of this pose.

Yeah, but then we’ll also need a different unit test for each.

tests/data/sample_poses.py
```diff
@@ -0,0 +1,86 @@
...
+CIRCLE_ALL_EDGES_RELATIVE_POSES = generate_relative_from_global(
+    CIRCLE_TWO_EDGES_GLOBAL_POSES, [(0, 1), (0, 2), (0, 3), (1, 2), (1, 3), (2, 3)]
+)
+
+"""3 poses on a line, simulating forward motion, with large translations and no relative rotation.
+
+For relative poses, we have a fully connected graph.
+"""
+LINE_LARGE_EDGES_GLOBAL_POSES = [
+    Pose3(Rot3(), np.array([0, 0, 0])),
```
akshay-krishnan, line 57: My suggestion was to use a non-zero constant rotation here, so something like Rot3.RzRyRx(rx, ry, rz), where rx, ry and rz are constants for all three poses. Also, might be better to have the translations start at a different origin, something like camera1 = [1,1,1], camera2 = [2,2,2], camera3 = [3,3,3]. 
I understand we lose simplicity if we do this. 
But we can check whether we are able to align GT poses to estimated poses using our alignment methods. Because in the estimated pose, camera1 will be at [0, 0, 0]. 
But if you want to keep it simple and do not want to test alignment, I am fine with this.

tests/data/sample_poses.py
akshay-krishnan: PinholeCamera offers a LookAt method, https://github.com/borglab/gtsam/blob/f5ff7aa49f4d6387e3d7a823db140710f343cccf/gtsam/gtsam.i#L1032
But you will need a dummy calibration :( 
Its quite straightforward if its just the x-y plane though.

akshay-krishnan: I also cant understand the camera frame here. If z is outwards, I don’t think they look at the center, do they? It does not matter much, but please clarify in the comment.

ayushbaid:
> > I was thinking of creating multiple functions, each with one of this pose.
> 
> Yeah, but then we’ll also need a different unit test for each.

I think its better if there is a unit test per configuration as we can catch the error pretty easily because of that.

tests/data/sample_poses.py
ayushbaid: image
This is what I have. The cameras point towards the center.

johnwlambert: @ayushbaid do you mind adding a legend with per-axis labels? Maybe we should check that in actually

johnwlambert: (a label on the matplotlib plot axes and also what red, green , and blue correspond to)

tests/data/sample_poses.py
ayushbaid: I did change this to have non-zero defaults but forgot to push the code. Is it better now?

I have some non-zero values, but not all coordinates are at a non-zero value.

tests/data/sample_poses.py
ayushbaid: Sure. Adding them now.

tests/data/sample_poses.py
```diff
@@ -0,0 +1,140 @@
...
+"""
+LINE_SMALL_EDGES_GLOBAL_POSES = [
+    Pose3(DEFAULT_ROTATION, DEFAULT_TRANSLATION + np.array([0, 0, 0])),
+    Pose3(DEFAULT_ROTATION, DEFAULT_TRANSLATION + np.array([0, 0, 1e-3])),
+    Pose3(DEFAULT_ROTATION, DEFAULT_TRANSLATION + np.array([0, 0, 5e-3])),
+]
+
+LINE_SMALL_EDGES_RELATIVE_POSES = generate_relative_from_global(LINE_SMALL_EDGES_GLOBAL_POSES, [(0, 1), (0, 2), (1, 2)])
+
+"""3 poses in a panorama (i.e. 3 pitch values being the same but large relative rotations)
```
johnwlambert, line 79: by pitch i meant rotation about the y axis -- so the pitch values would be all different, but translations are the same

How about:
```
"""3 poses in a panorama (i.e. 3 translation values being the same but large relative rotations i.e. pitch varies)
```

tests/data/sample_poses.py
johnwlambert: thanks

ayushbaid: image
Plot with labels.

ayushbaid: @johnwlambert looks good?

johnwlambert: Thanks Ayush. That's much easier to understand now :-)

Can we place the objects on the world xy plane? (which is xz camera for each camera)? Relative measurements will have to be generated in the camera coordinate frame, to be consistent with what DMV produces, right?

ayushbaid: I directly use an example in GTSAM to get the required configuration.

image

johnwlambert: I see -- the plot you just shared looks good. But doesn't seem the same as the earlier one you posted?

ayushbaid: Yeah its the cameras point slightly downwards. But it provides us the same thing we wanted to test so I think it is good enough?

johnwlambert: Oh sure, I didnt realize you were saying we should replace the old circle one with the GTSAM circle example

akshay-krishnan: Nice pictures, can we check these into the same directory? For each configuration here?

tests/data/sample_poses.py
akshay-krishnan: Yeah, looks good now, although a higher value might be better for default translation. Maybe 2 instead of 0.2, as estimates themselves could be off by 0.2.

akshay-krishnan (approved):
I left 2 other optional comments, but this can be checked in if you prefer not making those changes. :)

tests/data/sample_poses.py
ayushbaid: Done

tests/data/sample_poses.py
ayushbaid: Cool I made them 2.

ayushbaid:
Thanks Akshay and John. I'll merge this and follow up with PR for adding tests based on these configs.

PR #102, opened 2021-02-23
Fixing error in reading pose by using strides

ayushbaid:
In the Argoverse data loader, we use strides between images to get some relative pose difference between consecutive images. 

We had a bug where we applied the stride on images but not on the pose data. This PR fixes the bug.


johnwlambert (approved):
nice, thanks Ayush

PR #104, opened 2021-02-24
Dump some data association metrics

johnwlambert:
- Make RANSAC more modular in data association 
- Save json file with some basic DA stats
- Make it so 3d point outliers dont get plotted


johnwlambert:
bev
3d
poses_bev
poses_3d

gtsfm/data_association/point3d_initializer.py
```diff
@@ -159,20 +175,22 @@ def triangulate(self, track_2d: SfmTrack2d) -> Optional[SfmTrack]:
...
-            return None
+            return None, None
 
         # compute reprojection errors for each measurement
         reproj_errors = self.compute_track_reprojection_errors(inlier_track.measurements, triangulated_pt)
 
         # all the measurements should have error < threshold
         if not np.all(reproj_errors < self.reproj_error_thresh):
-            return None
+            return None, None
```
akshay-krishnan, line 185: Since we have reprojection errors, cant we return the mean here?

gtsfm/multi_view_optimizer.py
```diff
@@ -77,9 +78,21 @@ def create_computation_graph(
 
         init_cameras_graph = dask.delayed(init_cameras)(wRi_graph, wti_graph, intrinsics_graph)
 
-        ba_input_graph = self.data_association_module.create_computation_graph(
+        data_assoc_graph = self.data_association_module.create_computation_graph(
             init_cameras_graph, v_corr_idxs_graph, keypoints_graph
         )
+        ba_input_graph = data_assoc_graph[0]
+        data_assoc_metrics_graph = data_assoc_graph[1]
```
akshay-krishnan, line 85:
```
ba_input_graph, data_assoc_metrics_graph = self.data_association_module.create_computation_graph(
            init_cameras_graph, v_corr_idxs_graph, keypoints_graph
        )
```
Isn't that better?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -248,3 +248,31 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
     if wti1 is None or wti2 is None:
         return None
     return np.linalg.norm(wti1 - wti2)
+
+
+def get_points_within_radius_of_cameras(wTi_list: List[Pose3], points_3d: np.ndarray, radius: float = 50) -> Optional[np.ndarray]:
+    """
+    Args:
```
akshay-krishnan, line 255: Doesnt the function need a one-line description before the args section?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -248,3 +248,31 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
...
+        wTi_list: camera poses
+        points_3d: array of shape (N,3) 
+        radius: distance threshold, in meters
+
+    Returns:
+        nearby_points_3d: array of shape (N,3)
+    """
+    if len(wTi_list) == 0 or points_3d.size == 0 or radius < 0:
```
akshay-krishnan, line 263: radius <= 0 I guess?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -248,3 +248,31 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
...
+        return None
+
+    num_points = points_3d.shape[0]
+    num_poses = len(wTi_list)
+    # each row represents attributes for a single point
+    # each column represents
+    is_nearby_matrix = np.zeros((num_points, num_poses), dtype=bool)
+    for j,wTi in enumerate(wTi_list):
+        is_nearby_matrix[:,j] = np.linalg.norm(points_3d - wTi.translation(), axis=1) < radius
```
akshay-krishnan, line 272: I'm torn about this. I feel like we should only accept wti_list (translations) since the rotations are unused. But at the same time, you'll need another loop to separate the translations. So, yeah maybe this is better, but just want to point to the alternative.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -248,3 +248,31 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
...
+
+def get_points_within_radius_of_cameras(wTi_list: List[Pose3], points_3d: np.ndarray, radius: float = 50) -> Optional[np.ndarray]:
+    """
+    Args:
+        wTi_list: camera poses
+        points_3d: array of shape (N,3) 
+        radius: distance threshold, in meters
+
+    Returns:
+        nearby_points_3d: array of shape (N,3)
```
akshay-krishnan, line 261: From the unit test, it does not seem like points_3d and nearby_points_3d have the same shape.

johnwlambert: good catch, thanks

gtsfm/utils/geometry_comparisons.py
johnwlambert: good point, i'll update that

gtsfm/utils/geometry_comparisons.py
johnwlambert: yeah, that's fair. I think a bit simpler to avoid the extra for-loop though

gtsfm/utils/geometry_comparisons.py
johnwlambert: good point, I'll add it

gtsfm/data_association/point3d_initializer.py
johnwlambert: That's a good point. I was a bit torn whether we should dump statistics of *accepted* tracks, or of any *generated* track, even if rejected.

@ayushbaid any thoughts here? maybe we should dump both for completeness now, since DA's failures are a bit of a mystery?

gtsfm/data_association/data_assoc.py
```diff
@@ -107,16 +109,22 @@ def run(
...
+        mean_track_length, median_track_length = SfmResult(triangulated_data, None).get_track_length_statistics()
+
+        logger.debug("[Data association] output number of tracks: %s", num_tracks)
+        logger.debug("[Data association] output avg. track length: %s", mean_track_length)
+
+        points_3d = [list(triangulated_data.track(j).point3()) for j in range(num_tracks)]
+        data_assoc_metrics = {
+            "num_tracks": num_tracks,
+            "mean_track_length": mean_track_length,
+            "median_track_length": median_track_length,
```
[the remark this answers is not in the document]
johnwlambert: I was just trying to dump as much as possible now, since space is cheap and performance is not great yet :-)

Do you mean median error within each track, or the median of all avg track errors @akshay-krishnan ?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,7 +58,87 @@ class Point3dInitializer(NamedTuple):
...
+
+        for sample_idxs in samples:
+            k1, k2 = measurement_pairs[sample_idxs]
+
+            i1, uv1 = track_2d.measurements[k1]
+            i2, uv2 = track_2d.measurements[k2]
+
+            camera_estimates = CameraSetCal3Bundler()
+            # check for unestimated cameras
+            if self.track_camera_dict.get(i1) != None and self.track_camera_dict.get(i2) != None:
```
akshay-krishnan, line 94: Its generally preferred to write these corner cases without having to indent most of the code. 
What I mean is: 
```
if self.track_camera_dict.get(i1) is None or self.track_camera_dict.get(i2) is None:
    logger.warn()
    continue

... do the normal work ... 
```

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,7 +58,87 @@ class Point3dInitializer(NamedTuple):
...
+                camera_estimates.append(self.track_camera_dict.get(i1))
+                camera_estimates.append(self.track_camera_dict.get(i2))
+
+                img_measurements = Point2Vector()
```
akshay-krishnan, line 98: Cant we just do `img_measurements = [uv1, uv2]`

gtsfm/data_association/point3d_initializer.py
akshay-krishnan: You could also then move this to before `camera_estimates = CameraSetCal3Bundler()`

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,7 +58,87 @@ class Point3dInitializer(NamedTuple):
...
+            k1, k2 = measurement_pairs[sample_idxs]
+
+            i1, uv1 = track_2d.measurements[k1]
+            i2, uv2 = track_2d.measurements[k2]
+
+            camera_estimates = CameraSetCal3Bundler()
+            # check for unestimated cameras
+            if self.track_camera_dict.get(i1) != None and self.track_camera_dict.get(i2) != None:
+                camera_estimates.append(self.track_camera_dict.get(i1))
+                camera_estimates.append(self.track_camera_dict.get(i2))
```
akshay-krishnan, line 96: Cant we do `camera_estimates = [self.track_camera_dict.get(i1), self.track_camera_dict.get(i2)]`?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,7 +58,87 @@ class Point3dInitializer(NamedTuple):
...
+                        optimize=True,
+                    )
+                except RuntimeError:
+                    # TODO: handle cheirality exception properly?
+                    logger.info(
+                        "Cheirality exception from GTSAM's triangulatePoint3() likely due to outlier, skipping track"
+                    )
+                    continue
+
+                errors = self.compute_track_reprojection_errors(track_2d.measurements, triangulated_pt)
```
akshay-krishnan, line 117: I think it would be more modular and cleaner if you made the above changes, and moved this code to a different function `compute_reprojection_error_for_hypothesis(track_2d.measurements, k1, k2)`

gtsfm/multi_view_optimizer.py
```diff
@@ -77,9 +78,21 @@ def create_computation_graph(
...
+
+        auxiliary_graph_list = []
+        auxiliary_graph_list.append(
```
akshay-krishnan, line 88: nit: cant we create `auxiliary_graph_list` like this?
``` 
auxiliary_graph_list = [
 dask.delayed(io.save_json_file)(
                os.path.join("result_metrics", "data_association_metrics.json"), data_assoc_metrics_graph
            )
]
```

akshay-krishnan (changes requested):

gtsfm/data_association/point3d_initializer.py
ayushbaid: This looks better.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,7 +58,87 @@ class Point3dInitializer(NamedTuple):
...
+                try:
+                    triangulated_pt = gtsam.triangulatePoint3(
+                        camera_estimates,
+                        img_measurements,
+                        rank_tol=SVD_DLT_RANK_TOL,
+                        optimize=True,
+                    )
+                except RuntimeError:
+                    # TODO: handle cheirality exception properly?
+                    logger.info(
```
ayushbaid, line 112: do we have a way to log the counts of this exception?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,7 +58,87 @@ class Point3dInitializer(NamedTuple):
     reproj_error_thresh: float
     num_ransac_hypotheses: Optional[int] = None
 
-    def triangulate(self, track_2d: SfmTrack2d) -> Optional[SfmTrack]:
+
+    def execute_ransac_variant(self, track_2d: SfmTrack2d) -> np.ndarray:
```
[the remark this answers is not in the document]
xwu4lab: We randomly sample num_ransac_hypothesis pairs of measurements in one track, each pair is used to initialize the landmark and then use the votes of all measurements to score each hypothesis. The one with the best score survives and all the inliers are again feed into triangulation algorithm to recover the 3D structure of this landmark.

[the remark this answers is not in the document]
johnwlambert: There are 3 different sampling schemes that are supported here. RANSAC chooses one of them.

gtsfm/data_association/point3d_initializer.py
johnwlambert: this is a nice improvement, thanks. updated

gtsfm/utils/geometry_comparisons.py
```diff
@@ -248,3 +248,31 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
     if wti1 is None or wti2 is None:
         return None
     return np.linalg.norm(wti1 - wti2)
+
+
+def get_points_within_radius_of_cameras(wTi_list: List[Pose3], points_3d: np.ndarray, radius: float = 50) -> Optional[np.ndarray]:
+    """Return those 3d points that fall within a specified radius from any camera.
+    Args:
```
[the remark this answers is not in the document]
johnwlambert: updated, thanks

gtsfm/data_association/data_assoc.py
```diff
@@ -86,17 +86,19 @@ def run(
...
             # triangulate and filter based on reprojection error
-            sfm_track = point3d_initializer.triangulate(track_2d)
+            sfm_track, avg_track_reproj_error = point3d_initializer.triangulate(track_2d)
 
             if sfm_track is None:
                 continue
 
             if self.__validate_track(sfm_track):
                 triangulated_data.add_track(sfm_track)
+                per_track_avg_errors += [avg_track_reproj_error]
```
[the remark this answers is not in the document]
johnwlambert: done, thanks

gtsfm/data_association/point3d_initializer.py
johnwlambert: updated

gtsfm/data_association/point3d_initializer.py
johnwlambert: Cool, I added these extra metrics now

gtsfm/data_association/data_assoc.py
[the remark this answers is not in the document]
johnwlambert: We would need to change the SFMResult class to get the min and max stats, so i'd prefer to leave that for another PR

gtsfm/multi_view_optimizer.py
johnwlambert: Dask won't allow this unfortunately:
```python
  File "gtsfm/runner/run_scene_optimizer_argoverse.py", line 97, in <module>
    run_scene_optimizer(args)
  File "gtsfm/runner/run_scene_optimizer_argoverse.py", line 37, in run_scene_optimizer
    sfm_result_graph = scene_optimizer.create_computation_graph(
  File "/Users/johnlambert/Documents/gtsfm/gtsfm/scene_optimizer.py", line 184, in create_computation_graph
    (ba_input_graph, ba_output_graph, optimizer_metrics_graph, ) = self.multiview_optimizer.create_computation_graph(
  File "/Users/johnlambert/Documents/gtsfm/gtsfm/multi_view_optimizer.py", line 81, in create_computation_graph
    ba_input_graph, data_assoc_metrics_graph = self.data_association_module.create_computation_graph(
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v1/lib/python3.8/site-packages/dask/delayed.py", line 562, in __iter__
    raise TypeError("Delayed objects of unspecified length are not iterable")
TypeError: Delayed objects of unspecified length are not iterable
```

gtsfm/multi_view_optimizer.py
johnwlambert: nice, fixed that

johnwlambert:
Dumps something like:
```python
{
    "mean_2d_track_length": 2.8,
    "accepted_tracks_ratio": 0.977,
    "track_cheirality_failure_ratio": 0.007,
    "num_accepted_tracks": 2498,
    "mean_3d_track_length": 2.404,
    "median_3d_track_length": 2.0,
    "per_rejected_track_avg_errors": [
        null,
        null,
        null,
...
}
```

gtsfm/data_association/point3d_initializer.py
johnwlambert: @akshay-krishnan can you point me to which line you were referring to?

gtsfm/data_association/point3d_initializer.py
johnwlambert: GTSAM requires the `Point2Vector` type as argument

gtsfm/data_association/data_assoc.py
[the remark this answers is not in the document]
johnwlambert: On second thought, i added that.

But "mean track length" is a metric I've seen in published papers like COLMAP so I'd like to keep it for now

gtsfm/data_association/data_assoc.py
```diff
@@ -107,16 +114,37 @@ def run(
...
+        data_assoc_metrics = {
+            "mean_2d_track_length": np.round(mean_2d_track_length,3),
+            "accepted_tracks_ratio": np.round(accepted_tracks_ratio,3),
+            "track_cheirality_failure_ratio": np.round(track_cheirality_failure_ratio,3),
+            "num_accepted_tracks": num_accepted_tracks,
+            "median_3d_track_length": median_3d_track_length,
+            "mean_3d_track_length": mean_3d_track_length,
+            "min_3d_track_length": int(track_lengths_3d.min()),
+            "max_3d_track_length": int(track_lengths_3d.max()),
+            "num_len_2_tracks": int(np.sum(track_lengths_3d == 2)),
```
ayushbaid, line 133: `np.histogram` might make the code cleaner?

gtsfm/data_association/data_assoc.py
```diff
@@ -107,16 +114,37 @@ def run(
...
+            "num_len_4_tracks": int(np.sum(track_lengths_3d == 4)),
+            "num_len_5_tracks": int(np.sum(track_lengths_3d == 5)),
+            "num_len_6_tracks": int(np.sum(track_lengths_3d == 6)),
+            "num_len_7_tracks": int(np.sum(track_lengths_3d == 7)),
+            "num_len_8_tracks": int(np.sum(track_lengths_3d == 8)),
+            "num_len_9_tracks": int(np.sum(track_lengths_3d == 9)),
+            "num_len_10_tracks": int(np.sum(track_lengths_3d == 10)),
+            "per_rejected_track_avg_errors": per_rejected_track_avg_errors,
+            "per_accepted_track_avg_errors": per_accepted_track_avg_errors,
+            "points_3d": points_3d,
```
ayushbaid, line 144: why dump just 3d points? Why not dump the track?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,98 +58,123 @@ class Point3dInitializer(NamedTuple):
...
 
         Args:
             track: feature track from which measurements are to be extracted
 
         Returns:
             track with inlier measurements and 3D landmark. None returned if triangulation fails or has high error.
+            avg_track_reproj_error: reprojection error of 3d triangulated point to each image plane
+                Note: this may be "None" if the 3d point could not be triangulated successfully
+                due to a cheirality exception or insufficient number of RANSAC inlier measurements
+            is_cheirality_failure: boolean representing whether the selected 2d measurements lead
```
ayushbaid, line 156: nit: this is technically not needed, but I guess it helps avoid convoluted logic to figure out chierality exception?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,98 +58,123 @@ class Point3dInitializer(NamedTuple):
...
-                else:
-                    logger.warning("Unestimated cameras found at indices {} or {}. Skipping them.".format(i1, i2))
+            best_inliers = self.execute_ransac_variant(track_2d)
 
         elif self.mode == TriangulationParam.NO_RANSAC:
             best_inliers = np.ones(len(track_2d.measurements), dtype=bool)  # all marked as inliers
 
         inlier_idxs = (np.where(best_inliers)[0]).tolist()
 
+        is_cheirality_failure = False
```
ayushbaid, line 171: flag doesn't get read anytime. The only time it is set True, we return immediately. I think it can be skipped.

gtsfm/multi_view_optimizer.py
ayushbaid: This logic should be moved to `DataAssociation`'s `create_computation_graph`.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -248,3 +248,32 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
     if wti1 is None or wti2 is None:
         return None
     return np.linalg.norm(wti1 - wti2)
+
+
+def get_points_within_radius_of_cameras(wTi_list: List[Pose3], points_3d: np.ndarray, radius: float = 50) -> Optional[np.ndarray]:
```
ayushbaid, line 253: I guess this will be dependent on each scene? Or is 50 a high enough value that most datasets will fit within it?

gtsfm/data_association/data_assoc.py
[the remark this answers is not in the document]
akshay-krishnan: Then we probably won't need median. We'll ultimately end up using one number for the central tendency (bias) and one for the range/variance. But its upto you.

tests/utils/test_geometry_comparisons.py
```diff
@@ -144,5 +144,60 @@ def test_compute_points_distance_l2_is_nonzero(self):
...
+def test_get_points_within_radius_of_cameras_no_poses():
+    """Catch degenerate input."""
+    wTi_list = []
+    points_3d = np.array([[-15, 0, 0], [0, 15, 0], [-5, 0, 0], [15, 0, 0], [25, 0, 0]])
+    radius = 10.0
+
+    nearby_points_3d = geometry_comparisons.get_points_within_radius_of_cameras(wTi_list, points_3d, radius)
+    assert nearby_points_3d is None, "At least one camera pose must be provided"
+
+
```
ayushbaid, line 201: I think we should have these tests inside the class as we have done it in other tests.

gtsfm/data_association/point3d_initializer.py
akshay-krishnan: I was referring to all the lines I selected. Basically, the code for computing the errors for each sample_idx could be moved to a different function as this function is getting a little too big. But it's fine if you'd like to keep it this way.

gtsfm/data_association/data_assoc.py
johnwlambert: I was visualizing them in Mayavi on my local machine for debugging.

I recommend in a separate PR we dump the output in COLMAP-format. Then we could remove my `points_3d` dump later.

akshay-krishnan, line 144: I think there can be some more structure here to make this a bit more readable. If using multiple statistics for the same metric, could you use a nested dict? 
I mean
"3d_tracks_length": {
    "min": .. 
    "max": ..
    "median":...
    "mean": ...
}

gtsfm/data_association/point3d_initializer.py
johnwlambert: yeah, i thought returning `None, None, None` was not too readable though. Was just striving for readability

gtsfm/data_association/data_assoc.py
```diff
@@ -107,16 +114,37 @@ def run(
...
+            "max_3d_track_length": int(track_lengths_3d.max()),
+            "num_len_2_tracks": int(np.sum(track_lengths_3d == 2)),
+            "num_len_3_tracks": int(np.sum(track_lengths_3d == 3)),
+            "num_len_4_tracks": int(np.sum(track_lengths_3d == 4)),
+            "num_len_5_tracks": int(np.sum(track_lengths_3d == 5)),
+            "num_len_6_tracks": int(np.sum(track_lengths_3d == 6)),
+            "num_len_7_tracks": int(np.sum(track_lengths_3d == 7)),
+            "num_len_8_tracks": int(np.sum(track_lengths_3d == 8)),
+            "num_len_9_tracks": int(np.sum(track_lengths_3d == 9)),
+            "num_len_10_tracks": int(np.sum(track_lengths_3d == 10)),
```
akshay-krishnan, line 141: Theres probably a better way to do this. We could consider nesting. Compute the histogram in a loop. 
Also, you may consider using [np.unique](https://numpy.org/doc/stable/reference/generated/numpy.unique.html) or [np.histogram](https://numpy.org/doc/stable/reference/generated/numpy.histogram.html).

akshay-krishnan (approved):
Looks good John, adding some optional comments for improved readability.

gtsfm/utils/geometry_comparisons.py
johnwlambert: Technically I think we should be using `react-three-fiber` soon. So I think this will be a moot point then, and can be removed.

But a 50 m radius improved the plots a lot for now, and seems to help with debugging.

gtsfm/data_association/point3d_initializer.py
johnwlambert: Yeah, we have two types of cheirality exceptions to log: (1) those in RANSAC when selecting inliers, and (2) those when doing final triangulation.

I'm only logging those from part (2) here. To aggregate stats over all tracks, we have to do it in `DataAssociation`, instead of in the `Point3dInitializer`.

I'm not sure if those from (1) should be logged, or where it should be logged.

tests/utils/test_geometry_comparisons.py
johnwlambert: I'm fine either way. Technically it's not a class, so i was going more for the functional, free-function approach

gtsfm/data_association/data_assoc.py
johnwlambert: this is a nice suggestion -- thanks. Updated now

gtsfm/data_association/point3d_initializer.py
```diff
@@ -58,98 +58,123 @@ class Point3dInitializer(NamedTuple):
...
+            i1, uv1 = track_2d.measurements[k1]
+            i2, uv2 = track_2d.measurements[k2]
+
+            # check for unestimated cameras
+            if self.track_camera_dict.get(i1) is None or self.track_camera_dict.get(i2) is None:
+                logger.warning("Unestimated cameras found at indices {} or {}. Skipping them.".format(i1, i2))
+                continue
+
+            camera_estimates = CameraSetCal3Bundler()
+
```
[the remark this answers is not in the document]
johnwlambert: thanks, updated

gtsfm/data_association/data_assoc.py
johnwlambert: thanks, updated

gtsfm/data_association/data_assoc.py
johnwlambert: i like that, thanks. Updated

gtsfm/multi_view_optimizer.py
johnwlambert: I'm good with moving it, but is there a particular reason to put it in one place rather than the other?

gtsfm/data_association/data_assoc.py
```diff
@@ -114,36 +114,45 @@ def run(
...
-            "num_len_9_tracks": int(np.sum(track_lengths_3d == 9)),
-            "num_len_10_tracks": int(np.sum(track_lengths_3d == 10)),
-            "per_rejected_track_avg_errors": per_rejected_track_avg_errors,
-            "per_accepted_track_avg_errors": per_accepted_track_avg_errors,
-            "points_3d": points_3d,
+            "3d_tracks_length": {
+                "median_3d_track_length": median_3d_track_length,
+                "mean_3d_track_length": mean_3d_track_length,
+                "min_3d_track_length": int(track_lengths_3d.min()),
+                "max_3d_track_length": int(track_lengths_3d.max()),
```
akshay-krishnan, line 137: nit: "the 3d_track_length" in the nested keys is kind of redundant. But feel free to leave it this way if you prefer so.

gtsfm/data_association/data_assoc.py
```diff
@@ -114,36 +114,45 @@ def run(
...
+        data_assoc_metrics["3d_tracks_length"].update(
+            {f"num_len_{i}_tracks": int(histogram_track_lengths[i]) for i in range(2, 11)}
+        )
+
+        # placing long lists at the end of dictionary, since they are less easily interpreted
+        data_assoc_metrics.update(
+            {
+                "mean_accepted_track_avg_error": np.array(per_accepted_track_avg_errors).mean(),
+                "per_rejected_track_avg_errors": per_rejected_track_avg_errors,
+                "per_accepted_track_avg_errors": per_accepted_track_avg_errors,
```
akshay-krishnan, line 151: Just curious, why per_ here?

akshay-krishnan (approved):

gtsfm/data_association/data_assoc.py
johnwlambert: I don't have a good way of differentiating between names for 
(1) average error within a track
(2) average of per-track avg errors

but i'm open to suggestions

gtsfm/data_association/data_assoc.py
johnwlambert: good point, i just cleaned it up

gtsfm/data_association/data_assoc.py
```diff
@@ -114,36 +114,45 @@ def run(
...
+            triangulated_data, None
+        ).get_track_length_statistics()
 
         logger.debug("[Data association] output number of tracks: %s", num_accepted_tracks)
         logger.debug("[Data association] output avg. track length: %s", mean_3d_track_length)
 
         # dump the 3d point cloud before Bundle Adjustment for offline visualization
         points_3d = [list(triangulated_data.track(j).point3()) for j in range(num_accepted_tracks)]
+        # bin edges are halfway between each integer
+        histogram_track_lengths, _ = np.histogram(track_lengths_3d, bins=np.linspace(-0.5, 10.5, 12))
```
ayushbaid, line 127: `track_lengths_histogram`?

ayushbaid: Also, why bin it from -0.5. We can use the ` np.histogram(track_lengths_3d, bins=8, range=(2, 10))`

gtsfm/data_association/data_assoc.py
```diff
@@ -114,36 +114,45 @@ def run(
...
+            "3d_tracks_length": {
+                "median": median_3d_track_length,
+                "mean": mean_3d_track_length,
+                "min": int(track_lengths_3d.min()),
+                "max": int(track_lengths_3d.max()),
+            },
         }
 
+        # min possible track len is 2, above 10 is improbable
+        data_assoc_metrics["3d_tracks_length"].update(
```
ayushbaid, line 142: I'd prefer if you can create a histogram dictionary before the `data_assoc_metrics` and just add that sub-dictionary in "3d_tracks_length". Its very easy to follow visually.

ayushbaid: I prefer this to updating the dictionary.

gtsfm/multi_view_optimizer.py
ayushbaid: I think its more clear seeing the semantics on what to expect. If a module has three `Delayed` being returned, we can create richer docstrings.

ayushbaid (commented):
Left some small suggestions. Looks good otherwise.

ayushbaid (approved):

PR #108, opened 2021-03-10
Create CONTRIBUTING.md

johnwlambert:
Establish some contributing guidelines


ayushbaid:
I'll linting details and mytype too here.

akshay-krishnan:
Looks great, we should also have a conventions section, that explains all the symbol conventions (wTc, wTc_list, w_wUc, etc.). We should also request contributors to prefer gtsam types wherever possible unless it's not already wrapped and is a lot of work to do so, or there are good advantages to using other types (like np arrays).

PR #111, opened 2021-03-11
Removing custom serialization as GTSAM types are pickleable

ayushbaid:
We were relying on custom serialization for GTSAM types. However, that was not very reliable on clusters as dask was falling back on pickle and not using custom serialization.

As GTSAM now supports pickle (https://github.com/borglab/gtsam/pull/709), we will remove all custom serialization.

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -8,8 +8,9 @@ echo "Running .github/scripts/python.sh..."
 conda init
 conda info --envs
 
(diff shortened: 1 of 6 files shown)


johnwlambert:
Should we also delete this file `import gtsfm.utils.serialization as serialization_utils` @ayushbaid ?

johnwlambert (approved):

ayushbaid:
> Should we also delete this file `import gtsfm.utils.serialization as serialization_utils` @ayushbaid ?

Yup we should. I'll directly delete in master now.

PR #113, opened 2021-03-16
Adding expressive GTSFM data to allow non-consecutive cameras

ayushbaid:
The current SfmData structure stores the camera as a list, and hence needs all the image to have a valid camera. This is not the case when some upstream module drops cameras. 

The new data structure uses a dictionary to store the cameras, and hence can store non-contiguous cameras.


ayushbaid:
Some changes in the PR are just fixing the unit tests failing due to #104.

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,192 @@
...
+import numpy as np
+from gtsam import PinholeCameraCal3Bundler, SfmTrack
+
+import gtsfm.utils.graph as graph_utils
+
+EQUALITY_TOLERANCE = 1e-5
+
+
+class GtsfmData:
+    """Class containing cameras and tracks, essentially describing the complete 3D scene."""
```
johnwlambert, line 17: Thanks for the PR, @ayushbaid . Looks nice. It might be helpful here to explain why we need this class. I.e. why SfmData is not sufficient.

Just if we forget in 6 months or for other people who might want to change it in the future

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        camera_edges = []
+        for sfm_track in self._tracks:
+            cameras_in_use = []
+            for m_idx in range(sfm_track.number_measurements()):
+                i, _ = sfm_track.measurement(m_idx)
+                cameras_in_use.append(i)
+
+            # add all combinations of edges
+            camera_edges += list(itertools.combinations(cameras_in_use, 2))
```
johnwlambert, line 183: why do we need to create this combinations?  why cant we just use the pairs inside each track?

gtsfm/utils/io.py
```diff
@@ -85,3 +84,76 @@ def save_json_file(
...
+    with open(file_path, "w") as f:
+        f.write("# Number of cameras: {}\n".format(gtsfm_data.number_images()))
+
+        for i in gtsfm_data.get_valid_camera_indices():
+            camera = gtsfm_data.get_camera(i)
+            wRi_quaternion = camera.pose().rotation().quaternion()  # pylint: disable=unused-variable
+            wti = camera.pose().translation()  # pylint: disable=unused-variable
+
+            f.write("{i}")
+            f.write("{wRi_quaternion[0]} {wRi_quaternion[1]} {wRi_quaternion[2]} {wRi_quaternion[3]}")
```
johnwlambert, line 158: Thanks for adding this. I prefer to do 
```python
qx,qy,qz,qw = wRi_quaternion
f.write("{qx} {qy} {qz} {qw}")
```

gtsfm/utils/io.py
```diff
@@ -85,3 +84,76 @@ def save_json_file(
...
+
+        for i in gtsfm_data.get_valid_camera_indices():
+            camera = gtsfm_data.get_camera(i)
+            calibration = camera.calibration()
+
+            fx = calibration.fx()  # pylint: disable=unused-variable
+            u0 = calibration.px()  # pylint: disable=unused-variable
+            v0 = calibration.py()  # pylint: disable=unused-variable
+
+            f.write("{i} SIMPLE_PINHOLE {image_width} {image_height} {fx} {u0} {v0}\n")
```
johnwlambert, line 136: is `SIMPLE_PINHOLE` a COLMAP type? Wouldnt our camera be considered RADIAL?

gtsfm/utils/io.py
```diff
@@ -85,3 +84,76 @@ def save_json_file(
...
+
+    with open(file_path, "w") as f:
+        f.write("# Number of cameras: {}\n".format(gtsfm_data.number_images()))
+
+        for i in gtsfm_data.get_valid_camera_indices():
+            camera = gtsfm_data.get_camera(i)
+            wRi_quaternion = camera.pose().rotation().quaternion()  # pylint: disable=unused-variable
+            wti = camera.pose().translation()  # pylint: disable=unused-variable
+
+            f.write("{i}")
```
johnwlambert, line 157: don't we need a space after `i` before `qw`?

gtsfm/common/gtsfm_data.py
ayushbaid: Makes sense. Added it.

gtsfm/common/gtsfm_data.py
ayushbaid: There is a list of camera for a track.

For example: a track has cameras [0, 2, 5]. In that case we will add pairs (0, 2), (0, 5), (2, 5).

gtsfm/utils/io.py
ayushbaid: Yes thats correct. Thanks!

gtsfm/utils/io.py
ayushbaid: Thanks. Corrected!

gtsfm/utils/io.py
ayushbaid: done

gtsfm/utils/io.py
```diff
@@ -85,3 +84,80 @@ def save_json_file(
...
+    for i in range(num_images):
+        camera = sfm_data.camera(i)
+        gtsfm_data.add_camera(camera, i)
+    for j in range(sfm_data.number_tracks()):
+        gtsfm_data.add_track(sfm_data.track(j))
+
+    return gtsfm_data
+
+
+def write_cameras(gtsfm_data: GtsfmData, file_path: str) -> None:
```
johnwlambert, line 110: Thanks for the updates, @ayushbaid.

I think it would make more sense to pass the `save_dir` here, since by design the filename must be `cameras.txt` to match COLMAP

johnwlambert: (and not pass a file_path)

gtsfm/utils/io.py
```diff
@@ -85,3 +84,80 @@ def save_json_file(
...
+            fx = calibration.fx()  # pylint: disable=unused-variable
+            u0 = calibration.px()  # pylint: disable=unused-variable
+            v0 = calibration.py()  # pylint: disable=unused-variable
+            k1 = calibration.k1()  # pylint: disable=unused-variable
+            k2 = calibration.k2()  # pylint: disable=unused-variable
+
+            f.write("{i} SIMPLE_RADIAL {image_width} {image_height} {fx} {u0} {v0} {k1} {k2}\n")
+
+
+def write_images(gtsfm_data: GtsfmData, file_path: str) -> None:
```
johnwlambert, line 141: same note here as above. Let's pass in `save_dir` since the file must end in `images.txt`, and otherwise we obscure this

gtsfm/utils/io.py
```diff
@@ -85,3 +84,80 @@ def save_json_file(
...
+
+        for i in gtsfm_data.get_valid_camera_indices():
+            camera = gtsfm_data.get_camera(i)
+            wRi_quaternion = camera.pose().rotation().quaternion()
+            wti = camera.pose().translation()  # pylint: disable=unused-variable
+
+            qw, qx, qy, qz = wRi_quaternion  # pylint: disable=unused-variable
+            f.write("{i} ")
+            f.write("{qw} {qx} {qy} {qz} ")
+            f.write("{wti[0]} {wti[3]} {wti[2]}")
```
johnwlambert, line 162: is this cleaner?
```python
tx, ty, tz = wti
f.write("{tx} {ty} {tz}")
```

gtsfm/utils/io.py
```diff
@@ -85,3 +84,80 @@ def save_json_file(
...
+def write_images(gtsfm_data: GtsfmData, file_path: str) -> None:
+    """Writes the image data file in the COLMAP format.
+
+    Args:
+        gtsfm_data: scene data to write.
+        file_path: name of the file.
+    """
+
+    # TODO: get image shape somehow
+
```
johnwlambert, line 150: i believe to match COLMAP, we need also to dump 3 lines:
```python
+    #     f.write("# Image list with two lines of data per image:\n")
+    #     f.write("#   IMAGE_ID, QW, QX, QY, QZ, TX, TY, TZ, CAMERA_ID, NAME\n")
+    #     f.write("#   POINTS2D[] as (X, Y, POINT3D_ID)\n")
+    #     f.write(f"# Number of images: {num_imgs}, mean observations per image: {mean_obs_per_img}")
```

gtsfm/scene_optimizer.py
```diff
@@ -213,12 +211,20 @@ def create_computation_graph(
         if self._save_bal_files:
             # save the input to Bundle Adjustment (from data association)
             auxiliary_graph_list.append(
-                dask.delayed(write_sfmdata_to_disk)(ba_input_graph, os.path.join(RESULTS_PATH, "ba_input.bal"))
+                dask.delayed(io_utils.write_cameras)(ba_input_graph, os.path.join(RESULTS_PATH, "ba_input_cameras.txt"))
```
johnwlambert, line 214: i see why you named these differently, because we have multiple versions, and we want to differentiate between them.

But can we just put them in separate subdirectories? Per https://colmap.github.io/cli.html, we need to preserve the names to be compatible:

COLMAP as far as I know 
Screen Shot 2021-03-22 at 10 42 53 PM

gtsfm/scene_optimizer.py
```diff
@@ -213,12 +211,20 @@ def create_computation_graph(
...
+            )
+            auxiliary_graph_list.append(
+                dask.delayed(io_utils.write_images)(ba_input_graph, os.path.join(RESULTS_PATH, "ba_input_images.txt"))
```
johnwlambert, line 217: Can we save `"points3D.txt"` as well here?

gtsfm/utils/io.py
ayushbaid: Can I make a different PR to match the format of COLMAP? I feel it will make it easier to land?

johnwlambert: Sure.

gtsfm/utils/io.py
```diff
@@ -85,3 +84,81 @@ def save_json_file(
...
+            f.write("{i} SIMPLE_RADIAL {image_width} {image_height} {fx} {u0} {v0} {k1} {k2}\n")
+
+
+def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
+    """Writes the image data file in the COLMAP format.
+
+    Args:
+        gtsfm_data: scene data to write.
+        save_dir: folder to put the cameras.txt file in.
```
johnwlambert, line 147: should this say "images.txt"?

gtsfm/scene_optimizer.py
```diff
@@ -213,13 +211,17 @@ def create_computation_graph(
         if self._save_bal_files:
             # save the input to Bundle Adjustment (from data association)
             auxiliary_graph_list.append(
-                dask.delayed(write_sfmdata_to_disk)(ba_input_graph, os.path.join(RESULTS_PATH, "ba_input.bal"))
+                dask.delayed(io_utils.write_cameras)(ba_input_graph, os.path.join(RESULTS_PATH, "ba_input"))
```
johnwlambert, line 214: keyword arg might help readability, like 
```python
dask.delayed(io_utils.write_cameras)(ba_input_graph, save_dir=os.path.join(RESULTS_PATH, "ba_input"))
```

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -102,7 +101,7 @@ def run(self, initial_data: SfmData) -> SfmResult:
             result_values = lm.optimize()
         except Exception as e:
             logger.exception("LM Optimization failed")
-            return SfmResult(SfmData(), float("Nan"))
+            return SfmResult(GtsfmData(initial_data.number_images()), float("Nan"))
```
johnwlambert, line 104: keyword arg would help to understand what float("Nan") is being used for

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -52,47 +54,44 @@ def run(self, initial_data: SfmData) -> SfmResult:
...
+            initial.insert(C(cam_idx), camera)
 
-        j = 0
         # add each SfmTrack
         for t_idx in range(initial_data.number_tracks()):
-            track = initial_data.track(t_idx)
-            initial.insert(P(j), track.point3())
-            j += 1
+            track = initial_data.get_track(t_idx)
+            initial.insert(P(t_idx), track.point3())
```
johnwlambert, line 94: i prefer we use the variable `j` here, not `t_idx`, since it is an index of a 3d point

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -52,47 +54,44 @@ def run(self, initial_data: SfmData) -> SfmResult:
...
         # Create initial estimate
         initial = gtsam.Values()
 
-        i = 0
         # add each PinholeCameraCal3Bundler
-        for cam_idx in range(initial_data.number_cameras()):
-            camera = initial_data.camera(cam_idx)
-            initial.insert(C(i), camera)
-            i += 1
+        for cam_idx in valid_camera_indices:
```
johnwlambert, line 87: i prefer we use the variable `i` here, not `cam_idx`, since `i` is a camera index.
But an inline comment in the line above saying `# i is a camera index` would be nice

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+        for j in range(track.number_measurements()):
+            i, _ = track.measurement(j)
+
+            if i not in self._cameras:
+                return False
+
+        self._tracks.append(track)
+        return True
+
+    def add_camera(self, camera: PinholeCameraCal3Bundler, camera_index: int) -> None:
```
johnwlambert, line 130: this is a very minor nitpick, but i think having args in order (self, camera_index, camera) might make mores sense, since it symbolizes to me what this does:
`dict[cam_idx] = cam`

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+
+        if self._number_images != other.number_images():
+            return False
+
+        for i, cam in self._cameras.items():
+            other_cam = other.get_camera(i)
+            if not cam.equals(other_cam, EQUALITY_TOLERANCE):
+                return False
+
+        for k in range(self.number_tracks()):
```
johnwlambert, line 47: `j` should be each track, not `k`

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+            track = self.get_track(k)
+            other_track = other.get_track(k)
+
+            if track.number_measurements() != other_track.number_measurements():
+                return False
+
+            for j in range(track.number_measurements()):
```
johnwlambert, line 54: `k` should be here, not `j` -- measurement index inside a track

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+        """Add a track, after checking if all the cameras in the track are already added.
+
+        Args:
+            track: track to add.
+
+        Returns:
+            Flag indicating the success of adding operation.
+        """
+        # check if all cameras are already added
+        for j in range(track.number_measurements()):
```
johnwlambert, line 121: should be `k` here, not `j`

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+            raise ValueError("Camera cannot be None, should be a valid camera")
+        self._cameras[camera_index] = camera
+
+    def drop_cameras(self, camera_indices: List[int]) -> "GtsfmData":
+        """Drops cameras and tracks associated with them.
+
+        Returns:
+            New object with the requested cameras (and associated tracks) dropped.
+        """
+
```
johnwlambert, line 150: prefer no newline here for compactness

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+            camera_index: the index associated with this camera.
+
+        Raises:
+            ValueError: if the camera to be added is not a valid camera object.
+        """
+        if camera is None:
+            raise ValueError("Camera cannot be None, should be a valid camera")
+        self._cameras[camera_index] = camera
+
+    def drop_cameras(self, camera_indices: List[int]) -> "GtsfmData":
```
johnwlambert, line 144: is there a better name for this? It looks like a factory function. But its named like a method. So I'm not sure. Let's discuss before we change it, though. @akshay-krishnan any thoughts?

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+        for i, camera in self._cameras.items():
+            if i not in camera_indices:
+                new_data.add_camera(camera, i)
+
+        new_camera_indices = new_data.get_valid_camera_indices()
+
+        # add tracks which have all the camera present in new data
+        for track in self._tracks:
+            is_valid = True
+            for m_idx in range(track.number_measurements()):
```
johnwlambert, line 162: prefer `k` here

gtsfm/common/gtsfm_data.py
johnwlambert: I see. Makes sense. So there's no way around this really, to reconstruct the connectivity, without doing this

johnwlambert: an line comment would help:
```python
# Recreate track connectivity from track information
# For example: a track has cameras [0, 2, 5]. In that case we will add pairs (0, 2), (0, 5), (2, 5).
```

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+            # add all combinations of edges
+            camera_edges += list(itertools.combinations(cameras_in_use, 2))
+
+        if len(camera_edges) == 0:
+            return GtsfmData(self._number_images)
+
+        cameras_in_largest_cc = graph_utils.get_nodes_in_largest_connected_component(camera_edges)
+
+        # select cameras to drop and drop them
+        dropped_cameras = np.setdiff1d(list(self._cameras.keys()), cameras_in_largest_cc)
```
johnwlambert, line 195: @ayushbaid is it simpler to make the constructor with the "active cameras", instead of the "dropped ones"? @akshay-krishnan does that seem cleaner to you?

gtsfm/common/sfm_result.py
```diff
@@ -26,25 +28,22 @@ def __eq__(self, other: object) -> bool:
...
 
         # finally, compare reprojection error
-        return np.isclose(
-            self.total_reproj_error,
-            other.total_reproj_error,
-            rtol=1e-2,
-            atol=1e-1,
-            equal_nan=True,
-        )
+        return np.isclose(self.total_reproj_error, other.total_reproj_error, rtol=1e-2, atol=1e-1, equal_nan=True)
```
johnwlambert, line 35: i might just add in the inline comment why we consider NaNs to be equal

gtsfm/common/sfm_result.py
```diff
@@ -26,25 +28,22 @@ def __eq__(self, other: object) -> bool:
...
 
     def get_camera_poses(self) -> List[Pose3]:
         """Getter for camera poses wTi.
 
         Returns:
             camera poses as a list, each representing wTi
         """
-        return [self.sfm_data.camera(i).pose() for i in range(self.sfm_data.number_cameras())]
+        cameras = [self.gtsfm_data.get_camera(i) for i in range(self.gtsfm_data.number_images())]
```
johnwlambert, line 43: why not do:
```python
cameras = [ self.gtsfm_data.get_camera(i) for i in self.gtsfm_data.get_valid_camera_indices() ]
```
Isn't that simpler?

gtsfm/data_association/dummy_da.py
```diff
@@ -28,19 +29,21 @@ def __init__(self, reproj_error_thresh: float, min_track_len: int):
 
     def run(
         self,
+        num_images: int,
```
johnwlambert, line 32: i prefer we delete the `dummy_da` class at this time, as I don't believe it serves any purpose

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,196 @@
...
+
+
+class GtsfmData:
+    """Class containing cameras and tracks, essentially describing the complete 3D scene.
+    
+    This class is needed over GTSAM's SfmData type because GTSAM's type does not allow for non-contiguous cameras.
+    The situation of non-contiguous cameras can exists because of failures in front-end.
+    """
+
+    def __init__(self, number_images: int) -> None:
```
johnwlambert, line 23: why does this class need to know `number_images`?

gtsfm/utils/io.py
```diff
@@ -85,3 +84,79 @@ def save_json_file(
...
+            camera = gtsfm_data.get_camera(i)
+            calibration = camera.calibration()
+
+            fx = calibration.fx()
+            u0 = calibration.px()
+            v0 = calibration.py()
+            k1 = calibration.k1()
+            k2 = calibration.k2()
+
+            f.write("{} SIMPLE_RADIAL {} {} {} {} {} {} {}\n".format(i, image_width, image_height, fx, u0, v0, k1, k2))
```
johnwlambert, line 139: i think python f-strings would make this even simpler :-)

gtsfm/utils/io.py
```diff
@@ -85,3 +84,79 @@ def save_json_file(
         json.dump(data, f, indent=4)
 
 
+def read_bal(file_path: str) -> GtsfmData:
+    """Read a BAL file.
```
johnwlambert, line 88: Maybe we can add parts of the docstring from write_bal or read_bal from gtsam? I added a URL there showing what BAL means

gtsfm/utils/graph.py
```diff
@@ -0,0 +1,30 @@
...
+def get_nodes_in_largest_connected_component(edges: List[Tuple[int, int]]) -> List[int]:
+    """Finds the nodes in the largest connected component of the bidirectional graph defined by the input edges.
+
+    Args:
+        edges: edges of the bi-directional graph.
+
+    Returns:
+        Nodes in the largest connected component of the input graph.
+    """
+
```
johnwlambert, line 19: nit: i prefer no new line here

gtsfm/utils/graph.py
```diff
@@ -0,0 +1,30 @@
+"""Utilities for performing graph operations.
```
johnwlambert, line 1: Can we be more specific here?

Technically GTSFM is one big graph. So it's not clear what kind of graph this is.

Can we name the file something like `gtsfm/utils/connected_components.py`?

gtsfm/utils/io.py
johnwlambert: i found it:
```
    """Write SfmData object as a "Bundle Adjustment in the Large" (BAL) file
    See https://grail.cs.washington.edu/projects/bal/ for more details on the format.
```

gtsfm/scene_optimizer.py
```diff
@@ -213,12 +211,20 @@ def create_computation_graph(
         if self._save_bal_files:
             # save the input to Bundle Adjustment (from data association)
             auxiliary_graph_list.append(
-                dask.delayed(write_sfmdata_to_disk)(ba_input_graph, os.path.join(RESULTS_PATH, "ba_input.bal"))
+                dask.delayed(io_utils.write_cameras)(ba_input_graph, save_dir=os.path.join(RESULTS_PATH, "ba_input"))
```
johnwlambert, line 214: @ayushbaid i forgot to ask, does a keyword arg work on dask.delayed?

tests/bundle/test_bundle_adjustment.py
```diff
@@ -8,12 +8,13 @@
...
 import numpy as np
-from gtsam import SfmData, SfmTrack
 
+import gtsfm.utils.io as io_utils
 from gtsfm.bundle.bundle_adjustment import BundleAdjustmentOptimizer
+from gtsfm.common.gtsfm_data import GtsfmData
 
 GTSAM_EXAMPLE_FILE = "dubrovnik-3-7-pre"
-EXAMPLE_DATA = gtsam.readBal(gtsam.findExampleDataFile(GTSAM_EXAMPLE_FILE))
+EXAMPLE_DATA = io_utils.read_bal(gtsam.findExampleDataFile(GTSAM_EXAMPLE_FILE))
```
johnwlambert, line 17: don't we need to convert this to `GtsfmData`? It is `SfmData`?

tests/bundle/test_bundle_adjustment.py
```diff
@@ -33,14 +34,7 @@ def test_simple_scene(self):
...
 
-        self.assertTrue(
-            np.isclose(
-                expected_error,
-                computed_result.total_reproj_error,
-                atol=1e-2,
-                rtol=1e-2,
-            )
-        )
+        self.assertTrue(np.isclose(expected_error, computed_result.total_reproj_error, atol=1e-2, rtol=1e-2))
```
johnwlambert, line 37: do we need both atol and rtol? I prefer just atol

tests/common/test_sfm_result.py
```diff
@@ -7,13 +7,12 @@
...
+import gtsfm.utils.io as io_utils
+from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.common.sfm_result import SfmResult
 
 GTSAM_EXAMPLE_FILE = "dubrovnik-3-7-pre"
-EXAMPLE_RESULT = SfmResult(
-    gtsam.readBal(gtsam.findExampleDataFile(GTSAM_EXAMPLE_FILE)),
-    total_reproj_error=1.5e1,
-)
+EXAMPLE_RESULT = SfmResult(io_utils.read_bal(gtsam.findExampleDataFile(GTSAM_EXAMPLE_FILE)), total_reproj_error=1.5e1,)
```
johnwlambert, line 15: 15 is more readable than 1.5e1?

tests/common/test_sfm_result.py
```diff
@@ -29,8 +28,7 @@ def testEqualsWithDifferentObject(self):
         """Test the equality function with different object, expecting false result."""
         other_example_file = "dubrovnik-1-1-pre.txt"
         other_result = SfmResult(
-            gtsam.readBal(gtsam.findExampleDataFile(other_example_file)),
-            total_reproj_error=1.1e1,
+            io_utils.read_bal(gtsam.findExampleDataFile(other_example_file)), total_reproj_error=1.1e1,
```
johnwlambert, line 31: 11 more readable than 1.1e1?

tests/common/test_gtsfm_data.py
```diff
@@ -0,0 +1,149 @@
...
+        """Test dropping cameras."""
+
+        # drop the camera at index 1 in the example data
+        computed = EXAMPLE_DATA.drop_cameras([1])
+
+        self.assertEqual(computed.get_camera(0), computed.get_camera(0))
+        self.assertEqual(computed.get_camera(2), computed.get_camera(2))
+        self.assertEqual(computed.number_tracks(), 0)
+
+    @mock.patch.object(graph_utils, "get_nodes_in_largest_connected_component", return_value=[1, 2, 4])
```
johnwlambert, line 119: do you mind adding an inline comment explaining why we need to use mock here?

tests/common/test_gtsfm_data.py
```diff
@@ -0,0 +1,149 @@
...
+    def test_select_largest_connected_component(self, graph_largest_cc_mock):
+        """Test pruning to largest connected component according to tracks."""
+        gtsfm_data = GtsfmData(5)
+        cam = PinholeCameraCal3Bundler(Pose3(), Cal3Bundler())
+
+        # add the same camera at all indices
+        for cam_idx in range(gtsfm_data.number_images()):
+            gtsfm_data.add_camera(cam, cam_idx)
+
+        # add two tracks to define connected component
```
johnwlambert, line 129: inline comment might be nice, explaining that this track will be thrown away (and we should check that it is), and that these cameras will be dropped

tests/common/test_gtsfm_data.py
```diff
@@ -0,0 +1,149 @@
...
+        self.assertTrue(gtsfm_data.add_track(track_to_add))
+
+    def test_add_track_with_nonexistant_cameras(self):
+        """Testing track addition where some cameras are not in tracks, resulting in failure."""
+        gtsfm_data = copy.deepcopy(EXAMPLE_DATA)
+
+        # add a track on camera #0 and #1, which exists in the data
+        track_to_add = SfmTrack(np.array([0, -2.0, 5.0]))
+        track_to_add.add_measurement(0, np.array([20.0, 5.0]))
+        track_to_add.add_measurement(3, np.array([60.0, 50.0]))  # this camera does not exist
```
johnwlambert, line 105: keyword arg would help, to explain what 3 represents

tests/common/test_gtsfm_data.py
```diff
@@ -0,0 +1,149 @@
...
+
+import gtsam
+import numpy as np
+from gtsam import Cal3Bundler, PinholeCameraCal3Bundler, Pose3, SfmTrack
+
+import gtsfm.utils.graph as graph_utils
+import gtsfm.utils.io as io_utils
+from gtsfm.common.gtsfm_data import GtsfmData
+
+GTSAM_EXAMPLE_FILE = "dubrovnik-3-7-pre"
```
johnwlambert, line 17: it would help to explain what this scenario is here. is it 3 cameras? what does 7 represent?
these should go in a docstring or inline comment here

gtsfm/common/sfm_result.py
ayushbaid: That won't work because we need None for a camera if it wasn't optimized.

gtsfm/common/gtsfm_data.py
ayushbaid: Because we can have 5 images, but just 3 cameras optimized by SfM

gtsfm/utils/io.py
ayushbaid: Oh I forgot to put `f` earlier and that is why it was not working. Will revert and fix that.

gtsfm/utils/graph.py
ayushbaid: I wanted this file to be the place where we can put all graph (i.e. operations on nodes and edges) related operations. We currently just need connected_components, but in the future we might need some more.

For example, we might have a system to break the global graph into smaller densely connected components and join them using re-sectioning? Or figuring out which part of graph needs more edges for averaging and figuring out how to recover two view estimated on required edges better.

gtsfm/scene_optimizer.py
ayushbaid: I was not sure too. But i checked and it works.

tests/bundle/test_bundle_adjustment.py
ayushbaid: The users of gtsfm wont ever need to worry about gtsam's `read_bal`, so I do not think it is an issue.

If you think it needs to be renamed, I'd suggest `read_bal_to_gtsfm_data`.

tests/bundle/test_bundle_adjustment.py
ayushbaid: Larger scenes might have larger absolute errors?

gtsfm/scene_optimizer.py
johnwlambert: cool

gtsfm/common/sfm_result.py
johnwlambert: I see. Then can we make 2 changes:

The return signatures should be `List[Optional[Pose3]]`
Can we say that the length here should be equal to the number of images in `gtsfm_data`, even if they are None

tests/common/test_gtsfm_data.py
ayushbaid: is it better now?

tests/common/test_gtsfm_data.py
ayushbaid: better?

tests/common/test_gtsfm_data.py
ayushbaid: better?

tests/common/test_gtsfm_data.py
johnwlambert: much better!

tests/common/test_gtsfm_data.py
johnwlambert: yes, thanks!

tests/bundle/test_bundle_adjustment.py
johnwlambert: maybe. i dont know about that, though

tests/bundle/test_bundle_adjustment.py
johnwlambert: ok, i think its fine as is then

gtsfm/utils/graph.py
johnwlambert: ok. well i dont feel strongly about it. we can leave it then

gtsfm/common/gtsfm_data.py
johnwlambert: right. is it required information inside the class though? i.e. could we do without it?

gtsfm/common/gtsfm_data.py
johnwlambert: thanks

gtsfm/common/gtsfm_data.py
johnwlambert: @ayushbaid the changes looks great.

i think this one is still outstanding, though

tests/common/test_gtsfm_data.py
johnwlambert: I'm not too familiar with the mocking lib in Python or when we need it -- could you explain a little more about why a typical function call won't work?

gtsfm/common/gtsfm_data.py
```diff
@@ -0,0 +1,199 @@
...
+            camera_edges += list(itertools.combinations(cameras_in_use, 2))
+
+        if len(camera_edges) == 0:
+            return GtsfmData(self._number_images)
+
+        cameras_in_largest_cc = graph_utils.get_nodes_in_largest_connected_component(camera_edges)
+        return GtsfmData.pick_cameras(self, cameras_in_largest_cc)
+
+    @classmethod
+    def pick_cameras(cls, gtsfm_data: "GtsfmData", camera_indices: List[int]) -> "GtsfmData":
```
johnwlambert, line 169: looks great @ayushbaid . much better. only thing i might change is to name this
`def from_picked_cameras()`

I think class methods often start with `from_blah` to make their factory functionality clearer?

ayushbaid: updated

johnwlambert:
Thanks @ayushbaid !

PR #115, opened 2021-03-17
Fixing the loader for the door dataset 

ayushbaid:
Fixing the loader by reading the intrinsics and extrinsics from the reconstruction data.

This will only work for scenes with the same intrinsics (same camera and same orientation [portrait/landscape]).


johnwlambert:
Thanks for the PR, @ayushbaid . Did you mean to upload the .mat file?

I would probably prefer to have just the original file instead of the several .npy files

johnwlambert:
Should we make this an `OlssonLoader` or something like that?

johnwlambert:
@ayushbaid I prefer we delete the `FolderLoader` and replace it with an `OlssonLoader` and `ColmapLoader`. I agree with Frank that we already have too many formats and defining a new one is too confusing

The `ColmapLoader` should load from a cameras.txt file:
https://colmap.github.io/format.html#cameras-txt

PR #117, opened 2021-03-23
Olsson loader

johnwlambert:
- Replace old `FolderLoader` with OlssonLoader or ArgoverseDatasetLoader
- Eliminate need for intermediate .npy files, and just rely on original file
- Add unit tests on ArgoverseDatasetLoader
- Fix failing unit tests for SceneOptimizer and 1dsfm


gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+
+    Folder layout structure:
+    - RGB Images: images/
+    - Intrinsics + Extrinsics data (optional): data.mat
+
+    If explicit intrinsics are not provided, the exif data will be used.
+    """
+
+    def __init__(self, folder: str, image_extension: str = "jpg", use_gt_intrinsics: bool = True) -> None:
+        """Initializes to load from a specified folder on disk
```
akshay-krishnan, line 33: nit: missing period at the end of sentence.

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+
+        Args:
+            folder: the base folder for a given scene
+            image_extension: (optional)extension for the image files. Defaults to 'jpg'.
```
akshay-krishnan, line 37: I dont think we mention (optional) in the args section, do we? It's evident from the function signature.

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+            use_gt_intrinsics: whether to use ground truth intrinsics
+        """
+        self.use_gt_intrinsics = use_gt_intrinsics
+
+        # fetch all the file names in /images folder
+        search_path = os.path.join(folder, "images", f"*.{image_extension}")
```
akshay-krishnan, line 43: Do we follow a convention for strings? Double quotes vs single quotes? Its nice to remain consistent.

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+
+        self.image_paths = glob.glob(search_path)
+
+        # sort the file names
+        self.image_paths.sort()
+        self.num_imgs = len(self.image_paths)
+
+        file_path = os.path.join(folder, "data.mat")
```
akshay-krishnan, line 51: What file is this? Looks like intrinsics from the rest of the code, but would be useful to be specific in the variable name.

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+        # M = K [R | t]
+        # in GTSAM notation, M = K @ cTw
+        M_list = [data['P'][0][i] for i in range(self.num_imgs)]
+
+        # first pose is identity, so K is immediate given
+        self.K = M_list[0][:3,:3]
+        Kinv = np.linalg.inv(self.K)
+
+        # decode camera poses as:
+        # # K^{-1} @ M = cTw
```
akshay-krishnan, line 70: This can be on the same line right? Also cTw should be on the LHS. So `# decode camera poses as:  cTw = K^{-1} @ M`

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+        Returns:
+            intrinsics for the given camera.
+        """
+        if not self.use_gt_intrinsics:
+            # get intrinsics from exif
+            intrinsics = io_utils.load_image(self.image_paths[index]).get_intrinsics_from_exif()
+
+        else:
+            intrinsics = Cal3Bundler(
+                fx=min(self.K[0, 0], self.K[1, 1]),
```
akshay-krishnan, line 117: Are there cases where fx != fy ?

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+        """Get the camera pose (in world coordinates) at the given index.
+
+        Args:
+            index: the index to fetch.
+
+        Returns:
+            the camera pose w_T_index.
+        """
+        if not self.use_gt_intrinsics:
+            # poses also not available
```
akshay-krishnan, line 136: Poses *are* not available?

tests/loader/test_olsson_loader.py
```diff
@@ -0,0 +1,152 @@
...
+        # randomly check image loads from a few indices
+        np.testing.assert_allclose(
+            results[5].value_array, self.loader.get_image(5).value_array
+        )
+        np.testing.assert_allclose(
+            results[7].value_array, self.loader.get_image(7).value_array
+        )
+
+    def test_create_computation_graph_for_intrinsics(self):
+        """Tests the graph for all intrinsics."""
```
akshay-krishnan, line 137: Not sure what is being tested in these two unit tests. Arent tests for the graph supposed to be in the base class?

akshay-krishnan (changes requested):
Looks good, except for some minor changes

gtsfm/loader/olsson_loader.py
johnwlambert: Thanks, fixed

gtsfm/loader/olsson_loader.py
johnwlambert: Thanks, fixed

gtsfm/loader/olsson_loader.py
johnwlambert: Should be double quotes everywhere, enforced by `python black`

gtsfm/loader/olsson_loader.py
johnwlambert: contains 3x4 camera projection matrices -- updated as `cam_matrices_fpath`

gtsfm/loader/olsson_loader.py
johnwlambert: There are. In fact, fx!=fy in the provided ground truth

```python
        expected_fx = 2398.119
        expected_fy = 2393.952
```

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+            folder: the base folder for a given scene
+            image_extension: file extension for the image files. Defaults to 'jpg'.
+            use_gt_intrinsics: whether to use ground truth intrinsics
+        """
+        self.use_gt_intrinsics = use_gt_intrinsics
+
+        # fetch all the file names in /images folder
+        search_path = os.path.join(folder, "images", f"*.{image_extension}")
+
+        self.image_paths = glob.glob(search_path)
```
ayushbaid, line 45: these and other variables can be private. i.e. `self.foo` can be written as `self._foo`.

gtsfm/loader/olsson_loader.py
ayushbaid: In this case, is doing `min` the best idea?

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+        # in GTSAM notation, M = K @ cTw
+        M_list = [data['P'][0][i] for i in range(self.num_imgs)]
+
+        # first pose is identity, so K is immediate given
+        self.K = M_list[0][:3,:3]
+        Kinv = np.linalg.inv(self.K)
+
+        # decode camera poses as:
+        # # K^{-1} @ M = cTw
+        iTw_list = [ Kinv @ M_list[i] for i in range(self.num_imgs)]
```
ayushbaid, line 71: why not store it as a `Pose3` in this step?

johnwlambert: Hi Ayush, thanks for the PR. I would need to split the result into 3x3 and 3x1 inside the list comprehension, which gets quite messy

gtsfm/loader/olsson_loader.py
```diff
@@ -0,0 +1,154 @@
...
+    def get_camera_pose(self, index: int) -> Optional[Pose3]:
+        """Get the camera pose (in world coordinates) at the given index.
+
+        Args:
+            index: the index to fetch.
+
+        Returns:
+            the camera pose w_T_index.
+        """
+        if not self.use_gt_intrinsics:
```
ayushbaid, line 135: should we rename the boolean flag?

gtsfm/loader/olsson_loader.py
johnwlambert: The unit tests are now passing -- we can reconstruct the ground truth up to 
```python
            self.assertTrue(comp_utils.compare_global_poses(
                poses,
                expected_poses,
                rot_err_thresh=0.03,
                trans_err_thresh=0.35
            ))
```
so i think it is ok

gtsfm/scene_optimizer.py
```diff
@@ -301,12 +302,15 @@ def visualize_camera_poses(
     fig = plt.figure()
     ax = fig.gca(projection="3d")
 
-    viz_utils.plot_poses_3d(pre_ba_poses, ax, center_marker_color="c", label_name="Pre-BA")
-    viz_utils.plot_poses_3d(post_ba_poses, ax, center_marker_color="k", label_name="Post-BA")
     if gt_pose_graph is not None:
-        gt_pose_graph = comp_utils.align_poses_sim3(post_ba_poses, gt_pose_graph)
+        # ground truth is used as the reference
+        pre_ba_poses = comp_utils.align_poses_sim3(gt_pose_graph, copy.deepcopy(pre_ba_poses))
+        post_ba_poses = comp_utils.align_poses_sim3(gt_pose_graph, copy.deepcopy(post_ba_poses))
```
ayushbaid, line 308: why did you add `deepcopy`. Was it not working as expected earlier?

gtsfm/loader/olsson_loader.py
johnwlambert: I was wondering this too.

How about two separate variables, which both default to true, but get overridden if we cant find what we need?
````python
self._provide_gt_extrinsics = blah
self.provide_gt_intrinsics = blah
```

gtsfm/utils/geometry_comparisons.py
```diff
@@ -162,15 +163,24 @@ def compare_global_poses(
...
+        np.allclose(aTi.translation(), aTi_.translation(), atol=trans_err_thresh) for (aTi, aTi_) in zip(aTi_list, aTi_list_)
+    ])
+    if verbose:
+        # GTSAM Rot3.equals() compares rotation matrices by absolute tolerance
+        rotation_errors = np.array([
+            (aTi.rotation().matrix() - aTi_.rotation().matrix()).max() for (aTi, aTi_) in zip(aTi_list, aTi_list_)
+        ])
+        translation_errors = np.array([
+            np.linalg.norm(aTi.translation() - aTi_.translation()) for (aTi, aTi_) in zip(aTi_list, aTi_list_)
+        ])
```
ayushbaid, line 179: i think we should move to angular errors? Or L2 errors in quaternions/eular angles? 

Comparing matrices for rotations do not make much sense.

gtsfm/scene_optimizer.py
johnwlambert: so my concern before was that we were modifying the ground truth, to fit the estimated data. I prefer the other way around.

However, I don't want to modify the original data just for the sake of alignment (since we'll do that later when comparing with ground truth)

So I prefer to just work with a copy of the data

gtsfm/utils/geometry_comparisons.py
johnwlambert: I agree. But I haven't changed that code for now. How about I open an issue, and we address it later?

gtsfm/loader/olsson_loader.py
johnwlambert: good suggestion, I've updated it now

gtsfm/loader/olsson_loader.py
johnwlambert: it's a 3x4 array, not 4x4, so we don't immediately have a relevant constructor

gtsfm/loader/olsson_loader.py
johnwlambert: fixed, i use a different variable name now for clarity

gtsfm/loader/olsson_loader.py
johnwlambert: Since I had the following above

```python
        # M = K [R | t]
        # in GTSAM notation, M = K @ cTw
```
i thought the math looked  simpler to keep things on the same side:
```python
        # decode camera poses as:
        #  K^{-1} @ M = cTw
```

.github/scripts/python.sh
```diff
@@ -24,5 +24,5 @@ pip install -e .
 # Run GTSFM unit tests
 ##########################################################
 
-cd $GITHUB_WORKSPACE/tests
-python -m unittest discover
+cd $GITHUB_WORKSPACE
+pytest tests
```
ayushbaid, line 28: Yeah I think this is better as it will run all tests even when a failure occurs?

gtsfm/loader/olsson_loader.py
```diff
@@ -18,7 +18,7 @@
 
 
 class OlssonLoader(LoaderBase):
-    """Simple loader class that reads Carl Olsson's dataset from a folder on disk.
+    """Simple loader class that reads any of Carl Olsson's datasets from a folder on disk.
 
     Ref: http://www.maths.lth.se/matematiklth/personal/calle/dataset/dataset.html
```
ayushbaid, line 23: shouldn't this be better placed in the header docstring at the top?

tests/data/argoverse/train1/273c1883-673a-36bf-b124-88311b1a80be/poses/city_SE3_egovehicle_315975643412234000.json
```diff
@@ -0,0 +1 @@
+{"rotation": [-0.8553418671199027, -0.0007711494300289546, -0.01385911873546937, -0.5178779977062857], "translation": [-273.2461280730776, 3052.5016282247057, -19.100182530254255]}
```
ayushbaid, line 1: why did these argoverse files change?

tests/loader/test_argoverse_dataset_loader.py
```diff
@@ -0,0 +1,88 @@
...
+
+    def setUp(self) -> None:
+        """ """
+        self.loader = ArgoverseDatasetLoader(
+            dataset_dir=ARGOVERSE_DATA_ROOT_PATH / "train1",
+            log_id="273c1883-673a-36bf-b124-88311b1a80be",
+            stride=1,
+            max_num_imgs=2,
+            max_lookahead_sec=50,
+            camera_name="ring_front_center",
```
ayushbaid, line 24: as these params are used in the function `test_get_image_contents`, maybe make them global variables above this class?

gtsfm/utils/geometry_comparisons.py
ayushbaid: Makes sense. Thanks!

gtsfm/scene_optimizer.py
ayushbaid: Ok got it.

gtsfm/loader/olsson_loader.py
ayushbaid: Looks good

tests/data/argoverse/train1/273c1883-673a-36bf-b124-88311b1a80be/poses/city_SE3_egovehicle_315975643412234000.json
johnwlambert: @ayushbaid I added these tests before we implemented the ArgoverseLoader, so I generated some intermediate files for the FolderLoader back then. No I am just using the native dataset format in the ArgoverseLoader.

gtsfm/loader/olsson_loader.py
johnwlambert: sure, the URL link you mean?

.github/scripts/python.sh
johnwlambert: Before we actually weren't running almost 100 tests -- we were only running the unittest ones, not the other pytest ones

gtsfm/utils/geometry_comparisons.py
johnwlambert: I opened an issue

ayushbaid:
LGTM

ayushbaid (approved):

PR #123, opened 2021-03-26
Adding superpoint detector-descriptor

ayushbaid:


ayushbaid:
Hi @johnwlambert  and @codyly. I have not put the model weights into git tracking yet. The file is of size ~5MB. I wonder how I should proceed.

Also, how do I handle licensing issues.

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -0,0 +1,267 @@
+"""Superpoint detector+descriptor implementation.
+
+The network was proposed in 'SuperPoint: Self-Supervised Interest Point Detection and Description' and is implemented
+by wrapping over author's implementation.
+
+References:
+- https://arxiv.org/abs/1712.07629
+- https://github.com/magicleap/SuperGluePretrainedNetwork
```
johnwlambert, line 8: Thanks for the PR, Ayush.
this looks like the link for SuperGlue, not SuperPoint?

ayushbaid: This repo has the code and model weights for superpoint too.

johnwlambert: Got it

johnwlambert:
@ayushbaid I think no need to check in the weights. We can always just pull it down from 
```
https://github.com/magicleap/SuperGluePretrainedNetwork/raw/master/models/weights/superpoint_v1.pth
```
over wget when we need it, and we could put that path in the code, btw

johnwlambert:
Regarding derivatives, in their license I see:
```
DERIVATIVES: You may create derivatives of or make modifications to the Software, however, You agree that all and any such derivatives and modifications will be owned by Licensor and become a part of the Software licensed to You under this Agreement.  You may only use such derivatives and modifications for your own noncommercial internal research purposes, and you may not otherwise use, distribute or copy such derivatives and modifications in violation of this Agreement.
```

Let's keep their license in the file as you have it. I believe we can add type hints to it (modify it) and check that in

dellaert:
HMMMM `noncommercial internal research purposes` is problematic. We will have a free license, which allows commercial use. We might have to segregate parts of our code along license terms

ayushbaid:
> HMMMM `noncommercial internal research purposes` is problematic. We will have a free license, which allows commercial use. We might have to segregate parts of our code along license terms

Got it. So what should we do? Maybe organize it as OpenCV does? `gtsfm` and `gtsfm-nonfree`?

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -0,0 +1,267 @@
...
+            is_cuda (optional): flag controlling the use of GPUs via CUDA. Defaults to True.
+            weights_path (optional): Path to the model weights. Defaults to MODEL_WEIGHT_PATH.
+        """
+        super().__init__()
+        self._use_cuda = use_cuda
+        self._config = {"weights_path": weights_path}
+
+    def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
+        device = torch.device("cuda" if self._use_cuda else "cpu")
+        model = SuperPoint(self._config).to(device)
```
johnwlambert, line 45: add a TODO with the issue number

johnwlambert:
Questions: (1) can we modify their code within the license (e.g. type hints, tensor shapes)
(2) do we want noncommercial code in a separate folder within the repo, or in a separate repo as submodule?

johnwlambert:
Maybe the best strategy is through submodule as https://github.com/cvg/Hierarchical-Localization/tree/master/third_party

ayushbaid:
Superseded by other branches

PR #124, opened 2021-03-27
Add flake8 to CI

johnwlambert:
- Using flake8 to do some static analysis
- Dump codecov report


.github/workflows/test-python.yml
```diff
@@ -22,5 +22,6 @@ jobs:
         python-version: 3.8
     - shell: bash -l {0}
       run: |
+        pip install pytest-cov
```
ayushbaid, line 25: should we put this in the conda env?

johnwlambert: i think it might clutter the conda-env, since we only use it in CI. But i may be able to be persuaded otherwise

ayushbaid:
How does the coverage result work? How do we see it on github? And can we block merges based on the results?

johnwlambert:
@ayushbaid We can add `code-cov` if we want to block things. I havent added that yet since I think it's lower-priority for now

https://docs.codecov.io/docs/commit-status

johnwlambert:
Right now, you just need to look at the printed report in the CI output

ayushbaid:
LGTM!

ayushbaid (approved):

PR #125, opened 2021-03-27
Eval and visualize based on largest connected component

johnwlambert:
- Clean up JSON saving to specific number of sig figs
- Only send to alignment the ground truth poses that correspond to largest connected component


gtsfm/common/gtsfm_data.py
```diff
@@ -127,7 +130,11 @@ def add_track(self, track: SfmTrack) -> bool:
         self._tracks.append(track)
         return True
 
-    def add_camera(self, index: int, camera: PinholeCameraCal3Bundler,) -> None:
+    def add_camera(
+        self,
+        index: int,
+        camera: PinholeCameraCal3Bundler,
+    ) -> None:
```
ayushbaid, line 137: did you reformat with 80 line limit?

gtsfm/common/gtsfm_data.py
```diff
@@ -163,6 +170,11 @@ def select_largest_connected_component(self) -> "GtsfmData":
             return GtsfmData(self._number_images)
 
         cameras_in_largest_cc = graph_utils.get_nodes_in_largest_connected_component(camera_edges)
+        logger.info(
+            "Largest connected component contains {} of {} cameras".format(
+                len(cameras_in_largest_cc), self._number_images
```
ayushbaid, line 175: the original data itself might not have some cameras. I think you might want to print `len(self.get_valid_camera_indices())` here.

gtsfm/data_association/data_assoc.py
```diff
@@ -117,10 +117,10 @@ def run(
         track_cheirality_failure_ratio = num_tracks_w_cheirality_exceptions / len(tracks_2d)
 
         # pick only the largest connected component
-        result_data = triangulated_data.select_largest_connected_component()
+        connected_data = triangulated_data.select_largest_connected_component()
 
         mean_3d_track_length, median_3d_track_length, track_lengths_3d = SfmResult(
-            result_data, total_reproj_error=float("Nan")
+            triangulated_data, total_reproj_error=float("Nan")
```
ayushbaid, line 123: this should be `connected_data`?

gtsfm/utils/metrics.py
```diff
@@ -57,23 +57,23 @@ def count_correct_correspondences(
     return np.count_nonzero(epipolar_distances < epipolar_dist_threshold)
 
 
-def compute_errors_statistics(errors: List[Optional[float]]) -> StatsDict:
+def compute_errors_statistics(errors: List[Optional[float]], num_sig_figs: int = 2) -> StatsDict:
```
ayushbaid, line 60: Can we somehow make this a global param? I feel we might need to do this everywhere?

tests/data_association/test_data_association.py
```diff
@@ -252,13 +252,20 @@ def test_data_association_with_missing_camera(self):
...
+        # no track will be formed, and thus connected component will be empty
+        sfm_data, _ = da.run(
+            num_images=3,
+            cameras=cameras,
+            corr_idxs_dict=corr_idxs_dict,
+            keypoints_list=[keypoints_shared] * 3
+        )
 
-        self.assertEqual(len(sfm_data.get_valid_camera_indices()), 2)
+        self.assertEqual(len(sfm_data.get_valid_camera_indices()), 0)
```
ayushbaid, line 268: this test should check if 0->1->2, right? Maybe we need to have a new test for the cheirality exception.

ayushbaid (changes requested):

gtsfm/common/gtsfm_data.py
johnwlambert: @ayushbaid I didn't quite catch this -- we form the new data with `cameras_in_largest_cc`, so that should be equivalent to `get_valid_camera_indices())` right?

gtsfm/utils/metrics.py
johnwlambert: I think we could put it at the top of the file, perhaps.

But I think more than 2 sig figs would not be useful for anyone, and I think less than 2 sig figs is also two few. What if I make 

```python
PRINT_NUM_SIG_FIGS = 2
```
at the top of the file?

gtsfm/data_association/data_assoc.py
johnwlambert: I think we could go either way. We could save just the DA stats for the largest connected component, or for the whole DA output.

The advantage of the former is that it gives us a 1:1 correspondence with the BA stats. But the latter is more representative of how DA did on average.

Perhaps the former is best? @ayushbaid @akshay-krishnan

tests/data_association/test_data_association.py
johnwlambert: Hmm I didn't quite catch what you meant here @ayushbaid

gtsfm/common/gtsfm_data.py
johnwlambert: i did, strange but must be >120

gtsfm/data_association/data_assoc.py
akshay-krishnan: Sorry missing context here, what are connected data and triangulated data?

tests/data_association/test_data_association.py
johnwlambert: the cheirality exception is being caught here still

gtsfm/data_association/data_assoc.py
johnwlambert: So after data association, we make have a disconnected graph. So we take the largest connected component of cameras.

So we could report DA stats before or after pruning to largest CC>

gtsfm/data_association/point3d_initializer.py
```diff
@@ -21,7 +21,8 @@
 )
 
 import gtsfm.utils.logger as logger_utils
-from gtsfm.common.sfm_track import SfmMeasurement, SfmTrack2d
+from gtsfm.common.sfm_track import SfmTrack2d
+from gtsfm.utils.reprojection import compute_point_reprojection_errors
```
ayushbaid, line 25: lets not import functions from files.

tests/data_association/test_data_association.py
ayushbaid: I mean why did the excepted value change from 2 to 0.

tests/utils/test_reprojection.py
```diff
@@ -0,0 +1,88 @@
+
+import numpy as np
+from gtsam import Cal3Bundler, Rot3, PinholeCameraCal3Bundler, Pose3, SfmTrack
+
+from gtsfm.common.sfm_track import SfmMeasurement
+from gtsfm.utils.reprojection import compute_point_reprojection_errors, compute_track_reprojection_errors
+
+
+def test_compute_track_reprojection_errors():
```
ayushbaid, line 9: will these tests run from `unittest.discover()`?

tests/data_association/test_data_association.py
johnwlambert: thanks for the review @ayushbaid. Basically we have just one track, but since there is a cheirality exception when we try to process that one track, there is no edge between any of the cameras. This means that 0 cameras end up in the largest CC. This is the reason for the change.

gtsfm/common/gtsfm_data.py
```diff
@@ -197,3 +210,17 @@ def from_selected_cameras(cls, gtsfm_data: "GtsfmData", camera_indices: List[int
                 new_data.add_track(track)
 
         return new_data
+
+    def get_scene_avg_reprojection_error(self) -> float:
+        """Get average reprojection error for all 3d points in the entire scene
+
+        Returns:
+            scene_avg_repoj_error: average of reprojection errors for every 3d point to its 2d measurements
```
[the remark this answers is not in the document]
johnwlambert: good catch, thanks

gtsfm/common/gtsfm_data.py
```diff
@@ -197,3 +210,17 @@ def from_selected_cameras(cls, gtsfm_data: "GtsfmData", camera_indices: List[int
...
+        """
+        scene_errors = []
```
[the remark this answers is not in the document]
johnwlambert: we're actually extending the list though, since each element returned is a numpy array, and it is cast to a list in the `extend()` call. and i dont know how to do that with list comprehension

gtsfm/common/gtsfm_data.py
```diff
@@ -197,3 +210,17 @@ def from_selected_cameras(cls, gtsfm_data: "GtsfmData", camera_indices: List[int
...
+        for track in self._tracks:
+            track_errors, _ = compute_track_reprojection_errors(self._cameras, track)
+            scene_errors.extend(track_errors)
+
+        scene_avg_repoj_error = np.array(scene_errors).mean()
```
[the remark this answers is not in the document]
johnwlambert: thanks, updated

gtsfm/common/gtsfm_data.py
johnwlambert: i took out the comma and seems to be fixed now

tests/utils/test_reprojection.py
johnwlambert: they won't, but `pytest tests` will catch them

gtsfm/data_association/point3d_initializer.py
johnwlambert: maybe`import from gtsfm.utils.reprojection import as reproj_utils`?

tests/data_association/test_data_association.py
ayushbaid:
```
keypoints_list = [
            feature_utils.generate_random_keypoints(1, [200, 200]),
            feature_utils.generate_random_keypoints(1, [200, 200]),
            feature_utils.generate_random_keypoints(1, [200, 200]),
        ]
```

ayushbaid (approved):
LGTM. We have added 2 TODOs for follow up

gtsfm/data_association/point3d_initializer.py
johnwlambert: fixed

PR #126, opened 2021-03-28
Modified data_assoc.py to call functions from utils/ellipsoid.py

adisingh50:
**Note:** this PR is outdated. Once the rest of the React application code is merged into master, I will work make a new PR similar to this where I add ellipsoid fitting functionality.

Modified `data_association/data_assoc.py` to import some helper functions from `util/ellipsoid.py`. These helper functions serve to 

- fit an ellipsoid quadric surface to the generated point cloud
- center the point cloud with respect to the origin
- apply a rotation matrix to the point cloud to align it along the x, y, z axes

This new rotated point cloud is added as a key/vaue pair to the `data_assoc_metrics` dictionary (line 145), which ultimately gets saved into the `data_association_metrics.json`.

### DIFF
--- a/gtsfm/data_association/data_assoc.py
+++ b/gtsfm/data_association/data_assoc.py
@@ -16,6 +16,7 @@ import numpy as np
 from dask.delayed import Delayed
 from gtsam import PinholeCameraCal3Bundler, SfmTrack
 
+import gtsfm.utils.ellipsoid as ellipsoid_utils
 import gtsfm.utils.logger as logger_utils
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.common.keypoints import Keypoints
@@ -25,7 +26,6 @@ from gtsfm.data_association.point3d_initializer import (
     Point3dInitializer,
     TriangulationParam,
 )
-
 
 logger = logger_utils.get_logger()
 
@@ -128,6 +128,14 @@ class DataAssociation(NamedTuple):
 
         # dump the 3d point cloud before Bundle Adjustment for offline visualization
         points_3d = [list(triangulated_data.get_track(j).point3()) for j in range(num_accepted_tracks)]
+
+        #create the centered and rotated 3d point cloud - properly aligned with x,y,z axes
+        centered_pc = ellipsoid_utils.center_point_cloud(points)
+        hullSet_points = ellipsoid_utils.compute_convex_hull(centered_pc)
+        polynomial_params = ellipsoid_utils.fit_ls_ellipsoid(hullSet_points)
+        center, axes, rot_matrix = ellipsoid_utils.extract_params_from_poly(polynomial_params)
+        aligned_points_3d = ellipsoid_utils.apply_ellipsoid_rotation(rot_matrix, centered_pc)
+
         # bin edges are halfway between each integer
         track_lengths_histogram, _ = np.histogram(track_lengths_3d, bins=np.linspace(-0.5, 10.5, 12))
 
@@ -150,6 +158,7 @@ class DataAssociation(NamedTuple):
             "per_rejected_track_avg_errors": per_rejected_track_avg_errors,
             "per_accepted_track_avg_errors": per_accepted_track_avg_errors,
             "points_3d": points_3d,
+            "aligned_points_3d": aligned_points_3d
         }
 
         return triangulated_data, data_assoc_metrics
--- a/gtsfm/utils/ellipsoid.py
+++ b/gtsfm/utils/ellipsoid.py
@@ -1,0 +1,163 @@
+"""Algorithms to fit Ellisoid geometry to 3D points and extract the rotation matrix necessary to align point cloud with
+x, y, and z axes. Used in React Three Fiber Visualization Tool.
+
+Authors: Adi Singh
+"""
+
+import numpy as np
+from numpy.linalg import eig, inv
+from scipy.spatial import ConvexHull, convex_hull_plot_2d
+
+
+def center_point_cloud(pointsList: list) -> np.ndarray:
+    """Centers a point cloud with respect using mean values of x, y, and z.
+
+    Args:
+        pointCloud: list(length N) of lists(length 3) (representing the point cloud)
+
+    Returns:
+        Centered point cloud, of shape Nx3
+
+    Raises:
+        TypeError: if points list is not 3 dimensional
+    """
+    if len(pointsList[0]) != 3:
+        raise TypeError("Points list should be 3D")
+
+    points = np.array([np.array(p) for p in pointsList])
+
+    means = np.mean(points, axis=0)
+    points_centered = points - means
+    return points_centered
+
+
+def compute_convex_hull(points_np: np.ndarray) -> np.ndarray:
+    """Returns a set of vertices which represent the 3D Convex Hull (outer perimeter) of our point cloud.
+
+    Args:
+        points_np: point cloud of shape N x 3
+
+    Returns:
+        a reduced collection of points representing the convex hull, of shape M x 3 (M <= N)
+
+    Raises:
+        TypeError: if points_np is not 3 dimensional
+    """
+    if points_np.shape[1] != 3:
+        raise TypeError("Point Cloud should be 3D")
+
+    hull = ConvexHull(points_np)
+    hullSet = set()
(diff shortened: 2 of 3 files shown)


johnwlambert:
Thanks for the PR, @adisingh50 . We'll need to add the ellipsoid_utils.py file here, so CI can pass when we merge

adisingh50:
I've added the ellipsoid_utils.py file on this branch.

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
...
+    Args:
+        points_np: point cloud of shape N x 3
+
+    Returns:
+        a reduced collection of points representing the convex hull, of shape M x 3 (M <= N)
+
+    Raises:
+        TypeError: if points_np is not 3 dimensional
+    """
+    if points_np.shape[1] != 3:
```
akshay-krishnan, line 46: Check if the dimension is greater than 1 before accessing shape[1]

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
...
+    Args:
+        hull_set: the convex hull of the original point cloud, of shape N x 3
+
+    Returns:
+        Array, of length 9, for parameters A, B, ..., J
+
+    Raises:
+        TypeError: if hull_set is not 3 dimensional
+    """
+    if hull_set.shape[1] != 3:
```
akshay-krishnan, line 76: Same as above, check dim > 1 first.

akshay-krishnan: Dim == 2 to be fair

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
...
+    x = hull_set[:, 0].reshape(hull_set.shape[0], 1)
+    y = hull_set[:, 1].reshape(hull_set.shape[0], 1)
+    z = hull_set[:, 2].reshape(hull_set.shape[0], 1)
+
+    A = np.hstack((x * x, y * y, z * z, 2 * x * y, 2 * x * z, 2 * y * z, 2 * x, 2 * y, 2 * z))
+    b = np.ones_like(x)
+
+    # solve normal equations to find the 9 parameters
+    ATA = A.T @ A
+    ATA_inv = np.linalg.inv(ATA)
```
akshay-krishnan, line 88: What if the matrix is singular? I guess you need at least 9 points, but they should also be linearly independent? Maybe handle that exception?

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
+"""Algorithms to fit Ellisoid geometry to 3D points and extract the rotation matrix necessary to align point cloud with
+x, y, and z axes. Used in React Three Fiber Visualization Tool.
+
+Authors: Adi Singh
+"""
+
+import numpy as np
+from numpy.linalg import eig, inv
```
akshay-krishnan, line 8: For readability, we don’t import functions, could you please use np.linalg.inv()?

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
...
+            [params[0], params[3] / 2.0, params[4] / 2.0, params[6] / 2.0],
+            [params[3] / 2.0, params[1], params[5] / 2.0, params[7] / 2.0],
+            [params[4] / 2.0, params[5] / 2.0, params[2], params[8] / 2.0],
+            [params[6] / 2.0, params[7] / 2.0, params[8] / 2.0, params[9]],
+        ]
+    )
+
+    # find center of ellipsoid
+    A3 = M[0:3, 0:3]
+    A3_inv = inv(A3)
```
akshay-krishnan, line 121: I’m wondering if inv is guaranteed to exist here.

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
...
+
+    Returns:
+        a modified point cloud entirely in list form
+
+    Raises:
+        TypeError: if rot matrix isn't 3x3 or centered_pc is not of dimension 3
+    """
+    if rot.shape[0] != 3 or rot.shape[1] != 3:
+        raise TypeError("Rotation Matrix should be of shape 3x3")
+    if centered_pc.shape[1] != 3:
```
akshay-krishnan, line 158: Might be worth checking dimensions first here as well.

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,163 @@
...
+    hullSet = set()
+    for simplex in hull.simplices:
+        hullSet.add(tuple(points_np[simplex[0]]))
+        hullSet.add(tuple(points_np[simplex[1]]))
+
+    hull_set = np.array([np.array(p) for p in hullSet])
+    return hull_set
+
+
+def fit_ls_ellipsoid(hull_set: np.ndarray) -> np.ndarray:
```
akshay-krishnan, line 59: “LS” in the name is actually an implementation detail. It might also not be clear to everyone what it means. So I’d recommend calling this just “fit_ellipsoid”

gtsfm/data_association/data_assoc.py
```diff
@@ -128,6 +128,14 @@ def run(
 
         # dump the 3d point cloud before Bundle Adjustment for offline visualization
         points_3d = [list(triangulated_data.get_track(j).point3()) for j in range(num_accepted_tracks)]
+
+        #create the centered and rotated 3d point cloud - properly aligned with x,y,z axes
+        centered_pc = ellipsoid_utils.center_point_cloud(points)
+        hullSet_points = ellipsoid_utils.compute_convex_hull(centered_pc)
+        polynomial_params = ellipsoid_utils.fit_ls_ellipsoid(hullSet_points)
+        center, axes, rot_matrix = ellipsoid_utils.extract_params_from_poly(polynomial_params)
```
akshay-krishnan, line 136: It would be useful to have a higher level API that accepts points and provides the aligned_points_3d.

gtsfm/utils/test_ellipsoid_utils.py
```diff
@@ -0,0 +1,121 @@
...
+    def test_center_point_cloud_sample_points(self):
+        """Testing the center_point_cloud() function with 5 sample points."""
+        sample_points = [[6, 13, 9], [5, 9, 7], [6, 10, 9], [9, 13, 6], [5, 12, 9]]
+        computed = ellipsoid_utils.center_point_cloud(sample_points)
+        expected = np.array([[-0.2, 1.6, 1], [-1.2, -2.4, -1], [-0.2, -1.4, 1], [2.8, 1.6, -2], [-1.2, 0.6, 1]])
+        npt.assert_almost_equal(computed, expected, 6)
+
+    def test_center_point_cloud_wrong_dims(self):
+        """Testing the center_point_cloud() function with 5 sample points of 2 dimensions."""
+        sample_points = [[6, 13], [5, 9], [6, 10], [9, 13], [5, 12]]
```
akshay-krishnan, line 21: You could also add a test for some input like [1,2,3] after you handle invalid dimensions ;)

gtsfm/data_association/data_assoc.py
```diff
@@ -128,6 +128,14 @@ def run(
...
+        aligned_points_3d = ellipsoid_utils.apply_ellipsoid_rotation(rot_matrix, centered_pc)
```
akshay-krishnan, line 137: I am surprised we don’t have utils to translate and rotate a point cloud already. If we really don’t, we should expand the scope of “ellipsoid_utils” to something like “3d_geometry_utils” and provide those utils. Also, we should use gtsam types for Rotation and translation, along with naming conventions like “wRc” that are used in other parts of GTSfM.

akshay-krishnan (changes requested):

ayushbaid:
Why do we have this function in data association? From what I understand, this method is used only in the UI.

johnwlambert:
> Why do we have this function in data association? From what I understand, this method is used only in the UI.

I guess we want to save a version of the "aligned"-to-orthogonal-axes point cloud. And we want to save every output like that.

I guess this should just be called in the COLMAP points3d export?

johnwlambert:
@adisingh50 we know it's summer, but we'd love to get this PR landed : - )

PR #127, opened 2021-03-29
MVS Api

johnwlambert:
Add abstract class implementing the MVS api

Assumes that GtsfmData will have a method to compute the depth range (min and max distances from any camera to any 3d point), which we still need to implement.

### DIFF
--- a/gtsfm/densify/mvs_base.py
+++ b/gtsfm/densify/mvs_base.py
@@ -3,7 +3,7 @@
 Authors: John Lambert, Ren Liu
 """
 import abc
-from typing import Dict, Tuple
+from typing import Dict
 
 import dask
 import numpy as np
--- a/gtsfm/densify/mvs_base.py
+++ b/gtsfm/densify/mvs_base.py
@@ -46,8 +46,10 @@ class MvsBase(metaclass=abc.ABCMeta):
         Args:
             images_graph: computation graph for images.
             cameras_graph
+            min_distance_graph
+            max_distance_graph
 
         Returns:
-            Delayed task for detection on the input image.
+            Delayed task for MVS computation on the input images.
         """
         return dask.delayed(self.densify)(image_graph, cameras_graph, min_distance_graph, max_distance_graph)
--- a/gtsfm/densify/mvs_base.py
+++ b/gtsfm/densify/mvs_base.py
@@ -1,4 +1,4 @@
-"""Base class for the MVS stage of the back-end.
+"""Base class for the multi-view stereo (MVS) stage of the back-end.
 
 Authors: John Lambert, Ren Liu
 """
@@ -13,8 +13,8 @@ from gtsam import PinholeCameraCal3Bundler
 from gtsfm.common.image import Image
 
 
-class MvsBase(metaclass=abc.ABCMeta):
-    """Base class for all multi-view stereo hyperparameters."""
+class MVSBase(metaclass=abc.ABCMeta):
+    """Base class for all multi-view stereo implementations."""
 
     def __init__(self) -> None:
         """Initialize the MVS module """
@@ -27,18 +27,20 @@ class MvsBase(metaclass=abc.ABCMeta):
         cameras: Dict[int, PinholeCameraCal3Bundler],
         min_distance: float,
         max_distance: float,
-    ) -> Tuple[np.ndarray, Dict[int, np.ndarray]]:
-        """Densify a point cloud using multi-view stereo
+    ) -> np.ndarray:
+        """Densify a point cloud using multi-view stereo.
+        
+        Note: we do not return depth maps here per image, as they would need to be aligned to ground truth
+        before evaluation for completeness, etc.
 
         Args:
-            image: dictionary mapping image indices to input images.
+            images: dictionary mapping image indices to input images.
             cameras: dictionary mapping image indices to camera parameters
             min_distance: minimum distance from any camera to any 3d point
             max_distance: maximum distance from any camera to any 3d point
 
         Returns:
-            array of shape (N,3) representing dense point cloud
-            dictionary mapping integer index to depth map of shape (H,W)
+            Dense point cloud, as an array of shape (N,3)
         """
 
     def create_computation_graph(
@@ -52,9 +54,9 @@ class MvsBase(metaclass=abc.ABCMeta):
 
         Args:
             images_graph: computation graph for images.
-            cameras_graph
-            min_distance_graph
-            max_distance_graph
+            cameras_graph: computation graph for cameras
+            min_distance_graph: minimum distance from any camera to any 3d point, wrapped up as Delayed
+            max_distance_graph: minimum distance from any camera to any 3d point, wrapped up as Delayed
 
         Returns:
             Delayed task for MVS computation on the input images.
--- a/gtsfm/densify/mvs_base.py
+++ b/gtsfm/densify/mvs_base.py
@@ -24,7 +24,6 @@ class MvsBase(metaclass=abc.ABCMeta):
         self,
         images: Dict[int,Image],
         cameras: Dict[int, PinholeCameraCal3Bundler],
-        gtsfm_data: Optional[GtsfmData] = None,
         min_distance: Optional[float] = None,
         max_distance: Optional[float] = None,
     ) -> Tuple[np.ndarray, Dict[int,np.ndarray]]:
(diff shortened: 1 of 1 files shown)


gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
+"""Base class for the MVS stage of the back-end.
```
akshay-krishnan, line 1: Could you expand MVS here?

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+import dask
+import numpy as np
+from dask.delayed import Delayed
+from gtsam import PinholeCameraCal3Bundler
+
+from gtsfm.common.image import Image
+
+
+class MvsBase(metaclass=abc.ABCMeta):
+    """Base class for all multi-view stereo hyperparameters."""
```
akshay-krishnan, line 17: Is it a base class for hyperparams or for all MVS modules?

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+
+    @abc.abstractmethod
+    def densify(
+        self,
+        images: Dict[int,Image],
+        cameras: Dict[int, PinholeCameraCal3Bundler],
+        min_distance: float,
+        max_distance: float,
+    ) -> Tuple[np.ndarray, Dict[int, np.ndarray]]:
+        """Densify a point cloud using multi-view stereo
```
akshay-krishnan, line 31: nit: missing period at the end.

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+
+        Args:
+            image: dictionary mapping image indices to input images.
+            cameras: dictionary mapping image indices to camera parameters
+            min_distance: minimum distance from any camera to any 3d point
+            max_distance: maximum distance from any camera to any 3d point
```
akshay-krishnan, line 37: Is it depth or distance? If it is used to determine some cost volume in disparity space, it might be depth?

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+        self,
+        images: Dict[int,Image],
+        cameras: Dict[int, PinholeCameraCal3Bundler],
+        min_distance: float,
+        max_distance: float,
+    ) -> Tuple[np.ndarray, Dict[int, np.ndarray]]:
+        """Densify a point cloud using multi-view stereo
+
+        Args:
+            image: dictionary mapping image indices to input images.
```
akshay-krishnan, line 34: typo: parameter name - images instead of image.

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+            cameras: dictionary mapping image indices to camera parameters
+            min_distance: minimum distance from any camera to any 3d point
+            max_distance: maximum distance from any camera to any 3d point
+
+        Returns:
+            array of shape (N,3) representing dense point cloud
+            dictionary mapping integer index to depth map of shape (H,W)
```
akshay-krishnan, line 41: do we need the depth map?

akshay-krishnan:
@johnwlambert I'm surprised that it allows merging without a review. :)

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+
+import dask
+import numpy as np
+from dask.delayed import Delayed
+from gtsam import PinholeCameraCal3Bundler
+
+from gtsfm.common.image import Image
+
+
+class MvsBase(metaclass=abc.ABCMeta):
```
ayushbaid, line 16: nit: maybe name it `MVSBase`?

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+        pass
+
+    @abc.abstractmethod
+    def densify(
+        self,
+        images: Dict[int,Image],
+        cameras: Dict[int, PinholeCameraCal3Bundler],
+        min_distance: float,
+        max_distance: float,
+    ) -> Tuple[np.ndarray, Dict[int, np.ndarray]]:
```
ayushbaid, line 30: so this won't take sparse point cloud as input?

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+        """Densify a point cloud using multi-view stereo
+
+        Args:
+            image: dictionary mapping image indices to input images.
+            cameras: dictionary mapping image indices to camera parameters
+            min_distance: minimum distance from any camera to any 3d point
+            max_distance: maximum distance from any camera to any 3d point
+
+        Returns:
+            array of shape (N,3) representing dense point cloud
```
ayushbaid, line 40: maybe write it as `Dense point cloud, as array of shape (N, 3)`.

gtsfm/densify/mvs_base.py
ayushbaid: I do not get what this return type is. Depth map for each image?

Maybe the docstring can be `Depth map for each input image`.

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+            dictionary mapping integer index to depth map of shape (H,W)
+        """
+
+    def create_computation_graph(
+        self,
+        images_graph: Delayed,
+        cameras_graph: Delayed,
+        min_distance_graph: Delayed,
+        max_distance_graph: Delayed
+    ) -> Delayed:
```
ayushbaid, line 50: We should return a tuple here, same thing as we did in a PR last month.

gtsfm/densify/mvs_base.py
```diff
@@ -0,0 +1,62 @@
...
+        """Generates the computation graph for performing multi-view stereo.
+
+        Args:
+            images_graph: computation graph for images.
+            cameras_graph
+            min_distance_graph
+            max_distance_graph
```
ayushbaid, line 57: Missing docstrings

ayushbaid (changes requested):

gtsfm/densify/mvs_base.py
johnwlambert: That's right -- depth map for each image. @codyly do you think it is useful?

Technically, to calculate MVS completeness metrics, we need to align both point clouds, and project them into the same space. I prefer we align the estimate to the ground truth. So that means unless we pass in the ground truth here, we would need to re-compute the depth maps again afterwards. So maybe returning the depth maps as well here is not helpful.

gtsfm/densify/mvs_base.py
johnwlambert: sure, thanks

gtsfm/densify/mvs_base.py
johnwlambert: sure, that looks better

gtsfm/densify/mvs_base.py
johnwlambert: Good point, should be "Base class for all multi-view stereo implementations."

gtsfm/densify/mvs_base.py
johnwlambert: There are a few ways to go about this. One is compute a single min and max depth for the whole scene. But for a big scene, and a long trajectory, I guess that doesnt make sense, since that can vary dramatically per camera. So maybe we should be computing the min and max depth per camera.? @dellaert  @codyly

gtsfm/densify/mvs_base.py
johnwlambert: thanks, fixed

gtsfm/densify/mvs_base.py
johnwlambert: i think actually depth maps will not be helpful, so i changed it to just one return type now

gtsfm/densify/mvs_base.py
codyly: It should be the depth from any camera to the detect points in the sparse point cloud, which is in the same scale as points coordinates.

akshay-krishnan: John, I understand distance to be norm of the 3D position in camera frame and depth to be the distance along the optical axis. They’re not equivalent

johnwlambert: Good point, thanks for catching that Akshay

ayushbaid (approved):
The recent changes looks good! Except the discussion on the return type, it looks good to me!

ayushbaid (approved):
LGTM!

Issue #130, opened 2021-03-31
Missing info for colmap export


johnwlambert:
missing access to 
(1) image filenames
(2) point color
(3) unique 2d point indices

in colmap-format dump/ write

PR #131, opened 2021-03-31
start colmap export

johnwlambert:
Dump 
images.txt
cameras.txt
points3D.txt

per the COLMAP file specification

Screen Shot 2021-04-02 at 12 11 10 PM
Screen Shot 2021-04-02 at 12 11 06 PM
Screen Shot 2021-04-02 at 12 11 03 PM


gtsfm/utils/io.py
```diff
@@ -151,9 +154,19 @@ def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
...
+    mean_obs_per_img = 0
+
+    # TODO: compute this
+    img_fname = "dummy.jpg"
+
     file_path = os.path.join(save_dir, "images.txt")
     with open(file_path, "w") as f:
-        f.write("# Number of cameras: {}\n".format(gtsfm_data.number_images()))
+        f.write("# Image list with two lines of data per image:")
+        f.write("#   IMAGE_ID, QW, QX, QY, QZ, TX, TY, TZ, CAMERA_ID, NAME")
```
ayushbaid, line 167: Technically, camera ID and image ID can be different. How will we handle that? Maybe we need the camera-id image-id map in our pipeline too.

gtsfm/utils/io.py
```diff
@@ -162,4 +175,44 @@ def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
...
+        f.write("# 3D point list with one line of data per point:\n")
+        f.write("#   POINT3D_ID, X, Y, Z, R, G, B, ERROR, TRACK[] as (IMAGE_ID, POINT2D_IDX)\n")
+        f.write(f"# Number of points: {num_pts}, mean track length: {avg_track_length}\n")
+
+        # TODO: extract point color
+        r, g, b = 0, 0, 0
+        # TODO: coming in another PR
+        track_reproj_error = 0
+        # TODO: assign unique indices to all keypoints (2d points)
+        point2d_idx = 0
```
ayushbaid, line 207: These indices will be 0-n per image, right?

gtsfm/utils/io.py
johnwlambert: i think that is a more conceptual change that will require some re-thinking. I added it to an issue https://github.com/borglab/gtsfm/issues/130

gtsfm/utils/io.py
johnwlambert: I don't think so.

POINT2D_IDX defines the zero-based index of the keypoint in the images.txt file

https://colmap.github.io/format.html#points3d-txt

johnwlambert: Actually, maybe you are right. It's a bit unclear.

gtsfm/scene_optimizer.py
```diff
@@ -394,21 +405,19 @@ def aggregate_frontend_metrics(
...
-        "translation": {"success_count": int(success_count_unit3),},
-        "pose": {"success_count": int(success_count_pose),},
-        "correspondences": {"all_inliers": int(all_correct),},
+        "rotation": {"success_count": int(success_count_rot3)},
+        "translation": {"success_count": int(success_count_unit3)},
+        "pose": {"success_count": int(success_count_pose)},
+        "correspondences": {"all_inliers": int(all_correct)},
     }
 
     io_utils.save_json_file(os.path.join(METRICS_PATH, "frontend_summary.json"), front_end_result_info)
```
adisingh50, line 423: I noticed you were removing a lot of trailing commas that weren't needed. On lines 89 and 256, I think you have trailing commas which could be removed.

gtsfm/utils/io.py
```diff
@@ -151,9 +158,19 @@ def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
...
+    # TODO: compute this
+    img_fname = "dummy.jpg"
+
     file_path = os.path.join(save_dir, "images.txt")
     with open(file_path, "w") as f:
-        f.write("# Number of cameras: {}\n".format(gtsfm_data.number_images()))
+        f.write("# Image list with two lines of data per image:")
+        f.write("#   IMAGE_ID, QW, QX, QY, QZ, TX, TY, TZ, CAMERA_ID, NAME")
+        f.write("#   POINTS2D[] as (X, Y, POINT3D_ID)")
+        f.write(f"# Number of cameras: {num_imgs}, mean observations per image: {mean_obs_per_img}\n")
```
adisingh50, line 173: small change; based on the COLMAP format, I think this line should read `"Number of images: "`...

gtsfm/scene_optimizer.py
johnwlambert: Thanks for taking a look Adi. Python black adds those sometimes -- the ones I removed earlier I believe were due to us formatting and then re-formatting with different options, that led to the artifact being added.

I think the ones in lines 89 and 256, black will force us to keep for now

adisingh50 (approved):

adisingh50:
Thanks for adding all this. Once this PR is merged into master, the point clouds we render will look much cooler now that we can render colors and camera poses.

gtsfm/utils/io.py
johnwlambert: Good catch, I just fixed that

gtsfm/utils/io.py
johnwlambert: @adisingh50 would you mind looking into this issue here about what the 2d keypoint/point index means?

gtsfm/utils/images.py
```diff
@@ -122,3 +123,32 @@ def match_image_widths(
...
+
+def get_average_point_color(track: SfmTrack, images: List[Image]) -> Tuple[int, int, int]:
+    """
+    Args:
+        track: 3d point/landmark and its corresponding 2d measurements in various cameras
+        images: list of all images for this scene
+
+    Returns:
+        r: red color intensity, in range [0,255]
+        g: red color intensity, in range [0,255]
```
ayushbaid, line 136: green and blue, respectively.

gtsfm/utils/images.py
```diff
@@ -122,3 +123,32 @@ def match_image_widths(
...
+        b: red color intensity, in range [0,255]
+    """
+    rgb_measurements = []
+    for k in range(track.number_measurements()):
+
+        # process each measurement
+        i, uv_measured = track.measurement(k)
+        img_h, img_w, _ = images[i].value_array.shape
```
ayushbaid, line 144: you should use `height` and `width` property of image directly.

gtsfm/utils/images.py
```diff
@@ -122,3 +123,32 @@ def match_image_widths(
...
+
+        u, v = np.round(uv_measured).astype(np.int32)
+        # ensure round did not push us out of bounds
+        u = np.clip(u, 0, img_w - 1)
+        v = np.clip(v, 0, img_h - 1)
+        rgb_measurements += [images[i].value_array[v, u]]
+
+    r, g, b = np.array(rgb_measurements).mean(axis=0).astype(np.uint8)
```
ayushbaid, line 152: is this typecasting necessary?

gtsfm/utils/io.py
```diff
@@ -12,6 +12,8 @@
 from PIL import Image as PILImage
 from PIL.ExifTags import GPSTAGS, TAGS
 
+import gtsfm.utils.reprojection as reproj_utils
+import gtsfm.utils.images as image_utils
```
ayushbaid, line 16: this import should be placed above, following an alphabetical ordering.

gtsfm/common/gtsfm_data.py
```diff
@@ -145,6 +145,23 @@ def add_camera(self, index: int, camera: PinholeCameraCal3Bundler) -> None:
             raise ValueError("Camera cannot be None, should be a valid camera")
         self._cameras[index] = camera
 
+    def get_track_length_statistics(self) -> Tuple[float, float, np.ndarray]:
+        """Compute mean and median lengths of all the tracks.
+
+        Returns:
+            Mean track length.
+            Median track length.
+            Array containing all track lengths.
```
ayushbaid, line 154: I think we should not return the full array with this function. There should be two functions.

gtsfm/scene_optimizer.py
```diff
@@ -215,6 +215,12 @@ def create_computation_graph(
             auxiliary_graph_list.append(
                 dask.delayed(io_utils.write_images)(ba_input_graph, save_dir=os.path.join(RESULTS_PATH, "ba_input"))
```
ayushbaid, line 216: make a variable `save_dir` and use it in two places. This might lead to a bug creeping in later.

gtsfm/scene_optimizer.py
```diff
@@ -226,6 +232,11 @@ def create_computation_graph(
                     filtered_sfm_data_graph, save_dir=os.path.join(RESULTS_PATH, "ba_output")
```
ayushbaid, line 232: same point as above.

ayushbaid (changes requested):
Some minor comments

gtsfm/utils/images.py
johnwlambert: thanks, i'll fix it

gtsfm/utils/images.py
johnwlambert: thanks, i'll fix it

gtsfm/utils/images.py
johnwlambert: indeed it is:
```python
>>> import numpy as np
>>> arr = np.array([[50,60,40],[40,50,30]],dtype=np.uint8)
>>> arr
array([[50, 60, 40],
       [40, 50, 30]], dtype=uint8)
>>> arr.mean(axis=0)
array([45., 55., 35.])
>>> arr.mean(axis=0).dtype
dtype('float64')
>>> 
```

gtsfm/utils/io.py
johnwlambert: thanks, i'll fix it

gtsfm/common/gtsfm_data.py
johnwlambert: i think we often want all 3 things at once, though

gtsfm/utils/images.py
johnwlambert: fixed

gtsfm/utils/images.py
johnwlambert: fixed

gtsfm/utils/io.py
johnwlambert: fixed

gtsfm/scene_optimizer.py
johnwlambert: fixed

gtsfm/scene_optimizer.py
johnwlambert: fixed

gtsfm/common/gtsfm_data.py
ayushbaid: maybe rename it?

gtsfm/utils/images.py
ayushbaid: got it. did not expect this.

ayushbaid (approved):

PR #136, opened 2021-04-04
Increase output quality (increase number tracks and track length)

johnwlambert:
- Do not filter by reprojection error after DA
- Filter by reprojection error only after BA
- Do not use RANSAC in DA

- Save visualization of each track (patch for each measurement)
- Remove SfmResult class, as GtsfmData is sufficient


gtsfm/common/gtsfm_data.py
```diff
@@ -246,4 +261,46 @@ def get_scene_avg_reprojection_error(self) -> float:
             scene_reproj_errors.extend(track_errors)
 
         scene_avg_repoj_error = np.mean(scene_reproj_errors)
+
+        # DEBUG
```
ayushbaid, line 265: should be removed once debug is complete.

johnwlambert:
On Door:
Before PR:
```
Scene avg reproj error: 0.21
4966 points
3.95 avg track length
```

After PR
```
Scene avg reproj error: 0.18
9110 points
5.25 avg track length
```

    "rotation_averaging_angle_deg": {
Before PR
```
        "median_error": 0.09,
        "min_error": 0.02,
        "max_error": 0.16,
```
After PR
```
    "rotation_averaging_angle_deg": {
        "median_error": 0.1,
        "min_error": 0.03,
        "max_error": 0.15,
```
------------
    "translation_averaging_distance": {
Before PR
```
        "median_error": 0.08,
        "min_error": 0.03,
        "max_error": 0.14,
```
After PR
```
        "median_error": 0.08,
        "min_error": 0.01,
        "max_error": 0.34,
```
--------
    "translation_to_direction_angle_deg": {
Before PR
```
        "median_error": 0.5,
        "min_error": 0.07,
        "max_error": 1.96,
```
After PR:
```
        "median_error": 0.55,
        "min_error": 0.02,
        "max_error": 1.96,
```

gtsfm/data_association/data_assoc.py
```diff
@@ -55,6 +59,8 @@ def run(
         cameras: Dict[int, PinholeCameraCal3Bundler],
         corr_idxs_dict: Dict[Tuple[int, int], np.ndarray],
         keypoints_list: List[Keypoints],
+        images: Optional[List[Image]] = None,
+        viz_patch_sz: int = 100,
```
ayushbaid, line 63: I think  visualization should be separate from the validation function.

gtsfm/scene_optimizer.py
```diff
@@ -60,7 +60,8 @@ def __init__(
         feature_extractor: FeatureExtractor,
         two_view_estimator: TwoViewEstimator,
         multiview_optimizer: MultiViewOptimizer,
-        save_viz: bool,
+        save_image_pair_viz: bool,
```
ayushbaid, line 63: change to `save_two_view_correspondences_viz`?

gtsfm/utils/images.py
```diff
@@ -41,7 +41,7 @@ def rgb_to_gray_cv(image: Image) -> Image:
     return Image(output_array, image.exif_data)
 
 
-def vstack_images(image_i1: Image, image_i2: Image) -> Image:
+def vstack_image_pair(image_i1: Image, image_i2: Image) -> Image:
```
ayushbaid, line 44: can we just use `vstack_image_list`?

tests/bundle/test_bundle_adjustment.py
```diff
@@ -23,18 +23,19 @@ class TestBundleAdjustmentOptimizer(unittest.TestCase):
...
-        self.obj = BundleAdjustmentOptimizer()
+        output_reproj_error_thresh = 100
+        self.obj = BundleAdjustmentOptimizer(output_reproj_error_thresh)
 
         self.test_data = EXAMPLE_DATA
 
-    def test_simple_scene(self):
-        """Test the simple scene using the `run` API."""
+    # def test_simple_scene(self):
+    #     """Test the simple scene using the `run` API."""
```
ayushbaid, line 32: why are we commenting out this test?

tests/common/test_gtsfm_data.py
```diff
@@ -30,13 +32,25 @@ class TestGtsfmData(unittest.TestCase):
     """Unit tests for GtsfmData."""
 
     def test_equality_with_same_data(self):
-        """Test equality with the same data."""
+        """Test equality with the same data (same object)."""
```
ayushbaid, line 35: its same value but not same object, right? We do a deepcopy.

tests/test_scene_optimizer.py
```diff
@@ -56,13 +56,13 @@ def test_create_computation_graph(self):
...
 
-            self.assertIsInstance(sfm_result, SfmResult)
+            self.assertIsInstance(sfm_result, GtsfmData)
 
             # compare the camera poses
             poses = sfm_result.get_camera_poses()
 
             # get active cameras from largest connected component, may be <len(self.loader)
-            connected_camera_idxs = sfm_result.gtsfm_data.get_valid_camera_indices()
+            connected_camera_idxs = sfm_result.get_valid_camera_indices()
```
ayushbaid, line 65: why do we have `get_valid_camera_indices()` in `SfmResult`?

ayushbaid (commented):
Looks good overall. Some minor changes suggested.

gtsfm/data_association/data_assoc.py
johnwlambert: Thanks for taking a look, Ayush.

What did you mean by "validation function"?

tests/test_scene_optimizer.py
johnwlambert: maybe i should rename it -- here `sfm_result` is actually an instance of `GtsfmData`. Does that make sense?

gtsfm/utils/images.py
johnwlambert: Sure. The only difference though is that for the image pair, you had functionality to resize to the same width; for the stacking image list I assumed they were all of the same width. But do you think it is cleaner  to consolidate them?

gtsfm/scene_optimizer.py
johnwlambert: Sure, thanks

tests/common/test_gtsfm_data.py
johnwlambert: Good point, I'll update the comment

tests/bundle/test_bundle_adjustment.py
johnwlambert: We are no longer packaging the factor graph error with the result. So we would need to recompute that error, or save it in the `GtsfmData` class somehow. Any thoughts about that?

gtsfm/scene_optimizer.py
johnwlambert: fixed

tests/common/test_gtsfm_data.py
johnwlambert: fixed

tests/test_scene_optimizer.py
ayushbaid: Ah got it. Yeah I think renaming is better :)

tests/bundle/test_bundle_adjustment.py
ayushbaid: got it. Yeah for now we can skip the test.

gtsfm/utils/images.py
ayushbaid: Oh I got it. Yeah lets keep it separate for now.

gtsfm/data_association/data_assoc.py
ayushbaid: This function is being used to check when we have a track with minimum acceptable length. I would prefer we write a seperate function for visualization and not merge core logic and visualization in the same function.

johnwlambert: makes sense, good point

tests/test_scene_optimizer.py
johnwlambert: I do think `sfm_result` is the best name for the variable name -- as long as we remember that the SfmResult class is now non-existent, so there would be no confusion there

ayushbaid (approved):
LGTM

PR #141, opened 2021-04-08
Discuss license in readme

johnwlambert:


ayushbaid (approved):

ayushbaid:
Should this statement also go into our license file?

johnwlambert:
> Should this statement also go into our license file?

It should, good point.

On that note, @dellaert do you have a preference between Apache 2.0 or MIT?

PR #143, opened 2021-04-11
logging warning when E decomposition doesnt match

ayushbaid:
When the essential matrix is decomposed into rotation and unit-translation between two poses, the R and t do not have the same essential matrix as the input each time. 

Right now, we have tried decomposing `E` into `R` and `t` manually instead of using OpenCV's API, but it does not resolve the issue.

For the door dataset using RANSAC-E, 9/66 pairs have this warning. For degensac, all 66/66 pairs have this warning.

### DIFF
--- a/gtsfm/utils/verification.py
+++ b/gtsfm/utils/verification.py
@@ -9,6 +9,9 @@ import numpy as np
 from gtsam import Cal3Bundler, EssentialMatrix, Pose3, Rot3, Unit3
 
 import gtsfm.utils.features as feature_utils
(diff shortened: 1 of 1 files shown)


ayushbaid:
While we debug this issue, this warning might help the users understand the scope of the potential 'bug' as well as report their observations.

dellaert (approved):
I really don’t understand what is going on. The warning code looks OK, but we should try and get to the root cause soon :-)

swarrier246 (approved):
Maybe we should also open an issue so users see that it's being looked into and contribute to it if they want?

PR #154, opened 2021-04-16
Dump BA metrics to disk

johnwlambert:
File resembles:
```
{
    "before_filtering": {
        "number_tracks": 9011,
        "3d_track_lengths": {
            "min": 2.0,
            "mean": 5.245,
            "median": 4.0,
            "max": 12.0
        },
        "reprojection_errors": {
            "min": 0.0,
            "mean": 0.207,
            "median": 0.136,
            "max": 78.841
        }
    },
    "after_filtering": {
        "number_tracks": 8971,
        "3d_track_lengths": {
            "min": 2.0,
            "mean": 5.245,
            "median": 4.0,
            "max": 12.0
        },
        "reprojection_errors": {
            "min": 0.0,
            "mean": 0.171,
            "median": 0.135,
            "max": 2.945
        }
    }
}
```


gtsfm/bundle/bundle_adjustment.py
```diff
@@ -136,6 +145,37 @@ def create_computation_graph(self, sfm_data_graph: Delayed) -> Delayed:
...
+    Args:
+        ba_data: bundle adjustment result
+
+    Returns:
+        dictionary containing summary statistics of bundle adjustment result
+    """
+    track_lengths_3d = ba_data.get_track_lengths()
+    scene_reproj_errors = ba_data.get_scene_reprojection_errors()
+
+    convert_to_rounded_float = lambda x: float(np.round(x,3))
```
akshay-krishnan, line 160: why is this necessary? Arent they float already? Maybe the rounding could be done without a lambda?

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -2,16 +2,21 @@
...
 import gtsam
+import numpy as np
 from dask.delayed import Delayed
 from gtsam import GeneralSFMFactorCal3Bundler, SfmTrack, Values, symbol_shorthand
 
+import gtsfm.utils.io as io_utils
 import gtsfm.utils.logger as logger_utils
 from gtsfm.common.gtsfm_data import GtsfmData
 
+METRICS_PATH = "result_metrics"
```
akshay-krishnan, line 18: We have this hardcoded in too many places, can we specify the output directory as a command line arg? And then the metrics subdirectory could be a config.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -136,6 +145,37 @@ def create_computation_graph(self, sfm_data_graph: Delayed) -> Delayed:
         return dask.delayed(self.run)(sfm_data_graph)
 
 
+def aggregate_ba_stats(ba_data: GtsfmData) -> Dict[str,Any]:
```
akshay-krishnan, line 148: missing space after str

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -136,6 +145,37 @@ def create_computation_graph(self, sfm_data_graph: Delayed) -> Delayed:
...
+    """Create a dictionary of summary statistics for a bundle adjustment result.
+
+    Args:
+        ba_data: bundle adjustment result
+
+    Returns:
+        dictionary containing summary statistics of bundle adjustment result
```
akshay-krishnan, line 155: We probably also need to align on some terminology here. Summary, statistics and metrics. Could we define each one so we be precise across gtsfm? 
I have been using metric for a quantity, and statistic for min, max, etc. Technically the statistics are also metrics so its not super accurate.

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan, line 148: Maybe change Any to Dict[str, float]?

akshay-krishnan (changes requested):
The data association metrics are returned to the multi-view optimizer. Why is that pattern not being followed here?

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Thanks for the review Akshay.

We will need a fixed place to store the outputs, since Adi's React application will look in a specific place for where to load the metrics. So I believe we will have to fix this here. Although we could save it in two places, one from a config, and one to the fixed location.

gtsfm/bundle/bundle_adjustment.py
johnwlambert: It's a bit weird -- np.round() or np.mean() will return a float32 object that is not json-serializable, until we cast it to a python `float`. So I couldn't get it to work without this

gtsfm/bundle/bundle_adjustment.py
ayushbaid, line 148: should this be a function in the `GtsfmData class`?

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -136,6 +145,37 @@ def create_computation_graph(self, sfm_data_graph: Delayed) -> Delayed:
...
+    """
+    track_lengths_3d = ba_data.get_track_lengths()
+    scene_reproj_errors = ba_data.get_scene_reprojection_errors()
+
+    convert_to_rounded_float = lambda x: float(np.round(x,3))
+
+    stats_dict = {}
+    stats_dict["number_tracks"] = ba_data.number_tracks()
+    stats_dict["3d_track_lengths"] = {
```
ayushbaid, line 164: just `track_lengths`?

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -136,6 +145,37 @@ def create_computation_graph(self, sfm_data_graph: Delayed) -> Delayed:
...
+        "min": convert_to_rounded_float(track_lengths_3d.min()),
+        "mean": convert_to_rounded_float(np.mean(track_lengths_3d)),
+        "median": convert_to_rounded_float(np.median(track_lengths_3d)),
+        "max": convert_to_rounded_float(track_lengths_3d.max())
```
ayushbaid, line 168: should we prefer dumping all the decimal places to the json?

ayushbaid (commented):
Looks good, except some minor changes

gtsfm/bundle/bundle_adjustment.py
johnwlambert: I think above 3 decimal places makes the json somewhat unreadable, and perhaps is not statistically significant

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Not sure. Seems specific to bundle adjustment, but I see your point too

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: Two points: 
- If we want a fixed directory, I think it should be a constant somewhere, maybe the config file, or in the runner. And that could be used in every module. In the worst case, all modules could accept an output_dir argument, which we hardcode in the runner. The problem with each module re-defining this constant is that someone might choose to change it at some point within a module, without knowledge of the fact that this should not be changed. 
- A user of gtsfm might want output (and metrics) to be stored somewhere else. I don't think it's good practice to restrict this to a fixed folder. We could of course have fixed paths *relative* to a variable folder that is specified by the user. Is it that way already?

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: Seems specific to BA to me too, is GtsfmData internal to BA? Or is it more like the output of the whole pipeline?

johnwlambert: `GtsfmData` is the current data structure for 3d tracks + camera poses/parameters. We use it as output of DA, input to BA, output of BA, and input to MVS. So it is a more general structure.

gtsfm/bundle/bundle_adjustment.py
johnwlambert: I do like differentiating between the 2d and 3d tracks though (and we use the term "3d_track_lengths" in the DA output, and there is a 1:1 correspondence between these two

gtsfm/bundle/bundle_adjustment.py
johnwlambert: technically it is `Dict[str,Union[float,Dict[str,float]]]` but i thought that was a bit messy. the output is nested, so technically it is not `Dict[str, float]`

johnwlambert: thanks, should be updated now

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Good point. Currently we were working under the assumption that people would only run the scene optimizer at a certain level of the hierarchy, so it would be a fixed relative location if they always run `python gtsfm/runner/run_scene_optimizer.py` at that level of the hierarchy.

But you're right -- the proper way to do this is to use `Path(__file__).resolve().parent.parent.parent`, so I've updated it per that suggestion.

Good point about wanting to dump output to a specified directory -- that makes sense for experiments. I think we will have to dump in two different places, though, since Adi will need the data dumped to a specific location for React to load it.

Let me know what you all think about that @ayushbaid @akshay-krishnan .

gtsfm/bundle/bundle_adjustment.py
ayushbaid: ok cool.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -145,35 +146,37 @@ def create_computation_graph(self, sfm_data_graph: Delayed) -> Delayed:
...
+    convert_to_rounded_float = lambda x: float(np.round(x, 3))
 
     stats_dict = {}
     stats_dict["number_tracks"] = ba_data.number_tracks()
     stats_dict["3d_track_lengths"] = {
         "min": convert_to_rounded_float(track_lengths_3d.min()),
         "mean": convert_to_rounded_float(np.mean(track_lengths_3d)),
         "median": convert_to_rounded_float(np.median(track_lengths_3d)),
-        "max": convert_to_rounded_float(track_lengths_3d.max())
+        "max": convert_to_rounded_float(track_lengths_3d.max()),
```
ayushbaid, line 171: I feel our formatters interfere with each other

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Looks good to me for now

gtsfm/bundle/bundle_adjustment.py
ayushbaid: `Any` works for me

ayushbaid: If we have it in `GtsfmData` we can reuse it in non-BA context too. For example: loading colmap results and publishing metrics for comparison.

ayushbaid (approved):
Approved except a suggestion for moving the function to `GtsfmData`

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Thanks for the review, @ayushbaid 

That's a good point about being able to re-use the functionality for a COLMAP comparison. We should probably rename this function then, though, since GtsfmData is not necssarily a BA result. So we couldn't call it `aggregate_ba_stats()`

Maybe `aggregate_track_stats_dict()`?

ayushbaid (approved):

PR #156, opened 2021-04-21
Added Point Cloud Viewer for Bundle Adj. Output and GTSFM Graph Reformatting

adisingh50:
I have reformatted the GTSFM graph to reflect the most recent changes:
image

I have also added a second point cloud viewer to view the resulting output after Bundle Adjustment. So now, if you click on the `Data Association GtsfmData` node, this is rendered:
image

And if you click on the `Bundle Adjustment GtsfmData` node, this is rendered:
image


johnwlambert:
Very nice @adisingh50 . These figures look great. I'll take a look

johnwlambert:
Still seeing a few warnings when I run `npm start` -- are these still expected?
```
src/Components/AllFrustums.js
  Line 62:34:  Expected '!==' and instead saw '!='  eqeqeq

src/Components/CoordinateGrid.js
  Line 35:32:  Array.prototype.map() expects a return value from arrow function  array-callback-return

src/Components/PCViewer.js
  Line 46:8:  React Hook useEffect has missing dependencies: 'loadCOLMAPPointCloud' and 'props.filePath'. Either include them or remove the dependency array  react-hooks/exhaustive-deps

src/Components/frustum_classes/se3.js
  Line 19:31:  Expected '!==' and instead saw '!='    eqeqeq
  Line 19:57:  Expected '!==' and instead saw '!='    eqeqeq
  Line 19:63:  Expected an error object to be thrown  no-throw-literal
  Line 20:34:  Expected '!==' and instead saw '!='    eqeqeq
  Line 20:42:  Expected an error object to be thrown  no-throw-literal
  Line 33:18:  'row' is already defined               no-redeclare

Search for the keywords to learn more about each warning.
To ignore, add // eslint-disable-next-line to the line before.



```

johnwlambert:
The computational graph figure looks much better, btw

johnwlambert:
The BA cloud looks really nice. I noticed that sometimes the cloud seems to get stuck and I can't rotate it anymore inside the viewer -- is that expected?

The point cloud also takes a short while to load on my end -- can we speed that up later somehow?

rtf_vis_tool/src/Components/gtsfm_graph/gray_nodes.js
```diff
@@ -53,77 +53,72 @@ const gray_node_list = [
...
-        topOffset: 70,
-        leftOffset: 47
+        topOffset: 69,
+        leftOffset: 22
     },
     {
         text: 'SFMResult as files',
-        topOffset: 80,
-        leftOffset: 40
+        topOffset: 69,
```
johnwlambert, line 81: should be 'GtsfmData as files' i believe per our graph

rtf_vis_tool/src/Components/LandingPageGraph.js
```diff
@@ -130,23 +125,39 @@ function LandingPageGraph() {
...
+            {showBA_PC && <PCViewer title={'Bundle Adjustment Point Cloud'}
+                                    togglePC={toggleBA_PointCloud}
+                                    filePath={'results/ba_output/points3D.txt'}/>}
             {showFS && <FrontendSummary json={fs_json} toggleFS={toggleFrontEndSummaryDisplay}/>}
             {showMVO && <MVOSummary json={mvo_json} toggleMVO={toggleMVOMetrics}/>}
 
             <div className="gtsfm_graph">
 
-                {/* Render all Gray and Blue Nodes (43 combined). */}
+                {/* Render basic Gray and Blue Nodes (41 combined). */}
```
johnwlambert, line 139: just curious -- why 41 instead of 43 now?

rtf_vis_tool/src/Components/LandingPageGraph.js
```diff
@@ -105,22 +100,22 @@ function LandingPageGraph() {
...
 
-    function toggleRotSummaryDisplay(showDisplay) {
-        /*Toggles the display of the rotation averaging metrics.
+    function toggleBA_PointCloud(showDisplay) {
+        /*Toggles the display of the Bundle Adjustment Point Cloud.
 
         Args:
             showDisplay (boolean): Sets the display to be shown or not.
         */
-        setShowRSS(showDisplay);
```
johnwlambert, line 123: what does RSS stand for? and did we remove the `toggleRotSummaryDisplay` function?

johnwlambert:
Looks great, Adi. Left a few minor comments

rtf_vis_tool/src/Components/LandingPageGraph.js
adisingh50: I ended up removing it because it was an old variable that wasn't needed anymore. RSS just stood for "rotation success summary", but now that's being shown when you click on the plate.

rtf_vis_tool/src/Components/LandingPageGraph.js
adisingh50: So I am still rendering 43 nodes total. But the 2 that are not rendered in the for loop are the nodes `Data Association GtsfmData` and `Bundle Adjustment GtsfmData`. I wasn't able to get these two render within the for loop since they have the custom functionality of displaying the 2 point clouds. So 41 are rendered in the generic loop + 2 nodes that display custom point clouds.

johnwlambert: cool -- maybe you can just add that as an inline comment?

adisingh50: yup.

adisingh50:
I fixed all the warnings that were popping up when typing `npm start`.

rtf_vis_tool/src/Components/frustum_classes/se3.js
```diff
@@ -27,10 +27,8 @@ class SE3 {
             for (var col = 0; col < 3; col++) {
                 this.transform_matrix.set(row, col, rotation.get(row,col));
             }
-        }
 
-        //set last column to translation array
-        for (var row = 0; row < 3; row++) {
+            //set last column to translation array
```
johnwlambert, line 31: are we able to set the whole row at once now? we don't need the for loop?

rtf_vis_tool/src/Components/PCViewer.js
```diff
@@ -75,34 +73,12 @@ function PCViewer(props) {
                 <PointMesh  
                     position={[pointArr[1], pointArr[2], pointArr[3]]}  
                     color={`rgb(${pointArr[4]}, ${pointArr[5]}, ${pointArr[6]})`} 
-                    size={pointSizeArr}/>
+                    size={[pointRadius]}/>
```
johnwlambert, line 76: cool -- so we don't need the `updatePointSizes` function anymore?

rtf_vis_tool/src/Components/PCViewer.js
```diff
@@ -61,11 +61,9 @@ function PCViewer(props) {
             arrStringPoints.shift();
         }
 
-        /* Variable pointCloudRaw is an (N x 6) array, with the first 3 entries as (x,y,z) and the last
+        /* Variable arrNumPoints is an (N x 6) array, with the first 3 entries as (x,y,z) and the last
```
johnwlambert, line 64: naming it `arrNumPoints` sounds like it's an array. but it's an int, i believe. How about `num_points`? or `n_points`?

rtf_vis_tool/src/Components/CoordinateGrid.js
```diff
@@ -32,7 +32,7 @@ function getGridLines(lowerGridBound = -10, upperGridBound = 10) {
     }
 
     var finalLineGridSet = []
-    coordPairSet.map(coordPair => {
+    coordPairSet.forEach(coordPair => {
```
johnwlambert, line 35: out of curiosity, whats the difference between `map` and `forEach`?

rtf_vis_tool/src/Components/frustum_classes/se3.js
johnwlambert: before we were thinking that wasn't possible in `numjs` right?

rtf_vis_tool/src/Components/PCViewer.js
adisingh50: nope. the `useEffect` function will now run every time `pointRadius` is changed in the slider, meaning `loadCOLMAPPointCloud` will update the point sizes. So `updatePointSizes` isn't neccessary anymore.

rtf_vis_tool/src/Components/CoordinateGrid.js
adisingh50: `map` requires a return type, but in our case, we don't want anything returned in this for loop. So `forEach` is the more appropriate function to use, and the one I should have been using from the start.

rtf_vis_tool/src/Components/frustum_classes/se3.js
adisingh50: so unfortunately, we can still only edit matrices one element at a time. We cannot set the whole row at once. The change I made was so that I didn't use a second for loop to add the translation matrix elements.

rtf_vis_tool/src/Components/PCViewer.js
adisingh50: So actually `arrNumPoints` is supposed to be an (N x 6) array, but the name I have for it is actually misleading. I'll rename it to `arr_points`.

rtf_vis_tool/src/Components/frustum_classes/se3.js
johnwlambert: i think this is outside the 2nd loop though, right? so `row` is just defaulting to the last value from the previous loop?

johnwlambert: do we have a unit test for this constructor, to check that the transform matrix is set correctly?

e.g. for R= 3x3 identity matrix, and translation=[1,2,3]?

adisingh50: image

So for this code in the se3 constructor, you are right, `row` is just defaulting to the last value from the previous loop because it's setting (`row`, 3) to an element in the translation vector. So still, I'm restricted to only setting one element of the matrix at a time.

adisingh50: I'll work on a unit test for se3 constructor so we can make sure the transform matrix is set correctly.

johnwlambert: oh i see, i was reading it wrong. thanks

johnwlambert: cool, a unit test can go in another PR, i'll approve now

johnwlambert (approved):

PR #158, opened 2021-04-22
Adding smart union of tracks by checking reprojection errors

ayushbaid:


ayushbaid:
# Metrics

| Metric | Smart DA | Master |
| ------- | ------------ | ----------- |
| Input number of tracks | 9406 | 9103 |
| Input avg track length | 5.458 | 5.278 |
| output number of tracks | 8960 | 8966 |
| output avg. track length | 3.9 | 5.23 |
| final number of tracks | 8870 | 8927 |

johnwlambert:
Thanks for the PR, Ayush. How are the multi-view optimizer metrics affect (rotation and translation error?)

It looks like the avg track length is hurt (-25%), which might offset the offset the gains in # of tracks (+3.3%). Our current COLMAP target is 11k tracks with an average of 6.44 track length.

How does it affect avg reprojection error, and runtime?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -299,3 +284,102 @@ def extract_measurements(self, track: SfmTrack2d) -> Tuple[CameraSetCal3Bundler,
...
+
+        # initializer of 3D landmark for each track
+        point3d_initializer_simple = Point3dInitializer(
+            self.track_camera_dict,
+            TriangulationParam.NO_RANSAC,
+            reproj_error_thresh=1e6,  # just a large number, wont be used
+        )
+
+        if track_2d.number_measurements() == 2:
+            # with just two measurements, there is no iterative processing
```
akshay-krishnan, line 314: "iterative processing" is too generic, cant understand what it means.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -299,3 +284,102 @@ def extract_measurements(self, track: SfmTrack2d) -> Tuple[CameraSetCal3Bundler,
...
+
+            if track_2d.measurement(0).i != track_2d.measurement(1).i:
```
akshay-krishnan, line 316: is this check necessary? :O

akshay-krishnan (commented):
I think its more dangerous to have incorrect measurements in a track than to have shorter tracks. If a long track with all correct measurements breaks into two as a result of this, its fine. It will be treated as two tracks, and we will get 2 landmarks at the same location. We should look into merging these later.

johnwlambert:
Hi @ayushbaid, can you pull in master to this branch so we can check the latest metrics?

ayushbaid:
Closing this PR till we do further analysis of data association.

PR #165, opened 2021-04-26
Add colmap loader for Skydio Crane Mast dataset and Lorraine tree dataset

johnwlambert:
We don't have a proper loader for the Tree dataset. It is not an Olsson dataset.

However, we do have the COLMAP output for the Skydio Crane Mast and Tree datasets, so we can get intrinsics and "ground truth" pose from there, at the moment.


gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,137 @@
...
+        """Initializes to load from a specified folder on disk.
+
+        Args:
+            colmap_files_dirpath: path to directory containing COLMAP-exported data, with images.txt
+                and cameras.txt files
+            images_dir: path to directory containing images files
+            use_gt_intrinsics: whether to use ground truth intrinsics
+            use_gt_extrinsics: whether to use ground truth extrinsics
+            max_frame_lookahead: if images were sequentially captured, maximum number
+               of frames to consider for matching/co-visibility. Defaults to 1, i.e.
```
ayushbaid, line 44: I dont get why default value of 1 means that data is not sequentially captured. In such a case, wont we try to match an image with every other image.

gtsfm/runner/run_scene_optimizer_colmaploader.py
```diff
@@ -0,0 +1,62 @@
...
+DATA_ROOT = Path(__file__).resolve().parent.parent.parent / "tests" / "data"
+
+logger = logger_utils.get_logger()
+
+
+def run_scene_optimizer(args) -> None:
+    """ """
+    with initialize_config_module(config_module="gtsfm.configs"):
+        # config is relative to the gtsfm module
+        cfg = compose(config_name="default_lund_door_set1_config.yaml")
```
ayushbaid, line 22: maybe rename the default loader?

gtsfm/runner/run_scene_optimizer_colmaploader.py
```diff
@@ -0,0 +1,62 @@
+import argparse
```
ayushbaid, line 1: maybe have a config for the loader too, instead of creating copies of the file?

ayushbaid: or an if-else structure where we take the name of the loader from the command line args?

gtsfm/utils/io.py
```diff
@@ -116,6 +118,42 @@ def read_bal(file_path: str) -> GtsfmData:
...
+    Returns:
+        calibrations: calibration object for each camera
+    """
+    if not Path(fpath).exists():
+        return None
+
+    with open(fpath, "r") as f:
+        lines = f.readlines()
+
+    num_cams = int(lines[2].replace('# Number of cameras: ', '').strip())
```
ayushbaid, line 138: can we not just seek to the position instead of replacing?

gtsfm/utils/io.py
```diff
@@ -116,6 +118,42 @@ def read_bal(file_path: str) -> GtsfmData:
...
+def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
+    """Read camera calibrations from a COLMAP-formatted cameras.txt file.
+
+    Reference: https://colmap.github.io/format.html#cameras-txt
+
+    Args:
+        fpaths: path to cameras.txt file
+
+    Returns:
+        calibrations: calibration object for each camera
```
ayushbaid, line 130: we skip writing the name for return arguments

gtsfm/utils/io.py
```diff
@@ -116,6 +118,42 @@ def read_bal(file_path: str) -> GtsfmData:
...
+    """
+    if not Path(fpath).exists():
+        return None
+
+    with open(fpath, "r") as f:
+        lines = f.readlines()
+
+    num_cams = int(lines[2].replace('# Number of cameras: ', '').strip())
+    # should have one line per camera
+    assert len(lines[3:]) == num_cams
```
ayushbaid, line 140: maybe `assert len(lines)-3==num_cams`. Seems too inefficient to slice a list just to check its length.

gtsfm/utils/io.py
```diff
@@ -116,6 +118,42 @@ def read_bal(file_path: str) -> GtsfmData:
...
+    
+    calibrations = []
+    for line in lines[3:]:
+
+        cam_params = line.split()
+        # Note that u0 is px, and v0 is py
+        cam_id, model, img_w, img_h, fx, u0, v0 = cam_params[:7]
+        img_w, img_h, fx, u0, v0 = int(img_w), int(img_h), float(fx), float(u0), float(v0)
```
ayushbaid, line 148: maybe not read image width and height if we dont use it?

gtsfm/utils/io.py
```diff
@@ -153,6 +191,34 @@ def write_cameras(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) ->
...
+    with open(fpath, "r") as f:
+        lines = f.readlines()
+
+    wTi_list = []
+    img_fnames = []
+    # ignore first 4 lines of text -- they are a description of the file format
+    for line in lines[4::2]:
+        i, qw, qx, qy, qz, tx, ty, tz, i, img_fname = line.split()
+        wRi = Rot3(float(qw), float(qx), float(qy), float(qz))
+        wTi = Pose3(wRi, np.array([float(tx),float(ty),float(tz)]))
```
ayushbaid, line 215: `np.array([tx, ty, tz], dtype=np.float64)`?

johnwlambert: thanks -- yes this works and is cleaner

johnwlambert: updated

gtsfm/utils/io.py
johnwlambert: I think we may want it later; I didn't need it for now, so didn't return it.

gtsfm/utils/io.py
johnwlambert: good suggestion -- updated

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+            index: the index to fetch.
+
+        Raises:
+            IndexError: if an out-of-bounds image index is requested.
+
+        Returns:
+            Image: the image at the query index.
+        """
+
+        if index < 0 or index > self.__len__():
```
akshay-krishnan, line 118: Index >= self._ _ len_ _() right?

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+            raise IndexError("Image index is invalid")
+
+        img = io_utils.load_image(self._image_paths[index])
+        img = img_utils.resize_image(img, new_height=img.height//self._downsample_factor, new_width=img.width//self._downsample_factor)
```
akshay-krishnan, line 122: If this usage is common, I’d have the utils accept the scale as an argument.

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+        """
+        if not self._use_gt_extrinsics:
+            return None
+
+        wTi = self._wTi_list[index]
+        return wTi
+
+
+    def validate_pair(self, idx1: int, idx2: int) -> bool:
+        """Checks if (idx1, idx2) is a valid pair.
```
akshay-krishnan, line 169: I think that is obvious from the method name, it would be interesting to know what a valid pair is, if that can be explained briefly. 
Also `validate_pair` sounds like its doing something, maybe `is_valid_pair` is a better name. Ignore if it should be changed in the base class as well.

gtsfm/runner/run_scene_optimizer.py
```diff
@@ -42,5 +44,16 @@ def run_scene_optimizer() -> None:
...
-
+    parser = argparse.ArgumentParser(description="GTSFM with intrinsics and image names stored in COLMAP-format")
+    parser.add_argument(
+        "--dataset_root", type=str, default=os.path.join(DATA_ROOT, "set1_lund_door"), help=""
+    )
+    parser.add_argument(
+        "--image_extension", type=str, default="JPG", help=""
+    )
+    parser.add_argument(
+        "--max_frame_lookahead", type=int, default=20, help="maximum number of consecutively ordered frames to consider for matching/co-visibility"
```
akshay-krishnan, line 55: “Maximum number of consecutive frames” is probably simpler. Are we losing some meaning then?

gtsfm/utils/images.py
```diff
@@ -117,6 +117,29 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
...
+#     else:
+#         new_w = max_resolution
+#         scale = new_w / float(w)
+#         new_h = int(h * scale)
+
+#     downsampled_img = resize_image(img, new_height=new_h, new_width=new_w)
+#     downsample_factor = orig_h / downsampled_img.width
+#     target_h = downsampled_img.height
+#     target_w = downsampled_img.width
+#     return downsample_factor, target_h, target_w
```
akshay-krishnan, line 141: More commented code ...

gtsfm/utils/io.py
```diff
@@ -116,6 +118,42 @@ def read_bal(file_path: str) -> GtsfmData:
...
+    """
+    if not Path(fpath).exists():
+        return None
+
+    with open(fpath, "r") as f:
+        lines = f.readlines()
+
+    num_cams = int(lines[2].replace("# Number of cameras: ", "").strip())
+    # should have one line per camera
+    # assert len(lines) - 3 == num_cams
```
akshay-krishnan, line 140: Remove if unnecessary?

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -153,5 +154,73 @@ def test_pickleable(self):
             self.fail("Cannot dump rotation averaging object using pickle")
 
 
+class Test1dsfmAllOutliers(unittest.TestCase):
```
akshay-krishnan, line 157: Could you remove this file from this PR?

gtsfm/utils/io.py
```diff
@@ -116,6 +118,42 @@ def read_bal(file_path: str) -> GtsfmData:
     return gtsfm_data
 
 
+def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
+    """Read camera calibrations from a COLMAP-formatted cameras.txt file.
```
akshay-krishnan, line 122: Can we have unit tests for these methods?

akshay-krishnan (commented):
Nice work John, I think you pushed some changes while I was reviewing, so feel free to mark obsolete comments are fixed.

gtsfm/runner/run_scene_optimizer.py
johnwlambert: thanks -- updated

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -118,7 +117,7 @@ def run(
         # transforming the result to the list of Point3
         wti_list = [None] * num_images
         for i in range(num_images):
-            if wRi_list[i] is not None:
+            if wRi_list[i] is not None and wti_values.exists(i):
```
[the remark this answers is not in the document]
ayushbaid: Why? This goes in a different PR?

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -48,6 +48,12 @@ def run(self, initial_data: GtsfmData) -> GtsfmData:
         logger.info(
             f"Input: {initial_data.number_tracks()} tracks on {len(initial_data.get_valid_camera_indices())} cameras\n"
         )
+        if initial_data.number_tracks() == 0 or len(initial_data.get_valid_camera_indices()) == 0:
+            # no cameras or tracks to optimize, so bundle adjustment is not possible
+            logger.info(
```
ayushbaid, line 53: Is `error` log more appropriate?

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+
+        Args:
+            colmap_files_dirpath: path to directory containing COLMAP-exported data, with images.txt
+                and cameras.txt files
+            images_dir: path to directory containing images files
+            use_gt_intrinsics: whether to use ground truth intrinsics
+            use_gt_extrinsics: whether to use ground truth extrinsics
+            max_frame_lookahead: if images were sequentially captured, maximum number
+               of consecutive frames to consider for matching/co-visibility. Defaults to 1, i.e. 
+               assuming data is not sequentially captured.
```
ayushbaid, line 47: I do not get this. If the frames are not in a sequence, even the value 1 doesn't make sense.

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+            max_resolution: integer representing maximum length of image's short side
```
ayushbaid, line 48: `max_resolution` should be the property of `LoaderBase`?

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+               e.g. for 1080p (1920 x 1080), max_resolution would be 1080
+        """
+        self._use_gt_intrinsics = use_gt_intrinsics
+        self._use_gt_extrinsics = use_gt_extrinsics
+        self._max_frame_lookahead = max_frame_lookahead
+        self._max_resolution = max_resolution
+        #import pdb; pdb.set_trace()
```
ayushbaid, line 55: remove this

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+        for img_fname in img_fnames:
+            img_fpath = os.path.join(images_dir, img_fname)
+            from pathlib import Path
+            if not Path(img_fpath).exists():
+                continue
+            self._image_paths.append(img_fpath)
+
+
+        self._num_imgs = len(self._image_paths)
+        print(f"Has {self._num_imgs} images")
```
ayushbaid, line 82: print or log?

gtsfm/loader/colmap_loader.py
ayushbaid: `len(self)` looks cleaner to me.

gtsfm/loader/colmap_loader.py
ayushbaid: Yep. Should go in utils.

gtsfm/loader/colmap_loader.py
ayushbaid: It would involve a change in the base class too, but it sounds better.

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+        """Checks if (idx1, idx2) is a valid pair.
+
+        Args:
+            idx1: first index of the pair.
+            idx2: second index of the pair.
+
+        Returns:
+            validation result.
+        """
+        return idx1 < idx2 and abs(idx1 - idx2) <= self._max_frame_lookahead
```
ayushbaid, line 178: no need to have `abs` because `idx1` is less than `idx2`.

gtsfm/utils/io.py
```diff
@@ -227,26 +335,21 @@ def write_points(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> N
...
-
 def save_track_visualizations(
     tracks_2d: List[SfmTrack2d],
     images: List[Image],
     save_dir: str,
     viz_patch_sz: int = 100,
 ) -> None:
-    """
-    """
+    """"""
```
ayushbaid, line 344: write documentation

ayushbaid (changes requested):

gtsfm/utils/io.py
johnwlambert: removed, thanks

gtsfm/utils/io.py
johnwlambert: just added unit tests for them

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+        # import pdb; pdb.set_trace()
+        # self._wTi_list = [self._wTi_list[i] for i in [0, 8]]
+        # img_fnames = [img_fnames[i] for i in [0,8]] 
+        # self._calibrations = [self._calibrations[i] for i in [0,8]]
+
+        # preserve COLMAP ordering of images
+        self._image_paths = []
+        for img_fname in img_fnames:
+            img_fpath = os.path.join(images_dir, img_fname)
+            from pathlib import Path
```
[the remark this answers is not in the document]
johnwlambert: thanks -- fixed

gtsfm/runner/run_scene_optimizer_colmaploader.py
johnwlambert: i think each loader will have different set of required arguments (at least the way I've defined it). So I think it makes sense for them to be separate. Can you take a look at the 3 runner files in this branch and check it out?

gtsfm/loader/colmap_loader.py
johnwlambert: Thanks -- I agree that `is_valid_pair()` is more appropriate for the method name. I've updated it.

gtsfm/utils/io.py
johnwlambert: thanks, added

gtsfm/loader/colmap_loader.py
johnwlambert: technically we would need to do `idx2 - idx1` instead, then. I think `abs()` makes it easier for the reader so they dont have to think about negative number edge cases at all

gtsfm/utils/io.py
johnwlambert: thanks, updated

gtsfm/loader/colmap_loader.py
johnwlambert: Just to be clear -- we don't have a utils function for this at the present. We have to be a little explicit here since the requested aspect ratio may not be possible.

So we may want different scaling factors for focal length for each axis.

gtsfm/averaging/translation/averaging_1dsfm.py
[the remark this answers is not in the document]
johnwlambert: i think the other will get merged first : - )

gtsfm/bundle/bundle_adjustment.py
johnwlambert: indeed -- updated

gtsfm/loader/colmap_loader.py
johnwlambert: maybe that can go in a future PR? is it something that we want universally?

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+
+        self._wTi_list, img_fnames = io_utils.read_images_txt(fpath=os.path.join(colmap_files_dirpath, "images.txt"))
+        self._calibrations = io_utils.read_cameras_txt(fpath=os.path.join(colmap_files_dirpath,"cameras.txt"))
+        if len(self._calibrations) == 1:
+            # shared calibration!
+            import pdb; pdb.set_trace()
+            self._calibrations = [self._calibrations] * len(img_fnames)
+
+        if self._calibrations is None or len(img_fnames) != len(self._calibrations):
+            self._use_gt_intrinsics = False
```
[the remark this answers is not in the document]
johnwlambert: i think i understand what you meant -- updated

gtsfm/loader/colmap_loader.py
ayushbaid: True.

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,183 @@
...
+
+        Args:
+            colmap_files_dirpath: path to directory containing COLMAP-exported data, with images.txt
+                and cameras.txt files
+            images_dir: path to directory containing images files
+            use_gt_intrinsics: whether to use ground truth intrinsics. If COLMAP calibration is
+               not found on disk, then use_gt_intrinsics will be set to false automatically.
+            use_gt_extrinsics: whether to use ground truth extrinsics
+            max_frame_lookahead: if images were sequentially captured, maximum number
+               of consecutive frames to consider for matching/co-visibility. Defaults to 1, i.e.
```
ayushbaid, line 56: 1 doesn't imply data was sequentially captured. In fact, any value of max_frame_lookahead assumes data is sequentially captured.

johnwlambert: How about "1 implies data was sequentially captured by a fast moving camera with a low frame rate"?

ayushbaid: I think we do not need to add any comment. Your statement is true in general, but there might be other reasons why we use the value 1.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +40,52 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
     return [aRb.compose(bRi) for bRi in bRi_list]
 
 
+def align_poses_sim3_wrapper(aTi_list: List[Optional[Pose3]], bTi_list: List[Optional[Pose3]]) -> List[Optional[Pose3]]:
```
ayushbaid, line 43: can we make a separate PR for this?

gtsfm/utils/images.py
```diff
@@ -118,6 +118,45 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
     return Image(resized_value_array)
 
 
+def get_downsample_factor_per_axis(img: Image, max_resolution: int) -> Tuple[float, int, int]:
```
ayushbaid, line 121: This function returns 4 elements.

gtsfm/utils/io.py
```diff
@@ -116,6 +121,43 @@ def read_bal(file_path: str) -> GtsfmData:
...
+    Reference: https://colmap.github.io/format.html#cameras-txt
+
+    Args:
+        fpaths: path to cameras.txt file
+
+    Returns:
+        calibration object for each camera, or None if requested file is non-existent
+    """
+    if not Path(fpath).exists():
+        logger.info("%s does not exist", fpath)
```
ayushbaid, line 136: info or error?

gtsfm/utils/images.py
johnwlambert: thanks -- should be fixed now

gtsfm/loader/colmap_loader.py
johnwlambert: is this better?
```
            max_frame_lookahead: maximum number of consecutive frames to consider for
                matching/co-visibility. Any value of max_frame_lookahead less than the size of
                the dataset assumes data is sequentially captured
```

gtsfm/utils/io.py
johnwlambert: Should be fine, since we can fall back to just using the images in the dir, and using EXIF instead. So i think info is fine, not error

gtsfm/utils/io.py
```diff
@@ -153,6 +195,44 @@ def write_cameras(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) ->
...
+        quaternion and T is the translation vector. The local camera coordinate system
+        of an image is defined in a way that the X axis points to the right, the Y axis
+        to the bottom, and the Z axis to the front as seen from the image."
+
+    Args:
+        fpath: path to images.txt file
+
+    Returns:
+        wTi_list: list of camera poses for each image, or None if file path invalid
+        img_fnames: name of image file, for each image, or None if file path invalid
```
ayushbaid, line 213: we have not been writing the variable names in the return section of the docstring.

gtsfm/utils/io.py
```diff
@@ -180,13 +260,62 @@ def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
...
+
+        point = [float(x), float(y), float(z)]
+        point_cloud += [point]
+        rgb += [(int(r), int(g), int(b))]
+
+    point_cloud = np.array(point_cloud)
+    rgb = np.array(rgb).astype(np.uint8)
+
+    assert point_cloud.shape[0] == expected_num_pts
+    assert rgb.shape[0] == expected_num_pts
```
ayushbaid, line 317: should we remove it or keep it?

gtsfm/utils/io.py
ayushbaid: Cool. Thanks!

gtsfm/loader/yfcc_imb_loader.py
```diff
@@ -100,7 +100,7 @@ def get_camera_intrinsics(self, index: int) -> Optional[Cal3Bundler]:
         Returns:
             intrinsics for the given camera.
         """
-        if index < 0 or index > self.__len__():
+        if index < 0 or index > len(self):
```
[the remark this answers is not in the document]
akshay-krishnan: I believe len returns the size, hopefully I’m not wrong.

akshay-krishnan (changes requested):

gtsfm/utils/geometry_comparisons.py
johnwlambert: sure, done

gtsfm/utils/images.py
```diff
@@ -118,6 +118,45 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
     return Image(resized_value_array)
 
 
+def get_downsample_factor_per_axis(img: Image, max_resolution: int) -> Tuple[float, float, int, int]:
```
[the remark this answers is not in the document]
johnwlambert: We could, but the COLMAP loader is currently depending on it, so I might prefer otherwise

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,183 @@
...
+import gtsfm.utils.logger as logger_utils
+from gtsfm.common.image import Image
+from gtsfm.loader.loader_base import LoaderBase
+
+
+logger = logger_utils.get_logger()
+
+
+class ColmapLoader(LoaderBase):
+    """Simple loader class that reads a dataset with metadata formatted in the COLMAP style.
```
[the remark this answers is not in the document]
johnwlambert: Not ground truth — just COLMAP output files. Metadata includes camera output calibrations, image file names. I’ll clarify this.

gtsfm/runner/run_scene_optimizer_colmaploader.py
```diff
@@ -0,0 +1,65 @@
...
+
+if __name__ == "__main__":
+
+    parser = argparse.ArgumentParser(description="GTSFM with intrinsics and image names stored in COLMAP-format")
+    parser.add_argument(
+        "--images_dir", type=str, default="/Users/johnlambert/Downloads/Aishwarya_April21/10_test/images", help=""
+    )
+    parser.add_argument(
+        "--colmap_files_dirpath", type=str, default="/Users/johnlambert/Downloads/Aishwarya_April21", help=""
+    )
```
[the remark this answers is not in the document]
johnwlambert: Sure — good point

gtsfm/loader/olsson_loader.py
```diff
@@ -102,7 +104,7 @@ def get_image(self, index: int) -> Image:
             Image: the image at the query index.
         """
 
-        if index < 0 or index > self.__len__():
+        if index < 0 or index > len(self):
```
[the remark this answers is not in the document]
johnwlambert: You’re right — thanks

gtsfm/utils/io.py
johnwlambert: What would you prefer? Raise a RuntimeError?

gtsfm/loader/olsson_loader.py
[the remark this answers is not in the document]
johnwlambert: fixed, should be updated now

gtsfm/loader/yfcc_imb_loader.py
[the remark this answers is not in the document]
johnwlambert: fixed, should be updated now

gtsfm/runner/run_scene_optimizer_colmaploader.py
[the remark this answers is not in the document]
johnwlambert: should be fixed now

gtsfm/utils/io.py
johnwlambert: for a single return variable, i see Google python style doesn't list the variable name

but for many return variables, i think listing the name makes it easier to read

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+
+        # import pdb; pdb.set_trace()
+        # # read one image, to check if we need to downsample the images
+        # img = io_utils.load_image(self._image_paths[0])
+
+        # sample_h, sample_w = img.height, img.width
+        # if min(sample_h, sample_w) > self._max_resolution:
+        #     self._downsample_factor, self.target_h, self.target_w = img_utils.get_exact_downsample_factor(img, self._max_resolution)
+        # else:
+        #     self._downsample_factor = 1
```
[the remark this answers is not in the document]
johnwlambert: removed

gtsfm/loader/colmap_loader.py
johnwlambert: fixed

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,179 @@
...
+            import pdb; pdb.set_trace()
+            self._calibrations = [self._calibrations] * len(img_fnames)
+
+        if self._calibrations is None or len(img_fnames) != len(self._calibrations):
+            self._use_gt_intrinsics = False
+
+        # import pdb; pdb.set_trace()
+        # self._wTi_list = [self._wTi_list[i] for i in [0, 8]]
+        # img_fnames = [img_fnames[i] for i in [0,8]] 
+        # self._calibrations = [self._calibrations[i] for i in [0,8]]
```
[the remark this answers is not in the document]
johnwlambert: removed

gtsfm/loader/colmap_loader.py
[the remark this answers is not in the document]
akshay-krishnan: Not sure what exactly this class is doing. Correct me if I’m wrong, I think it accepts a path to Colmap output directory, and a different images directory, and uses the colmap output as ground truth? 
All of this is not obvious from “metadata formatted in colmap style”.

gtsfm/loader/colmap_loader.py
```diff
@@ -0,0 +1,187 @@
...
+    - Extrinsics data (optional): images.txt (see https://colmap.github.io/format.html#images-txt)
+
+    If explicit intrinsics are not provided, the exif data will be used.
+    """
+
+    def __init__(
+        self,
+        colmap_files_dirpath: str,
+        images_dir: str,
+        use_gt_intrinsics: bool = True,
```
akshay-krishnan, line 41: If true, does it use colmap output as ground truth? In that case, I think it should be called use_colmap_result_as_gt.

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: Remove changes from here?

gtsfm/utils/images.py
akshay-krishnan, line 121: Nit: maybe this function should accept only the dimensions of the image?

gtsfm/utils/images.py
```diff
@@ -118,6 +118,45 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
...
+        new_h = max_resolution
+        # compute scaling that will be applied to original image
+        scale = new_h / float(h)
+        new_w = int(w * scale)
+    else:
+        new_w = max_resolution
+        scale = new_w / float(w)
+        new_h = int(h * scale)
+
+    downsampled_img = resize_image(img, new_height=new_h, new_width=new_w)
```
akshay-krishnan, line 150: I don’t get why we are doing the resizing here..

gtsfm/utils/images.py
johnwlambert: I was thinking about that, but the way I've implemented it, we need to feed a dummy image through the resize function to see how OpenCV will round the dims. So rather than make a dummy image, I figured a single argument would give it to us, along with the dimensions.

gtsfm/utils/images.py
johnwlambert: So the main idea is that OpenCV will do some rounding on its end -- and unless we dig into its source to see its protocol, I can't predict it

```
def test_get_downsample_factor_per_axis() -> None:
    """Ensure that max resolution constraint is met, when downsampling image.
    Resize a 700x1500 image, so that the shorter image side is at most 600 px.
    """
    img = Image(np.zeros((700, 1500, 3), dtype=np.uint8))
    max_resolution = 600
    downsample_u, downsample_v, new_h, new_w = image_utils.get_downsample_factor_per_axis(img, max_resolution)

    # 6/7 will not give a clean integer division
    assert new_h == 1.167
    assert new_w == 1.166
    assert downsample_u == 600
    assert downsample_v == 1285
```

gtsfm/loader/colmap_loader.py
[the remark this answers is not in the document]
johnwlambert: let me know if this is clearer now -- ii updated it now

gtsfm/loader/colmap_loader.py
johnwlambert: the data could be actual ground truth, just in the colmap format. So I think it's better this way. It's up to the user if we want to use COLMAP output as pseudo-ground truth.

For example, we save our GTSFM output in the COLMAP-format.

gtsfm/utils/metrics.py
```diff
@@ -193,3 +184,19 @@ def get_rotations_translations_from_poses(
     metrics["translation_averaging_distance"] = compute_translation_distance_metrics(wti_aligned_list, gt_wti_list)
     metrics["translation_to_direction_angle_deg"] = compute_translation_angle_metrics(i2Ui1_dict, wTi_aligned_list)
     return metrics
+
+
```
johnwlambert, line 188: moved this callback to be an outright function, since i'll use it elsewhere in a future PR

gtsfm/utils/images.py
akshay-krishnan: I understand this finally after our offline discussion. The function returns the scaling factors and the new height and width. It also calls the OpenCV resize function to check that this new height and width is actually "feasible". 

Firstly, I dont think the function documentation states this. But more importantly, I dont think it is good design to have a computationally expensive check inside an otherwise computationally cheap util function. Maybe the name and outputs of the function can be changed? 

Curious to know what @ayushbaid thinks to see if this is just my personal opinion.

gtsfm/utils/images.py
```diff
@@ -118,6 +118,45 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
...
+    shorter_size = min(h, w)
+    if shorter_size == h:
+        new_h = max_resolution
+        # compute scaling that will be applied to original image
+        scale = new_h / float(h)
+        new_w = int(w * scale)
+    else:
+        new_w = max_resolution
+        scale = new_w / float(w)
+        new_h = int(h * scale)
```
akshay-krishnan, line 148: This would work even if max_resolution is greater than the input image dimensions right? Can we add a check to make sure it is not? I dont think it is expected to work, since the function mentions "downsampling".

gtsfm/utils/images.py
```diff
@@ -118,6 +118,45 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
...
+    Resizing an image by a specific downsample ratio may not be possible due to lack of a clean
+    divisor. However, we can still determine the exact downsampling ratio.
+
+    Args:
+        img: image to be downsampled
+        max_resolution: integer representing maximum length of image's short side
+            e.g. for 1080p (1920 x 1080), max_resolution would be 1080
+
+    Returns:
+        downsample_u: downsampling factor for u coordinate
```
akshay-krishnan, line 132: nit: the variable name does not seem to reflect it is a factor..

gtsfm/utils/images.py
```diff
@@ -118,6 +118,79 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
...
+
+
+def get_downsampling_factor_per_axis(img_h: int, img_w: int, max_resolution: int) -> Tuple[float, float, int, int]:
+    """Resize an image such that the shorter image side is *less than or equal to* the max_resolution.
+
+    This will always downsample the image or leave it intact.
+
+    Note: Different from COLMAP's `Downsize()`, which instead checks if ANY side is
+        larger than the max resolution.
+        See: https://github.com/colmap/colmap/blob/dev/src/mvs/image.cc#L83
```
akshay-krishnan, line 167: do we need this note?

gtsfm/utils/io.py
akshay-krishnan: I think we should follow a consistent style. :)

akshay-krishnan (approved):

akshay-krishnan:
Thanks @johnwlambert  for updating the rescaling function. I hope you'll agree that the outcome was well worth the extra effort :)

gtsfm/utils/images.py
johnwlambert: i think it is a bit of a nice parallel with our approach vs. other lib, especially if we decide to change it later, so it might be useful to keep in mind/ keep as a note. But not required.

PR #166, opened 2021-04-27
Auto Align Point Cloud and Camera Frustums using SVD

adisingh50:
In this PR, I've added an `ellipsoid.py` file which contains all the helper functions neccessary to align and center the point cloud with respect to the x, y, and z axes. I've also edited `io.py` to use a wrapper function to call the COLMAP export functions (`write_points()`, `write_images()`, and `write_cameras()`) in that order. Finally, I've added a unit test to ensure that the `se3.js` constructor properly creates a pose matrix given rotation and translation matrices. 

image

### DIFF
--- a/gtsfm/utils/ellipsoid.py
+++ b/gtsfm/utils/ellipsoid.py
@@ -5,7 +5,7 @@ Authors: Adi Singh
 """
 
 import numpy as np
-from numpy.linalg import eig, inv, norm, svd
+from numpy.linalg import inv, norm, svd
 
 
 def center_point_cloud(point_cloud: np.ndarray) -> np.ndarray:
--- a/gtsfm/utils/ellipsoid.py
+++ b/gtsfm/utils/ellipsoid.py
@@ -50,7 +50,8 @@ def center_point_cloud(point_cloud: np.ndarray) -> Tuple[np.ndarray, np.ndarray]
         point_cloud: N 3D points to center, of shape N x 3.
 
     Returns:
-        The centered point cloud of shape N x 3. Also an array of shape (3.) representing the mean x,y,z coordinates of point cloud.
+        The centered point cloud of shape N x 3. Also an array of shape (3.) representing the mean x,y,z 
+        coordinates of point cloud.
 
     Raises:
         TypeError: if points are not of shape N x 3.
--- a/gtsfm/utils/ellipsoid.py
+++ b/gtsfm/utils/ellipsoid.py
@@ -5,21 +5,55 @@ Authors: Adi Singh
 """
 
 import numpy as np
-from numpy.linalg import inv, norm, svd
+import numpy.linalg
+from scipy.spatial.transform import Rotation as R
+from typing import List, Tuple
+
+from gtsfm.common.gtsfm_data import GtsfmData
 
 
-def center_point_cloud(point_cloud: np.ndarray) -> np.ndarray:
+def transform_point_cloud_wrapper(gtsfm_data: GtsfmData) -> np.ndarray:
+    """Wrapper function for all the functions in ellipsoid.py. Transforms the point cloud contained within gtsfm_Data
+    to be aligned with the x, y, and z axes.
+
+    Args:
+        gtsfm_data: scene data to write to transform.
+
+    Returns:
+        The transformed, aligned point cloud of shape N x 3. Also the mean x,y,z coordinates and alignment rotation
+        matrix.
+    """
+
+    # Iterate through each track to gather a list of 3D points forming the point cloud.
+    point_cloud_list = []
+    num_pts = gtsfm_data.number_tracks()
+
+    for j in range(num_pts):
+        track = gtsfm_data.get_track(j)  # TEMP: might need an extra import for track
+        x, y, z = track.point3()
+        point_cloud_list.append([x, y, z])
+
+    # Transform the point cloud to be aligned with x,y,z axes using SVD.
+    point_cloud = np.array([np.array(p) for p in point_cloud_list])  # point_cloud has shape Nx3
+    points_centered, means = center_point_cloud(point_cloud)
+    points_filtered = remove_outlier_points(points_centered)
+    wuprightRw = get_rotation_matrix(points_filtered)
+    aligned_points = apply_ellipsoid_rotation(wuprightRw, points_centered)
+
+    return aligned_points, means, wuprightRw
+
+
+def center_point_cloud(point_cloud: np.ndarray) -> Tuple[np.ndarray, np.ndarray]:
     """Centers a point cloud using mean values of x, y, and z.
 
     Args:
-        pointCloud: array representing the point cloud, of shape N x 3.
+        point_cloud: N 3D points to center, of shape N x 3.
 
     Returns:
-        points_centered: centered point cloud of shape Nx3.
-        means: the means of the x,y,z coordinates, array of length 3.
+        The centered point cloud of shape N x 3. Also an array of shape (3.) representing the mean x,y,z coordinates of point cloud.
 
     Raises:
-        TypeError: if points list is not 3 dimensional.
+        TypeError: if points are not of shape N x 3.
     """
     if point_cloud.shape[1] != 3:
         raise TypeError("Points list should be 3D")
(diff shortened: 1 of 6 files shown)


gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+
+import numpy as np
+from numpy.linalg import inv, norm, svd
+
+
+def center_point_cloud(point_cloud: np.ndarray) -> np.ndarray:
+    """Centers a point cloud using mean values of x, y, and z.
+
+    Args:
+        pointCloud: array representing the point cloud, of shape N x 3.
```
johnwlambert, line 15: Thanks for the PR, Adi.

pointCloud should be point_cloud

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+
+    Returns:
+        points_centered: centered point cloud of shape Nx3.
+        means: the means of the x,y,z coordinates, array of length 3.
```
johnwlambert, line 19: maybe prefer `mean: array of shape (3.) representing the mean x,y,z, coordinates of point cloud`

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+    if centered_pc.shape[1] != 3:
+        raise TypeError("Point Cloud should be 3 dimensional")
+
+    # find mean and standard deviation of magnitudes of points to determine the threshold magnitude to filter by
+    mags = norm(centered_pc, axis=1)
+    mean_mag = np.mean(mags)
+    std_mag = np.std(mags)
+
+    # z = 1.65 represents the 95th percentile. Assuming a normal distribution of point magnitudes.
+    cutoff_mag = 1.65 * (std_mag) + mean_mag
```
johnwlambert, line 53: one other approach would be `np.percentile(ranges, 95)`. Which is better? i'm not sure

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+        The filtered point cloud of shape M x 3 (M = 0.95*N).
+
+    Raises:
+        TypeError: if centered point cloud is not 3 dimensional.
+    """
+    if centered_pc.shape[1] != 3:
+        raise TypeError("Point Cloud should be 3 dimensional")
+
+    # find mean and standard deviation of magnitudes of points to determine the threshold magnitude to filter by
+    mags = norm(centered_pc, axis=1)
```
johnwlambert, line 48: we generally only import classes, not functions. so `np.linalg.norm()` is actually clearer to me

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+    if point_cloud.shape[1] != 3:
+        raise TypeError("Points list should be 3D")
+
+    means = np.mean(point_cloud, axis=0)
+    points_centered = point_cloud - means
+    return points_centered, means
+
+
+def filter_outlier_points(centered_pc: np.ndarray) -> np.ndarray:
+    """Filters the top 5% of points with greatest distance from origin.
```
johnwlambert, line 33: "Filters out the top..."?

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+    Returns:
+        The rotation matrix, shape 3 x 3, required to align points with the x, y, and z axes.
+
+    Raises:
+        TypeError: if centered point cloud is not 3 dimensional.
+    """
+    if filtered_pc.shape[1] != 3:
+        raise TypeError("Point Cloud should be 3 dimensional")
+
+    u, s, v = svd(filtered_pc, full_matrices=True)
```
johnwlambert, line 75: prefer np.linalg.svd

numpy svd should return U, S, Vt. prefer upper case as are matrices (as we write on paper)

we should express the rotation as a rotation between two frames

e.g. `wuprightRw`

https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md#coordinate-system-conventions

transpose is inverse, so you dont need to invert -- you can just say `wuprightRw = Vt.T`

What does U represent semantically? s would be length of major axes i believe

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
...
+    Raises:
+        TypeError: if rot matrix isn't 3x3 or centered_pc is not of dimension 3
+    """
+    if rot.shape[0] != 3 or rot.shape[1] != 3:
+        raise TypeError("Rotation Matrix should be of shape 3 x 3")
+    if centered_pc.shape[1] != 3:
+        raise TypeError("Point Cloud shoud be 3 dimensional")
+
+    # First apply the rotation matrix to the point cloud. Then swap x and z axes for visual alignment.
+    swap_x_z = np.array([[0, 0, 1], [0, 1, 0], [-1, 0, 0]])
```
johnwlambert, line 100: it just so happens that for our point cloud, z is the longest axis, and so we want it to be the first axis (since vertical scene)

But we could have a horizontal scene -- See Figure 1 of https://openaccess.thecvf.com/content_cvpr_2016/papers/Schonberger_Structure-From-Motion_Revisited_CVPR_2016_paper.pdf

So we need to do like blender, where the user toggles on their end to see which axis they want to see from

gtsfm/utils/io.py
```diff
@@ -182,10 +200,40 @@ def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
...
+
+            center_pose = np.eye(4)
+            center_pose[:3, 3] = -1 * means
+
+            alignment_pose = np.eye(4)
+            alignment_pose[:3, :3] = align_rot
+
+            swap_x_z = np.array([[0, 0, -1, 0], [0, -1, 0, 0], [1, 0, 0, 0], [0, 0, 0, 1]])
+
+            final_pose = swap_x_z @ alignment_pose @ center_pose @ gtsfm_pose
```
johnwlambert, line 221: i'd prefer to keep the IO as pure IO, and just have `export_as_colmap_txt()` generate a new GtsfmData instance, and then make the pure 3 io calls

tests/utils/test_ellipsoid_utils.py
```diff
@@ -0,0 +1,84 @@
...
+import unittest
+
+
+class TestEllipsoidUtils(unittest.TestCase):
+    """Class containing all unit tests for ellipsoid utils."""
+
+    def test_center_point_cloud(self):
+        """Tests the center_point_cloud() function with 5 sample points."""
+
+        sample_points = np.array([[6, 13, 9], [5, 9, 7], [6, 10, 9], [9, 13, 6], [5, 12, 9]])
```
johnwlambert, line 17: could probably choose even simpler points, and add an inline comment showing the mental math

like just 3 points

tests/utils/test_ellipsoid_utils.py
```diff
@@ -0,0 +1,84 @@
...
+    def test_filter_outlier_points_wrong_dims(self):
+        """Tests the filter_point_cloud() function with 5 sample points of 2 dimensions."""
+
+        sample_points = np.array([[6, 13], [5, 9], [6, 10], [9, 13], [5, 12]])
+        self.assertRaises(TypeError, ellipsoid_utils.filter_outlier_points, sample_points)
+
+    def test_get_rotation_matrix(self):
+        """Tests the get_rotation_matrix() function with 5 sample points."""
+
+        sample_points = np.array([[1, 1, 0], [2, 2, 0], [-1, -1, 0], [-2, -2, 0], [-1, 1, 0], [1, -1, 0]])
```
johnwlambert, line 53: maybe an explanation of what these points represent? you could make 5 points -- 4 points in a cross (like biblical cross shape), and then 1 point for height in the last dim -- and see what the ellpse looks like?

gtsfm/utils/io.py
```diff
@@ -153,13 +169,15 @@ def write_cameras(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) ->
             f.write(f"{i} {camera_model} {image_width} {image_height} {fx} {u0} {v0} {k1} {k2}\n")
 
 
-def write_images(gtsfm_data: GtsfmData, save_dir: str) -> None:
+def write_images(gtsfm_data: GtsfmData, means: np.ndarray, align_rot: np.ndarray, save_dir: str) -> None:
```
ayushbaid, line 172: I dont think we should have a transform in the write function. We should have a different function to transform into a new `GtsfmData` object.

gtsfm/utils/io.py
```diff
@@ -213,19 +265,39 @@ def write_points(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> N
...
+            x, y, z = track.point3()
+            point_cloud_list.append([x, y, z])
+
+        # Transform the point cloud to be aligned with x,y,z axes using SVD.
+        point_cloud = np.array([np.array(p) for p in point_cloud_list])  # point_cloud has shape Nx3
+        points_centered, means = ellipsoid_utils.center_point_cloud(point_cloud)
+        points_filtered = ellipsoid_utils.filter_outlier_points(points_centered)
+        rot = ellipsoid_utils.get_rotation_matrix(points_filtered)
+        aligned_points = ellipsoid_utils.apply_ellipsoid_rotation(rot, points_centered)
+
```
ayushbaid, line 281: maybe have the transform function which returns a `GtsfmData`, and the write function just write the object.

rtf_vis_tool/src/Components/PCViewer.js
```diff
@@ -63,16 +63,16 @@ function PCViewer(props) {
 
         /* Variable arr_points is an (N x 6) array, with the first 3 entries as (x,y,z) and the last
            3 entries as (R,G,B). */
-        const arr_points = arrStringPoints.map(point => point.split(" ").map(Number));
+        const arr_points = arrStringPoints.map(point => point.split(" ").map(Number).slice(1,7));
```
ayushbaid, line 66: why is this needed now?

ayushbaid (changes requested):

gtsfm/utils/ellipsoid.py
adisingh50: oh.. I didn't even think to check if np had a percentile function. I'll look into it a bit closer to see if it's cleaner.

gtsfm/utils/ellipsoid.py
adisingh50: ok so then instead of `from numpy.linalg import inv, norm, svd`, I'll use `np.linalg.norm`, `np.linalg.inv`, and `np.linalg.svd`

rtf_vis_tool/src/Components/PCViewer.js
adisingh50: Because before, the array `arr_points` was actually not of length 6. It contained all fields of information shows in the picture below. But now, I've decided to only focus on indices 1 through 7, which are x,y,z and r,g,b.

image

gtsfm/utils/ellipsoid.py
adisingh50: So since I'm applying SVD on a N x 3 matrix, U would be of shape NxN. In the case of the lund door dataset, that means it's around 8000x8000. So I'm not really sure what U represents or if it represents anything meaningful here. S is definitely the length of major axes and V is the normalized direction vectors of said axes.

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,103 @@
+"""Algorithms to fit Ellisoid geometry to 3D points and extract the rotation matrix necessary to align point cloud 
+with x, y, and z axes. Used in React Three Fiber Visualization Tool.
+
+Authors: Adi Singh
+"""
+
+import numpy as np
+from numpy.linalg import inv, norm, svd
```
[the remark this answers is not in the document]
adisingh50: yup, I have changed it to just say `import numpy.linalg`

gtsfm/utils/io.py
adisingh50: sounds good. I'll make some sort of wrapper function in `ellipsoid.py` which contains all this transformation code. That way, `write_points()` within `io.py` can just call the wrapper function and write the object.

gtsfm/utils/io.py
adisingh50: yup. I'll create another function in `ellipsoid.py` which will perform the camera frustum transformations. That way, `write_images()` inside `io.py` can just call that before writing the quaternions and translations to file

gtsfm/utils/ellipsoid.py
adisingh50: actually, i think doing `np.percentile(ranges, 95)` would not be simpler since it only takes the percentile of individual numbers in the array. Whereas in our case, we wish to take the 95th percentile of the magnitudes of each 3d point. So I'll probably leave this `cutoff_mag` code as it is.

johnwlambert: Would it be identical to 3 lines
```python
ranges = np.linalg.norm(point_cloud, axis=1)
cutoff_range = np.percentile(ranges, 95)
point_cloud = point_cloud[ mags < cutoff_range]
```

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,179 @@
...
+
+def get_rotation_matrix(point_cloud: np.ndarray) -> np.ndarray:
+    """Applies SVD on a point cloud. The resulting V contains the rotation matrix required to align the 3 major axes of
+    the point cloud's ellipsoid with the x, y, z coordinate axes.
+
+    Args:
+        point_cloud: point cloud of shape N x 3.
+
+    Returns:
+        The rotation matrix, shape 3 x 3, required to align points with the x, y, and z axes.
```
johnwlambert, line 103: maybe we can define what w and wupright are also.

Maybe `w` and `waligned` are better, since we have no guarantee its upright -- only guaranteed to be aligned to orthonormal axes.

i.e. World and axis-aligned world frames

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,179 @@
...
+
+    Raises:
+        TypeError: if point cloud is not of shape N x 3.
+    """
+    if point_cloud.shape[1] != 3:
+        raise TypeError("Point Cloud should be 3 dimensional")
+
+    U, S, Vt = np.linalg.svd(point_cloud, full_matrices=True)
```
johnwlambert, line 111: you could mention the shapes in an inline comment:
```python
# U is NxN, S has 3 values, and Vt is 3x3
# S represents the scaling (lengths) of each of the three ellipsoid semi-axes
# ref: https://en.wikipedia.org/wiki/Singular_value_decomposition#Rotation,_coordinate_scaling,_and_reflection
```

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,179 @@
...
+def transform_camera_frustums(
+    wti: np.ndarray, wRi_quaternion: np.ndarray, means: np.ndarray, wuprightRw: np.ndarray
+) -> Tuple[List[float], List[float]]:
+    """Transforms the camera frustums in a similar manner as the point cloud.
+
+    Args:
+        wti: camera pose translation, list of length 3.
+        wRi_quaternion: camera pose rotation, list of length 4.
+        means: the mean x,y,z coordinates of point cloud, array of shape (3.).
+        wuprightRw: rotation matrix to align point cloud with x,y,z axes, shape 3 x 3.
```
johnwlambert, line 150: on second thought, here also `walignedRw` seems better

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,179 @@
...
+    """
+
+    tx, ty, tz = wti
+    qw, qx, qy, qz = wRi_quaternion
+    gtsfm_rot = R.from_quat([qx, qy, qz, qw]).as_matrix()
+    gtsfm_tran = np.array([tx, ty, tz])
+
+    gtsfm_pose = np.eye(4)
+    gtsfm_pose[:3, :3] = gtsfm_rot
+    gtsfm_pose[:3, 3] = gtsfm_tran
```
johnwlambert, line 164: we shouldn't name it as pose -- we can make it clearer to the reader per the gtsam conventions
`wTi` means it's the i'th camera's pose in the world frame

gtsfm/utils/ellipsoid.py
```diff
@@ -0,0 +1,179 @@
...
+
+    wcenterTw = np.eye(4)
+    wcenterTw[:3, 3] = -1 * means
+
+    walignRw = np.eye(4)
+    walignRw[:3, :3] = wuprightRw
+
+    w_xzswapRw = np.array([[0, 0, -1, 0], [0, -1, 0, 0], [1, 0, 0, 0], [0, 0, 0, 1]])
+
+    final_pose = w_xzswapRw @ walignRw @ wcenterTw @ gtsfm_pose
```
johnwlambert, line 174: this would be clearer as
```python
walignedTi = walignedRw @ wcenteredTw @ wTi
```

tests/utils/test_ellipsoid_utils.py
```diff
@@ -0,0 +1,92 @@
...
+        self.assertRaises(TypeError, ellipsoid_utils.remove_outlier_points, sample_points)
+
+    def test_get_rotation_matrix(self):
+        """Tests the get_rotation_matrix() function with 5 sample points.
+
+        4 points create a cross in the xy plane. The 5th point lies on the z-axis to provide some depth to ellipsoid.
+        The resulting major axes will be along the x, y, and z axes.
+        """
+
+        sample_points = np.array([[1, 1, 0], [-1, 1, 0], [-1, -1, 0], [1, -1, 0], [0, 0, 1]])
```
johnwlambert, line 63: i think it's a bit clearer to show the scaling, e.g.
```python
point_cloud = np.array([ [0,-4,0], [0,4,0], [1,0,0], [-1,0,0], [0,0,1] ])
```

johnwlambert: or actually the 45 degree rotation is better, i like that. but maybe we can stretch the axes a bit

PR #167, opened 2021-04-27
adding pylint to CI

ayushbaid:
Adding the pylint lint check to the CI

### DIFF
--- a/.github/scripts/python.sh
+++ b/.github/scripts/python.sh
@@ -33,8 +33,9 @@ pytest tests --cov gtsfm
 coverage report
 
 ##########################################################
-# Test with flake8
+# Lint with flake8 and pylint
 ##########################################################
 
 pip install flake8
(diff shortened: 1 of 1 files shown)


johnwlambert:
Thanks for the PR, @ayushbaid. What does pylint check for, that `black` and `flake8` dont catch?

It looked like most of the warnings it raised in the CI run were false alarms, so I'm a bit averse to adding this.

akshay-krishnan:
Yeah, could you let me know what it’s doing? Not much code to review anyway.

ayushbaid:
I do not understand it completely, but it looks like pylint is a superset of the flake8 checks. It is more thorough (and slow), and somewhat outdated too.

Pylint is more popular than flake8 though.

References:
1. https://www.reddit.com/r/Python/comments/82hgzm/any_advantages_of_flake8_over_pylint/

johnwlambert:
I see. I would prefer to not merge this unless there is a clear and compelling use case over flake8 and black.

I looked through several dozen of the pylint errors in CI and they all looked to be raising false alarms, so I'd consider it to not be helpful as used presently.

PR #168, opened 2021-04-28
Add gtsfm-nonfree as subtree

ayushbaid:
Adding the [gtsfm-nonfree](https://github.com/borglab/gtsfm-nonfree) repository as a subtree.

With the subtree setup, we can just continue adding code in both the *main* gtsfm and the nonfree repo from a single place.

### DIFF
--- a/nonfree/README.md
+++ b/nonfree/README.md
@@ -1,2 +1,2 @@
 # gtsfm-nonfree
-3rd party algorithms to be plugged into GTSFM which are not free for commercial use due to licensing issues
+3rd party algorithms to be plugged into GTSFM which are not free for commercial.
--- a/nonfree/README.md
+++ b/nonfree/README.md
@@ -1,0 +1,2 @@
+# gtsfm-nonfree
+3rd party algorithms to be plugged into GTSFM which are not free for commercial use due to licensing issues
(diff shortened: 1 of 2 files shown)


ayushbaid:
Commands used:

```
git remote add nonfree git@github.com:borglab/gtsfm-nonfree.git
git subtree add --prefix=nonfree/ nonfree main
```

nonfree/README.md
```diff
@@ -0,0 +1,2 @@
+# gtsfm-nonfree
+3rd party algorithms to be plugged into GTSFM which are not free for commercial use due to licensing issues
```
akshay-krishnan, line 2: " due to licensing issues" is probably unnecessary.

akshay-krishnan (approved):

nonfree/README.md
ayushbaid: Thanks. removed it.

johnwlambert:
Proposition: to revert this commit, and to use a very explicit license. Force users to change the config themselves if they any to use noncommercial code is the safest way. move the thirdparty stuff into the repo. change the default to be noncommercial for best results, and say if people want commercial, they need to switch to different config option.

Pros of subtree:
- standalone package which can be reused, but we will not use nonfree as a standalone package

Cons of subtree:
- two repos is harder to keep track of than one. added complexity for developers is added complexity for all future maintainers though.
- if we pull in code directly to our repo from the other, then it doesn’t seem to be separate from a legal perspective. the 3rd party code gets pulled into the commit history so it’s basically the same licensing wise as just putting it into a separate directory

PR #169, opened 2021-04-29
Adding matcher for the three view estimator

ayushbaid:
In this review, we add the three view estimator, and perform the matching operation. For three view estimators, we process triplets of images at a time.


gtsfm/three_view_estimator.py
```diff
@@ -0,0 +1,110 @@
...
+import numpy as np
+from dask.delayed import Delayed
+
+import gtsfm.utils.logger as logger_utils
+from gtsfm.frontend.matcher.matcher_base import MatcherBase
+from gtsfm.frontend.verifier.verifier_base import VerifierBase
+
+logger = logger_utils.get_logger()
+
+mpl_logger = logging.getLogger("matplotlib")
```
johnwlambert, line 17: thanks for the PR, Ayush.

Maybe we should configure the matplotlib logger settings inside `get_logger()`? are they needed here?

johnwlambert:
Hi Ayush, could you add a little more context about the intended use case for the files? for cycle consistency, or trifocal tensor estimation?

gtsfm/loader/loader_base.py
```diff
@@ -122,10 +141,28 @@ def get_valid_pairs(self) -> List[Tuple[int, int]]:
...
         return indices
+
+    def get_valid_triplets(self) -> List[Tuple[int, int, int]]:
+        """Get the valid triplets of images.
+
+        Returns:
+            list of indices of valid triplets.
+        """
+        indices = []
+        for idx1 in range(self.__len__()):
```
johnwlambert, line 158: any way to do this with itertools, instead of a triple for loop?

gtsfm/three_view_estimator.py
```diff
@@ -0,0 +1,110 @@
+"""Estimator which operates on a triplet of images to compute relative pose and verified indices.
```
johnwlambert, line 1: i saw this is placed at ` gtsfm/three_view_estimator.py`. Should it be inside gtsfm/frontend? Is it intended as a verifier or as a matcher? Is this more of an abstract class?

ayushbaid:
> Hi Ayush, could you add a little more context about the intended use case for the files? for cycle consistency, or trifocal tensor estimation?

I intend to use it for trifocal tensor estimation. This three-view-geometry will serve as an alternative to the two-view case.

gtsfm/three_view_estimator.py
ayushbaid: This is an alternative to the `TwoViewEstimator` plate. This class reuses the two-view's matcher (e.g. TwoWay matcher/SuperGlue) and performs cycle-consistency check to output matches for triplets.

I will need to implement a trifocal based verifier, and that will go into the frontend folder.

gtsfm/loader/loader_base.py
johnwlambert: @ayushbaid I think I can mark this as resolved, since you switched to itertools?

gtsfm/loader/loader_base.py
```diff
@@ -122,10 +142,27 @@ def get_valid_pairs(self) -> List[Tuple[int, int]]:
...
         return indices
+
+    def get_valid_triplets(self) -> List[Tuple[int, int, int]]:
+        """Get the valid triplets of images.
+
+        Returns:
+            list of indices of valid triplets.
+        """
+        indices = []
+        for idx1, idx2 in itertools.product(range(len(self)), range(len(self))):
```
johnwlambert, line 159: are permutations considered unique?
```python
>>> list(itertools.product( range(3), range(3) ))
[(0, 0), (0, 1), (0, 2), (1, 0), (1, 1), (1, 2), (2, 0), (2, 1), (2, 2)]
```
like (0,1) and (1,0)?

gtsfm/three_view_estimator.py
```diff
@@ -0,0 +1,104 @@
...
+
+        i2_to_i3_dict = {}
+        for idx_i2, idx_i3 in match_indices_i2i3:
+            i2_to_i3_dict[idx_i2] = idx_i3
+
+        i1_to_i3_dict = {}
+        for idx_i1, idx_i3 in match_indices_i1i3:
+            i1_to_i3_dict[idx_i1] = idx_i3
+
+        # for the triplet matches, chaining i1->i2 with i2->i3 should be present in i1->i3
```
johnwlambert, line 62: seems like we are putting 3-cycle consistency code here. is it how Theia implements their cycle consistency here? 


(https://github.com/sweeneychris/TheiaSfM/blob/master/src/theia/sfm/filter_view_graph_cycles_by_rotation.h or https://github.com/sweeneychris/TheiaSfM/blob/master/src/theia/sfm/filter_view_pairs_from_relative_translation.h

gtsfm/loader/loader_base.py
ayushbaid: Yes they would be considered unique, because I am relying on the `validate_pair` function. That should return true only for one of (0, 1)/(1, 0).

johnwlambert: got it -- do you mind adding that to the docstring?

johnwlambert: (or as an inline comment)

gtsfm/three_view_estimator.py
ayushbaid: That's a different type of cycle consistency. They check the cumulative translation/rotation errors between triplets to throw away cameras.

Here, we do a triplet matching of a correspondence. If point x2 in image i2 is matched with point x1 in image i1, and x3 in image i3, then x1 has to match to x3 too.

johnwlambert: I see. Do you mind adding this distinction (and references) to the docstring?

johnwlambert:
Thanks again for the PR, @ayushbaid. I think this approach has potential, but have we seen it used in the literature anywhere?

Could you run an experiment with the matching aspect you had in mind (be it with trifocal tensor estimation, or with the track cycle consistency), and show how it affects performance? Maybe those should be part of this PR, as well?

I fear otherwise we may just be adding more code to maintain in the future.

ayushbaid:
3-view estimator is not our priority right now. Closing this PR for later.

PR #170, opened 2021-05-16
Small translation averaging fix - return None if no inlier edges exist to a node instead of crashing

akshay-krishnan:
We need to check the output values from gtsam.TranslationRecovery contains an index before using it. If there were no inlier edges to this index, it will not be contained in the output values. 
Currently the program crashes for this case, but after this change we get a None output.


tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -153,5 +154,73 @@ def test_pickleable(self):
...
+            (1, 3): np.array([0.986072, -0.00746304, 0.166149]),
+            (1, 4): np.array([-0.996558, 0.00911097, -0.0823996]),
+            (2, 3): np.array([0.990546, -0.0294894, 0.133976]),
+            (2, 4): np.array([0.998932, -0.0300599, -0.035099]),
+            (3, 4): np.array([0.994791, -0.033332, -0.0963361]),
+        }
+        i2Ui1_input = {(i, j): Unit3(t) for (i, j), t in i2Ui1_input.items()}
+        wti_computed = self.obj.run(len(wRi_input), i2Ui1_input, wRi_input)
+
+        assert len(wti_computed) == 5
```
akshay-krishnan, line 222: Maybe we should also assert the last one is None?

johnwlambert: good idea -- updated

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -118,7 +118,7 @@ def run(
         # transforming the result to the list of Point3
         wti_list = [None] * num_images
         for i in range(num_images):
-            if wRi_list[i] is not None:
+            if wRi_list[i] is not None and wti_values.exists(i):
```
ayushbaid, line 121: `wRi_list[i] is not None` is redundant?

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -153,5 +154,74 @@ def test_pickleable(self):
             self.fail("Cannot dump rotation averaging object using pickle")
 
 
+class Test1dsfmAllOutliers(unittest.TestCase):
```
ayushbaid, line 157: why have a new class? We can just add it as a function in the old class?

ayushbaid (changes requested):

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: true, but i think it's a bit more readable?

tests/averaging/translation/test_averaging_1dsfm.py
johnwlambert: main reason here was to force the case w/ data that would make 1dsfm fail previously

gtsfm/averaging/translation/averaging_1dsfm.py
ayushbaid: I think it makes it more confusing. But i am fine with it.

johnwlambert (approved):

PR #179, opened 2021-05-21
Support for Sim(3) alignment with missing poses

johnwlambert:
After the MuitlviewOptimizer, we align the poses that come out of Shonan and 1dsfm to the ground truth, and measure the pose errors.

1dsfm may return None for some poses (if the Value is missing). Previous Sim(3) alignment required a valid pose for every frame, in GT and estimate.

### DIFF
--- a/gtsfm/utils/geometry_comparisons.py
+++ b/gtsfm/utils/geometry_comparisons.py
@@ -40,6 +40,52 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
     return [aRb.compose(bRi) for bRi in bRi_list]
 
 
+def align_poses_sim3_wrapper(aTi_list: List[Optional[Pose3]], bTi_list: List[Optional[Pose3]]) -> List[Optional[Pose3]]:
+    """Align by similarity transformation, but allow missing estimated poses in the input.
+
+    Note: this is a wrapper for align_poses_sim3() that allows for missing poses/dropped cameras.
+    This is necessary, as align_poses_sim3() requires a valid pose for every input pair.
+
+    We force SIM(3) alignment rather than SE(3) alignment.
+    We assume the two trajectories are of the exact same length.
+
+    Args:
+        aTi_list: reference poses in frame "a" which are the targets for alignment
+        bTi_list: input poses which need to be aligned to frame "a"
+
+    Returns:
+        aTi_list_: transformed input poses previously "bTi_list" but now which
+            have the same origin and scale as reference (now living in "a" frame)
+    """
+    assert len(aTi_list) == len(bTi_list)
+
+    # only choose target poses for which there is a corresponding estimated pose
+    corresponding_aTi_list = []
+    valid_camera_idxs = []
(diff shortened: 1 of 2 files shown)


gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +40,52 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
...
+    valid_bTi_list = []
+    for i, bTi in enumerate(bTi_list):
+        if bTi is None:
+            dropped_camera_idxs.append(i)
+        else:
+            valid_camera_idxs.append(i)
+            valid_bTi_list.append(bTi)
+            corresponding_aTi_list.append(aTi_list[i])
+
+    if len(dropped_camera_idxs) > 0:
```
akshay-krishnan, line 75: I think this should be logged in whichever module is dropping these (if necessary), unnecessary here.

akshay-krishnan (changes requested):
I don’t think its a good idea to have a _wrapper function for this. I think we should add this code to the align_poses_sim3() itself. Is there a problem with doing that? 

Also what does Similarity3 do if it gets a None?

johnwlambert:
Thanks for the review @akshay-krishnan. So my concern was that align_poses_sim3() is already quite long, and this would make it less readable. I don't think this code is relevant to alignment, its more just a wrapper around it

gtsfm/utils/geometry_comparisons.py
ayushbaid: I agree that logging here is not very useful. But I don't think we need to log it anywhere. We dont care about the exact indices anyway. We can just log (and I think we are already doing it) the number of valid cameras.

ayushbaid (approved):
LGTM!

gtsfm/utils/geometry_comparisons.py
johnwlambert: sure, i'll remove it

akshay-krishnan:
> Thanks for the review @akshay-krishnan. So my concern was that align_poses_sim3() is already quite long, and this would make it less readable. I don't think this code is relevant to alignment, its more just a wrapper around it

Okay in that case, can we give it a better name? Wrapper doesn’t say anything. Maybe something that includes “missing poses”? align_poses_sim3_ignore_missing()? I don’t feel strongly about it, so feel free to leave as is.

akshay-krishnan (approved):

johnwlambert:
> > Thanks for the review @akshay-krishnan. So my concern was that align_poses_sim3() is already quite long, and this would make it less readable. I don't think this code is relevant to alignment, its more just a wrapper around it
> 
> Okay in that case, can we give it a better name? Wrapper doesn’t say anything. Maybe something that includes “missing poses”? align_poses_sim3_ignore_missing()? I don’t feel strongly about it, so feel free to leave as is.

Good point -- the name can be better. I like `align_poses_sim3_ignore_missing()`

johnwlambert:
I updated the name

PR #181, opened 2021-05-23
Add SuperGlueMatcher class for interface w/ 3rd party lib

johnwlambert:
- Calls the author's SuperPoint and SuperGlue code directly, without modifications.

Also fixes `matplotlib` errors and warnings from a new version release.

Also, fixes a few old artifacts, where something that could have fit in 120 chars, was split into 3 lines, due to the original 80 char reformatting. Python black will not remove the trailing comma after the last argument, meaning it cannot recover the fact that the line could have fit on 1 line.


gtsfm/frontend/matcher/superglue_matcher.py
```diff
@@ -0,0 +1,78 @@
...
+        }
+        self.superglue = SuperGlue(config).eval()
+
+    def match(
+        self,
+        keypoints_i1: Keypoints,
+        keypoints_i2: Keypoints,
+        descriptors_i1: np.ndarray,
+        descriptors_i2: np.ndarray,
+        img_height: int,
```
johnwlambert, line 29: @ayushbaid these 2 arguments are new, but i believe necessary. So I guess we'll add them to the other matcher files as well?

johnwlambert: They could be `Optional[int]` elsewhere, but cannot be optional here

ayushbaid: You are right. They need to be present here. However, we should have a consistent API for all matchers. Hence we should make it compulsory for all matchers.

I already have this change sitting in this branch: https://github.com/borglab/gtsfm/compare/feature/superglue

ayushbaid:
I have the code in a branch for superpoint and superglue too. We can probably try to merge this branch with that?

gtsfm/frontend/matcher/superglue_matcher.py
akshay-krishnan: well, thats kinda sad, should we continue reviewing this PR?

akshay-krishnan: @ayushbaid my 2 cents: I dont think its a good idea to make it compulsory for all matchers if most matchers are not going to need it. People reading the code would think "why is this even here?" unless they read the supergluematcher. Its kind of a case of [YAGNI](https://martinfowler.com/bliki/Yagni.html). Exceptions to the base API are okay where they make sense. If at some point we have many matchers that need it, the base class could have it. 
But I'll leave the choice to you. :)

johnwlambert: I think we need the exact same interface though, for inter-operability of the modules.

johnwlambert: The YAGNI article is cool -- thanks for sharing

johnwlambert: @ayushbaid i'll merge your branch into this one

download_model_weights.sh
```diff
@@ -0,0 +1,30 @@
+
+# Download model weights for different modules available for use in GTSFM
```
johnwlambert, line 2: included in a separate PR, that we should merge first

environment_linux.yml
```diff
@@ -21,7 +21,7 @@ dependencies:
   - dask # same as dask[complete] pip distribution
   - python-graphviz
   # core functionality and APIs
-  - matplotlib>=3.3.3
+  - matplotlib==3.4.2
```
johnwlambert, line 24: matplotlib API has changed, so we have to force the version now, to latest, for consistent behavior

download_model_weights.sh
```diff
@@ -28,3 +28,4 @@ mkdir -p $PATCHMATCHNET_WEIGHTS_DIR
 PATCHMATCHNET_URL="https://github.com/FangjinhuaWang/PatchmatchNet/raw/main/checkpoints/model_000007.ckpt"
 
 wget -O $PATCHMATCHNET_WEIGHTS_DIR/model_000007.ckpt $PATCHMATCHNET_URL
+
```
ayushbaid, line 31: can be removed

gtsfm/frontend/verifier/ransac.py
```diff
@@ -23,9 +23,7 @@
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.frontend.verifier.verifier_base import VerifierBase
 
-
-NORMALIZED_COORD_RANSAC_THRESH = 0.001  # TODO: hyperparameter to tune
-PIXEL_COORD_RANSAC_THRESH = 0.5  # TODO: hyperparameter to tune
+PIXEL_COORD_RANSAC_THRESH = 4  # TODO: hyperparameter to tune
```
ayushbaid, line 26: any reason for increasing the threshold?

gtsfm/frontend/verifier/ransac.py
```diff
@@ -62,13 +60,15 @@ def verify(
...
+            # use stricter threshold, among the two choices
+            fx = max(camera_intrinsics_i1.K()[0,0], camera_intrinsics_i2.K()[0,0])
             i2Ei1, inlier_mask = cv2.findEssentialMat(
                 uv_norm_i1[match_indices[:, 0]],
                 uv_norm_i2[match_indices[:, 1]],
                 K,
                 method=cv2.RANSAC,
-                threshold=NORMALIZED_COORD_RANSAC_THRESH,
-                prob=DEFAULT_RANSAC_SUCCESS_PROB,
+                threshold=PIXEL_COORD_RANSAC_THRESH / fx,
```
ayushbaid, line 70: thanks for this change. 1 hyperparam is better than 2!

gtsfm/utils/io.py
```diff
@@ -36,16 +39,18 @@ def load_image(img_path: str) -> Image:
     """
     original_image = PILImage.open(img_path)
 
-    exif_data = original_image.getexif()
+    exif_data = original_image._getexif()
```
ayushbaid, line 42: is there an alternate for this?

tests/frontend/matcher/test_matcher_base.py
```diff
@@ -53,16 +53,21 @@ def __assert_one_to_one_constraint(self, match_idxs: np.ndarray):
     def test_empty_input(self):
         """Tests the matches when there are no descriptors."""
 
-        nonempty_keypoints, _, nonempty_descriptors, _, = generate_random_input()
+        nonempty_keypoints, _, nonempty_descriptors, _, _, _ = generate_random_input()
         empty_keypoints = Keypoints(coordinates=np.array([]))
         empty_descriptors = np.array([])
+        im_shape = (300, 200)
```
ayushbaid, line 59: maybe create two variables im1_shape and im2_shape to prevent confusion?

download_model_weights.sh
johnwlambert: thanks, done

gtsfm/frontend/verifier/ransac.py
johnwlambert: @ayushbaid so I noticed that COLMAP uses 3 or 4 px as its threshold. I think this should go in a global config, since it's such an important hyperparameter. COLMAP seems to use the same pixel threshold when fitting both E, F, and H

gtsfm/utils/io.py
johnwlambert: There is. This was more a hotfix. The question is whether we want to work with "flattened" or hierarchical EXIF: https://github.com/python-pillow/Pillow/pull/4947/files

May require more discussion as to what would be cleanest. The new Pillow EXIF API is not well documented, IMHO.

tests/frontend/matcher/test_matcher_base.py
johnwlambert: sure, updated now

gtsfm/configs/deep_front_end.yaml
```diff
@@ -0,0 +1,45 @@
...
+    corr_metric_dist_threshold: 3 # in px
+    matcher:
+      _target_: gtsfm.frontend.matcher.superglue_matcher.SuperGlueMatcher
+
+    verifier:
+      _target_: gtsfm.frontend.verifier.ransac.Ransac
+      use_intrinsics_in_verification: True
+    
+
+  multiview_optimizer:
```
akshay-krishnan, line 25: Is some form of inheritance supported in these YAML files? That way, we could have a default configuration in default.yaml, include it in deep_front_end.yaml and change/override only the feature_extractor and two_view_estimator.

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -0,0 +1,78 @@
...
+import gtsfm.utils.images as image_utils
+from gtsfm.common.image import Image
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.frontend.detector_descriptor.detector_descriptor_base import DetectorDescriptorBase
+from thirdparty.SuperGluePretrainedNetwork.models.superpoint import SuperPoint
+
+ROOT_PATH = Path(__file__).resolve().parent.parent.parent.parent
+MODEL_WEIGHTS_PATH = (
+    ROOT_PATH / "thirdparty" / "SuperGluePretrainedNetwork" / "models" / "weights" / "superpoint_v1.pth"
+)
```
akshay-krishnan, line 27: should this not be an option in the yaml file?

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -0,0 +1,78 @@
...
+        # sort by scores
+        sort_idx = np.argsort(-scores)
+        features_points = features_points[sort_idx]
+        scores = scores[sort_idx]
+        descriptors = descriptors[sort_idx]
+
+        # limit the number of keypoints
+        features_points = features_points[: self.max_keypoints]
+        scores = scores[: self.max_keypoints]
+        descriptors = descriptors[: self.max_keypoints]
```
akshay-krishnan, line 74: there is no need to sort and copy the other keypoints then. but sadly I could not find a partial sort in python. this is the C++ one https://www.cplusplus.com/reference/algorithm/partial_sort/

gtsfm/frontend/matcher/superglue_matcher.py
```diff
@@ -0,0 +1,108 @@
...
+from typing import Tuple
+
+import numpy as np
+import torch
+
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.frontend.matcher.matcher_base import MatcherBase
+from thirdparty.SuperGluePretrainedNetwork.models.superglue import SuperGlue
+
+SUPERGLUE_DESC_DIM = 256
```
akshay-krishnan, line 23: another yaml parameter?

johnwlambert: so this one is actually rigidly fixed, because we are using their pre-trained weights, which are set to 256-dimensional. So it's more an implementation detail on the 3rd party end, rather than something we can change. I put it up here for clarity, so we didn't have a random magic number "256" laying around.

gtsfm/frontend/detector_descriptor/superpoint.py
johnwlambert: So we expect the weights to be in a specific place, with our `download_model_weights.sh` script. So I think it's cleaner to put them in an expected place and use that location here.

It also allows us to use the authors' original SuperPoint API, without changing it at all

gtsfm/configs/deep_front_end.yaml
ayushbaid: +1 to this.

gtsfm/frontend/verifier/ransac.py
ayushbaid: Makes sense.

gtsfm/frontend/matcher/superglue_matcher.py
```diff
@@ -0,0 +1,108 @@
...
+
+class SuperGlueMatcher(MatcherBase):
+    """Implements the SuperGlue matcher -- a pretrained graph neural network using attention."""
+
+    def __init__(self, use_cuda: bool = True, use_outdoor_model: bool = True):
+        """Initialize the configuration and the parameters."""
+        super().__init__()
+
+        self._config = {
+            "descriptor_dim": 256,
```
ayushbaid, line 34: Use the variable here

gtsfm/frontend/matcher/superglue_matcher.py
```diff
@@ -0,0 +1,108 @@
...
+            "weights": "outdoor" if use_outdoor_model else "indoor",
+            "sinkhorn_iterations": 20,
```
ayushbaid, line 36: Maybe make it variable too. Also comment that we are using Author's params.

gtsfm/frontend/verifier/ransac.py
ayushbaid: Maybe create an issue for this?

gtsfm/frontend/matcher/superglue_matcher.py
johnwlambert: good catch -- updated

gtsfm/frontend/matcher/superglue_matcher.py
johnwlambert: good catch -- updated

gtsfm/frontend/verifier/ransac.py
johnwlambert: sure -- will do

tests/frontend/matcher/test_superglue_matcher.py
```diff
@@ -0,0 +1,28 @@
+
```
akshay-krishnan, line 1: docstring here?

akshay-krishnan (changes requested):
I see a lot of formatting changes. If these are intentional, could you document why in the PR description? It is nice to use the description to document everything a PR changes.

tests/frontend/verifier/test_verifier_base.py
```diff
@@ -274,8 +290,8 @@ def sample_points_on_plane(
         raise ValueError("z-coefficient for the plane should not be zero")
 
     # sample x and y coordinates randomly
-    x = np.random.uniform(low=range_x_coordinate[0], high=range_x_coordinate[1], size=(num_points, 1),)
-    y = np.random.uniform(low=range_y_coordinate[0], high=range_y_coordinate[1], size=(num_points, 1),)
+    x = np.random.uniform(low=range_x_coordinate[0],high=range_x_coordinate[1],size=(num_points, 1))
```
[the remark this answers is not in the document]
johnwlambert: forgot to re-format -- should be fixed now

tests/frontend/matcher/test_superglue_matcher.py
johnwlambert: thanks -- should be added now

tests/frontend/verifier/test_verifier_argoverse.py
```diff
@@ -124,6 +130,4 @@ def setUp(self):
         self.translation_err_tol = 0.02
 
     def testRecoveredPoseError(self):
-        check_verifier_output_error(
-            self.verifier, self.euler_angle_err_tol, self.translation_err_tol,
-        )
+        check_verifier_output_error(self.verifier, self.euler_angle_err_tol, self.translation_err_tol)
```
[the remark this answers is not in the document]
johnwlambert: Here I fixed an old artifact, where something that could have fit in 120 chars, was split into 3 lines, due to the original 80 char reformatting. Python black will not remove the trailing comma after the last argument, meaning it cannot recover the fact that the line could have fit on 1 line.

tests/frontend/verifier/test_verifier_base.py
```diff
@@ -193,9 +199,11 @@ def __verify_random_inputs(self,) -> Tuple[Optional[Rot3], Optional[Unit3], np.n
...
         ) = generate_random_input_for_verifier()
 
-        (i2Ri1, i2Ui1, verified_indices,) = self.verifier.verify(
-            keypoints_i1, keypoints_i2, match_indices, intrinsics_i1, intrinsics_i2
-        )
+        (
+            i2Ri1,
+            i2Ui1,
+            verified_indices,
+        ) = self.verifier.verify(keypoints_i1, keypoints_i2, match_indices, intrinsics_i1, intrinsics_i2)
```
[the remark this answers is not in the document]
johnwlambert: i believe this was a trailing comma that when removed, trigger new formatting

tests/frontend/verifier/test_verifier_base.py
```diff
@@ -53,15 +53,21 @@ def __execute_test(
         """Execute the verification and compute results."""
 
         i2Ri1_computed, i2Ui1_computed, verified_indices_computed = self.verifier.verify(
-            keypoints_i1, keypoints_i2, match_indices, camera_intrinsics_i1, camera_intrinsics_i2,
+            keypoints_i1,
+            keypoints_i2,
+            match_indices,
+            camera_intrinsics_i1,
+            camera_intrinsics_i2,
```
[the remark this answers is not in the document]
johnwlambert: i believe this was a trailing comma that when removed, trigger new formatting

tests/frontend/verifier/test_verifier_argoverse.py
```diff
@@ -89,10 +89,18 @@ def check_verifier_output_error(verifier: VerifierBase, euler_angle_err_tol: flo
...
+    gt_euler_angles_str = str(np.round(gt_euler_angles, 1))
+    euler_angles_str = str(np.round(euler_angles, 1))
+    assert np.allclose(
+        gt_euler_angles, euler_angles, atol=euler_angle_err_tol
+    ), f"GT {gt_euler_angles_str} vs. Est. {euler_angles_str} w/ tol {euler_angle_err_tol:.1f}"
 
     gt_i1ti2 = np.array([0.21, -0.0024, 0.976])
-    assert np.allclose(gt_i1ti2, i1ti2, atol=translation_err_tol)
+    gt_i1ti2_str = str(np.round(gt_i1ti2, 1))
+    i1ti2_str = str(np.round(i1ti2, 1))
```
akshay-krishnan, line 100: there should be a way to do this with python string formatting itself I think, like the 0.1f format.

akshay-krishnan (approved):

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -0,0 +1,78 @@
...
+
+class SuperPointDetectorDescriptor(DetectorDescriptorBase):
+    """Superpoint Detector+Descriptor implementation."""
+
+    def __init__(self, use_cuda: bool = True, weights_path: Union[Path,str] = MODEL_WEIGHTS_PATH) -> None:
+        """Configures the object.
+
+        Args:
+            max_keypoints: max keypoints to detect in an image.
+            is_cuda (optional): flag controlling the use of GPUs via CUDA. Defaults to True.
```
ayushbaid, line 38: the function has argument `use_cuda` and the docstring has `is_cuda`.

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -0,0 +1,78 @@
...
+            weights_path (optional): Path to the model weights. Defaults to MODEL_WEIGHT_PATH.
+        """
+        super().__init__()
+        self._use_cuda = use_cuda and torch.cuda.is_available()
+        self._config = {"weights_path": weights_path}
+
+    def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
+        """Jointly generate keypoint detections and their associated descriptors from a single image."""
```
ayushbaid, line 46: Should we copy the docstring from the parent class?

gtsfm/frontend/detector_descriptor/superpoint.py
ayushbaid: We can just pick the top entries from the sort_idx.

```
sort_idx = sort_idx[:self.max_keypoints]
```

This should be slightly better?

gtsfm/frontend/verifier/ransac.py
```diff
@@ -21,17 +21,32 @@
...
+                                            intrinsics are known as opposed to approximating them from exif data.
+            px_threshold: epipolar distance threshold (measured in pixels)
+        """
+        self._use_intrinsics_in_verification = use_intrinsics_in_verification
+        self._px_threshold = px_threshold
+        self._min_matches = (
+            NUM_MATCHES_REQ_E_MATRIX if self._use_intrinsics_in_verification else NUM_MATCHES_REQ_F_MATRIX
+        )
+
+        self._failure_result = (None, None, np.array([], dtype=np.uint64))
```
ayushbaid, line 48: this can be a global variable? I don't mind it here too btw.

gtsfm/two_view_estimator.py
```diff
@@ -159,7 +168,9 @@ def compute_correspondence_metrics(
 
 
 def compute_relative_pose_metrics(
-    i2Ri1_computed: Optional[Rot3], i2Ui1_computed: Optional[Unit3], i2Ti1_expected: Pose3,
```
ayushbaid, line 162: i dont think this violates the 120 width limit on the line.

tests/frontend/matcher/test_superglue_matcher.py
```diff
@@ -0,0 +1,31 @@
...
+Authors: John Lambert
+"""
+import numpy as np
+
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.frontend.matcher.superglue_matcher import SuperGlueMatcher
+
+
+def test_SuperGlueMatcher() -> None:
```
ayushbaid, line 11: Superglue should inherit tests from the base matcher tests?

ayushbaid (changes requested):

tests/frontend/matcher/test_superglue_matcher.py
johnwlambert: So this unfortunately won't work since only 256-dimensional descriptors work with SuperGlue, while the other unit tests use descriptors of various sizes.

gtsfm/two_view_estimator.py
johnwlambert: just updated it

gtsfm/frontend/detector_descriptor/superpoint.py
johnwlambert: So the base class says "Match descriptor vectors.", but that isn't accurate in this case, since we use keypoints and descriptors both to match.

gtsfm/frontend/detector_descriptor/superpoint.py
johnwlambert: thanks -- fixed

ayushbaid (approved):

PR #185, opened 2021-05-26
Add unit tests for ColmapLoader

johnwlambert:


tests/loader/test_colmap_loader.py
```diff
@@ -0,0 +1,78 @@
...
+        # should be shared intrinsics
+        np.testing.assert_allclose(self.loader.get_camera_intrinsics(0).K(), self.loader.get_camera_intrinsics(1).K())
+
+    def test_image_resolution(self) -> None:
+        """Ensure that the image is downsampled properly to a max resolution of 500 px.
+
+        Note: native resolution is (1936, 1296) for (H,W)
+        """
+        assert self.loader._scale_u == 500 / 1296
+        assert np.isclose(self.loader._scale_v, 500 / 1296, atol=1e-4)
```
akshay-krishnan, line 58: Was going to say it should be 500.0 / 1296.0, but I just learned python gives float for int / int. Even then, it might make it more obvious to the reader if you use decimal point.

akshay-krishnan (approved):

tests/loader/test_colmap_loader.py
```diff
@@ -0,0 +1,78 @@
...
+
+
+class TestColmapLoader(unittest.TestCase):
+
+    def setUp(self):
+        """Set up the loader for the test."""
+        super().setUp()
+
+        colmap_files_dirpath = TEST_DATA_ROOT / "set1_lund_door/colmap_ground_truth"
+        images_dir = TEST_DATA_ROOT / "set1_lund_door/images"
```
ayushbaid, line 21: Should we define these globally instead of the setup function?

tests/loader/test_colmap_loader.py
```diff
@@ -0,0 +1,78 @@
...
+        # there are 12 images in Lund Door set 1
+        assert len(self.loader) == 12
+        assert len(self.loader._calibrations) == 12
+        assert self.loader._num_imgs == 12
+        assert len(self.loader._image_paths) == 12
+
+    def test_get_camera_intrinsics(self) -> None:
+        """Ensure that for shared calibration case, GT intrinsics are identical across frames."""
+        # should be shared intrinsics
+        np.testing.assert_allclose(self.loader.get_camera_intrinsics(0).K(), self.loader.get_camera_intrinsics(1).K())
```
ayushbaid, line 50: Maybe the test data should have unique intrinsics for each camera so we can check if the indexing works as expected?

tests/loader/test_colmap_loader.py
```diff
@@ -0,0 +1,78 @@
...
+
+    def test_get_image(self) -> None:
+        """Ensure a downsampled image can be successfully provided."""
+        img0 = self.loader.get_image(0)
+        assert isinstance(img0, Image)
+
+    def test_get_camera_pose(self) -> None:
+        """Ensure a camera pose can be successfully provided"""
+        wT0 = self.loader.get_camera_pose(0)
+        assert isinstance(wT0, Pose3)
```
ayushbaid, line 75: Why not test against actual values too?

tests/loader/test_colmap_loader.py
ayushbaid, line 50: Also test if resizing operation was performed on the intrinsics?

ayushbaid (changes requested):

tests/loader/test_colmap_loader.py
johnwlambert: i was planning to add more unit tests in the future (perhaps with invalid paths and other directories) once we decide what we want the expected behavior to be. Maybe we can discuss that at our next meeting.

So I prefer not global for now.

tests/loader/test_colmap_loader.py
johnwlambert: We could add that in a future PR. This specific test case was just for the shared calibration situation.

tests/loader/test_colmap_loader.py
johnwlambert: testing against them now

tests/loader/test_colmap_loader.py
ayushbaid: Should we still check for values?

johnwlambert: good idea -- added a value test

johnwlambert: just added it

ayushbaid (approved):

PR #192, opened 2021-06-08
Reorder indices for input to Shonan so that N keys are consecutive [0,...,N-1]

johnwlambert:
GTSAM's ShonanAveraging.cpp assumes that N nodes are all connected, and are ordered consecutively from [0,...,N-1]

This assumption goes deeper than ShonanAveraging::NrUnknowns(), which just checks the assumption. Several other functions that construct matrices related to the convergence certificate also use the fact that the graph is connected and keys are 0..N-1.

There are about 10 places where nrUnknowns is used to dimension a sparse or dense matrix.

Modifying GTSAM would require a major philosophical overhaul, so we perform the re-ordering in GTSFM instead, in a sort of "wrapper".

See https://github.com/borglab/gtsam/issues/784 for more details

### DIFF
--- a/gtsfm/averaging/rotation/shonan.py
+++ b/gtsfm/averaging/rotation/shonan.py
@@ -30,20 +30,33 @@ class ShonanRotationAveraging(RotationAveragingBase):
     """Performs Shonan rotation averaging."""
 
     def __init__(self) -> None:
+        """
+        Note: `p_min` and `p_max` describe the minimum and maximum relaxation rank.
+        """
         self._p_min = 5
         self._p_max = 30
 
-    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
-        """Run the rotation averaging.
+    def run_w_consecutive_ordering(
+        self, num_connected_nodes: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
+    ) -> List[Optional[Rot3]]:
+        """Run the rotation averaging on a connected graph w/ N keys ordered consecutively [0,...,N-1].
+
+        Note: the assumption that Shonan keys are ordered as [0,...,N-1] goes deeper than ShonanAveraging::NrUnknowns(),
+        which just checks the assumption. Several other functions that construct matrices related to the
+        convergence certificate also use the fact that the graph is connected and keys are 0..N-1.
+        There are about 10 places where nrUnknowns is used to dimension a sparse or dense matrix.
+        Modifying GTSAM would require a major philosophical overhaul, so we perform the re-ordering
+        here in a sort of "wrapper". See: https://github.com/borglab/gtsam/issues/784
 
         Args:
-            num_images: number of poses.
-            i2Ri1_dict: relative rotations as dictionary (i1, i2): i2Ri1.
+            num_connected_nodes: number of unique connected nodes in the graph (can be far less the number of images in the dataset)
+            i2Ri1_dict: relative rotations for each image pair-edge as dictionary (i1, i2): i2Ri1.
 
         Returns:
-            Global rotations for each camera pose, i.e. wRi, as a list. The number of entries in the list is
-                `num_images`. The list may contain `None` where the global rotation could not be computed (either
+            Global rotations for each **CONNECTED** camera pose, i.e. wRi, as a list. The number of entries in the list is
+                `num_connected_nodes`. The list may contain `None` where the global rotation could not be computed (either
                 underconstrained system or ill-constrained system).
+
         """
         lm_params = LevenbergMarquardtParams.CeresDefaults()
         shonan_params = ShonanAveragingParameters3(lm_params)
@@ -56,23 +69,58 @@ class ShonanRotationAveraging(RotationAveragingBase):
 
         for (i1, i2), i2Ri1 in i2Ri1_dict.items():
             if i2Ri1 is not None:
-                between_factors.append(
-                    BetweenFactorPose3(
-                        i2,
-                        i1,
-                        Pose3(
-                            i2Ri1,
-                            np.zeros(
-                                3,
-                            ),
-                        ),
-                        noise_model,
-                    )
-                )
+                # ignore translation during rotation averaging
+                i2Ti1 = Pose3(i2Ri1, np.zeros(3))
+                between_factors.append(BetweenFactorPose3(i2, i1, i2Ti1, noise_model))
 
         obj = ShonanAveraging3(between_factors, shonan_params)
 
         initial = obj.initializeRandomly()
         result_values, _ = obj.run(initial, self._p_min, self._p_max)
 
-        return [result_values.atRot3(i) for i in range(num_images)]
+        return [result_values.atRot3(i) if result_values.exists(i) else None for i in range(num_connected_nodes)]
+
+    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
+        """Wrapper that re-orders keys to prepare a connected graph, w/ N keys ordered [0,...,N-1].
+
+        Args:
+            num_images: number of poses.
+            i2Ri1_dict: relative rotations for each image pair-edge as dictionary (i1, i2): i2Ri1.
+
+        Returns:
+            Global rotations for each camera pose, i.e. wRi, as a list. The number of entries in the list is
+                `num_images`. The list may contain `None` where the global rotation could not be computed (either
+                underconstrained system or ill-constrained system), or where the camera pose had no valid observation
+                in the input to run().
+        """
+        connected_nodes = set()
+        for (i1, i2) in i2Ri1_dict.keys():
+            connected_nodes.add(i1)
+            connected_nodes.add(i2)
+
+        connected_nodes = sorted(list(connected_nodes))
+
+        # given original index, get back new temporary index
+        reordered_idx_map = {}
+        for i in range(num_images):
+            if i in connected_nodes:
+                reordered_idx_map[i] = connected_nodes.index(i)
+
+        # now, re-order all of the indices
+        i2Ri1_dict_reordered = {}
+        for (i1, i2), i2Ri1 in i2Ri1_dict.items():
+            i1_ = reordered_idx_map[i1]
+            i2_ = reordered_idx_map[i2]
+            i2Ri1_dict_reordered[(i1_, i2_)] = i2Ri1
+
+        wRi_list_subset = self.run_w_consecutive_ordering(
+            num_connected_nodes=len(connected_nodes), i2Ri1_dict=i2Ri1_dict_reordered
+        )
+
+        wRi_list = [None] * num_images
+        for i in range(num_images):
+            if i in connected_nodes:
+                i_ = reordered_idx_map[i]
+                wRi_list[i] = wRi_list_subset[i_]
+
+        return wRi_list
--- a/gtsfm/averaging/rotation/shonan.py
+++ b/gtsfm/averaging/rotation/shonan.py
@@ -8,7 +8,7 @@ References:
 - https://arxiv.org/abs/2008.02737
 - https://gtsam.org/
 
-Authors: Jing Wu, Ayush Baid
+Authors: Jing Wu, Ayush Baid, John Lambert
 """
 from typing import Dict, List, Optional, Tuple
 
@@ -41,22 +41,22 @@ class ShonanRotationAveraging(RotationAveragingBase):
     ) -> List[Optional[Rot3]]:
         """Run the rotation averaging on a connected graph w/ N keys ordered consecutively [0,...,N-1].
 
-        Note: the assumption that Shonan keys are ordered as [0,...,N-1] goes deeper than ShonanAveraging::NrUnknowns(),
-        which just checks the assumption. Several other functions that construct matrices related to the
-        convergence certificate also use the fact that the graph is connected and keys are 0..N-1.
-        There are about 10 places where nrUnknowns is used to dimension a sparse or dense matrix.
+        Note: the assumption that Shonan keys are ordered as [0,...,N-1] goes deeper than
+        ShonanAveraging::NrUnknowns(), which just checks the assumption. Several other functions that construct
+        matrices related to the convergence certificate also use the fact that the graph is connected and
+        keys are 0..N-1. There are about 10 places where nrUnknowns is used to dimension a sparse or dense matrix.
         Modifying GTSAM would require a major philosophical overhaul, so we perform the re-ordering
         here in a sort of "wrapper". See: https://github.com/borglab/gtsam/issues/784
 
         Args:
-            num_connected_nodes: number of unique connected nodes in the graph (can be far less the number of images in the dataset)
+            num_connected_nodes: number of unique connected nodes in the graph (can be far less than the number
+                of images in the dataset)
             i2Ri1_dict: relative rotations for each image pair-edge as dictionary (i1, i2): i2Ri1.
 
         Returns:
-            Global rotations for each **CONNECTED** camera pose, i.e. wRi, as a list. The number of entries in the list is
-                `num_connected_nodes`. The list may contain `None` where the global rotation could not be computed (either
-                underconstrained system or ill-constrained system).
-
+            Global rotations for each **CONNECTED** camera pose, i.e. wRi, as a list. The number of entries in
+                the list is num_connected_nodes`. The list may contain `None` where the global rotation could
+                not be computed (either underconstrained system or ill-constrained system).
         """
         lm_params = LevenbergMarquardtParams.CeresDefaults()
         shonan_params = ShonanAveragingParameters3(lm_params)
--- a/gtsfm/averaging/rotation/shonan.py
+++ b/gtsfm/averaging/rotation/shonan.py
@@ -49,10 +49,11 @@ class ShonanRotationAveraging(RotationAveragingBase):
             num_connected_nodes: number of unique connected nodes (i.e. images) in the graph
                 (<= the number of images in the dataset)
             i2Ri1_dict: relative rotations for each edge between nodes as dictionary (i1, i2): i2Ri1.
+                Note: i1 < num_connected_nodes, and also i2 < num_connected_nodes.
 
         Returns:
             Global rotations for each **CONNECTED** camera pose, i.e. wRi, as a list. The number of entries in
-                the list is num_connected_nodes`. The list may contain `None` where the global rotation could
+                the list is `num_connected_nodes`. The list may contain `None` where the global rotation could
                 not be computed (either underconstrained system or ill-constrained system).
         """
         lm_params = LevenbergMarquardtParams.CeresDefaults()
@@ -83,10 +84,11 @@ class ShonanRotationAveraging(RotationAveragingBase):
         return wRi_list_consecutive
 
     def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
-        """Run the rotation averaging with arbitrary keys, where each key is a image/pose index.
+        """Run the rotation averaging on a connected graph with arbitrary keys, where each key is a image/pose index.
 
-        Note: run() functions as a wrapper that re-orders keys to prepare a connected graph,
-        w/ N keys ordered [0,...,N-1].
+        Note: run() functions as a wrapper that re-orders keys to prepare a graph w/ N keys ordered [0,...,N-1].
+        All input nodes must belong to a single connected component, in order to obtain an absolute pose for each
+        camera in a single, global coordinate frame.
 
         Args:
             num_images: number of images. Since we have one pose per image, it is also the number of poses.
--- a/gtsfm/averaging/rotation/shonan.py
+++ b/gtsfm/averaging/rotation/shonan.py
@@ -36,17 +36,14 @@ class ShonanRotationAveraging(RotationAveragingBase):
         self._p_min = 5
         self._p_max = 30
 
-    def run_w_consecutive_ordering(
+    def run_with_consecutive_ordering(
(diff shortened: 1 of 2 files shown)


gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,33 @@ class ShonanRotationAveraging(RotationAveragingBase):
...
     def __init__(self) -> None:
+        """
+        Note: `p_min` and `p_max` describe the minimum and maximum relaxation rank.
+        """
         self._p_min = 5
         self._p_max = 30
 
-    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
-        """Run the rotation averaging.
+    def run_w_consecutive_ordering(
```
akshay-krishnan, line 39: I think the name run_with_consecutive_ordering is more readable.

gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,33 @@ class ShonanRotationAveraging(RotationAveragingBase):
...
+        self, num_connected_nodes: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
+    ) -> List[Optional[Rot3]]:
+        """Run the rotation averaging on a connected graph w/ N keys ordered consecutively [0,...,N-1].
+
+        Note: the assumption that Shonan keys are ordered as [0,...,N-1] goes deeper than
+        ShonanAveraging::NrUnknowns(), which just checks the assumption. Several other functions that construct
+        matrices related to the convergence certificate also use the fact that the graph is connected and
+        keys are 0..N-1. There are about 10 places where nrUnknowns is used to dimension a sparse or dense matrix.
+        Modifying GTSAM would require a major philosophical overhaul, so we perform the re-ordering
+        here in a sort of "wrapper". See: https://github.com/borglab/gtsam/issues/784
```
akshay-krishnan, line 49: Why is this note useful here? To someone reading the code, it does not make much sense. NrUnknowns() is not used in this function. Maybe we could add a brief note like "GTSAM requires the nodes to be ordered from [0 ... N-1].  See https://github.com/borglab/gtsam/issues/784 for details. "

gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,33 @@ class ShonanRotationAveraging(RotationAveragingBase):
...
 
         Args:
-            num_images: number of poses.
-            i2Ri1_dict: relative rotations as dictionary (i1, i2): i2Ri1.
+            num_connected_nodes: number of unique connected nodes in the graph (can be far less than the number
```
akshay-krishnan, line 52: could you change to ( <= number of images in the dataset)? its more concise.

gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,33 @@ class ShonanRotationAveraging(RotationAveragingBase):
...
+                of images in the dataset)
+            i2Ri1_dict: relative rotations for each image pair-edge as dictionary (i1, i2): i2Ri1.
```
akshay-krishnan, line 54: We need to be consistent with the usage of nodes and images. "relative rotations for each edge between nodes as dictionary (i1, i2): i2Ri1"

akshay-krishnan, line 54: Do both i1 and i2 have to be less than num_connected_nodes? Could we document that? Do these edges have to form a connected component? Can there be "extra" edges?

gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,33 @@ class ShonanRotationAveraging(RotationAveragingBase):
...
 
         Returns:
-            Global rotations for each camera pose, i.e. wRi, as a list. The number of entries in the list is
-                `num_images`. The list may contain `None` where the global rotation could not be computed (either
-                underconstrained system or ill-constrained system).
+            Global rotations for each **CONNECTED** camera pose, i.e. wRi, as a list. The number of entries in
+                the list is num_connected_nodes`. The list may contain `None` where the global rotation could
+                not be computed (either underconstrained system or ill-constrained system).
```
akshay-krishnan, line 59: If the input edges belong to a connected component, it seems like this function returns wRi for all input nodes, except when the system is underconstrained.

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
-                        Pose3(
-                            i2Ri1,
-                            np.zeros(
-                                3,
-                            ),
-                        ),
-                        noise_model,
-                    )
-                )
+                # ignore translation during rotation averaging
```
akshay-krishnan, line 72: nice comment

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+                i2Ti1 = Pose3(i2Ri1, np.zeros(3))
+                between_factors.append(BetweenFactorPose3(i2, i1, i2Ti1, noise_model))
 
         obj = ShonanAveraging3(between_factors, shonan_params)
 
         initial = obj.initializeRandomly()
         result_values, _ = obj.run(initial, self._p_min, self._p_max)
 
-        return [result_values.atRot3(i) for i in range(num_images)]
+        return [result_values.atRot3(i) if result_values.exists(i) else None for i in range(num_connected_nodes)]
```
akshay-krishnan, line 81: It becomes less readable to have both the if ...  else .. and the for in the same line. Would be better to indent them instead of list composition here.

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+
+    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
+        """Wrapper that re-orders keys to prepare a connected graph, w/ N keys ordered [0,...,N-1].
```
akshay-krishnan, line 84: I think this function also requires documentation like "Runs Shonan averaging on ... ", and then you could add that keys can be arbitrary.

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+
+        Args:
+            num_images: number of poses.
```
akshay-krishnan, line 87: why dont we call it num_poses then? The documentation seems odd.

akshay-krishnan: "number of images" would be okay

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+            i2Ri1_dict: relative rotations for each image pair-edge as dictionary (i1, i2): i2Ri1.
+
+        Returns:
+            Global rotations for each camera pose, i.e. wRi, as a list. The number of entries in the list is
+                `num_images`. The list may contain `None` where the global rotation could not be computed (either
+                underconstrained system or ill-constrained system), or where the camera pose had no valid observation
+                in the input to run().
+        """
+        connected_nodes = set()
+        for (i1, i2) in i2Ri1_dict.keys():
```
akshay-krishnan, line 97: is the .keys() necessary?

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+            connected_nodes.add(i1)
+            connected_nodes.add(i2)
+
+        connected_nodes = sorted(list(connected_nodes))
+
+        # given original index, get back new temporary index
+        reordered_idx_map = {}
+        for i in range(num_images):
+            if i in connected_nodes:
+                reordered_idx_map[i] = connected_nodes.index(i)
```
akshay-krishnan, line 107: could you change to the following? 
```
for (new_idx, idx) in enumerate(connected_nodes): 
   reordered_idx_map[idx] = new_idx
```
I think its simpler and faster.

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +69,58 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+            i2_ = reordered_idx_map[i2]
+            i2Ri1_dict_reordered[(i1_, i2_)] = i2Ri1
+
+        wRi_list_subset = self.run_w_consecutive_ordering(
+            num_connected_nodes=len(connected_nodes), i2Ri1_dict=i2Ri1_dict_reordered
+        )
+
+        wRi_list = [None] * num_images
+        for i in range(num_images):
+            if i in connected_nodes:
```
akshay-krishnan, line 122: please check `if i in reordered_idx_map` instead. Since that is the one being used below. 
Also we might not be losing much readability by `wRi_list[i] = wRi_list_subset[reordered_idx_map[i]]`, but thats upto you.

tests/averaging/rotation/test_shonan.py
```diff
@@ -119,6 +119,37 @@ def test_pickleable(self):
...
 
+    def test_nonconsecutive_indices(self):
+        """Run rotation averaging on a graph with indices that are not ordered as [0,...,N-1].
+
+        Note: Test would fail if Shonan keys were not temporarily re-ordered inside the implementation.
+        See https://github.com/borglab/gtsam/issues/784
+        """
+        num_images = 4
+
+        wTi0 = Pose3(Rot3(), np.array([1, 1, 0]))
```
akshay-krishnan, line 130: this is unused

tests/averaging/rotation/test_shonan.py
```diff
@@ -119,6 +119,37 @@ def test_pickleable(self):
         except TypeError:
             self.fail("Cannot dump rotation averaging object using pickle")
 
+    def test_nonconsecutive_indices(self):
```
akshay-krishnan, line 122: might also be useful to add a test with edges that dont belong to a connected component.

akshay-krishnan (changes requested):

tests/averaging/rotation/test_shonan.py
johnwlambert: So rotation averaging runs after pruning to the largest component in `MultiviewOptimizer`, so that case should already be handled actually (in other words, we won't allow such input).

gtsfm/averaging/rotation/shonan.py
johnwlambert: good point, should be fixed now

gtsfm/averaging/rotation/shonan.py
johnwlambert: technically it's not, but I find it a bit more readable/clearer

gtsfm/averaging/rotation/shonan.py
johnwlambert: nice -- i like that. thanks

gtsfm/averaging/rotation/shonan.py
johnwlambert: good point, updated.

gtsfm/averaging/rotation/shonan.py
johnwlambert: Good point, updated

gtsfm/averaging/rotation/shonan.py
johnwlambert: updated, thanks

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +66,65 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+                `num_images`. The list may contain `None` where the global rotation could not be computed (either
+                underconstrained system or ill-constrained system), or where the camera pose had no valid observation
+                in the input to run().
+        """
+        connected_nodes = set()
+        for (i1, i2) in i2Ri1_dict.keys():
+            connected_nodes.add(i1)
+            connected_nodes.add(i2)
+
+        connected_nodes = sorted(list(connected_nodes))
```
ayushbaid, line 106: any reason to sort them?

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +66,65 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+
+        # given original index, this map gives back a new temporary index, starting at 0
+        reordered_idx_map = {}
+        for (new_idx, i) in enumerate(connected_nodes):
+            reordered_idx_map[i] = new_idx
+
+        # now, re-order all of the indices
```
ayushbaid, line 113: i think the statement "map the original indices to reordered indices" is better as we have already performed the reordering operation.

gtsfm/averaging/rotation/shonan.py
ayushbaid: We can directly loop over the `connected_nodes`?

```
for remapped_i, original_i in enumerate(connected_nodes):
    wRi_list[original_i] = wRi_list_subset[remapped_i]
```

gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,30 @@ class ShonanRotationAveraging(RotationAveragingBase):
...
     def __init__(self) -> None:
+        """
+        Note: `p_min` and `p_max` describe the minimum and maximum relaxation rank.
+        """
         self._p_min = 5
         self._p_max = 30
 
-    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
-        """Run the rotation averaging.
+    def run_with_consecutive_ordering(
```
ayushbaid, line 39: should be a private function?

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +66,65 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+                i2Ti1 = Pose3(i2Ri1, np.zeros(3))
+                between_factors.append(BetweenFactorPose3(i2, i1, i2Ti1, noise_model))
 
         obj = ShonanAveraging3(between_factors, shonan_params)
 
         initial = obj.initializeRandomly()
         result_values, _ = obj.run(initial, self._p_min, self._p_max)
 
-        return [result_values.atRot3(i) for i in range(num_images)]
+        wRi_list_consecutive = [None] * num_connected_nodes
```
ayushbaid, line 78: I dont see the need to name the variable as `consecutive`?

gtsfm/averaging/rotation/shonan.py
```diff
@@ -30,20 +30,30 @@ class ShonanRotationAveraging(RotationAveragingBase):
     """Performs Shonan rotation averaging."""
 
     def __init__(self) -> None:
+        """
+        Note: `p_min` and `p_max` describe the minimum and maximum relaxation rank.
+        """
```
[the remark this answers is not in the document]
johnwlambert: I think it can be helpful to alert the reader as to what the hyperparameters below signify.

gtsfm/averaging/rotation/shonan.py
johnwlambert: it should -- good point. i'll updated

gtsfm/averaging/rotation/shonan.py
johnwlambert: I guess I didn't want to confuse the reader if they saw the same variable name in `run()` and `__run_with_consecutive_ordering()`, since they symbolize very different quantities

gtsfm/averaging/rotation/shonan.py
johnwlambert: mainly just to preserve a monotonically increasing property, so no nodes got swapped out of relative order. Not strictly necessary, but I thought might convey to the reader that there is an implicit relative order.

gtsfm/averaging/rotation/shonan.py
johnwlambert: good point -- thanks. updated

gtsfm/averaging/rotation/shonan.py
johnwlambert: Hmm didn't quite follow this here -- so the distinction is about underconstrained vs. ill-constrained?

gtsfm/averaging/rotation/shonan.py
johnwlambert: sure, updated

gtsfm/averaging/rotation/shonan.py
johnwlambert: sure, updated

tests/averaging/rotation/test_shonan.py
johnwlambert: good point -- I moved it to an inline comment now

gtsfm/averaging/rotation/shonan.py
johnwlambert: @akshay-krishnan  So technically Shonan will still return results if the graph is disconnected. For example, if we give i1Ri0, and i3Ri2, Shonan will give back 4 absolute rotations. But they won't make sense, e.g.
```
[R: [
	-0.00275314, -0.883536, 0.468356;
	-0.605804, -0.371157, -0.703735;
	0.795609, -0.285669, -0.534228
]
, R: [
	-0.00275314, -0.883536, 0.468356;
	-0.605804, -0.371157, -0.703735;
	0.795609, -0.285669, -0.534228
]
, R: [
	-0.112206, -0.929598, 0.351081;
	-0.94733, 0.206725, 0.244601;
	-0.299958, -0.305144, -0.903832
]
, R: [
	-0.112206, -0.929598, 0.351081;
	-0.94733, 0.206725, 0.244601;
	-0.299958, -0.305144, -0.903832
]
]
```

johnwlambert:
```python
        wTi0 = Pose3(Rot3(), np.array([1, 1, 0]))
        wTi1 = Pose3(Rot3(), np.array([3, 1, 0]))
        wTi2 = Pose3(Rot3(), np.array([3, 3, 0]))
        wTi3 = Pose3(Rot3(), np.array([1, 3, 0]))

        # generate i2Ri1 rotations
        # (i1,i2) -> i2Ri1
        i2Ri1_input = {
            (0, 1): wTi1.between(wTi0).rotation(),
            (2, 3): wTi3.between(wTi2).rotation(),
        }
```

gtsfm/averaging/rotation/shonan.py
```diff
@@ -56,23 +67,64 @@ def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]
...
+
+        Args:
+            num_images: number of images. Since we have one pose per image, it is also the number of poses.
+            i2Ri1_dict: relative rotations for each image pair-edge as dictionary (i1, i2): i2Ri1.
+
+        Returns:
+            Global rotations for each camera pose, i.e. wRi, as a list. The number of entries in the list is
+                `num_images`. The list may contain `None` where the global rotation could not be computed (either
+                underconstrained system or ill-constrained system), or where the camera pose had no valid observation
+                in the input to run().
```
akshay-krishnan, line 101: in other words, no edges to or from the camera in i2Ri1_dict? Maybe that is more clear, or its just me.

tests/averaging/rotation/test_shonan.py
akshay-krishnan: okay

gtsfm/averaging/rotation/shonan.py
akshay-krishnan: I dont know the difference... what is it? 

I did not know num_images was num_connected_nodes. I thought it was num_nodes (>= num_connected_nodes). Hence the confusion.

tests/averaging/rotation/test_shonan.py
```diff
@@ -119,6 +119,37 @@ def test_pickleable(self):
...
+            (1, 3): wTi3.between(wTi1).rotation(),
+        }
+
+        wRi_expected = [None, wTi1.rotation(), wTi2.rotation(), wTi3.rotation()]
+
+        wRi_computed = self.obj.run(num_images, i2Ri1_input)
+
+        self.assertTrue(
+            geometry_comparisons.compare_rotations(wRi_computed, wRi_expected, angular_error_threshold_degrees=0.1)
+        )
```
akshay-krishnan: nit: too many empty lines.

akshay-krishnan (approved):

tests/averaging/rotation/test_shonan.py
johnwlambert: good point -- i'll clean that up

johnwlambert:
@akshay-krishnan so indeed ` num_images >= num_connected_nodes.` is that not clear from the docstrings?

suppose I have nodes 0,1,2,3,4,5

but the connected component is 2,3,4

so num_images is 6, but num_connected_nodes is 3, here

gtsfm/averaging/rotation/shonan.py
ayushbaid: cool

gtsfm/averaging/rotation/shonan.py
ayushbaid: ok cool.

ayushbaid: ok cool.

ayushbaid (approved):
LGTM!

PR #194, opened 2021-06-23
Loader - get image dictionary

codyly:
Add `get_image_dictionary` to `loader_base` to get image dictionary directly from loader.


gtsfm/loader/loader_base.py
```diff
@@ -143,3 +143,12 @@ def get_valid_pairs(self) -> List[Tuple[int, int]]:
                     indices.append((idx1, idx2))
 
         return indices
+
+    def get_image_dictionary(self) -> Dict[int, Image]:
```
johnwlambert, line 147: looks good @codyly . Can you add a unit test on the Lund door set 1 data in tests/data?

gtsfm/loader/loader_base.py
```diff
@@ -143,3 +143,12 @@ def get_valid_pairs(self) -> List[Tuple[int, int]]:
...
+        """Get the dictionary containing all loaded images. The keys are the images' indices.
+
+        Returns:
+            Dict[int, Image]: image dictionary
```
johnwlambert, line 151: No need to repeat the type her, so I would prefer
```python

Returns:
    dictionary mapping image indices [0,...,N-1] to their associated image
```

gtsfm/loader/loader_base.py
johnwlambert: actually, we already have this functionality in 
```
            image_graph=loader.create_computation_graph_for_images(),
```
right?

johnwlambert: so I dont think we need this PR, after all, if we are working within Dask

codyly: Ah...I think so. I am not very familiar with Dask so I didn't realize it.

johnwlambert: where are you hoping to use this function @codyly ?

codyly: in the densify implementation at #195

codyly: After we get `sfm_result` in `run_scene_optimizer`, `sfm_result` together with `images_dict` will be passed to `MVSPatchmatchnet.densify` (implemented in #195)

codyly: @johnwlambert As the `MVSBase` defines the densify API as
```def densify(self, images: Dict[int, Image], sfm_result: GtsfmData) -> np.ndarray:```
Is there a simple way to transfter image list from `loader.create_computation_graph_for_images()` to image dictionary, such as
```dict(zip(list(range(len(loader))), loader.create_computation_graph_for_images()))```

**Comment by ayushbaid** [comment]
@codyly do we need this PR?

**Comment by codyly** [comment]
@ayushbaid @johnwlambert perhaps we need to find a proper approach to invoke PatchmatchNet in gtsfm runners. Currently it can be
```python
from gtsfm.densify.mvs_patchmatchnet import MVSPatchmatchNet

img_dict = {i: self.loader.get_image(i) for i in range(len(self.loader))}
points = MVSPatchmatchNet().densify(img_dict, sfm_result)
```

ayushbaid:
> @ayushbaid @johnwlambert perhaps we need to find a proper approach to invoke PatchmatchNet in gtsfm runners. Currently it can be
> ```python
> from gtsfm.densify.mvs_patchmatchnet import MVSPatchmatchNet
> 
> img_dict = {i: self.loader.get_image(i) for i in range(len(self.loader))}
> points = MVSPatchmatchNet().densify(img_dict, sfm_result)
> ```

Yep this should be good!

PR #195, opened 2021-06-23
Inference using PatchmatchNet

codyly:
Inference using PatchmatchNet


gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,247 @@
...
+PATCHMATCHNET_WEIGHTS_PATH = "thirdparty/patchmatchnet/checkpoints/model_000007.ckpt"
+DEFAULT_VIEW_NUMBER = 5
+DEFAULT_GEOMETRIC_PIXEL_THRESH = 1.0
+DEFAULT_GEOMETRIC_DEPTH_THRESH = 0.01
+DEFAULT_PHOTOMETRIC_THRESH = 0.8
+DEFAULT_BATCH_SIZE = 1
+DEFAULT_WORKERS_NUMBER = 4
+
+
+class MVSPatchmatchNet(MVSBase):
```
johnwlambert, line 35: thanks for the PR, @codyly.

Do you mind adding a unit test for this?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,247 @@
...
+        photo_thres: float,
+    ) -> np.ndarray:
+        """Filter depth map and get filtered dense point cloud
+        Ref: Wang et al. https://github.com/FangjinhuaWang/PatchmatchNet/blob/main/eval.py
+
+        Args:
+            dataset (PatchmatchNetData): an instance of PatchmatchData as the inference dataset
+            depth_list (Dict[int, np.ndarray]): list of 2D depth map (H, W) from each view
+            confidence_list (Dict[int, np.ndarray]): list of 2D confidence map (H, W) from each view
+            geo_pixel_thres (float): geometric pixel threshold
```
johnwlambert, line 164: no need to repeat the types down here, since they're already clear from the function signature

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,247 @@
...
+        # for each reference view and the corresponding source views
+        for ref_view, src_views in pair_data:
+            # load the camera parameters
+            ref_intrinsics, ref_extrinsics = dataset.get_camera_params(ref_view)
+
+            # load the reference image
+            ref_img = dataset.get_image(ref_view)
+            # load the estimated depth of the reference view
+            ref_depth_est = depth_list[ref_view][0]
+            # load the photometric mask of the reference view
```
johnwlambert, line 187: how do we define "photometric mask"? would you mind adding that as an inline comment?

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: @codyly could you take another look at my comments here?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,247 @@
...
+        return dense_point_cloud
+
+    def filter_depth(
+        self,
+        dataset: PatchmatchNetData,
+        depth_list: Dict[int, np.ndarray],
+        confidence_list: Dict[int, np.ndarray],
+        geo_pixel_thres: float,
+        geo_depth_thres: float,
+        photo_thres: float,
```
johnwlambert, line 155: Prefer thresh to thres

johnwlambert:
Hi @codyly, could you pull master into this branch, if this is the next one to merge?

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: marking as resolved, thanks

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+                for idx, depth_est, photometric_confidence in zip(
+                    ids, outputs["refined_depth"]["stage_0"], outputs["photometric_confidence"]
+                ):
+
+                    idx = idx.cpu().numpy().tolist()
+                    depth_est_list[idx] = depth_est.copy()
+                    confidence_est_list[idx] = photometric_confidence.copy()
+
+                logger.info(
+                    "[Densify::PatchMatchNet] Iter {}/{}, time = {:.3f}".format(
```
johnwlambert, line 131: we usually use the python logger formatting:
```python
logger.info("[Densify::PatchMatchNet] Iter %d/%d, time = %.3f", batch_idx + 1, len(loader), time.time() - start_time)
```

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+    def filter_depth(
+        self,
+        dataset: PatchmatchNetData,
+        depth_list: Dict[int, np.ndarray],
+        confidence_list: Dict[int, np.ndarray],
+        geo_pixel_thresh: float,
+        geo_depth_thresh: float,
+        photo_thresh: float,
+    ) -> np.ndarray:
+        """Filter depth map and get filtered dense point cloud
```
johnwlambert, line 157: what do we filter based on? I think some more details are needed about the criteria / methodology.

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+            depth_list=depth_est_list,
+            confidence_list=confidence_est_list,
+            geo_pixel_thresh=thresholds[0],
+            geo_depth_thresh=thresholds[1],
+            photo_thresh=thresholds[2],
+        )
+
+        return dense_point_cloud
+
+    def filter_depth(
```
johnwlambert, line 148: this function is pretty long (over 100 lines). I would like us to try to stick to 10-20 line functions where we can. Is there any intuitive way to decompose it that you can think of?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+            # Initialize coordinate grids
+            height, width = depth_est_averaged.shape[:2]
+            x, y = np.meshgrid(np.arange(0, width), np.arange(0, height))
+
+            # Get valid points filtered by photometric and geometric thresholds
+            valid_points = final_mask
+            x, y, depth = x[valid_points], y[valid_points], depth_est_averaged[valid_points]
+
+            # Get the point coordinates in world frame
+            xyz_ref = np.matmul(np.linalg.inv(ref_intrinsics), np.vstack((x, y, np.ones_like(x))) * depth)
```
johnwlambert, line 238: i think 
```python
def cart_to_homogenous(non_homogenous_coordinates: np.ndarray,) -> Optional[np.ndarray]:
    """Convert cartesian coordinates to homogenous system (by appending a column of ones). 
   Args: 
         non_homogenous_coordinates: 3d non-homogenous coordinates, of shape Nx3.
    
    Returns: 
          4d homogenous coordinates, of shape Nx4. 

    Raises: TypeError: if input is not 2 dimensional. 
     """
     n = non_homogenous_coordinates.shape[0]
     return np.hstack([non_homogenous_coordinates, np.ones((n, 1))])  

# get the point coordinates inside the reference view's camera frame
xyz_ref = np.linalg.inv(ref_intrinsics) @ np.vstack([u, v, np.ones_like(u])) * depth)
# get the point coordinates inside the world frame
xyz_world = wTref @ cart_to_homogeneous(xyz_ref)
```
is more readabile

johnwlambert: let's call it u, v, not x, y?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+        return dense_point_cloud
+
+    def filter_depth(
+        self,
+        dataset: PatchmatchNetData,
+        depth_list: Dict[int, np.ndarray],
+        confidence_list: Dict[int, np.ndarray],
+        geo_pixel_thresh: float,
+        geo_depth_thresh: float,
+        photo_thresh: float,
```
johnwlambert, line 155: is the photometric threshold actually comparing RGB intensities?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+        confidence_est_list = {}
+        with torch.no_grad():
+            for batch_idx, sample in enumerate(loader):
+                start_time = time.time()
+
+                # Check if cuda devices is supported, and store the inference data to the target device
+                if torch.cuda.is_available():
+                    sample_device = tocuda(sample)
+                else:
+                    sample_device = copy.copy(sample)
```
johnwlambert, line 106: why do we need a copy?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+        loader = DataLoader(
+            dataset=dataset,
+            batch_size=DEFAULT_BATCH_SIZE,
+            shuffle=False,
+            num_workers=DEFAULT_WORKERS_NUMBER,
+            drop_last=False,
+        )
+
+        model = PatchmatchNet(
+            patchmatch_interval_scale=[0.005, 0.0125, 0.025],
```
johnwlambert, line 76: nit: these seem like magic numbers / defaults that should go at the top of the file

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+from thirdparty.patchmatchnet.models.net import PatchmatchNet
+from thirdparty.patchmatchnet.utils import tensor2numpy, tocuda
+from thirdparty.patchmatchnet.eval import check_geometric_consistency
+
+logger = logger_utils.get_logger()
+
+torch.backends.cudnn.benchmark = True
+PATCHMATCHNET_WEIGHTS_PATH = "thirdparty/patchmatchnet/checkpoints/model_000007.ckpt"
+DEFAULT_VIEW_NUMBER = 5
+DEFAULT_GEOMETRIC_PIXEL_THRESH = 1.0
```
johnwlambert, line 28: maybe we can add an inline comment explaining where these threshold values come from? paper / link / url?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+import torch.nn as nn
+from torch.utils.data import DataLoader
+
+from gtsfm.common.image import Image
+from gtsfm.common.gtsfm_data import GtsfmData
+from gtsfm.densify.mvs_base import MVSBase
+from gtsfm.densify.patchmatchnet_data import PatchmatchNetData
+from gtsfm.utils import logger as logger_utils
+from thirdparty.patchmatchnet.models.net import PatchmatchNet
+from thirdparty.patchmatchnet.utils import tensor2numpy, tocuda
```
johnwlambert, line 20: we only import classes directly from modules, not function names

gtsfm/densify/mvs_patchmatchnet.py
codyly: PatchmatchNet only produces the depth maps and the confidence maps of each view. Some points should be filtered out because:
1. too low confidence (not fit the RGB image well), filtered by `photo_thresh`
2. cannot keep a consistent geometry among depth maps of ref views and source views, filtered by `geo_*_thresh`

gtsfm/densify/mvs_patchmatchnet.py
codyly: shall I put `cart_to_homogenous` into `utils/geometry_comparison`?

johnwlambert: Let's just put it as a helper in this file for now. I don't think it's so related to geometric comparisons

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: Cool, how about we say something like:
```python
"""Create a dense point cloud by filtering depth maps based on estimated confidence maps and consistent geometry.

Consistent geometry is defined as X% of Y metric matching as ABC
"""
```

gtsfm/densify/mvs_patchmatchnet.py
codyly: I keep it because in the original repo the `filter_depth` method looks just like this. So I guess in this way we can easily match with the original code. And there are a lot of comments in the function, too.

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,251 @@
...
+import time
+from typing import Dict, List
+
+import numpy as np
+import torch
+import torch.nn as nn
+from torch.utils.data import DataLoader
+
+from gtsfm.common.image import Image
+from gtsfm.common.gtsfm_data import GtsfmData
```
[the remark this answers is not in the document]
johnwlambert: marking as resolved, thanks

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,280 @@
...
+
+from gtsfm.common.gtsfm_data import GtsfmData
+from gtsfm.common.image import Image
+from gtsfm.densify.mvs_base import MVSBase
+import gtsfm.densify.mvs_utils as mvs_utils
+from gtsfm.densify.patchmatchnet_data import PatchmatchNetData
+from gtsfm.utils import logger as logger_utils
+from thirdparty.patchmatchnet.models.net import PatchmatchNet
+import thirdparty.patchmatchnet.utils as patchmatchnet_utils
+import thirdparty.patchmatchnet.eval as patchmatchnet_eval_funcs
```
johnwlambert, line 21: `patchmatchnet_eval_funcs` seems too long to me. How about just `mvs_eval` or `patchmatchnet_eval`

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: marking as resolved, thanks

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: thanks, marking as resolved

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: resolved, thanks

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: I saw you removed the copy, but I just want to confirm that the copy wasnt needed. Could you post a screenshot of the result?

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: how about `photometric_conf_thresh`, or just `conf_thresh`?

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: much better now, thanks

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: marking as resolved, thanks

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
...
+                # color the track point with (r, g, b) = (255, 255, 255)
+                self._img_dict[i].value_array[int(v + 0.5), int(u + 0.5), :] = 255
+
+                track_to_add.add_measurement(idx=i, m=[u, v])
+            sfm_result.add_track(track_to_add)
+
+        return sfm_result
+
+    def test_patchmatchnet_result(self) -> None:
+        """Test whether the result point cloud calculated by PatchmatchNet correctly reflects the original shape"""
```
johnwlambert, line 101: could you share an image of the result depth maps and point clouds to explain this a bit more?

codyly: Designed model for unit test
dummy_gt

Generated images from 8 default cameras
gen_images

Output point cloud

codyly: do we need to output depth maps?

gtsfm/densify/mvs_patchmatchnet.py
codyly: make sense, if we don't have to follow the original name in the paper

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,280 @@
...
+
+logger = logger_utils.get_logger()
+
+torch.backends.cudnn.benchmark = True
+PATCHMATCHNET_WEIGHTS_PATH = "thirdparty/patchmatchnet/checkpoints/model_000007.ckpt"
+
+# all default values are assigned by Wang et al. https://github.com/FangjinhuaWang/PatchmatchNet/blob/main/eval.py
+DEFAULT_BATCH_SIZE = 1
+DEFAULT_VIEW_NUMBER = 5
+DEFAULT_WORKERS_NUMBER = 4
```
johnwlambert, line 31: @codyly DEFAULT_NUM_WORKERS sounds better to me

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
...
+import numpy as np
+from gtsam import Cal3_S2, PinholeCameraCal3_S2, Point3
+from gtsam.examples import SFMdata
+
+from gtsfm.common.image import Image
+from gtsfm.common.gtsfm_data import GtsfmData, SfmTrack
+from gtsfm.densify.mvs_patchmatchnet import MVSPatchmatchNet
+
+# set the default image size as 800x600, with 3 channels
+DEFAULT_IMAGE_W = 400
```
johnwlambert, line 23: 800x600 or 400x300? Code seems to contradict comment

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
...
+DEFAULT_IMAGE_H = 300
+DEFAULT_IMAGE_C = 3
+
+# set default track points, the coordinates are in the world frame
+DEFAULT_NUM_TRACKS = 30
+DEFAULT_TRACK_POINTS = [Point3(0, 0.5 * float(i), float(i)) for i in range(DEFAULT_NUM_TRACKS)]
+
+# set default camera intrinsics
+DEFAULT_CAMERA_INTRINSICS = Cal3_S2(
```
johnwlambert, line 32: Cal3Bundler is our default

johnwlambert (changes requested):

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,280 @@
...
+                )
+                geo_mask_sum += geo_mask.astype(np.int32)
+                all_srcview_depth_ests.append(depth_reprojected)
+
+            depth_est_averaged = (sum(all_srcview_depth_ests) + ref_depth_est) / (geo_mask_sum + 1)
+            # Valid points requires at least 3 source views validated under geometric threshoulds
+            geo_mask = geo_mask_sum >= DEFAULT_MINIMUM_CONSISTENT_VIEW_NUMBER
+
+            # Combine geometric mask and photometric mask
+            final_mask = np.logical_and(photo_mask, geo_mask)
```
johnwlambert, line 251: i probably prefer the name `joint_mask` instead of `final_mask`, since it's more descriptive

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,280 @@
...
+DEFAULT_PHOTOMETRIC_THRESH = 0.8
+
+DEFAULT_INTERVAL_SCALE = [0.005, 0.0125, 0.025]
+DEFAULT_PROPAGATION_RANGE = [6, 4, 2]
+DEFAULT_ITERATION_NUMBER = [1, 2, 2]
+DEFAULT_SAMPLE_NUMBER = [8, 8, 16]
+DEFAULT_PROPAGATE_NEIGHBORS = [0, 8, 16]
+DEFAULT_EVALUATE_NEIGHBORS = [9, 9, 9]
+
+DEFAULT_MINIMUM_CONSISTENT_VIEW_NUMBER = 3
```
johnwlambert, line 44: I prefer `DEFAULT_MIN_NUM_CONSISTENT_VIEWS`. Can you add a 1-liner inline comment explaining what this means, too?

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,280 @@
...
+logger = logger_utils.get_logger()
+
+torch.backends.cudnn.benchmark = True
+PATCHMATCHNET_WEIGHTS_PATH = "thirdparty/patchmatchnet/checkpoints/model_000007.ckpt"
+
+# all default values are assigned by Wang et al. https://github.com/FangjinhuaWang/PatchmatchNet/blob/main/eval.py
+DEFAULT_BATCH_SIZE = 1
+DEFAULT_VIEW_NUMBER = 5
+DEFAULT_WORKERS_NUMBER = 4
+
```
johnwlambert, line 32: inline 1-liner explaining what `DEFAULT_GEOMETRIC_PIXEL_THRESH` means mathematically would be great

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
...
+
+        return sfm_result
+
+    def test_patchmatchnet_result(self) -> None:
+        """Test whether the result point cloud calculated by PatchmatchNet correctly reflects the original shape"""
+
+        # test the dense point cloud shape is correct
+        n, d = self._dense_points.shape
+
+        self.assertTrue(n > 0 and d == 3)
```
johnwlambert, line 106: what value of `n` do we expect @codyly?

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
+"""Unit tests for using Patchmatchnet method for dense point cloud reconstruction
+
+In this unit test, we make a simple scenario that eight cameras are around a circle (radius = 40.0).
+Every camera's pose is towards the center of the circle (0, 0, 0).
+
+There are 30 toy points uniformly located at the same line:
+  /  z - 2y = 0
```
johnwlambert, line 7: i prefer the syntax `ax + by + cz + d = 0`. that's what we use elsewhere. So maybe we could say
```python
#     0x + -2y + 1z = 0, i.e. x=0
```

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
+"""Unit tests for using Patchmatchnet method for dense point cloud reconstruction
```
johnwlambert, line 1: nit: missing period at end of line

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
...
+    def test_patchmatchnet_result(self) -> None:
+        """Test whether the result point cloud calculated by PatchmatchNet correctly reflects the original shape"""
+
+        # test the dense point cloud shape is correct
+        n, d = self._dense_points.shape
+
+        self.assertTrue(n > 0 and d == 3)
+
+        # calculate mean relative error for z - 2y = 0 as avg(|(z-2y)/z|)
+        yz_error = np.abs(
```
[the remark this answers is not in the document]
johnwlambert: why not say
```python
x = self._dense_points[:, 0]
y = self._dense_points[:, 1]
z = self._dense_points[:, 2]
# per the line `0x + -2y + 1z = 0`
yz_errors = -2 * y + z
mean_yz_error = yz_errors.mean()

# should be close to 0
x_error = np.abs(x).mean()
```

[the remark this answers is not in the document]
johnwlambert: why use 
```
|(z-2y)/z|
```
isn't that (z - 2y) * z = 0? i dont understand the significance of that quantity

tests/densify/test_mvs_patchmatchnet.py
codyly: But `SFMdata.createPoses` in `gtsam/examples/SFMdata.py` requires `Cal3_S2`

tests/densify/test_mvs_patchmatchnet.py
codyly: I think the size of the reconstructed dense point cloud is not predictable.

tests/densify/test_mvs_patchmatchnet.py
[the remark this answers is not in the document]
codyly: Because z and y can be large, I revised the code to test the slope between y and z

tests/densify/test_mvs_patchmatchnet.py
codyly: I think the number of output points cannot be precisely estimated

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -0,0 +1,288 @@
...
+            DEFAULT_CONFIDENCE_THRESH,
+        ],
+    ) -> np.ndarray:
+        """Get dense point cloud using PatchmatchNet from GtsfmData. The method implements the densify method in MVSBase
+        Ref: Wang et al. https://github.com/FangjinhuaWang/PatchmatchNet/blob/main/eval.py
+
+        Args:
+            images: image dictionary obtained from loaders
+            sfm_result: result of GTSFM after bundle adjustment
+            num_views: maximum number of views, containing 1 reference view and (num_views-1) source views
```
johnwlambert, line 70: i think this should be `max_num_views` to be consistent with function signature @codyly ?

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: resolved, thanks

tests/densify/test_mvs_patchmatchnet.py
```diff
@@ -0,0 +1,120 @@
...
+from gtsfm.densify.mvs_patchmatchnet import MVSPatchmatchNet
+
+# set the default image size as 800x600, with 3 channels
+DEFAULT_IMAGE_W = 400
+DEFAULT_IMAGE_H = 300
+DEFAULT_IMAGE_C = 3
+
+# set default track points, the coordinates are in the world frame
+DEFAULT_NUM_TRACKS = 30
+DEFAULT_TRACK_POINTS = [Point3(0, 0.5 * float(i), float(i)) for i in range(DEFAULT_NUM_TRACKS)]
```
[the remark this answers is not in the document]
johnwlambert: resolved, thanks

gtsfm/densify/mvs_patchmatchnet.py
johnwlambert: resolved, thanks

PR #200, opened 2021-06-27
add extra unit test on compute_relative_rotation_angle

johnwlambert:
Add a unit test to make sure relative rotation angles are computed correctly.

The test uses rotation matrices generated from known Euler angles using scipy.spatial.transform.

### DIFF
--- a/gtsfm/utils/geometry_comparisons.py
+++ b/gtsfm/utils/geometry_comparisons.py
@@ -1,371 +1,359 @@
-"""Utility functions for comparing different types related to geometry.
+"""Unit tests for comparison functions for geometry types.
 
 Authors: Ayush Baid, John Lambert
 """
-from typing import List, Optional
+import unittest
+from typing import List
+from unittest.mock import patch
 
 import numpy as np
-from gtsam import Point3, Pose3, Pose3Pairs, Rot3, Similarity3, Unit3
+from gtsam import Cal3_S2, Point3, Pose3, Rot3, Similarity3, Unit3
+from gtsam.examples import SFMdata
 from scipy.spatial.transform import Rotation
 
-from gtsfm.utils.logger import get_logger
+import gtsfm.utils.geometry_comparisons as geometry_comparisons
 
-EPSILON = np.finfo(float).eps
+import tests.data.sample_poses as sample_poses
 
-logger = get_logger()
+POSE_LIST = SFMdata.createPoses(Cal3_S2())
 
-
-def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
-    """Aligns the list of rotations to the reference list by shifting origin.
-
-    TODO (John): replace later with Karcher mean to account for noisy estimates
-
-    Args:
-        aRi_list: reference rotations in frame "a" which are the targets for alignment
-        bRi_list: input rotations which need to be aligned to frame "a"
-
-    Returns:
-        aRi_list_: transformed input rotations previously "bRi_list" but now which
-            have the same origin as reference (now living in "a" frame)
-    """
-    aRi0 = aRi_list[0]
-
-    bRi0 = bRi_list[0]
-    i0Rb = bRi0.inverse()
-
-    # origin_transform -- map the origin of the input list to the reference list
-    aRb = aRi0.compose(i0Rb)
-
-    # apply the coordinate shift to all entries in input
-    return [aRb.compose(bRi) for bRi in bRi_list]
+ROT3_EULER_ANGLE_ERROR_THRESHOLD = 1e-2
+POINT3_RELATIVE_ERROR_THRESH = 1e-1
+POINT3_ABS_ERROR_THRESH = 1e-2
 
 
-def align_poses_sim3_ignore_missing(
-    aTi_list: List[Optional[Pose3]], bTi_list: List[Optional[Pose3]]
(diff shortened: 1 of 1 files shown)


tests/utils/test_geometry_comparisons.py
```diff
@@ -217,6 +219,24 @@ def test_compute_relative_rotation_angle(self):
 
         np.testing.assert_allclose(computed_deg, expected_deg, rtol=1e-3, atol=1e-3)
 
+    def test_compute_relative_rotation_angle_euler(self) -> None:
+        """"Sample 100 rotations about each individual Euler angle, and ensure Scipy and GTSAM result match."""
+        num_samples = 100
+        for axis in ["x", "y", "z"]:
+            for _ in range(num_samples):
```
ayushbaid, line 226: nit: I'd prefer the loops to be swapped, i.e. loop over samples first and then the axes for each sample.

tests/utils/test_geometry_comparisons.py
```diff
@@ -217,6 +219,24 @@ def test_compute_relative_rotation_angle(self):
...
+                theta = np.random.rand() * 360
+                R_1 = Rot3(Rotation.from_euler(axis, 0, degrees=True).as_matrix())
+                R_2 = Rot3(Rotation.from_euler(axis, theta, degrees=True).as_matrix())
```
ayushbaid, line 229: Why do we just test on the rotations with non-zero value on just one axis?

tests/utils/test_geometry_comparisons.py
```diff
@@ -217,6 +219,24 @@ def test_compute_relative_rotation_angle(self):
 
         np.testing.assert_allclose(computed_deg, expected_deg, rtol=1e-3, atol=1e-3)
 
+    def test_compute_relative_rotation_angle_euler(self) -> None:
```
ayushbaid, line 222: Is there any specific bug which we catch from this test? Or does this test help increase coverage?

johnwlambert: thanks for the review Ayush. was just trying to increase coverage.

At one point I was wondering if there was a bug here, but there wasn't. But I figured i'd add the extra test anyways.

tests/utils/test_geometry_comparisons.py
johnwlambert: just to make sure i understood your question -- you were wondering why i am generating a second rotation with only a single Euler angle?

johnwlambert: i was doing that just because it made the axis-angle computation simple.

tests/utils/test_geometry_comparisons.py
ayushbaid: Got it. Makes sense!

tests/utils/test_geometry_comparisons.py
ayushbaid: Do you think we miss out on some piece of logic when we have non-zero angle in just one axis?

I think we should add one more test which is very comprehensive (i.e. non-zero values in each axis for both R1 and R2)

tests/utils/test_geometry_comparisons.py
```diff
@@ -217,6 +219,24 @@ def test_compute_relative_rotation_angle(self):
...
+        """"Sample 100 rotations about each individual Euler angle, and ensure Scipy and GTSAM result match."""
+        num_samples = 100
```
akshay-krishnan, line 224: I think we can reduce num_samples. I found the scipy rotation function to be very time consuming when I used it before. I think num_samples = 10 or 5 should suffice.

tests/utils/test_geometry_comparisons.py
```diff
@@ -217,6 +219,24 @@ def test_compute_relative_rotation_angle(self):
...
+        for axis in ["x", "y", "z"]:
+            for _ in range(num_samples):
+                theta = np.random.rand() * 360
+                R_1 = Rot3(Rotation.from_euler(axis, 0, degrees=True).as_matrix())
```
akshay-krishnan, line 228: Why are we using scipy? Doesn’t gtsam rot3 have a constructor from axis angle?

akshay-krishnan (commented):
Sorry for the late review but I had a couple important questions.

tests/utils/test_geometry_comparisons.py
johnwlambert: indeed GTSAM does. Scipy is just a reliable package (and a dependency we already use) so I figured it could be useful to add more coverage

tests/utils/test_geometry_comparisons.py
johnwlambert: Interesting -- maybe they've pushed some optimizations since.

The test completes in 0.0569 sec (56 milliseconds), so I think the speed is ok currently.

tests/utils/test_geometry_comparisons.py
akshay-krishnan: Are you implying GTSAM isn’t reliable? ;)

akshay-krishnan: I’m not concerned about the dependency, I’m concerned about doing extra work in code. Constructing the scipy rotation is unnecessary. Whoever reads this code in future would just stare and think “why would the author do that…”

travisdriver:
@johnwlambert Can we close this?

PR #202, opened 2021-06-29
make runner for colmap loader configurable from CLI

johnwlambert:
Allowing passing parameters like number of workers and desired config file to the runner that ingests COLMAP-formatted data.


akshay-krishnan (approved):

ayushbaid (approved):
LGTM

PR #203, opened 2021-06-29
Produce report from each two-view estimation result to allow in analysis

johnwlambert:
- Specify estimation threshold and evaluation threshold for epipolar distances inside the config, instead of within each file, to make experimentation easier.

- Reject two-view results where the inlier ratio is absurdly low (estimated model not consistent with much of the data)

- Create a new class called `TwoViewEstimationReport` to assist with the objectives above, in the spirit of COLMAP's `Report` struct [here](https://github.com/colmap/colmap/blob/9f3a75ae9c72188244f2403eb085e51ecf4397a8/src/optim/ransac.h#L78):

```c++
template <typename Estimator, typename SupportMeasurer = InlierSupportMeasurer,
          typename Sampler = RandomSampler>
class RANSAC {
 public:
  struct Report {
    EIGEN_MAKE_ALIGNED_OPERATOR_NEW

    // Whether the estimation was successful.
    bool success = false;

    // The number of RANSAC trials / iterations.
    size_t num_trials = 0;

    // The support of the estimated model.
    typename SupportMeasurer::Support support;

    // Boolean mask which is true if a sample is an inlier.
    std::vector<char> inlier_mask;

    // The estimated model.
    typename Estimator::M_t model;
  };
```

### DIFF
--- a/gtsfm/configs/deep_front_end.yaml
+++ b/gtsfm/configs/deep_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: False
+  save_two_view_correspondences_viz: True
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
@@ -25,6 +25,7 @@ SceneOptimizer:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: True
       estimation_threshold_px: 4
+      min_allowed_inlier_ratio_est_model: 0.1
 
   multiview_optimizer:
     _target_: gtsfm.multi_view_optimizer.MultiViewOptimizer
--- a/gtsfm/configs/deep_front_end.yaml
+++ b/gtsfm/configs/deep_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: True
+  save_two_view_correspondences_viz: False
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
@@ -14,7 +14,6 @@ SceneOptimizer:
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
     eval_threshold_px: 3 # in px
-    estimation_threshold_px: 4 # for homography estimator
     min_num_inliers_acceptance: 15
 
     matcher:
@@ -24,7 +23,7 @@ SceneOptimizer:
     verifier:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: True
-      estimation_threshold_px: 4
+      estimation_threshold_px: 4 # for H/E/F estimators
       min_allowed_inlier_ratio_est_model: 0.1
 
   multiview_optimizer:
--- a/gtsfm/configs/deep_front_end.yaml
+++ b/gtsfm/configs/deep_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: True
+  save_two_view_correspondences_viz: False
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
--- a/gtsfm/configs/deep_front_end.yaml
+++ b/gtsfm/configs/deep_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: False
+  save_two_view_correspondences_viz: True
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
@@ -13,14 +13,18 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    corr_metric_dist_threshold: 3 # in px
+    eval_threshold_px: 3 # in px
+    estimation_threshold_px: 4 # for homography estimator
+    min_num_inliers_acceptance: 15
+
     matcher:
       _target_: gtsfm.frontend.matcher.superglue_matcher.SuperGlueMatcher
+      use_outdoor_model: True
 
     verifier:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: True
-    
+      estimation_threshold_px: 4
 
   multiview_optimizer:
     _target_: gtsfm.multi_view_optimizer.MultiViewOptimizer
--- a/gtsfm/configs/scene_optimizer_unit_test_config.yaml
+++ b/gtsfm/configs/scene_optimizer_unit_test_config.yaml
@@ -13,14 +13,15 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    corr_metric_dist_threshold: 3 # in px
+    eval_threshold_px: 3 # in px
+    estimation_threshold_px: 0.5 # for homography estimator
     matcher:
       _target_: gtsfm.frontend.matcher.twoway_matcher.TwoWayMatcher
 
     verifier:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: False
-    
+      estimation_threshold_px: 0.5
 
   multiview_optimizer:
     _target_: gtsfm.multi_view_optimizer.MultiViewOptimizer
--- a/gtsfm/configs/scene_optimizer_unit_test_config.yaml
+++ b/gtsfm/configs/scene_optimizer_unit_test_config.yaml
@@ -3,7 +3,7 @@ SceneOptimizer:
   save_gtsfm_data: True
   save_two_view_correspondences_viz: False
   save_3d_viz: True
-  pose_angular_error_thresh: 10 # degrees
+  pose_angular_error_thresh: 5 # degrees
 
   feature_extractor:
     _target_: gtsfm.feature_extractor.FeatureExtractor
@@ -15,6 +15,8 @@ SceneOptimizer:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
     eval_threshold_px: 3 # in px
     estimation_threshold_px: 0.5 # for homography estimator
+    min_num_inliers_acceptance: 15
+    
     matcher:
       _target_: gtsfm.frontend.matcher.twoway_matcher.TwoWayMatcher
 
@@ -22,6 +24,7 @@ SceneOptimizer:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: False
       estimation_threshold_px: 0.5
+      min_allowed_inlier_ratio_est_model: 0.1
 
   multiview_optimizer:
     _target_: gtsfm.multi_view_optimizer.MultiViewOptimizer
--- a/gtsfm/configs/scene_optimizer_unit_test_config.yaml
+++ b/gtsfm/configs/scene_optimizer_unit_test_config.yaml
@@ -14,7 +14,6 @@ SceneOptimizer:
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
     eval_threshold_px: 3 # in px
-    estimation_threshold_px: 0.5 # for homography estimator
     min_num_inliers_acceptance: 15
     
     matcher:
@@ -23,7 +22,7 @@ SceneOptimizer:
     verifier:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: False
-      estimation_threshold_px: 0.5
+      estimation_threshold_px: 0.5 # for H/E/F estimators
       min_allowed_inlier_ratio_est_model: 0.1
 
   multiview_optimizer:
--- a/gtsfm/configs/sift_front_end.yaml
+++ b/gtsfm/configs/sift_front_end.yaml
@@ -1,9 +1,9 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: False
+  save_two_view_correspondences_viz: True
   save_3d_viz: True
-  pose_angular_error_thresh: 10 # degrees
+  pose_angular_error_thresh: 5 # degrees
 
   feature_extractor:
     _target_: gtsfm.feature_extractor.FeatureExtractor
@@ -13,13 +13,17 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    corr_metric_dist_threshold: 3 # in px
+    eval_threshold_px: 3 # in px
+    estimation_threshold_px: 0.5 # for homography estimator
+    min_num_inliers_acceptance: 15
+    
     matcher:
       _target_: gtsfm.frontend.matcher.twoway_matcher.TwoWayMatcher
 
     verifier:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
-      use_intrinsics_in_verification: False
+      use_intrinsics_in_verification: True
+      estimation_threshold_px: 4.0
     
 
   multiview_optimizer:
--- a/gtsfm/configs/sift_front_end.yaml
+++ b/gtsfm/configs/sift_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: False
+  save_two_view_correspondences_viz: True
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
@@ -24,7 +24,7 @@ SceneOptimizer:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: True
       estimation_threshold_px: 4.0
-    
+      min_allowed_inlier_ratio_est_model: 0.1
 
   multiview_optimizer:
     _target_: gtsfm.multi_view_optimizer.MultiViewOptimizer
--- a/gtsfm/configs/sift_front_end.yaml
+++ b/gtsfm/configs/sift_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: True
+  save_two_view_correspondences_viz: False
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
@@ -14,7 +14,6 @@ SceneOptimizer:
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
     eval_threshold_px: 3 # in px
-    estimation_threshold_px: 0.5 # for homography estimator
     min_num_inliers_acceptance: 15
     
     matcher:
@@ -23,7 +22,7 @@ SceneOptimizer:
     verifier:
       _target_: gtsfm.frontend.verifier.ransac.Ransac
       use_intrinsics_in_verification: True
-      estimation_threshold_px: 4.0
+      estimation_threshold_px: 4 # for H/E/F estimators
       min_allowed_inlier_ratio_est_model: 0.1
 
   multiview_optimizer:
--- a/gtsfm/configs/sift_front_end.yaml
+++ b/gtsfm/configs/sift_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: True
+  save_two_view_correspondences_viz: False
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
--- a/gtsfm/frontend/verifier/degensac.py
+++ b/gtsfm/frontend/verifier/degensac.py
@@ -9,7 +9,7 @@ References:
 - http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.466.2719&rep=rep1&type=pdf
 - https://github.com/ducha-aiki/pyransac
 
-Authors: Ayush Baid
+Authors: Ayush Baid, John Lambert
 """
 
 from typing import Optional, Tuple
@@ -37,12 +37,13 @@ class Degensac(VerifierBase):
 
         Args:
             use_intrinsics_in_verification: Flag to perform keypoint normalization and compute the essential matrix
-                                            instead of fundamental matrix. This should be preferred when the exact
-                                            intrinsics are known as opposed to approximating them from exif data.
-            estimation_threshold_px: maximum distance (in pixels) to consider a match an inlier, under squared Sampson distance.
+                instead of fundamental matrix. This should be preferred when the exact intrinsics are known as opposed
+                to approximating them from exif data.
+            estimation_threshold_px: maximum distance (in pixels) to consider a match an inlier, under squared
+                Sampson distance.
             min_allowed_inlier_ratio_est_model: minimum allowed inlier ratio w.r.t. the estimated model to accept
-                the verification result and use the image pair, i.e. the lowest allowed ratio of #final RANSAC inliers/ #putatives.
-                A lower fraction indicates less consistency among the result.
+                the verification result and use the image pair, i.e. the lowest allowed ratio of
+                #final RANSAC inliers/ #putatives. A lower fraction indicates less agreement among the result.
         Raises:
             NotImplementedError: when configured to compute essential matrices.
         """
@@ -85,7 +86,6 @@ class Degensac(VerifierBase):
             keypoints_i2.coordinates[match_indices[:, 1]],
             px_th=self._estimation_threshold_px,
         )
-
         inlier_idxs = np.where(inlier_mask.ravel() == 1)[0]
 
         v_corr_idxs = match_indices[inlier_idxs]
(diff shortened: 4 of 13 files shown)


gtsfm/frontend/verifier/degensac.py
```diff
@@ -23,23 +23,25 @@
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.frontend.verifier.verifier_base import VerifierBase
 
-PIXEL_COORD_RANSAC_THRESH = 0.5
-
 logger = logger_utils.get_logger()
 
+MAX_TOLERATED_POLLUTION_INLIER_RATIO_EST_MODEL = 0.1
```
akshay-krishnan, line 28: Too long name, and its not clear what it means either. What is “tolerated polluted”? Is MIN_INLIER_CORRESPONDENCES_RATIO better? Also, would you like to move that to config?

gtsfm/frontend/verifier/degensac.py
```diff
@@ -71,24 +73,33 @@ def verify(
...
-            camera_intrinsics_i1,
-            camera_intrinsics_i2,
+            px_th=self._estimation_threshold_px,
         )
 
-        return i2Ri1, i2Ui1, match_indices[inlier_idxes]
+        inlier_idxs = np.where(inlier_mask.ravel() == 1)[0]
+
+        v_corr_idxs = match_indices[inlier_idxs]
+        inlier_ratio_est_model = np.mean(inlier_mask)
```
akshay-krishnan, line 85: I think inlier_ratio is sufficient for the name.

gtsfm/frontend/verifier/ransac.py
```diff
@@ -23,29 +23,32 @@
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.frontend.verifier.verifier_base import VerifierBase, NUM_MATCHES_REQ_E_MATRIX, NUM_MATCHES_REQ_F_MATRIX
 
-PIXEL_COORD_RANSAC_THRESH = 4  # TODO: hyperparameter to tune
-DEFAULT_RANSAC_SUCCESS_PROB = 0.9999
+
+DEFAULT_RANSAC_SUCCESS_PROB = 0.99999
+DEFAULT_RANSAC_MAX_ITERS = 20000
```
akshay-krishnan, line 28: Isn’t that a lot? How many features do we typically detect?

gtsfm/frontend/verifier/ransac.py
```diff
@@ -23,29 +23,32 @@
...
+MAX_TOLERATED_POLLUTION_INLIER_RATIO_EST_MODEL = 0.1
```
akshay-krishnan, line 29: Same comment as in ransac

gtsfm/two_view_estimator.py
```diff
@@ -26,20 +27,67 @@
...
+MAX_H_INLIER_RATIO = 0.8
+
+EPSILON = 1e-6
+
+
+@dataclass(frozen=False)
+class TwoViewEstimationReport:
+    """
+    Args:
+        num_inliers_est_model: #correspondences consistent with estimated model (not necessarily "correct")
```
akshay-krishnan, line 47: Can we rename to num_inliers and inlier_ratio? 
We could use num_inliers_gt or gt_inlier_ratio for ground truth.

gtsfm/two_view_estimator.py
```diff
@@ -26,20 +27,67 @@
...
+        inlier_ratio_est_model: measures how polluted the putative matches were
+        num_inliers_gt_model: measures how well the verification worked, w.r.t. GT
+        inlier_ratio_gt_model: Only defined if GT relative pose provided
+        R_error_deg: relative pose error. Only defined if GT poses provided
+        U_error_deg
```
akshay-krishnan, line 52: Missing explanation.

gtsfm/two_view_estimator.py
```diff
@@ -26,20 +27,67 @@
...
+        min_num_inliers_acceptance: int,
+    ) -> None:
         """Initializes the two-view estimator from matcher and verifier.
 
         Args:
             matcher: matcher to use.
             verifier: verifier to use.
-            corr_metric_dist_threshold: distance threshold for marking a correspondence pair as inlier.
+            eval_threshold_px: distance threshold for marking a correspondence pair as inlier during evaluation (not estimation).
+            estimation_threshold_px: distance threshold for marking a correspondence pair as inlier during estimation
```
akshay-krishnan, line 83: Should that not be called inlier_threshold_px? Why do we need different thresholds for eval vs estimation?

gtsfm/two_view_estimator.py
```diff
@@ -92,7 +140,7 @@ def create_computation_graph(
 
         # verification on putative correspondences to obtain relative pose
         # and verified correspondences
-        (i2Ri1_graph, i2Ui1_graph, v_corr_idxs_graph,) = self._verifier.create_computation_graph(
+        (i2Ri1_graph, i2Ui1_graph, v_corr_idxs_graph, inlier_ratio_est_model) = self._verifier.create_computation_graph(
```
akshay-krishnan, line 143: Rename to inlier_ratio?

gtsfm/two_view_estimator.py
```diff
@@ -114,19 +162,88 @@ def create_computation_graph(
                 i2Ti1_expected_graph,
                 self._corr_metric_dist_threshold,
             )
+            number_correct, inlier_ratio = corr_error_graph[0], corr_error_graph[1]
```
akshay-krishnan, line 165: Rename to inlier_ratio_gt?

gtsfm/two_view_estimator.py
```diff
@@ -114,19 +162,88 @@ def create_computation_graph(
...
+        two_view_report: TwoViewEstimationReport,
+        i2Ri1: Optional[Rot3],
+        i2Ui1: Optional[Unit3],
+        v_corr_idxs: np.ndarray,
+    ) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray]:
+        """ """
+        insufficient_inliers = two_view_report.num_inliers_est_model < self._min_num_inliers_acceptance
+
+        H_EF_inlier_ratio = two_view_report.num_H_inliers / (two_view_report.num_inliers_est_model + EPSILON)
+        is_planar_or_panoramic = H_EF_inlier_ratio > MAX_H_INLIER_RATIO
```
akshay-krishnan, line 199: Should it be MAX_H_EF_INLIER_RATIO then?

gtsfm/two_view_estimator.py
```diff
@@ -114,19 +162,88 @@ def create_computation_graph(
...
+
+        # TODO: technically this should almost always be non-zero, just need to move up to earlier
+        valid_model = two_view_report.num_inliers_est_model > 0
```
akshay-krishnan, line 202: Ideally should we not check for a minimum inlier ratio instead? Something very low like 0.1? If we had 1 inlier, is that a valid model?

akshay-krishnan: You also seem to be logging the inlier ratio afterwards not the num inliers.

gtsfm/two_view_estimator.py
```diff
@@ -114,19 +162,88 @@ def create_computation_graph(
...
+def generate_two_view_report(
+    inlier_ratio_est_model: float,
+    R_error_deg: float,
+    U_error_deg: float,
+    number_correct: int,
+    inlier_ratio: float,
+    v_corr_idxs: np.ndarray,
+) -> TwoViewEstimationReport:
+    """ """
+    two_view_report = TwoViewEstimationReport(
```
akshay-krishnan, line 236: Do we need a helper for this?

gtsfm/two_view_estimator.py
```diff
@@ -149,22 +266,22 @@ def compute_correspondence_metrics(
...
-        Number of correct correspondences.
-        Inlier Ratio, i.e. ratio of correspondences which are correct.
+        Number of inlier correspondences to ground truth epipolar geometry, i.e. #correct correspondences.
+        Inlier Ratio, i.e. ratio of correspondences which are correct w.r.t. given relative pose.
     """
     if corr_idxs_i1i2.size == 0:
         return 0, float("Nan")
 
-    number_correct = metric_utils.count_correct_correspondences(
+    num_inliers_gt_model = metric_utils.count_correct_correspondences(
```
akshay-krishnan, line 275: Rename to num_inliers_gt for brevity?

akshay-krishnan (changes requested):

gtsfm/two_view_estimator.py
johnwlambert: Based on our other conversation on Slack, can I resolve this comment @akshay-krishnan, and leave as-is?

gtsfm/two_view_estimator.py
johnwlambert: thanks -- docstring should be cleaned up now

gtsfm/two_view_estimator.py
johnwlambert: So for evaluation, we might want to be very strict, e.g. look for cases where we get matches under 2 px in epipolar distance.

But for fitting, we might want a different threshold, e.g. 4 px, like in COLMAP.

This is actually what we have already been doing before.

gtsfm/two_view_estimator.py
johnwlambert: so the problem is that this is in the dask loop, so we have to wrap with `dask.delayed()` any function we call inside `create_computational_graph()`.

gtsfm/configs/deep_front_end.yaml
```diff
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: False
+  save_two_view_correspondences_viz: True
```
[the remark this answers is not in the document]
johnwlambert: `save_two_view_correspondences_viz` will save image pairs with correspondences on top of them. But I'll turn it off; it's just for debugging

gtsfm/configs/deep_front_end.yaml
```diff
@@ -13,14 +13,18 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    corr_metric_dist_threshold: 3 # in px
+    eval_threshold_px: 3 # in px
+    estimation_threshold_px: 4 # for homography estimator
```
ayushbaid, line 17: why just for homography estimation? What about epipolar geometry cases? Also is there a better name than `estimation_threshold_px`?

gtsfm/configs/deep_front_end.yaml
```diff
@@ -13,14 +13,18 @@ SceneOptimizer:
...
+    min_num_inliers_acceptance: 15
```
ayushbaid, line 18: `min_num_inliers_for_success`?

gtsfm/configs/deep_front_end.yaml
```diff
@@ -13,14 +13,18 @@ SceneOptimizer:
...
+
     matcher:
       _target_: gtsfm.frontend.matcher.superglue_matcher.SuperGlueMatcher
+      use_outdoor_model: True
```
ayushbaid, line 22: do we need this as exposed to the user right now?

gtsfm/configs/deep_front_end.yaml
ayushbaid: This homography based check should be verifier specific? Because in the future we might have a deep-learning based verifier which performs the model selection via networks too?

gtsfm/configs/sift_front_end.yaml
```diff
@@ -3,7 +3,7 @@ SceneOptimizer:
   save_gtsfm_data: True
   save_two_view_correspondences_viz: False
   save_3d_viz: True
-  pose_angular_error_thresh: 10 # degrees
+  pose_angular_error_thresh: 5 # degrees
```
ayushbaid, line 6: this name could be made more clear, now that we have so many thresholds

gtsfm/frontend/verifier/degensac.py
ayushbaid: maybe `MIN_RATIO_OF_OUTPUT_TO_INPUT_POINTS`? if we want to be explicitly clear?

johnwlambert: it should probably go in the config, under verifier.

gtsfm/frontend/verifier/degensac.py
```diff
@@ -23,23 +23,25 @@
...
 class Degensac(VerifierBase):
-    def __init__(self, use_intrinsics_in_verification: bool = False) -> None:
+    def __init__(self, use_intrinsics_in_verification: bool, estimation_threshold_px: float) -> None:
         """Initializes the verifier.
 
         Args:
             use_intrinsics_in_verification: Flag to perform keypoint normalization and compute the essential matrix 
                                             instead of fundamental matrix. This should be preferred when the exact
                                             intrinsics are known as opposed to approximating them from exif data.
+            estimation_threshold_px
```
ayushbaid, line 39: add description

gtsfm/frontend/verifier/degensac.py
ayushbaid: This is not a ratio. This should be `inlier_sum`?

gtsfm/frontend/verifier/ransac.py
ayushbaid, line 28: i think we should do away with `DEFAULT_` prefix?

gtsfm/frontend/verifier/ransac.py
```diff
@@ -54,7 +57,7 @@ def verify(
         match_indices: np.ndarray,
         camera_intrinsics_i1: Cal3Bundler,
         camera_intrinsics_i2: Cal3Bundler,
-    ) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray]:
+    ) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray, float]:
```
ayushbaid, line 60: need to add description of the new returned variable

gtsfm/two_view_estimator.py
```diff
@@ -25,21 +26,69 @@
...
+
+    In the spirit of COLMAP's Report class:
+    https://github.com/colmap/colmap/blob/dev/src/optim/ransac.h#L82
+
+    Inlier ratio is defined in Heinly12eccv: https://www.cs.unc.edu/~jheinly/publications/eccv2012-heinly.pdf
+    or in Slide 59: https://www.cc.gatech.edu/~afb/classes/CS4495-Fall2014/slides/CS4495-Ransac.pdf
+
+    Args:
+        v_corr_idxs: verified correspondence indices.
+        num_inliers_est_model: #correspondences consistent with estimated model (not necessarily "correct")
```
ayushbaid, line 45: I think this should not be num_inliers. The results should be called num_verified.

gtsfm/two_view_estimator.py
ayushbaid: Yep this is what we should do

ayushbaid (changes requested):

gtsfm/configs/deep_front_end.yaml
akshay-krishnan: I don’t think estimation_threshold_px is a great name either. How is estimation measured in px?

gtsfm/two_view_estimator.py
akshay-krishnan: Resolving

gtsfm/frontend/verifier/ransac.py
johnwlambert: i'm cool with that, @akshay-krishnan does that sound alright to you?

gtsfm/frontend/verifier/degensac.py
johnwlambert: @ayushbaid thanks for the review. I didn't quite follow here -- I believe this is a ratio -- #ransac inliers / #putative matches.

gtsfm/frontend/verifier/degensac.py
johnwlambert: thanks -- just added it.

gtsfm/configs/sift_front_end.yaml
johnwlambert: Hmm I'm having trouble thinking of something clearer. DId you have something specific in mind?

gtsfm/configs/deep_front_end.yaml
johnwlambert: I was testing on some indoor and outdoor data, so I think we should expose it

gtsfm/configs/deep_front_end.yaml
johnwlambert: So technically epipolar distance (squared Sampson) is measured in pixels (or pixels^2).

But unfortunately we measure epipolar distance twice -- once in estimation, once in evaluation. So we need to differentiate between the two.

gtsfm/frontend/verifier/ransac.py
johnwlambert: thanks -- should be fixed now

gtsfm/two_view_estimator.py
johnwlambert: Thanks for the review, Ayush. I think `num_verified` is more confusing since it could refer to images or points or correspondences. I think `num_inliers` is more unambiguous in referring to correspondences

gtsfm/two_view_estimator.py
johnwlambert: marking as resolved from our earlier conversation

gtsfm/configs/deep_front_end.yaml
akshay-krishnan: How about `estimation_error_threshold_px` and `eval_error_threshold_px`?

gtsfm/configs/deep_front_end.yaml
akshay-krishnan: I find num_ kinda redundant when theres a min_. min_inliers_acceptance is shorter with no meaning lost.

gtsfm/configs/scene_optimizer_unit_test_config.yaml
```diff
@@ -13,14 +13,15 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    corr_metric_dist_threshold: 3 # in px
+    eval_threshold_px: 3 # in px
+    estimation_threshold_px: 0.5 # for homography estimator
```
akshay-krishnan, line 17: same renaming suggestion as above.

gtsfm/frontend/verifier/ransac.py
akshay-krishnan: Yeah, +1

gtsfm/two_view_estimator.py
akshay-krishnan: And `TwoViewEstimationReport` cannot be the function?

gtsfm/configs/sift_front_end.yaml
ayushbaid: I meant conveying that pose angular error is max of rotation and direction angular error. Or maybe its too verbose?

gtsfm/frontend/verifier/degensac.py
ayushbaid: Oh I got it now. Thanks!

gtsfm/configs/sift_front_end.yaml
johnwlambert: yeah i fear it would be too verbose

gtsfm/two_view_estimator.py
johnwlambert: @ayushbaid thanks for the review -- let me know what you think about my comment above when you get a chance

gtsfm/two_view_estimator.py
johnwlambert: it's strange, but if i wrap the class constructor in dask.delayed, I get a nasty error:
```
distributed.worker - WARNING - Compute Failed
Function:  execute_task
args:      ((<function apply at 0x104155b80>, <class 'type'>, (), (<class 'dict'>, [['inlier_ratio_est_model', None], ['num_inliers_gt_model', None], ['inlier_ratio_gt_model', None], ['R_error_deg', None], ['U_error_deg', None], ['i2Ri1', None], ['i2Ui1', None]])))
kwargs:    {}
Exception: TypeError('type() takes 1 or 3 arguments')

  File "gtsfm/runner/run_scene_optimizer_colmaploader.py", line 102, in <module>
    run_scene_optimizer(args)
  File "gtsfm/runner/run_scene_optimizer_colmaploader.py", line 49, in run_scene_optimizer
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/dask/base.py", line 285, in compute
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/dask/base.py", line 567, in compute
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/distributed/client.py", line 2672, in get
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/distributed/client.py", line 1981, in gather
    return self.sync(
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/distributed/client.py", line 852, in sync
    return sync(
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/distributed/utils.py", line 354, in sync
    raise exc.with_traceback(tb)
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/distributed/utils.py", line 337, in f
    result[0] = yield future
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/distributed/client.py", line 1846, in _gather
  File "/Users/johnlambert/anaconda3/envs/gtsfm-v2/lib/python3.8/site-packages/dask/utils.py", line 34, in apply
    return func(*args, **kwargs)
TypeError: type() takes 1 or 3 arguments
```
So I prefer to just use the helper.

gtsfm/configs/deep_front_end.yaml
johnwlambert: i'm cool with that, i'll update accordingly

gtsfm/configs/deep_front_end.yaml
johnwlambert: Thanks for the suggestions. I worry that those names are a bit too longwinded, though.

gtsfm/configs/deep_front_end.yaml
johnwlambert: On second though, I actually think `min_num_inliers_acceptance` is perhaps better, since otherwise `min_inliers_acceptance` might make people think these are the actual inliers. Inded this is consistent with COLMAP: https://github.com/colmap/colmap/blob/dev/src/estimators/essential_matrix.h#L60
```
  // The minimum number of samples needed to estimate a model.
  static const int kMinNumSamples = 5;
```

gtsfm/two_view_estimator.py
johnwlambert: so we are chekcing the minimum inlier ratio also, but that is in the `Verifier` class, not the `TwoViewEstimator` class. I was thinking that belonged in the verifier, since there would be no reason to find a F or E matrix if we already no the inlier ratio w.r.t. est model is too low.

gtsfm/configs/deep_front_end.yaml
johnwlambert: > why just for homography estimation? What about epipolar geometry cases? Also is there a better name than `estimation_threshold_px`?

if we want to compare the # homography inliers vs. # E or F inliers in the future, like COLMAP does, then we would want to share the same threshold between homography and E/F fitting
4:21
that would mean that estimation threshold would need to be passed to the TwoViewEstimator, since we’ve defined our verifiers to operate on E/F
4:21
that would mean that we would need to pass that threshold to the verify() method from the TwoViewEstimator every time, or provide the same threshold in two places — the TwoViewEstimator config options, and the Verifier config options

gtsfm/frontend/verifier/degensac.py
johnwlambert: i moved it to the config and renamed it to `min_allowed_inlier_ratio_est_model`

gtsfm/frontend/verifier/ransac.py
johnwlambert: i changed it back to our previous defaults, since OpenCV is sufficiently buggy that this argument doesn't affect our results anyways.

gtsfm/frontend/verifier/ransac.py
johnwlambert: should be fixed now, I renamed it to `min_allowed_inlier_ratio_est_model` in the config.

gtsfm/two_view_estimator.py
johnwlambert: renamed, should be fixed

gtsfm/two_view_estimator.py
johnwlambert: i removed this, the homography comparison code will come in a future PR

gtsfm/frontend/verifier/ransac.py
johnwlambert: i removed the DEFAULT_ prefix

gtsfm/configs/deep_front_end.yaml
ayushbaid: I have one more followup comment. Why not use `min_num_verified_corr`? I prefer to call the output of verification as verified correspondences instead of inlier correspondence. **Inlier** should be reserved for things which are correct according to the ground truth.

gtsfm/configs/sift_front_end.yaml
ayushbaid: Do you think we should add the keyword `metric` in the name. We might have thresholds based on cycle consistency or consistency between global-local pose values. Maybe something like `pose_angular_error_metric_thresh`?

gtsfm/two_view_estimator.py
ayushbaid: I still think verified is better. But I am fine with inliers too as it is an established nomenclature.

gtsfm/two_view_estimator.py
akshay-krishnan: I see, okay.

gtsfm/configs/sift_front_end.yaml
johnwlambert: Thanks for the review Ayush.

I think `pose_angular_error_metric_thresh` is a bit too long. I think we could later add `max_cycle_rot_deviation_deg` and `max_global_relative_deviation_deg`. But I prefer the current names.

gtsfm/configs/deep_front_end.yaml
johnwlambert: resolved by using `self.verifier._estimation_threshold_px` within TwoViewEstimator, for homography, in the future

PR #204, opened 2021-06-30
Add cycle consistency checks before Shonan

johnwlambert:
To eliminate noisy measurements, use cycle error as a proxy for what the pose errors w.r.t. GT would be.

Reduces error significantly:
```bash
python gtsfm/runner/run_scene_optimizer_colmaploader.py --config_name sift_front_end.yaml --max_frame_lookahead 8 --images_dir /Users/johnlambert/Downloads/skydio_crane_mast_8imgs_with_exif/images --colmap_files_dirpath /Users/johnlambert/Downloads/skydio_crane_mast_8imgs_with_exif/crane_mast_8imgs_colmap_output
```

and

```bash
python gtsfm/runner/run_scene_optimizer_colmaploader.py --config_name deep_front_end.yaml --max_frame_lookahead 8 --images_dir /Users/johnlambert/Downloads/skydio_crane_mast_8imgs_with_exif/images --colmap_files_dirpath /Users/johnlambert/Downloads/skydio_crane_mast_8imgs_with_exif/crane_mast_8imgs_colmap_output
```


johnwlambert:
Hi @travisdriver, if you have time, feel free to take a look at this as well

.github/scripts/python.sh
```diff
@@ -34,7 +34,7 @@ cd $GITHUB_WORKSPACE
 cd $GITHUB_WORKSPACE
 
 # check that main script executes on toy Door dataset
-python gtsfm/runner/run_scene_optimizer.py
+python gtsfm/runner/run_scene_optimizer.py --config_name sift_front_end.yaml
```
johnwlambert, line 37: CI took an hour with the default front-end (deep_front_end.yaml), so switching to one that takes 15 minutes to not burn up our CI quota.

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+
+    I_3x3 = Rot3()
+    cycle_error = comp_utils.compute_relative_rotation_angle(I_3x3, i0Ri0)
+
+    # form 3 edges between fully connected subgraph (nodes i,j,k)
+    e_i = (i0, i1)
+    e_j = (i1, i2)
+    e_k = (i0, i2)
+
+    rot_errors = [two_view_reports_dict[e].R_error_deg for e in [e_i, e_j, e_k]]
```
akshay-krishnan, line 117: Is it possible for two_view_reports_dict to have an edge in a different order? like (i1, i0)?

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+        max_trans_error = np.max(trans_errors)
+    else:
+        # ground truth unknown, so cannot estimate error w.r.t. GT
+        max_rot_error = None
+        max_trans_error = None
+
+    if verbose:
+        i1Ri0_euler = Rotation.from_matrix(i1Ri0.matrix()).as_euler(seq="xyz", degrees=True).tolist()
+        i2Ri1_euler = Rotation.from_matrix(i2Ri1.matrix()).as_euler(seq="xyz", degrees=True).tolist()
+        i0Ri2_euler = Rotation.from_matrix(i0Ri2.matrix()).as_euler(seq="xyz", degrees=True).tolist()
```
akshay-krishnan, line 132: why cant we use gtsam functionality directly? https://github.com/borglab/gtsam/blob/6d6d129d05f64216cc8814dc7f24fcf05fd67f44/gtsam/geometry/Rot3.h#L464

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+        euler_x = [i1Ri0_euler[0], i2Ri1_euler[0], i0Ri2_euler[0]]
+        euler_y = [i1Ri0_euler[1], i2Ri1_euler[1], i0Ri2_euler[1]]
+        euler_z = [i1Ri0_euler[2], i2Ri1_euler[2], i0Ri2_euler[2]]
+
+        logger.info("\n")
+        logger.info(f"{i0},{i1},{i2} --> Cycle error is: {cycle_error:.1f}")
+        if gt_known:
+            logger.info(f"Triplet: w/ max. R err {max_rot_error:.1f}, and w/ max. t err {max_trans_error:.1f}")
+
+        logger.info("X: (0->1) %.1f deg., (1->2) %.1f deg., (2->0) %.1f deg.", euler_x[0], euler_x[1], euler_x[2])
```
akshay-krishnan, line 143: why not directly use i1Ri0_euler[0], i2Ri1_euler[0], i0Ri2_euler[0] here?

tests/utils/test_cycle_consistency.py
```diff
@@ -0,0 +1,299 @@
...
+    Imagine 3 poses, all centered at the origin, at different orientations.
+
+    Ground truth poses:
+       Let i0 face along +x axis (0 degrees in yaw)
+       Let i2 have a 30 degree rotation from the +x axis.
+       Let i4 have a 90 degree rotation from the +x axis.
+
+    However, suppose one edge measurement is corrupted (from i0 -> i4) by 5 degrees.
+    """
+    i2Ri0 = Rot3(Rotation.from_euler("y", 30, degrees=True).as_matrix())
```
akshay-krishnan, line 92: John please use gtsam functionality directly. :)

tests/utils/test_cycle_consistency.py
```diff
@@ -0,0 +1,299 @@
...
+    cycle_nodes = [0, 2, 4]
+    i2Ri1_dict = {
+        (0, 2): i2Ri0,  # edge i
+        (2, 4): i4Ri2,  # edge j
+        (0, 4): i4Ri0,  # edge k
+    }
+
+    two_view_reports_dict = {}
+    # rest of attributes will default to None
+    two_view_reports_dict[(0, 4)] = TwoViewEstimationReport(
```
akshay-krishnan, line 105: code is repeated too much. You could define a helper inside this unit test. 
def get_report(r_error, u_error): 
and use it 3 times.

tests/utils/test_cycle_consistency.py
```diff
@@ -0,0 +1,299 @@
...
+    Imagine 3 poses, all centered at the origin, at different orientations.
+
+    Ground truth poses:
+       Let i0 face along +x axis (0 degrees in yaw)
+       Let i2 have a 30 degree rotation from the +x axis.
+       Let i4 have a 90 degree rotation from the +x axis.
+
+    However, suppose one edge measurement is corrupted (from i0 -> i4) by 5 degrees.
+    """
+    i2Ri0 = Rot3(Rotation.from_euler("y", 30, degrees=True).as_matrix())
```
akshay-krishnan, line 147: Same, gtsam doesnt need a matrix to construct a rotation.

tests/utils/test_cycle_consistency.py
```diff
@@ -0,0 +1,299 @@
...
+
+
+def test_filter_to_cycle_consistent_edges() -> None:
+    """Ensure correct edges are kept in a 2-triplet scenario.
+
+    Scenario Ground Truth: consider 5 camera poses in a line, connected as follows, all with identity rotations:
+
+      _________    ________
+     /         \\ /        \
+    i4 -- i3 -- i2 -- i1 -- i0
```
akshay-krishnan, line 191: the fact that they are on a line is irrelevant. A better visualization might be two triangles. 
```
i4                 i1
           i2
i3                 i0
```

akshay-krishnan (changes requested):
As we discussed offline, I dont think the translation error is meaningful for directions. Could you remove all the translation error bits?

gtsfm/two_view_estimator.py
```diff
@@ -26,6 +27,40 @@
 pil_logger.setLevel(logging.INFO)
 
 
+@dataclass(frozen=False)
+class TwoViewEstimationReport:
```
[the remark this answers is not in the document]
johnwlambert: So it is in the other PR. But since we import it in `cycle_consistency.py`, we need to include it here, while we wait for the other PR to be ready for merge

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
+"""
```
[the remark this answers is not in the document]
johnwlambert: good point, i agree. I'll move it under `averaging`

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+    for (i1, i2), i2Ri1 in i2Ri1_dict.items():
+        if i2Ri1 is None:
+            continue
+
+        nodes_from_i1 = adj_list[i1]
+        nodes_from_i2 = adj_list[i2]
+
+        node_intersection = (nodes_from_i1).intersection(nodes_from_i2)
+
+        for node in node_intersection:
```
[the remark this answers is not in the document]
johnwlambert: no particular reason, I'll get rid of them. i agree it's too much whitespace

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+
+def compute_cycle_error(
+    i2Ri1_dict: Dict[Tuple[int, int], Rot3],
+    cycle_nodes: Tuple[int, int, int],
+    two_view_reports_dict: Dict[Tuple[int, int], TwoViewEstimationReport],
+    verbose: bool = True,
+) -> Tuple[float, Optional[float], Optional[float]]:
+    """Compute the cycle error by the magnitude of the axis-angle rotation after composing 3 rotations.
+
+    Note: i1 < i2 for every valid edge, by construction.
```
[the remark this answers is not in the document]
johnwlambert: i mean this in a more general case, that a<b for every edge. would `a < b` be less confusing?

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+    # should compose to identity, with ideal measurements
+    i0Ri0 = i0Ri2.compose(i2Ri1).compose(i1Ri0)
+
+    I_3x3 = Rot3()
+    cycle_error = comp_utils.compute_relative_rotation_angle(I_3x3, i0Ri0)
+
+    # form 3 edges between fully connected subgraph (nodes i,j,k)
+    e_i = (i0, i1)
+    e_j = (i1, i2)
+    e_k = (i0, i2)
```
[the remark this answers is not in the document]
johnwlambert: good point, i like that. i'll do that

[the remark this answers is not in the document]
johnwlambert: updated.

gtsfm/utils/cycle_consistency.py
johnwlambert: i thought this was a bit more readable, since otherwise the line gets quite wrong. but i'm fine either way

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+        http://people.inf.ethz.ch/pomarc/pubs/ZachCVPR10.pdf
+
+        Enqvist, Olof; Kahl, Fredrik; Olsson, Carl. Non-Sequential Structure from Motion. ICCVW, 2011.
+        https://portal.research.lu.se/ws/files/6239297/2255278.pdf
+
+    Args:
+        i2Ri1_dict: mapping from image pair indices to relative rotation.
+        i2Ui1_dict: smapping from image pair indices to relative translation direction.
+            Should have same keys as i2Ri1_dict.
+        two_view_reports_dict
```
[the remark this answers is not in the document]
johnwlambert: good catch, thanks. updated.

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+    logger.info("cycle_consistent_keys: " + str(cycle_consistent_keys))
+
+    i2Ri1_dict_consistent, i2Ui1_dict_consistent = {}, {}
+    for (i1, i2) in cycle_consistent_keys:
+        i2Ri1_dict_consistent[(i1, i2)] = i2Ri1_dict[(i1, i2)]
+        i2Ui1_dict_consistent[(i1, i2)] = i2Ui1_dict[(i1, i2)]
+
+    num_consistent_rotations = len(i2Ri1_dict_consistent)
+    logger.info("Found %d consistent rel. rotations from %d original edges.", num_consistent_rotations, n_valid_edges)
+    assert len(i2Ui1_dict_consistent) == num_consistent_rotations
```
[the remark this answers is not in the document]
johnwlambert: good point, yeah it's unnecessary. i'll remove it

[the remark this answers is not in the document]
johnwlambert: removed

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+        plt.savefig(os.path.join("plots", "cycle_error_vs_GT_trans_error.jpg"), dpi=200)
+
+    logger.info("cycle_consistent_keys: " + str(cycle_consistent_keys))
+
+    i2Ri1_dict_consistent, i2Ui1_dict_consistent = {}, {}
+    for (i1, i2) in cycle_consistent_keys:
+        i2Ri1_dict_consistent[(i1, i2)] = i2Ri1_dict[(i1, i2)]
+        i2Ui1_dict_consistent[(i1, i2)] = i2Ui1_dict[(i1, i2)]
+
+    num_consistent_rotations = len(i2Ri1_dict_consistent)
```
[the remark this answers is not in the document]
johnwlambert: good point, since i've removed the assert, i agree that's cleaner. updated.

gtsfm/utils/cycle_consistency.py
johnwlambert: so it is not, since the keys for `two_view_reports_dict` will be used from the loader class, and in the loader, we always ensure `a < b` for any image indices `a, b`

tests/utils/test_cycle_consistency.py
johnwlambert: agreed, good point. let me know if it's cleaner now, I updated it.

.github/scripts/python.sh
johnwlambert: so this 1-line change I had to make as a hotfix to the Github Actions CI file here in master, to prevent us from running out of CI $$. So it's in master already now.

gtsfm/utils/cycle_consistency.py
johnwlambert: @akshay-krishnan would like to chat more about this. So here I'm assuming we have a camera with +z pointing out of the camera, and we rotate it by 30 degrees about the rotate axis:
```python
>>> Rotation.from_euler("y", 30, degrees=True).as_matrix()
array([[ 0.8660254,  0.       ,  0.5      ],
       [ 0.       ,  1.       ,  0.       ],
       [-0.5      ,  0.       ,  0.8660254]])
```

The equivalent in GTSAM is:

```python
>>> Rot3.Pitch(np.deg2rad(30))
R: [
	0.866025, 0, 0.5;
	0, 1, 0;
	-0.5, 0, 0.866025
]
```

But pitch only makes sense if it's a vehicle frame with vehicle pointing in the +x direction, with +y to the left, and +z upwards.

So I feel like using the term "Pitch" is less accurate.

tests/utils/test_cycle_consistency.py
johnwlambert: How about I add both figures in the docstring? One that shows the spatial layout, and one that shows the topology?

gtsfm/utils/cycle_consistency.py
johnwlambert: I guess there is
```python
>>> Rot3.Ry(np.deg2rad(30))
R: [
	0.866025, 0, 0.5;
	0, 1, 0;
	-0.5, 0, 0.866025
]
```

johnwlambert: I don't see a way to decode "Ry" though from a Rot3 instance, without calling `.Pitch()`, which I feel like isn't accurate. So I'm fine with using GTSAM for the constructor, but I think when decoding the euler angles, I prefer the scipy.spatial.transform

tests/utils/test_cycle_consistency.py
johnwlambert: I updated it now to use `Rot3.Ry(np.deg2rad(angle_deg))` here

tests/utils/test_cycle_consistency.py
johnwlambert: I updated it now to use `Rot3.Ry(np.deg2rad(angle_deg))`

gtsfm/utils/cycle_consistency.py
```diff
@@ -0,0 +1,233 @@
...
+        plt.xlabel("Cycle error")
+        plt.ylabel("Avg. Rot3 error over cycle triplet")
+        plt.savefig(os.path.join("plots", "cycle_error_vs_GT_rot_error.jpg"), dpi=200)
+
+        plt.scatter(cycle_errors, max_trans_errors)
+        plt.xlabel("Cycle error")
+        plt.ylabel("Avg. Unit3 error over cycle triplet")
+        plt.savefig(os.path.join("plots", "cycle_error_vs_GT_trans_error.jpg"), dpi=200)
+
+    logger.info("cycle_consistent_keys: " + str(cycle_consistent_keys))
```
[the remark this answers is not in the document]
johnwlambert: I say we leave it now for debugging, and can remove later. While development is ongoing, I prefer more verbosity

tests/averaging/rotation/test_cycle_consistency.py
```diff
@@ -0,0 +1,304 @@
...
+
+    0 ---- 1
+          /|
+         / |
+        /  |
+      2 -- 3
+           |
+           |
+           4
+    """
```
ayushbaid, line 28: Can we add 2 more cycles: one which shares an edge with the current cycle and the other which just shares a node.

tests/utils/test_cycle_consistency.py
akshay-krishnan: But the spatial layout is irrelevant no? We only care about rotations and that is not evident from it.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,229 @@
+"""
```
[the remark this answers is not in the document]
johnwlambert: So @akshay-krishnan and I thought that this is specific to rotation averaging (outlier removal), so it was better suited for the averaging/ folder. I think we have the 1dsfm outlier removal code in averaging/ also. Wouldn’t be used anywhere else I think.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,229 @@
...
+
+def compute_cycle_error(
+    i2Ri1_dict: Dict[Tuple[int, int], Rot3],
+    cycle_nodes: Tuple[int, int, int],
+    two_view_reports_dict: Dict[Tuple[int, int], TwoViewEstimationReport],
+    verbose: bool = True,
+) -> Tuple[float, Optional[float], Optional[float]]:
+    """Compute the cycle error by the magnitude of the axis-angle rotation after composing 3 rotations.
+
+    Note: a < b for every valid edge (a,b), by construction inside the image loader class.
```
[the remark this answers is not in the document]
johnwlambert: I think it’s a good convention for us to enforce. How do you think we should go about enforcing it? Putting a check in the base loader class?

would there ever be a time where this is undesired? I can’t think of one.

tests/averaging/rotation/test_cycle_consistency.py
johnwlambert: So I think the other unit test includes the case where we share an edge with the current cycle.

I can add another that just shares a node.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,229 @@
...
+            If ground truth is not known for a scene, None will be returned instead.
+    """
+    cycle_nodes = list(cycle_nodes)
+    cycle_nodes.sort()
+
+    i0, i1, i2 = cycle_nodes
+
+    i1Ri0 = i2Ri1_dict[(i0, i1)]
+    i2Ri1 = i2Ri1_dict[(i1, i2)]
+    i0Ri2 = i2Ri1_dict[(i0, i2)].inverse()
```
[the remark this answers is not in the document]
johnwlambert: I think the code will be much more complex if we don't have this assumption. So I'm all for enforcing this assumption.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,229 @@
...
+    Args:
+        i2Ri1_dict: mapping from image pair indices to relative rotation.
+        cycle_nodes: 3-tuples of nodes that form a cycle. Nodes of are provided in sorted order.
+        two_view_reports_dict:
+        verbose: whether to dump to logger information about error in each Euler angle
+
+    Returns:
+        cycle_error: deviation from 3x3 identity matrix, in degrees. In other words,
+            it is defined as the magnitude of the axis-angle rotation of the composed transformations.
+        max_rot_error: maximum rotation error w.r.t. GT across triplet edges, in degrees.
```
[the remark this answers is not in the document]
johnwlambert: I found it useful to plot for a scene, cycle error in each triplet vs. max error w.r.t. GT in triplet, for all triplets.

I think we need to study that plot on more scenes to determine the threshold.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,229 @@
...
+    connected to `a` and the nodes connected to `b`.
+
+    Args:
+        i2Ri1_dict: mapping from image pair indices to relative rotation.
+
+    Returns:
+        triplets: 3-tuples of nodes that form a cycle. Nodes of each triplet are provided in sorted order.
+    """
+    # only want to keep the unique ones
+    triplets = set()
```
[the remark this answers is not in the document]
johnwlambert: Yeah, they are always sorted before they are inserted into the set.

tests/averaging/rotation/test_cycle_consistency.py
johnwlambert: I added another unit test to check the shared node case.

tests/utils/test_cycle_consistency.py
johnwlambert: I think the spatial layout is helpful too. I would probably prefer to keep both

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,232 @@
...
+    if gt_known:
+        max_rot_error = np.max(rot_errors)
+        max_trans_error = np.max(trans_errors)
+    else:
+        # ground truth unknown, so cannot estimate error w.r.t. GT
+        max_rot_error = None
+        max_trans_error = None
+
+    if verbose:
+        i1Ri0_euler = Rotation.from_matrix(i1Ri0.matrix()).as_euler(seq="xyz", degrees=True).tolist()
```
akshay-krishnan, line 130: I think I had a previous comment about directly using GTSAM functionality here. Is that not possible?

tests/averaging/rotation/test_cycle_consistency.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+if __name__ == "__main__":
+
+    # test_extract_triplets_adjacency_list_intersection1()
+    # test_extract_triplets_adjacency_list_intersection2()
+
+    # test_compute_cycle_error_known_GT()
+    # test_compute_cycle_error_unknown_GT()
+    # test_filter_to_cycle_consistent_edges()
```
akshay-krishnan, line 337: should these not be implemented as `TestCase`s? Why are they being called in main?

johnwlambert: so `pytest` can discover them all as long as they start with `test_*`

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,232 @@
...
+from gtsfm.two_view_estimator import TwoViewEstimationReport
+
+
+logger = logger_utils.get_logger()
+
+
+CYCLE_ERROR_THRESHOLD = 5.0
+
+
+def extract_triplets_adjacency_list_intersection(i2Ri1_dict: Dict[Tuple[int, int], Rot3]) -> List[Tuple[int, int, int]]:
```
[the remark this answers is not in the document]
johnwlambert: good point. just renamed them.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,232 @@
...
+    verbose: bool = True,
+) -> Tuple[float, Optional[float], Optional[float]]:
+    """Compute the cycle error by the magnitude of the axis-angle rotation after composing 3 rotations.
+
+    Note: a < b for every valid edge (a,b), by construction inside the image loader class.
+
+    Args:
+        i2Ri1_dict: mapping from image pair indices to relative rotation.
+        cycle_nodes: 3-tuples of nodes that form a cycle. Nodes of are provided in sorted order.
+        two_view_reports_dict:
```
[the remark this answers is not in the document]
johnwlambert: good catch, thanks. just fixed it.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,232 @@
+"""
+Utilities for cycle triplet extraction and cycle error computation.
+
+In short, one can check the cumulative rotation errors between triplets to throw away cameras.
```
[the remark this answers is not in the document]
johnwlambert: agreed. updated.

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,232 @@
+"""
+Utilities for cycle triplet extraction and cycle error computation.
```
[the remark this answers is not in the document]
johnwlambert: agreed. updated.

gtsfm/averaging/rotation/cycle_consistency.py
johnwlambert: I left some comments in the thread up above, if you could take a look at it: https://github.com/borglab/gtsfm/pull/204#discussion_r663117516

gtsfm/utils/cycle_consistency.py
akshay-krishnan: .xyz()[1] ? 
https://github.com/borglab/gtsam/blob/6d6d129d05f64216cc8814dc7f24fcf05fd67f44/gtsam/geometry/Rot3.h#L464

akshay-krishnan: I dont like the idea of redundant calculation and memory transfer in a library, it's kinda acceptable in a unit test.

gtsfm/averaging/rotation/cycle_consistency.py
akshay-krishnan: replied in that comment.

tests/averaging/rotation/test_cycle_consistency.py
akshay-krishnan: Do we still need the main?

johnwlambert: good point -- I just removed the main

gtsfm/averaging/rotation/cycle_consistency.py
johnwlambert: I switched to Rot3.xyz(), it's equivalent to scipy.spatial.transform's .as_euler("xyz")

gtsfm/utils/cycle_consistency.py
johnwlambert: Updated and removed the scipy code here

gtsfm/averaging/rotation/cycle_consistency.py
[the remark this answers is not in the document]
ayushbaid: I agree that this is a good condition to enforce. We should update the docstring of the base loaders (and consequently all the loaders to reflect this)?

gtsfm/averaging/rotation/cycle_consistency.py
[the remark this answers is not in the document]
ayushbaid: I see. To figure out what kind of two view errors give rise to high triplet errors?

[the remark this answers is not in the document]
ayushbaid: Even if we have these extra params, do you think having all of them in the same function is better?

tests/averaging/rotation/test_cycle_consistency.py
```diff
@@ -0,0 +1,327 @@
...
+                if len(cycle_nodes) == 3 and edges_are_unique:
+                    triplets.add(cycle_nodes)
+
+    return list(triplets)
+
+
+def test_triplet_extraction_correctness_runtime() -> None:
+    """Ensure that for large graphs, the adjacency-list-based algorithm is faster and still correct,
+    when compared with the brute-force O(n^3) implementation.
+    """
```
ayushbaid, line 301: The time aspect can create false alarms. I am a bit wary of checking this in.

johnwlambert: The brute force method is unbelievably slow, with cubic complexity vs linear for the release implementation. Were you concerned about it being faster than the other impl and raising an assert?

gtsfm/averaging/rotation/cycle_consistency.py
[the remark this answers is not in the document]
johnwlambert: Hmm what do you mean?

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,234 @@
...
+
+    for (i1, i2), i2Ri1 in i2Ri1_dict.items():
+        if i2Ri1 is None:
+            continue
+
+        adj_list[i1].add(i2)
+        adj_list[i2].add(i1)
+
+    # find intersections
+    for (i1, i2), i2Ri1 in i2Ri1_dict.items():
```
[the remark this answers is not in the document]
akshay-krishnan: The errors accumulate as the cycle length increases. I think there are papers that don’t recommend using larger cycles for outlier rejection. We would end up rejecting more edges than what we added (by increasing look-ahead window). If we really want to do that we should weigh the errors by the inverse of their cycle length, which increases hand-tuned parameters. Shonan is also quite robust already I think. 

I agree we must use more edges than the hard-coded look ahead window, but even if we do, it may not be worth using cycles with length > 3 for outlier rejection.

tests/utils/test_cycle_consistency.py
akshay-krishnan: Sure, its just a nit. I found the topology useful as this is a rotation-only problem.

gtsfm/utils/cycle_consistency.py
akshay-krishnan: I don’t quite like the idea of doing redundant work for logging. Also don’t think any readability is lost. You’ll save on the other lines even if these take up 2 lines. Also, is to_list() necessary?

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -0,0 +1,234 @@
...
+
+    # form adjacency list
+    adj_list = defaultdict(set)
+
+    for (i1, i2), i2Ri1 in i2Ri1_dict.items():
+        if i2Ri1 is None:
+            continue
+
+        adj_list[i1].add(i2)
+        adj_list[i2].add(i1)
```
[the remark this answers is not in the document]
johnwlambert: sure, i like that idea. just added it.

gtsfm/averaging/rotation/cycle_consistency.py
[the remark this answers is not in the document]
johnwlambert: I added a runtime check during execution for invalid edges, and a unit test that checks for it.

I also opened an issue to add this to documentation more thoroughly, and to the base loader class.

gtsfm/utils/cycle_consistency.py
johnwlambert: Good points. I removed it

gtsfm/averaging/rotation/cycle_consistency.py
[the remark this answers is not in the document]
travisdriver: I wasn't saying to increase the cycle length; the cycle length would still be three. I'm saying to check _all_ of the valid triplets cycles in the graph based on matches between the frames, not just the ones within the look ahead window.

akshay-krishnan (approved):
Looks like all of my comments have been addressed. Approving the PR, but please do not merge until the comments from others are addressed, and until the other PR that contains `gtsfm/two_view_estimator.py` is merged, so that it can be removed from this one.

tests/averaging/rotation/test_cycle_consistency.py
ayushbaid: For some reason yeah. I think we do not need to enforce it via a test. But keep it as a note to future developers.

ayushbaid (approved):
Approved, as I am happy with the PR. Just one point of concern being the unit test on time complexity.

johnwlambert:
Thanks Ayush, I removed the time test, since as you noted it is dependent upon the state of the external environment, which we cannot assume to be fixed.

PR #208, opened 2021-07-03
Log to stdout summary about SfM run

johnwlambert:


gtsfm/utils/io.py
```diff
@@ -101,6 +101,19 @@ def save_json_file(
         json.dump(data, f, indent=4)
 
 
+def read_json_file(fpath: Union[str, Path]) -> Any:
```
ayushbaid, line 104: We need this wrapper for dask?

ayushbaid (commented):
Why do we read jsons from disk instead of data being passed as inputs?

gtsfm/utils/io.py
akshay-krishnan, line 104: I am surprised we dont have this yet .. I feel like I saw it somewhere.

gtsfm/utils/metrics.py
```diff
@@ -215,3 +223,36 @@ def compute_pose_errors(gt_wTi_list: List[Pose3], wTi_list: List[Pose3]) -> Dict
...
+    averaging_metrics_fpath = REPO_ROOT / "result_metrics" / "multiview_optimizer_metrics.json"
+    averaging_metrics = io_utils.read_json_file(averaging_metrics_fpath)
+
+    logger.info("Averaging median_rot_err_deg: %.2f", averaging_metrics["rotation_averaging_angle_deg"]["median_error"])
+    logger.info("Averaging max_rot_err_deg: %.2f", averaging_metrics["rotation_averaging_angle_deg"]["max_error"])
+
+    logger.info(
+        "Averaging median_trans_dist_err: %.2f", averaging_metrics["translation_averaging_distance"]["median_error"]
+    )
+    logger.info("Averaging max_trans_dist_err: %.2f", averaging_metrics["translation_averaging_distance"]["max_error"])
```
akshay-krishnan, line 258: I cant stress enough how hardcoding the metrics names here is so hacky. If we change the metric in a module we have to change all this.

akshay-krishnan:
> Why do we read jsons from disk instead of data being passed as inputs?

That is what I've been advocating for :)

akshay-krishnan:
I dont have strong opinions (though Im not much in favor) with merging this. We can do it if you need it, but I see it as something we'd have to change in the future (i.e adding more work).

johnwlambert:
Thanks for the review @akshay-krishnan @ayushbaid. So currently we dont have a simple way to aggregate info to put in an experiments sheet without something like this. Otherwise we need to manually inspect output in a few different files.

If we want to do the more generic metrics class in the future, I'm all for it. But this is a bit of a stopgap in the meantime.

gtsfm/utils/io.py
johnwlambert: I included it just to avoid writing `with(blah): json.load()` over and over again.

Purely for convenience.

johnwlambert: I thought so too, but I grepped the repo and couldnt find it in `master`.

gtsfm/utils/io.py
```diff
@@ -101,6 +101,19 @@ def save_json_file(
...
+    """Load dictionary from JSON file.
+
+    Args:
+        fpath: Path to JSON file.
+
+    Returns:
+        Deserialized Python dictionary or list.
+    """
+    with open(fpath, "rb") as f:
```
ayushbaid, line 113: should we read/write in plaintext?

gtsfm/utils/metrics.py
ayushbaid: So you think we should just loop over the keys of the dictionary and print them? in other words, should the organization of the metric in the dictionary be structured so that we can print any dictionary?

ayushbaid (approved):

gtsfm/utils/metrics.py
johnwlambert: So some or the key value pairs are very verbose and not informative. I prefer in a future PR we come up with a more long term solution for printing out the summary statistics of interest with a generic metric class

akshay-krishnan (approved):
Yeah, I understand this is a temporary solution.

gtsfm/utils/io.py
johnwlambert: plaintext should be sufficient. i'll update it.

PR #209, opened 2021-07-05
Add python-based viewer/visualization

johnwlambert:
Adds a python-based viewer for GTSFM results. Supported rendering backends are Mayavi mlab or Open3d:

### as Mayavi spheres:
Screen Shot 2021-07-05 at 7 22 28 PM

### as Open3d points:
Screen Shot 2021-07-05 at 7 19 57 PM

### as Open3d spheres:
Screen Shot 2021-07-05 at 7 21 23 PM


view_scene.py
```diff
@@ -0,0 +1,347 @@
...
+    Args:
+        point_cloud: array of shape (N,3) representing 3d points.
+
+    Returns:
+        mean_pt: coordinates of central point, ignoring outliers.
+    """
+    ranges = np.linalg.norm(point_cloud, axis=1)
+    outlier_thresh = np.percentile(ranges, 75)
+    mean_pt = point_cloud[ranges < outlier_thresh].mean(axis=0)
+    return mean_pt
```
johnwlambert, line 215: let me know if you think there is a better way to do this. Perhaps largest CC of density-based clustering?

view_scene.py
```diff
@@ -0,0 +1,112 @@
...
+
+    Args:
+        args: rendering options.
+    """
+    points_fpath = f"{args.output_dir}/points3D.txt"
+    images_fpath = f"{args.output_dir}/images.txt"
+    cameras_fpath = f"{args.output_dir}/cameras.txt"
+
+    wTi_list, img_fnames = io_utils.read_images_txt(images_fpath)
+    calibrations = io_utils.read_cameras_txt(cameras_fpath)
```
akshay-krishnan, line 48: Unrelated to this PR: shouldnt wTi_list be returned by read_cameras semantically?

view_scene.py
```diff
@@ -0,0 +1,112 @@
...
+
+    point_cloud, rgb = io_utils.read_points_txt(points_fpath)
+
+    mean_pt = compute_point_cloud_center_robust(point_cloud)
+
+    # Zero-center the point cloud (about estimated center)
+    zcwTw = Pose3(Rot3(np.eye(3)), -mean_pt)
+    n = point_cloud.shape[0]
+    for j in range(n):
+        point_cloud[j] = zcwTw.transformFrom(point_cloud[j])
```
akshay-krishnan, line 61: Is it easier to simply do `point_cloud -= mean_pt`?

view_scene.py
akshay-krishnan: clustering sounds like unnecessary computation for visualization. I think this is fine. Alternatively, you could filter out points beyond 2 x sigma or 3 x sigma.

view_scene.py
```diff
@@ -0,0 +1,112 @@
+"""
+Script to render a GTSFM scene using either Open3d or Mayavi mlab.
```
akshay-krishnan, line 2: move to visualization?

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,116 @@
...
+from colour import Color
+from gtsam import Cal3Bundler, Pose3
+from mayavi import mlab
+
+
+from gtsfm.common.view_frustum import ViewFrustum
+
+
+def draw_point_cloud_mayavi(
+    args: argparse.Namespace, fig: mayavi.core.scene.Scene, point_cloud: np.ndarray, rgb: np.ndarray
```
akshay-krishnan, line 21: Is only sphere_radius required? No need pass all arguments.

akshay-krishnan: same for other usages.

view_scene.py
johnwlambert: hmm what do you mean by "return semantically"?

view_scene.py
johnwlambert: so since `point_cloud` is passed by reference, I didn't want to modify it. Every row has a corresponding row in `rgb` that would need to be modified together

view_scene.py
johnwlambert: thanks, updated

visualization/mayavi_vis_utils.py
johnwlambert: i changed it in just two functions. let me know if you think i should change it in more places also

view_scene.py
akshay-krishnan: Is the wTi_list ground truth?

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,117 @@
...
+def draw_point_cloud_mayavi(
+    sphere_radius: float, fig: mayavi.core.scene.Scene, point_cloud: np.ndarray, rgb: np.ndarray
+) -> None:
+    """Render a point cloud as a collection of spheres, using Mayavi.
+
+    Args:
+        sphere_radius: radius of each rendered sphere.
+        fig: Mayavi figure object.
+        point_cloud: array of shape (N,3) representing 3d points.
+        rgb: uint8 array of shape (N,3) representing colors in RGB order, in the range [0,255]
```
akshay-krishnan, line 30: I think point_cloud and rgb should come before sphere_radius in args list.

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,117 @@
...
+def draw_cameras_mayavi(
+    zcwTw: Pose3, fig: mayavi.core.scene.Scene, calibrations: List[Cal3Bundler], wTi_list: List[Pose3]
+) -> None:
+    """Render camera frustums as collections of line segments, using Mayavi mlab.
+
+    Args:
+        zcwTw: transforms world points to a new world frame where the point cloud is zero-centered
+        fig: Mayavi mlab figure object.
+        calibrations: calibration object for each camera
+        wTi_list: list of camera poses for each image
```
akshay-krishnan, line 54: Same here. I honk wTi_list and calibrations should come before, as they are the most important.

visualization/open3d_vis_utils.py
```diff
@@ -0,0 +1,149 @@
...
+) -> List[open3d.geometry.TriangleMesh]:
+    """Create a colored sphere mesh for every point inside the point cloud, using Open3d.
+
+    Note: this is quite computationally expensive.
+
+    Args:
+        sphere_radius: radius of each rendered sphere.
+        point_cloud: array of shape (N,3) representing 3d points.
+        rgb: uint8 array of shape (N,3) representing colors in RGB order, in the range [0,255]
+
```
akshay-krishnan, line 48: Move point_cloud and rgb to first two args?

akshay-krishnan: Because you could potentially have a default radius.

visualization/open3d_vis_utils.py
```diff
@@ -0,0 +1,149 @@
...
+def create_all_frustums_open3d(
+    zcwTw: Pose3, calibrations: List[Cal3Bundler], wTi_list: List[Pose3]
+) -> List[open3d.geometry.LineSet]:
+    """Render camera frustums as collections of line segments, using Open3d.
+
+    Args:
+        zcwTw: transforms world points to a new world frame where the point cloud is zero-centered
+        calibrations: calibration object for each camera
+        wTi_list: list of camera poses for each image
```
akshay-krishnan, line 77: Args order: wTi_list, calibrations, zwcTw?

visualization/open3d_vis_utils.py
```diff
@@ -0,0 +1,149 @@
...
+                lines=open3d.utility.Vector2iVector(lines),
+            )
+            line_set.colors = open3d.utility.Vector3dVector(colors)
+            line_sets.append(line_set)
+
+    return line_sets
+
+
+def draw_scene_open3d(
+    args: argparse.Namespace,
```
akshay-krishnan, line 124: Shouldn’t the args come at the end? I think that is standard.

akshay-krishnan (approved):
Should be good to merge after the above comments are addressed.

view_scene.py
akshay-krishnan: You could make a copy of point_cloud. Broadcasting can be much faster than an explicit for loop.

view_scene.py
johnwlambert: no, it just represents poses stored in a COLMAP-format file.
It could be GT, or it could be the output of GTSFM.

view_scene.py
johnwlambert: oh i see what you mean. good point; i had misunderstood

visualization/mayavi_vis_utils.py
johnwlambert: good point, updated

visualization/mayavi_vis_utils.py
johnwlambert: Sure. I moved `zcwTw` to the end, but still have `fig` as the first arg. let me know if this is what you had in mind

visualization/open3d_vis_utils.py
johnwlambert: good point, updated

visualization/open3d_vis_utils.py
johnwlambert: updated

visualization/open3d_vis_utils.py
johnwlambert: thanks, updated

visualization/open3d_vis_utils.py
akshay-krishnan: documentation has not been updated.

visualization/open3d_vis_utils.py
akshay-krishnan: docs not updated.

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    rgb: np.ndarray,
+    wTi_list: List[Pose3],
+    calibrations: List[Cal3Bundler],
+    zcwTw: Pose3,
+    args: argparse.Namespace,
+) -> None:
+    """Render camera frustums and a 3d point cloud against a white background, using Mayavi.
+
+    Args:
+        point_cloud
```
akshay-krishnan, line 104: missing docstring for point_cloud

visualization/mayavi_vis_utils.py
akshay-krishnan: should be fine

visualization/mayavi_vis_utils.py
johnwlambert: thanks, should be fixed now

visualization/open3d_vis_utils.py
johnwlambert: just fixed it, thanks

visualization/open3d_vis_utils.py
johnwlambert: just updated it ,thanks

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,115 @@
...
+    Args:
+        point_cloud: array of shape (N,3) representing 3d points.
+        rgb: uint8 array of shape (N,3) representing colors in RGB order, in the range [0,255].
+        wTi_list: list of camera poses for each image.
+        calibrations: calibration object for each camera.
+        zcwTw: transforms world points to a new world frame where the point cloud is zero-centered.
+        args: rendering options.
+    """
+    bgcolor = (1, 1, 1)
+    fig = mlab.figure(figure=None, bgcolor=bgcolor, fgcolor=None, engine=None, size=(1600, 1000))  # type: ignore
```
ayushbaid, line 112: Is 1000x1000 size too big?

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,115 @@
...
+def draw_cameras_mayavi(
+    fig: mayavi.core.scene.Scene, wTi_list: List[Pose3], calibrations: List[Cal3Bundler], zcwTw: Pose3
+) -> None:
+    """Render camera frustums as collections of line segments, using Mayavi mlab.
+
+    Args:
+        fig: Mayavi mlab figure object.
+        wTi_list: list of camera poses for each image.
+        calibrations: calibration object for each camera.
+        zcwTw: transforms world points to a new world frame where the point cloud is zero-centered.
```
ayushbaid, line 52: why not just accept points which are zero-centered? I dont see the point of performing centering in this function. This function should just be concerned with viewing.

visualization/view_scene.py
```diff
@@ -0,0 +1,111 @@
...
+def compute_point_cloud_center_robust(point_cloud: np.ndarray) -> np.ndarray:
+    """Robustly estimate the point cloud center.
+
+    Args:
+        point_cloud: array of shape (N,3) representing 3d points.
+
+    Returns:
+        mean_pt: coordinates of central point, ignoring outliers.
+    """
+    ranges = np.linalg.norm(point_cloud, axis=1)
```
ayushbaid, line 31: is there some reference for this?

visualization/view_scene.py
```diff
@@ -0,0 +1,111 @@
...
+    parser = argparse.ArgumentParser(description="Visualize GTSFM result with Mayavi or Open3d.")
+    parser.add_argument(
+        "--rendering_library",
+        type=str,
+        default="mayavi",
+        choices=["mayavi", "open3d"],
+        help="3d rendering library to use.",
+    )
+    parser.add_argument(
+        "--output_dir", type=str, default=os.path.join(REPO_ROOT, "results", "ba_output"), help="Path to GTSFM output"
```
ayushbaid, line 85: I think we should take the full path of the GTSFMData object as input. I don't find `output` term as a clear one.

ayushbaid (changes requested):

visualization/view_scene.py
johnwlambert: So we use 3 file paths inside this directory, to create the GTSFMData. I think providing 1 directory is easier than providing the 3 file paths:
```
    points_fpath = f"{args.output_dir}/points3D.txt"
    images_fpath = f"{args.output_dir}/images.txt"
    cameras_fpath = f"{args.output_dir}/cameras.txt"
```
Maybe I should clarify the argparse docstring for this arg more @ayushbaid ?

visualization/mayavi_vis_utils.py
johnwlambert: so my main idea is here -- we apply a transformation to the points, so we need to apply the same transformation to the cameras.

Would it be simpler if we just transformed the `wTI_list` before plotting the cameras?

visualization/view_scene.py
johnwlambert: so the main idea was to just that the worst outliers were far-away points. So i restrict the range when estimating the center of the point cloud.

visualization/mayavi_vis_utils.py
johnwlambert: on my screen, 1600x1000 took up the whole screen, which made it easier to see in detail the scene.

i think a smaller window is harder to parse visually.

visualization/view_scene.py
ayushbaid: Makes sense. Maybe use '--gtsfm-data-dir' or '--scene-data-dir'?

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,115 @@
+"""
+Utilities for rendering camera frustums and 3d point clouds using the Mayavi mlab library.
+
+Author: John Lambert
```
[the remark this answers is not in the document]
johnwlambert: thanks, updated

visualization/mayavi_vis_utils.py
```diff
@@ -0,0 +1,115 @@
...
+def draw_point_cloud_mayavi(
+    fig: mayavi.core.scene.Scene, point_cloud: np.ndarray, rgb: np.ndarray, sphere_radius: float
+) -> None:
+    """Render a point cloud as a collection of spheres, using Mayavi.
+
+    Args:
+        fig: Mayavi figure object.
+        point_cloud: array of shape (N,3) representing 3d points.
+        rgb: uint8 array of shape (N,3) representing colors in RGB order, in the range [0,255].
+        sphere_radius: radius of each rendered sphere.
```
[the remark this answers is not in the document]
johnwlambert: so we have a default value in the argparse defaults. i would prefer to just keep it there, so we dont have to update defaults in two places (here and in the argparse defaults). the argparse default will flow down to here

visualization/mayavi_vis_utils.py
johnwlambert: cool, so we don't have an explicit GtsfmData instance here, but I moved the camera pose shift to the main loop in `view_scene.py`, before passing it to the Open3d or Mayavi vis utilities.

visualization/view_scene.py
johnwlambert: so when I think of "data dir", I think of the input image directory. And the same for scene data dir. But I cleaned up the `help` description. Let me know if this is clearer now.

ayushbaid (approved):
LGTM!

PR #215, opened 2021-07-28
Adding check for pair index ordering (Closes #210)

### LINKED ISSUE #210: Enforce in base loader that image pairs i.e. edges (i1,i2) are ordered as i1 < i2.
We need to enforce in the base loader that image pairs i.e. edges (i1,i2) are ordered as i1 < i2.

We also need to add this in documentation for each loader.

ayushbaid:
This PR introduces the check of idx1 < idx2 in an (idx1, idx2) pair. This is already being enforced by all loader implementations, and now will be introduced in the base class too.

### DIFF
--- a/gtsfm/loader/colmap_loader.py
+++ b/gtsfm/loader/colmap_loader.py
@@ -96,13 +96,9 @@ class ColmapLoader(LoaderBase):
         img = io_utils.load_image(self._image_paths[0])
         sample_h, sample_w = img.height, img.width
         # no downsampling may be required, in which case scale_u and scale_v will be 1.0
-        (
-            self._scale_u,
-            self._scale_v,
-            self._target_h,
-            self._target_w,
-        ) = img_utils.get_downsampling_factor_per_axis(sample_h, sample_w, self._max_resolution)
-
+        (self._scale_u, self._scale_v, self._target_h, self._target_w,) = img_utils.get_downsampling_factor_per_axis(
+            sample_h, sample_w, self._max_resolution
+        )
 
     def __len__(self) -> int:
         """The number of images in the dataset.
@@ -178,7 +174,7 @@ class ColmapLoader(LoaderBase):
         return wTi
 
     def is_valid_pair(self, idx1: int, idx2: int) -> bool:
-        """Checks if (idx1, idx2) is a valid pair.
+        """Checks if (idx1, idx2) is a valid pair. idx1 < idx2 is required.
 
         Args:
             idx1: first index of the pair.
@@ -187,4 +183,4 @@ class ColmapLoader(LoaderBase):
         Returns:
             validation result.
         """
-        return idx1 < idx2 and abs(idx1 - idx2) <= self._max_frame_lookahead
+        return super().is_valid_pair(idx1, idx2) and abs(idx1 - idx2) <= self._max_frame_lookahead
--- a/gtsfm/loader/loader_base.py
+++ b/gtsfm/loader/loader_base.py
@@ -69,9 +69,10 @@ class LoaderBase(metaclass=abc.ABCMeta):
             the camera pose w_P_index.
         """
 
-    @abc.abstractmethod
     def is_valid_pair(self, idx1: int, idx2: int) -> bool:
-        """Checks if (idx1, idx2) is a valid pair.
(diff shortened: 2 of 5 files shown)


gtsfm/loader/loader_base.py
```diff
@@ -80,6 +81,7 @@ def is_valid_pair(self, idx1: int, idx2: int) -> bool:
         Returns:
             validation result.
         """
+        return idx1 < idx2
```
johnwlambert, line 84: thanks for the PR, Ayush. Should we add a unit test on this?

gtsfm/loader/yfcc_imb_loader.py
```diff
@@ -129,7 +125,10 @@ def is_valid_pair(self, idx1: int, idx2: int) -> bool:
         Returns:
             validation result.
         """
-        return (idx1, idx2) in self._image_pairs
+        if not super().is_valid_pair(idx1, idx2):
```
johnwlambert, line 128: will this affect the results across their benchmark in any way? i.e. how is their GT provided?

gtsfm/loader/yfcc_imb_loader.py
```diff
@@ -129,7 +125,10 @@ def is_valid_pair(self, idx1: int, idx2: int) -> bool:
...
+            return False
+
+        return (idx1, idx2) in self._image_pairs or (idx2, idx1) in self._image_pairs
```
johnwlambert, line 131: why do we need both ways now?

ayushbaid: because IMB dataset does not have any implied ordering.

A pair between image (1, 5) can be either stored as (1, 5) or (5, 1) in IMB, but not both. I'll modify the logic to just flip the pairs in the initialization step, so that this confusing check if not needed.

gtsfm/loader/loader_base.py
johnwlambert: @ayushbaid gentle reminder about this : - )

gtsfm/loader/yfcc_imb_loader.py
ayushbaid: No it should not. We are just being consistent in how we store the pairs internally in this class.

johnwlambert:
Hi Ayush, this PR looks good -- could we add just one unit test, and pull in master?

tests/loader/test_olsson_loader.py
```diff
@@ -148,6 +141,20 @@ def test_create_computation_graph_for_intrinsics(self) -> None:
...
 
+    @patch("gtsfm.loader.loader_base.LoaderBase.is_valid_pair", return_value=True)
+    def test_is_valid_pair_within_lookahead(self, base_is_valid_pair_mock) -> None:
+        i1 = 1
+        i2 = 10
+        self.assertTrue(self.loader.is_valid_pair(i1, i2))
+        base_is_valid_pair_mock.assert_called_once_with(i1, i2)
+
+    @patch("gtsfm.loader.loader_base.LoaderBase.is_valid_pair", return_value=True)
+    def test_is_valid_pair_outside_lookahead(self, base_is_valid_pair_mock) -> None:
```
johnwlambert, line 152: nit: type hint

gtsfm/loader/loader_base.py
```diff
@@ -83,9 +83,26 @@ def get_camera_pose(self, index: int) -> Optional[Pose3]:
             the camera pose w_P_index.
         """
 
+    def get_camera(self, index: int) -> Optional[PinholeCameraCal3Bundler]:
```
johnwlambert: this shouldn't be new in this branch, right @ayushbaid ?

gtsfm/loader/colmap_loader.py
```diff
@@ -93,14 +93,6 @@ def __init__(
         self._num_imgs = len(self._image_paths)
         logger.info("Colmap image loader found and loaded %d images", self._num_imgs)
 
-        # read one image, to check if we need to downsample the images
-        img = io_utils.load_image(self._image_paths[0])
-        sample_h, sample_w = img.height, img.width
-        # no downsampling may be required, in which case scale_u and scale_v will be 1.0
```
johnwlambert: this was changed a while back -- is the branch base stale?

PR #223, opened 2021-07-31
Generic GtsfmMetric class to store metrics

akshay-krishnan:
Adding two classes `GtsfmMetric` and `GtsfmMetricsGroup` in the evaluation directory. These classes can be used to store, save, load and plot metrics, all of which will be done in future PRs. 

The added tests illustrate the intended usage of these classes.


gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+class GtsfmMetric:
+    """Class to store a metric computed in a GTSfM module."""
+
+    class PlotType(Enum):
+        BAR = 1  # For scalars
+        BOX = 2  # For 1D distributions
+        HISTOGRAM = 3  # For 1D distributions
+
+    def _get_plot_types_for_dim(self, dim) -> List[PlotType]:
```
johnwlambert, line 26: Thanks for the PR, Akshay. Looks great. Missing type hint on dim

johnwlambert: Double underscore for private methods?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+
+        Returns:
+            summary as a dict that can be serialized to JSON for storage.
+        """
+        if data.ndim != 1:
+            raise ValueError("Metric must be a 1D distribution to get summary.")
+        summary = {
+            "min": np.min(data).tolist(),
+            "max": np.max(data).tolist(),
+            "median": np.median(data).tolist(),
```
johnwlambert, line 88: Should it be nanmean and nanmedian?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+        return output
+
+    def __init__(
+        self,
+        name: str,
+        data: Optional[Union[np.array, float, List[Union[int, float]]]] = None,
+        summary: Optional[Dict[str, Any]] = None,
+        store_full_data: bool = True,
+        plot_type: PlotType = None,
+    ):
```
johnwlambert, line 121: Missing return type?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+            Quartiles of the data as a dict where keys are q0, q1, q2, q3, and q4
+        """
+        query = list(range(0, 101, 25))
+        quartiles = np.percentile(data, query)
+        output = {}
+        for i, q in enumerate(query):
+            output["q" + str(i)] = quartiles[i].tolist()
+        return output
+
+    def __init__(
```
johnwlambert, line 114: Should the constructor be the first method listed in the class?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+    def extend(self, metrics_group: GtsfmMetricsGroup):
+        self._metrics.extend(metrics_group.metrics)
+
+    def get_metrics_as_dict(self) -> Dict[str, Dict[str, Any]]:
+        metrics_dict = {}
+        for metric in self._metrics:
+            metrics_dict.update(metric.get_metric_as_dict())
+        return {self._name: metrics_dict}
+
+    def save_to_json(self, path: str):
```
johnwlambert, line 269: Missing return type

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,134 @@
+"""Tests for GtsfmMetric and GtsfmMetricsGroup. 
+
+Authors: Akshay Krishnan
+"""
+import copy
+import numpy as np
+import unittest
+import tempfile
+import os
```
johnwlambert, line 9: Needs to be listed in three groups. I think not sorted alphabetically currently.

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,134 @@
...
+        self._metric_dict_no_data = copy.deepcopy(self._metric_dict_histogram)
+        del self._metric_dict_no_data["bar_metric"]["full_data"]
+
+    def test_create_scalar_metric(self):
+        metric = GtsfmMetric("a_scalar", 2)
+        self.assertEqual(metric.name, "a_scalar")
+        np.testing.assert_equal(metric.data, np.array([2]))
+        self.assertEqual(metric.plot_type, GtsfmMetric.PlotType.BAR)
+
+    def test_create_1d_distribution_metric(self):
```
johnwlambert, line 65: Could you explain in a comment why box is the only valid Plot type here?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+from __future__ import annotations
+
+import json
+import numpy as np
+from enum import Enum
+from typing import Any, Dict, List, Union
+
+import gtsfm.utils.io as io
+
+DATA_KEY = "full_data"
```
johnwlambert, line 14: Could you provide some high level comments explaining what these keys mean? How is full data defined, and when is it preferred?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+        metrics_dict = {}
+        for metric in self._metrics:
+            metrics_dict.update(metric.get_metric_as_dict())
+        return {self._name: metrics_dict}
+
+    def save_to_json(self, path: str):
+        io.save_json_file(path, self.get_metrics_as_dict())
+
+    @classmethod
+    def parse_from_dict(cls, metrics_group_dict) -> GtsfmMetricsGroup:
```
johnwlambert, line 273: Could you add A docstring explaining where/why this would be used in a GTSFM workflow? Parsing from json makes sense for later vis,  i I was confused about this  one

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,287 @@
...
+            "mean": np.mean(data).tolist(),
+            "stddev": np.std(data).tolist(),
+        }
+        if self._plot_type == self.PlotType.BOX:
+            summary.update({"quartiles": self._get_distribution_quartiles(data)})
+        elif self._plot_type == self.PlotType.HISTOGRAM:
+            summary.update({"histogram": self._get_distribution_histogram(data)})
+        return summary
+
+    def _get_distribution_quartiles(self, data: np.ndarray) -> Dict[int, float]:
```
johnwlambert, line 98: Where were you thinking we would use quartile info? Just for dumping to stdout or json?

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,134 @@
...
+
+class TestGtsfmMetricsGroup(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self._metrics_list = []
+        self._metrics_list.append(GtsfmMetric("metric1", 2))
+        self._metrics_list.append(GtsfmMetric("metric2", np.array([1, 2, 3])))
+        self._metrics_group = GtsfmMetricsGroup("test_metrics", self._metrics_list)
+
+    def test_get_metric_as_dict(self):
```
johnwlambert, line 113: Missing return types here

gtsfm/evaluation/metrics.py
akshay-krishnan: doesnt return anything, what do we do then?

gtsfm/evaluation/metrics.py
akshay-krishnan: The quartiles are used for the box plots if raw data is not stored. Raw data can be really big for things like track lengths or reprojection errors, and should not be stored.

gtsfm/evaluation/metrics.py
akshay-krishnan: this is only used for viz, I am adding documentation to the method, I dont think the documentation should say where its used though, thats upto the user.

gtsfm/evaluation/metrics.py
akshay-krishnan: It could also be used for logging if thats necessary.

gtsfm/evaluation/metrics.py
akshay-krishnan: Looks like single underscores are the recommended choice, do we have a gtsfm specific convention for this? https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

gtsfm/evaluation/metrics.py
akshay-krishnan: done

gtsfm/evaluation/metrics.py
akshay-krishnan: done

tests/evaluation/metrics_test.py
akshay-krishnan: updated, please let me know if the order is incorrect.

tests/evaluation/metrics_test.py
akshay-krishnan: Its not the only valid one, but the default one.

akshay-krishnan: explained this further in the constructor description.

tests/evaluation/metrics_test.py
johnwlambert: thanks. I think it should go
```python
import copy
import os
import tempfile
import unittest
```
right?

gtsfm/evaluation/metrics.py
johnwlambert: We can just do 
```python
def save_to_json(self, path: str) -> None:
```

https://stackoverflow.com/questions/48223558/typehints-none-or-leave-blank
example here from PEP: https://www.python.org/dev/peps/pep-0484/#callable

gtsfm/evaluation/metrics.py
johnwlambert: i think all of our data is still pretty small for now. Weren't you using full data to show all the plotly plots this past week?

johnwlambert:
Things are looking good. I think it might be helpful to add some of the places where we plan to use this (the viz code and the actual changes to code files) in this PR as well, so we can use it to tweak the API a bit now. Bit hard to decide in isolation

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,357 @@
...
+   If the metric is scalar, it is stored simply as {metric_name: value}. 
+"""
+DATA_KEY = "full_data"
+SUMMARY_KEY = "summary"
+
+
+class GtsfmMetric:
+    """Class to store a metric computed in a GTSfM module."""
+
+    class PlotType(Enum):
```
johnwlambert, line 34: could you add a brief inline comment explaining why we have plot code here, whenever thing else is agnostic, and doesn't mention visualization? is the idea that we have to choose apriori which type of plot we prefer for each metric, rather than rendering all possible ones, like only one of box plot or histogram would be shown?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,357 @@
...
+        If the data is float, the keys of the dictionary are interval buckets, and if the data is int, the keys are also int.
+
+        Args:
+            data: 1D array of all values of the metric
+
+        Returns:
+            Histogram of data as a dict from bucket to count.
+        """
+        if data.size == 0:
+            print("Requested histogram for empty data metric, returning None.")
```
johnwlambert, line 138: can we use the logger instead of printing this?

gtsfm/evaluation/metrics.py
johnwlambert: wouldn't we parse from JSON, not from a dict, in the viz code? or is it that we first read in from JSON as a dict, and then parse it?

gtsfm/evaluation/metrics.py
johnwlambert: this could be a good assumption to add as an inline comment, if true

gtsfm/evaluation/metrics.py
johnwlambert: ok i looked and we call `parse_from_dict()` inside of `parse_from_json()`. But would we ever use `parse_from_dict()` besides from calling it inside the json classmethod? I guess i'm wondering if it's necessary

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,135 @@
...
+        metric = GtsfmMetric("to_be_written_metric", np.arange(10.0))
+        with tempfile.TemporaryDirectory() as tempdir:
+            metric.save_to_json(os.path.join(tempdir, "test_metrics.json"))
+
+
+class TestGtsfmMetricsGroup(unittest.TestCase):
+    def setUp(self):
+        super().setUp()
+        self._metrics_list = []
+        self._metrics_list.append(GtsfmMetric("metric1", 2))
```
johnwlambert, line 110: keyword arg would be helpful here:
```python
self._metrics_list.append(GtsfmMetric(name="metric1", data=2))
self._metrics_list.append(GtsfmMetric(name="metric2", data=np.array([1, 2, 3])))
```

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,357 @@
...
+            "stddev": np.nanstd(data).tolist(),
+        }
+        if self._plot_type == self.PlotType.BOX:
+            summary.update({"quartiles": self._get_distribution_quartiles(data)})
+        elif self._plot_type == self.PlotType.HISTOGRAM:
+            summary.update({"histogram": self._get_distribution_histogram(data)})
+        return summary
+
+    def _get_distribution_quartiles(self, data: np.ndarray) -> Dict[int, float]:
+        """Computes quartiles for the provided 1D data distribution.
```
johnwlambert, line 192: it would be helpful to explain that quartiles are needed for the box plot here, and this is why we compute and save the quartiles.

gtsfm/evaluation/metrics.py
johnwlambert: or i guess you were just saving this histogram bin counts a priori, and quartiles for the box plots a priori? this could be helpful to put in the module description as a high-level summary of the file's purpose, in line 2 of this file

gtsfm/evaluation/metrics.py
akshay-krishnan: I dont think functions should describe how they are being used. That is to be done at the function call site.

gtsfm/evaluation/metrics.py
akshay-krishnan: I was not too satisfied with the plot type either, its basically used to decide whether to store histogram or quartiles. Let me know if you think theres a better way to do that.

gtsfm/evaluation/metrics.py
akshay-krishnan: > i think all of our data is still pretty small for now.

No, I dont think the reprojection errors are small. data association and/or bundle adjustment metric JSON files are not human readable because there are too many lines. Also, I found some metrics (probably in bundle adjustment) where we store the mean, min, max etc of the same metric without storing the full data.

akshay-krishnan: > or i guess you were just saving this histogram bin counts a priori, and quartiles for the box plots a priori? this could be helpful to put in the module description as a high-level summary of the file's purpose, in line 2 of this file

Yeah, that is what I was doing when we dont have full data. I think should be documented in the module that does the plotting, not here.

gtsfm/evaluation/metrics.py
akshay-krishnan: I think its a useful abstraction. It would be a valid point if rewrote the parsing logic for both parse_from_json() and parse_from_dict(), but parse_from_json() is only calling parse_from_dict() right? 
I actually wanted to remove parse_from_json() and let the caller handle the IO, but this helper might reduce some json loading boilerplate.

tests/evaluation/metrics_test.py
akshay-krishnan: Yeah, my bad. Not sure why sublack did not format it that way though.

akshay-krishnan:
@johnwlambert I'd actually prefer to add the use of this class in another PR. That PR will touch many files, and a single PR might be too big for all this. We can always tweak the API after merging this class if necessary.

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
...
+        return self._summary
+
+    def _get_plot_types_for_dim(self, dim: int) -> List[PlotType]:
+        if dim == 0:
+            return [self.PlotType.BAR]
+        if dim == 1:
+            return [self.PlotType.BOX, self.PlotType.HISTOGRAM]
+        return []
+
+    def _get_distribution_histogram(self, data: np.ndarray) -> Dict[str, Union[float, int]]:
```
ayushbaid, line 131: this does not need to be a function in the class.

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
...
+        histogram = {}
+        for i in range(len(count)):
+            if discrete:
+                key = str(int(bins_lower[i]))
+            else:
+                key = "%.2f-%.2f" % (bins_lower[i], bins_upper[i])
+            histogram[key] = count[i]
+        return histogram
+
+    def _create_summary(self, data: np.ndarray) -> Dict[str, Any]:
```
ayushbaid, line 167: same comment here. Does not need to be in the class.

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
...
+            if DATA_KEY in metric_value:
+                data = metric_value[DATA_KEY]
+            if SUMMARY_KEY in metric_value:
+                summary = metric_value[SUMMARY_KEY]
+            return cls(metric_name, data=data, summary=summary)
+
+        # Scalar metrics
+        return cls(metric_name, metric_value)
+
+
```
ayushbaid, line 279: this can be a named tuple?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
...
+            self._summary = summary
+            self._plot_type = self.PlotType.HISTOGRAM if "histogram" in summary else self.PlotType.BOX
+            self._data = None
+
+    @property
+    def name(self) -> str:
+        return self._name
+
+    @property
+    def data(self) -> np.array:
```
johnwlambert, line 109: should be `-> np.ndarray`

gtsfm/evaluation/metrics.py
johnwlambert: i think i prefer it in the class, since it references the PlotTypes, which are part of the class?

gtsfm/evaluation/metrics.py
johnwlambert: it would need to be a `dataclass` not `NamedTuple` if we want it to be mutable, so we could call `.extend()`

johnwlambert: i think it's fine currently, though

tests/evaluation/metrics_test.py
johnwlambert: no problem

gtsfm/evaluation/metrics.py
johnwlambert: agreed, i would prefer this to be in this file as a free function

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
...
+        For example: box plots store quartiles, where as histogram stores a histogram.
+        """
+        BAR = 1  # For scalars
+        BOX = 2  # For 1D distributions
+        HISTOGRAM = 3  # For 1D distributions
+
+    def __init__(
+        self,
+        name: str,
+        data: Optional[Union[np.array, float, List[Union[int, float]]]] = None,
```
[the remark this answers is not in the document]
johnwlambert: I like that:
```
Distribution1d = List[Union[int, float]]
```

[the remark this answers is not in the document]
johnwlambert: or should it be 
```python
Distribution1d = Union[np.ndarray, List[Union[int, float]]]
```
?

gtsfm/evaluation/metrics.py
johnwlambert: i agree it's not strictly required. But I think it would help with readability to just add a 1-liner about why these might be useful.

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
+"""Class to store metrics computed in different GTSfM modules.
```
johnwlambert, line 1: can we add up here that if the full data is too large to be stored, we save only summary statistics like histogram bin counts or quartiles (e.g. for box plots)? It would make this file easier to understand.

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,362 @@
...
+                     Has to be provided if data = None.
+            store_full_data: Whether all the values have to be stored and saved or only summary is required. True by default.
+            plot_type: The plot to use for visualization of the metric. 
+                       Defaults:
+                          PlotType.BAR if data is a scalar
+                          PlotType.BOX if data is a distribution (other option is PlotType.HISTOGRAM)
+                        It is inferred from the summary if plot_type is not provided and summary is.
+    """
+        if summary is None and data is None:
+            raise ValueError("Data and summary cannot both be None.")
```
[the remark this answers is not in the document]
akshay-krishnan: We actually don't need both for plotting. But sometimes we also use the full_data for debugging. Like camera pose errors in rotation and translation averaging. 
It's just impossible (impractical?) to debug full_data that scales with the number of features/tracks like reprojection error. Not sure if anyone has ever done that... not on realistic datasets at least :)

gtsfm/evaluation/metrics.py
akshay-krishnan: But then users could import it, which we dont want.

akshay-krishnan: The same argument could be used for all internal methods in this class, I made them so because I did not want them to be imported without the class.

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+
+import gtsfm.utils.io as io
+import gtsfm.utils.logger as logger_utils
+
+# Keys to access data and summary in the dictionary representation of metrics.
+DATA_KEY = "full_data"
+SUMMARY_KEY = "summary"
+
+# Type hint for a 1D distribution
+Distribution1D = Union[np.ndarray, float, List[Union[int, float]]]
```
johnwlambert, line 26: did we mean to include `float` as part of `Distribution1D`?

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+from typing import Any, Dict, List, Union
+
+import numpy as np
+
+import gtsfm.utils.io as io
+import gtsfm.utils.logger as logger_utils
+
+# Keys to access data and summary in the dictionary representation of metrics.
+DATA_KEY = "full_data"
+SUMMARY_KEY = "summary"
```
johnwlambert, line 23: would it be clearer to say `FULL_DATA_KEY` and `SUMMARY_KEY`? technically its all "data"

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+            if self._dim == 1:
+                self._summary = self._create_summary(data)
+
+            # Store full data only if its a scalar or if asked to.
+            if self._dim == 0 or store_full_data:
+                self._data = data
+            else:
+                self._data = None
+        else:
+            self._dim = 1
```
johnwlambert, line 119: if data is None, why can't dim be 0-d or 1-d? why only 1d? inline comment would help

gtsfm/evaluation/metrics.py
johnwlambert: this fn doesnt access any private class attributes, though, right? i think we prefer functional when possible, since it makes unit testing so much easier

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,396 @@
...
+    def save_to_json(self, json_filename: str) -> None:
+        """Saves this metric's dict representation to a JSON file.
+
+        Args:
+            Path to the json file.
+        """
+        io.save_json_file(json_filename, self.get_metric_as_dict())
+
+    @classmethod
+    def parse_from_dict(cls, metric_dict: Dict[str, Any]) -> GtsfmMetric:
```
johnwlambert, line 270: again i think clearer to just make this part of `parse_from_json`, if we will never call `parse_from_dict()` on its own anywhere. kind of confusing to figure out where this dict would come from. but `parse_from_json()` is very clear in terms of provenance

akshay-krishnan: GtsfmMetric.parse_from_dict is used in GtsfmMetricsGroup.parse_from_dict. 
Your point would be valid for GtsfmMetricsGroup.parse_from_dict. That I do not plan on using anywhere outside this class. But I felt both classes could have a similar structure. I could remove GtsfmMetricsGroup.parse_from_dict if you'd like that.

gtsfm/evaluation/metrics.py
akshay-krishnan: There is no concept of summary for 0-d. Please refer to the class documentation. Adding a comment nevertheless.

gtsfm/evaluation/metrics.py
akshay-krishnan: no, good catch.

gtsfm/evaluation/metrics.py
akshay-krishnan: Even if we are not going to write a unit test for that function? We are not quite violating the functional paradigm if we have internal methods that do not change any data member of the class. This is still functional, the only difference is where we have the function. Since its not going to be used externally I thought it makes sense to have it as an internal member with an _ in the name indicating this. 
It also makes semantic sense, as these functions are not meaningful outside the context of this class.

akshay-krishnan: While its not my preference, I dont mind moving these functions outside the class. I would move _get_distribution_histogram, _get_distribution_quartiles, and _create_summary, since they are all similar. Let me know if you disagree with my motivation above. :)

gtsfm/evaluation/metrics.py
johnwlambert: ok, i dont feel too strongly about it

gtsfm/evaluation/metrics.py
johnwlambert: could you add that as an inline comment again in line 119, just for clarity again?

gtsfm/evaluation/metrics.py
johnwlambert: thanks for updating this

gtsfm/evaluation/metrics.py
johnwlambert: thanks for updating it

gtsfm/evaluation/metrics.py
akshay-krishnan: I added a comment where we set self._dim
# Metrics created from summary alone are 1D distribution metrics
Not sure if you saw this...

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,399 @@
+"""Classes to store metrics computed in different GTSfM modules.
+
+The GtsfmMetric class stores a single metric, and the GtsfmMetricsGroup stores a list of metrics.
+These classes are used to compute statistics on the metrics (min, max, etc), 
+save them to JSON, parse metrics from JSON, and plot them.
+
+Authors: Akshay Krishnan
+"""
+
+from __future__ import annotations
```
johnwlambert, line 10: any particular reason to include these? i think past 3.5.0 type hints are included, and 3.8 is our supported platform

gtsfm/evaluation/metrics.py
```diff
@@ -0,0 +1,399 @@
...
+    @property
+    def summary(self) -> Dict[str, Any]:
+        return self._summary
+
+    def _get_plot_types_for_dim(self, dim: int) -> List[PlotType]:
+        if dim == 0:
+            return [self.PlotType.BAR]
+        if dim == 1:
+            return [self.PlotType.BOX, self.PlotType.HISTOGRAM]
+        return []
```
johnwlambert, line 152: is a value error better, if dim is not 0 or 1?

gtsfm/evaluation/metrics.py
johnwlambert: i prefer just this one function as a free function, since it doesnt use any aspect of the class.

unit test would be nice too : - )

gtsfm/evaluation/metrics.py
johnwlambert: prefer not a free function

gtsfm/evaluation/metrics.py
johnwlambert: this would be nice as a free function with a unit test

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,137 @@
...
+        np.testing.assert_equal(parsed_metric.data, 2)
+
+    def test_parses_from_dict_1D_distribution(self) -> None:
+        parsed_metric = GtsfmMetric.parse_from_dict(self._metric_dict_quartiles)
+        self.assertEqual(parsed_metric.name, "foo_metric")
+        self.assertEqual(parsed_metric.plot_type, GtsfmMetric.PlotType.BOX)
+        self.assertIn("quartiles", parsed_metric.summary)
+        self.assertIn("full_data", parsed_metric.get_metric_as_dict()[parsed_metric.name])
+
+    def test_parses_from_dict_1D_distribution_histogram(self) -> None:
```
johnwlambert, line 86: simple 1-line docstrings would make this even easier to follow, for each unit test

johnwlambert:
flake8 complaints:
```
gtsfm/evaluation/metrics.py:69:15: F821 undefined name 'Optional'
gtsfm/evaluation/metrics.py:70:18: F821 undefined name 'Optional'
gtsfm/evaluation/metrics.py:80:121: E501 line too long (126 > 120 characters)
gtsfm/evaluation/metrics.py:158:121: E501 line too long (125 > 120 characters)
```

johnwlambert (approved):
approving once a few last changes are addressed.

feel free to merge after addressing those

gtsfm/evaluation/metrics.py
akshay-krishnan: we are using the class name as a type hint for methods in the class itself. 
https://stackoverflow.com/questions/33533148/how-do-i-type-hint-a-method-with-the-type-of-the-enclosing-class

gtsfm/evaluation/metrics.py
akshay-krishnan: I don't think it is an error, if it's not a supported dimension, it does not return any plot types. That is expected behavior.

gtsfm/evaluation/metrics.py
akshay-krishnan: Moved to free function, current usage is covered by the tests in the class, so have not added more tests. The only untested behavior is the histogram for float data, which is not used by any metric currently.

gtsfm/evaluation/metrics.py
akshay-krishnan: please excuse the large font before the edit, that was markdown :/

tests/evaluation/metrics_test.py
johnwlambert: thanks for adding these

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,150 @@
...
+                    "max": 7,
+                    "median": 3.5,
+                    "mean": 3.5,
+                    "stddev": 2,
+                    "quartiles": {
+                        "q0": 0,
+                        "q1": 1,
+                        "q2": 3,
+                        "q3": 5,
+                        "q4": 7,
```
johnwlambert, line 34: should it be:
```
>>> np.percentile( [0, 1, 2, 3, 4, 5, 6, 7], 25 )
1.75
>>> np.percentile( [0, 1, 2, 3, 4, 5, 6, 7], 50 )
3.5
>>> np.percentile( [0, 1, 2, 3, 4, 5, 6, 7], 75 )
5.25
```
?

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,150 @@
...
+        }
+        # A metric with histogram in its summary.
+        self._metric_dict_histogram = {
+            "bar_metric": {
+                "summary": {
+                    "min": 0,
+                    "max": 3,
+                    "median": 1.5,
+                    "mean": 1.5,
+                    "stddev": 2,
```
johnwlambert, line 48: should it be
```
>>> np.std( [0, 1, 0, 2, 2, 1, 3, 3] )
1.118033988749895
```

tests/evaluation/metrics_test.py
```diff
@@ -0,0 +1,150 @@
...
+        super().setUp()
+        # A metric with quartiles in its summary.
+        self._metric_dict_quartiles = {
+            "foo_metric": {
+                "summary": {
+                    "min": 0,
+                    "max": 7,
+                    "median": 3.5,
+                    "mean": 3.5,
+                    "stddev": 2,
```
johnwlambert, line 28: should it be
```
>>> np.std( [0, 1, 2, 3, 4, 5, 6, 7] )
2.29128784747792
```
?

akshay-krishnan:
Thanks John, updated them. I had approximated these values, was not checking them.

PR #224, opened 2021-08-03
Thirdparty ETH3D multi-view-evaluation

codyly:
Polygon file (.ply) dense point cloud evaluation tool implemented by ETH3D program. Source code in https://github.com/ETH3D/multi-view-evaluation.


johnwlambert:
Hi @codyly, thanks for the PR. Could you discuss a bit more about the functionality these files provide? Could we re-implement it in under 100 lines of numpy?

codyly:
Hi John @johnwlambert, thanks for your comments. These codes implement functions including:
1. Read a MeshLab project (contains several ground-truth scans with corresponding global transformation matrices);
2. Initialize spherical coordinates grid for evaluation;
3. Perform coordinates transformation from cartesian (XYZ) to spherical coordinates (radius, azimuth, and inclination);
4. Find closest points on the same scan ray (with certain azimuth and inclination) in ground-truth scans and reconstruction results;
5. Aggregate valid point counts for calculating Accuracy/Completeness and visualize the results.

The most inconvenient place to use the original implementation is that we have to generate the polygon files and MeshLab project first before we invoke the binary file. Things will be much simpler if we can implement a NumPy version. It's a little bit complicated but I'll try first.

travisdriver (changes requested):
Why not just add this as a submodule through git as opposed to copying all of the source code into GTSfM?

PR #70, opened 2021-01-07
Adding robust noise model for measurement noise

ayushbaid:
The input data for bundle adjustment occasionally contains outliers in tracks or 3d points. A robust noise model will help us tackle these outliers.

In this PR, I add a boolean option to switch to huber function based noise for 2d measurement factors. 

Right now, this PR has merged in changes from #65. I'll add the changes from that PR too.

PR #225, opened 2021-08-04
Adding 2-view Bundle adjustment post verification

ayushbaid:
This PR introduces a 2-view bundle adjustment step which refines the pose estimate after verification.

# Metrics

## skydio-crane dataset with 8 images (using SIFT frontend and max lookahead of 8):
2-view metrics:
- We achieve one more relative rotation success (12 vs 11) at an angular error threshold of 10 degrees.
- Relative unit translation success remains the same (14)
- Pose success increases by 1 (11 vs 10)

BA metrics:
- Output number of tracks:
  - prefiltering goes down 662 vs 800
  - post filtering goes up 281 vs 228


gtsfm/two_view_estimator.py
```diff
@@ -25,6 +30,13 @@
 pil_logger = logging.getLogger("PIL")
 pil_logger.setLevel(logging.INFO)
 
+BA_NOISE_MODEL_EPIPOLAR_ALGEBRAIC_ERROR = gtsam.noiseModel.Isotropic.Sigma(1, 0.01)
```
johnwlambert, line 33: Thanks for the PR, Ayush. Very cool to see the improvement.

Do we use `BA_NOISE_MODEL_EPIPOLAR_ALGEBRAIC_ERROR` anywhere?

tests/test_two_view_estimator.py
```diff
@@ -0,0 +1,71 @@
...
+import gtsfm.utils.geometry_comparisons as comp_utils
+import gtsfm.utils.io as io_utils
+from gtsfm.two_view_estimator import TwoViewEstimator
+
+GTSAM_EXAMPLE_FILE = "18pointExample1.txt"
+EXAMPLE_DATA = io_utils.read_bal(gtsam.findExampleDataFile(GTSAM_EXAMPLE_FILE))
+
+
+class TestTwoViewEstimator(unittest.TestCase):
+    """Unit tests for the 2-view estimator"""
```
johnwlambert, line 21: thanks for adding the test. could you add a sentence about the data you're using for the test, and why you chose it?

tests/test_two_view_estimator.py
```diff
@@ -0,0 +1,71 @@
...
+            camera_intrinsics_i1=Cal3Bundler(),
+            camera_intrinsics_i2=Cal3Bundler(),
+            i2Ri1_initial=i2Ei1.rotation(),
+            i2Ui1_initial=i2Ei1.direction(),
+        )
+
+        self.assertLessEqual(comp_utils.compute_relative_rotation_angle(i2Ri1_optimized, i2Ei1.rotation()), 1)
+        self.assertLessEqual(comp_utils.compute_relative_unit_translation_angle(i2Ui1_optimized, i2Ei1.direction()), 1)
+
+        # self.assertEqual(i2Ri1_optimized, i2Ei1.rotation())
```
johnwlambert, line 65: commented out code? what sort of tolerance do we expect here?

tests/test_two_view_estimator.py
```diff
@@ -0,0 +1,71 @@
...
+            keypoints_i2=Keypoints(normalized_coordinates_i2),
+            verified_corr_idxes=np.hstack([np.arange(normalized_coordinates_i1.shape[0]).reshape(-1, 1)] * 2),
+            camera_intrinsics_i1=Cal3Bundler(),
+            camera_intrinsics_i2=Cal3Bundler(),
+            i2Ri1_initial=i2Ei1.rotation(),
+            i2Ui1_initial=i2Ei1.direction(),
+        )
+
+        self.assertLessEqual(comp_utils.compute_relative_rotation_angle(i2Ri1_optimized, i2Ei1.rotation()), 1)
+        self.assertLessEqual(comp_utils.compute_relative_unit_translation_angle(i2Ui1_optimized, i2Ei1.direction()), 1)
```
johnwlambert, line 63: i think this would be simpler as 
```python
translation_deviation = comp_utils.compute_relative_unit_translation_angle(i2Ui1_optimized, i2Ei1.direction())
self.assertLessEqual(translation_deviation, 1)
```

tests/test_two_view_estimator.py
```diff
@@ -0,0 +1,71 @@
...
+            pA = track.measurement(0)[1]
+            pB = track.measurement(1)[1]
+
+            normalized_coordinates_i1.append(pA)
+            normalized_coordinates_i2.append(pB)
+
+        normalized_coordinates_i1 = np.array(normalized_coordinates_i1)
+        normalized_coordinates_i2 = np.array(normalized_coordinates_i2)
+
+        print(normalized_coordinates_i1.shape)
```
johnwlambert, line 50: print statement likely can be removed

gtsfm/two_view_estimator.py
```diff
@@ -41,6 +53,46 @@ def __init__(self, matcher: MatcherBase, verifier: VerifierBase, corr_metric_dis
...
+    def bundle_adjust(
+        cls,
+        keypoints_i1: Keypoints,
+        keypoints_i2: Keypoints,
+        verified_corr_idxes: np.ndarray,
+        camera_intrinsics_i1: Cal3Bundler,
+        camera_intrinsics_i2: Cal3Bundler,
+        i2Ri1_initial: Optional[Rot3],
+        i2Ui1_initial: Optional[Unit3],
+    ) -> Tuple[Optional[Rot3], Optional[Unit3]]:
```
johnwlambert, line 66: missing docstring, but structure looks quite good

gtsfm/two_view_estimator.py
```diff
@@ -41,6 +53,46 @@ def __init__(self, matcher: MatcherBase, verifier: VerifierBase, corr_metric_dis
...
+        ba_output = BUNDLE_ADJUST_2VIEW.run(ba_input)
+
+        # extract the camera poses
+        wPi1, wPi2 = ba_output.get_camera_poses()
+
+        if wPi1 is None or wPi2 is None:
+            logger.warning("2-view BA failed")
+            return i2Ri1_initial, i2Ui1_initial
+
+        i2Pi1_optimized = wPi2.between(wPi1)
```
johnwlambert, line 92: i prefer `i2Ti1` to `i2Pi1` since I think that's the convention we've generally been using

johnwlambert:
could you check it on Skydio-CraneMast-32 also? Looks like there is a small merge conflict also

gtsfm/two_view_estimator.py
```diff
@@ -25,6 +30,13 @@
 pil_logger = logging.getLogger("PIL")
 pil_logger.setLevel(logging.INFO)
 
+BA_NOISE_MODEL_EPIPOLAR_ALGEBRAIC_ERROR = gtsam.noiseModel.Isotropic.Sigma(1, 0.01)
+
+DATA_ASSOC_2VIEW = DataAssociation(reproj_error_thresh=3, min_track_len=2, mode=TriangulationParam.NO_RANSAC)
+BUNDLE_ADJUST_2VIEW = BundleAdjustmentOptimizer(
+    output_reproj_error_thresh=100
```
johnwlambert, line 37: 100 should be a module-level global var

gtsfm/two_view_estimator.py
```diff
@@ -25,6 +30,13 @@
 pil_logger = logging.getLogger("PIL")
 pil_logger.setLevel(logging.INFO)
 
+BA_NOISE_MODEL_EPIPOLAR_ALGEBRAIC_ERROR = gtsam.noiseModel.Isotropic.Sigma(1, 0.01)
+
+DATA_ASSOC_2VIEW = DataAssociation(reproj_error_thresh=3, min_track_len=2, mode=TriangulationParam.NO_RANSAC)
```
johnwlambert, line 35: 3 should be a named module-level global var

gtsfm/two_view_estimator.py
```diff
@@ -25,6 +30,13 @@
...
+BUNDLE_ADJUST_2VIEW = BundleAdjustmentOptimizer(
+    output_reproj_error_thresh=100
+)  # we dont care about output error threshold
```
travisdriver, line 38: If the reprojection error after BA is high, it's likely that the correspondences between the views are contaminated with outliers. Maybe this view pair should be rejected, or it at least needs further filtering?

johnwlambert:
with loransac front-end, 2 deg baseline thresh:
```

[2021-08-07 16:36:34,818 INFO cycle_consistency.py line 259 1005] Found 20 consistent rel. rotations from 25 original edges.
[2021-08-07 16:36:34,833 INFO scene_optimizer.py line 435 1005] R error (deg) before 2-view BA: 11.86
[2021-08-07 16:36:34,833 INFO scene_optimizer.py line 436 1005] U error (deg) before 2-view BA: 20.29
[2021-08-07 16:36:34,834 INFO scene_optimizer.py line 438 1005] R error (deg) after 2-view BA: 11.55
[2021-08-07 16:36:34,834 INFO scene_optimizer.py line 439 1005] U error (deg) after 2-view BA: 20.16
[2021-08-07 16:36:34,834 DEBUG scene_optimizer.py line 468 1005] [Two view optimizer] [Summary] Rotation success: 19/25/28
[2021-08-07 16:36:34,835 DEBUG scene_optimizer.py line 475 1005] [Two view optimizer] [Summary] Translation success: 20/25/28
[2021-08-07 16:36:34,835 DEBUG scene_optimizer.py line 482 1005] [Two view optimizer] [Summary] Pose success: 18/25/28
```



4 deg baseline thresh (worse than 2 deg thresh)
```
[2021-08-07 16:17:45,093 INFO scene_optimizer.py line 435 679] R error (deg) before 2-view BA: 11.86
[2021-08-07 16:17:45,112 INFO scene_optimizer.py line 436 679] U error (deg) before 2-view BA: 20.29
[2021-08-07 16:17:45,115 INFO scene_optimizer.py line 438 679] R error (deg) after 2-view BA: 11.65
[2021-08-07 16:17:45,134 INFO scene_optimizer.py line 439 679] U error (deg) after 2-view BA: 20.29
[2021-08-07 16:17:45,135 DEBUG scene_optimizer.py line 468 679] [Two view optimizer] [Summary] Rotation success: 19/25/28
[2021-08-07 16:17:45,136 DEBUG scene_optimizer.py line 475 679] [Two view optimizer] [Summary] Translation success: 20/25/28
[2021-08-07 16:17:45,136 DEBUG scene_optimizer.py line 482 679] [Two view optimizer] [Summary] Pose success: 18/25/28
```

no baseline thresh (rot worse, trans better than 2 deg)
```
[2021-08-07 16:23:40,151 INFO scene_optimizer.py line 435 828] R error (deg) before 2-view BA: 11.86
[2021-08-07 16:23:40,151 INFO scene_optimizer.py line 436 828] U error (deg) before 2-view BA: 20.29
[2021-08-07 16:23:40,152 INFO scene_optimizer.py line 438 828] R error (deg) after 2-view BA: 11.59
[2021-08-07 16:23:40,152 INFO scene_optimizer.py line 439 828] U error (deg) after 2-view BA: 20.10
[2021-08-07 16:23:40,152 DEBUG scene_optimizer.py line 468 828] [Two view optimizer] [Summary] Rotation success: 19/25/28
[2021-08-07 16:23:40,153 DEBUG scene_optimizer.py line 475 828] [Two view optimizer] [Summary] Translation success: 20/25/28
[2021-08-07 16:23:40,153 DEBUG scene_optimizer.py line 482 828] [Two view optimizer] [Summary] Pose success: 18/25/28
```

johnwlambert:
2-view BA should likely come only after the degeneracy checks?

gtsfm/two_view_estimator.py
ayushbaid: removed it.

ayushbaid:
Summary of metrics: the sift metrics seem to indicate that there is not much of an impact from this change. A leading hypothesis is the presence of outliers in the 2-view BA input.

Maybe merging #70 will help with outliers?

travisdriver:
> Summary of metrics: the sift metrics seem to indicate that there is not much of an impact from this change. A leading hypothesis is the presence of outliers in the 2-view BA input.
> 
> Maybe merging #70 will help with outliers?

Have you tested this on other datasets other than crane mast 8?

ayushbaid:
# Notre dame deep front end

**Frontend metrics**

Metric | Frontend - Baseline | Frontend - New | Cycle consistent frontend - Baseline | Cycle consistent frontend - New
-- | -- | -- | -- 
Rotation success count | 147 | 156 | 147 | 156 
Translation success count | 105 | 123 | 105 | 123 
Mean rotation error | 3.17295 | 2.55973  | 2.7296  | 2.01607
Mean translation error | 7.15367  | 6.11253  | 6.4117  | 5.14451 

I however see max errors around the same values. Maybe we can use the reprojection error to throw away some pairs?

ayushbaid:
We are seeing improvements in front-end metrics, but not after BA.

ayushbaid:
# Skydio-32 deep front end

**Frontend metrics**

Metric | Frontend - Baseline | Frontend - New | Cycle consistent frontend - Baseline | Cycle consistent frontend - New
-- | -- | -- | -- | -- 
Rotation success count | 289 | 294 | 289 | 294
Translation success count | 243 | 254 | 233 | 247 
Mean rotation error |  13.9309  |  13.4918  |  7.93553  |  7.42545 
Mean translation error |  33.1835   |  32.3973   |  23.8825   |  24.3372

ayushbaid:
# Astronet sift front end

**Frontend metrics**

Metric | Frontend - Baseline | Frontend - New | Cycle consistent frontend - Baseline | Cycle consistent frontend - New
-- | -- | -- | -- | -- 
Rotation success count | 240 | 311 | 236 | 294
Translation success count | 286 | 328 | 278 | 247 
Mean rotation error |   9.54955   |   7.54702   |   5.25442   |   3.21033  
Mean translation error |   21.5324    |   20.319   |   12.0799    |   9.23876 

Improvement in the number of tracks after BA

ayushbaid:
# Palace sift front end

**Frontend metrics**

Metric | Frontend - Baseline | Frontend - New | Cycle consistent frontend - Baseline | Cycle consistent frontend - New
-- | -- | -- | -- | -- 
Rotation success count | 3085 | 3086  | 3076  | 3075 
Translation success count | 369 | 375 | 354 | 361 
Mean rotation error |    8.92478    |    8.64571   |    7.36674    |    7.24579 
Mean translation error |    38.8359     |    38.6161   |    38.2916     |    37.9658 

Slight improvements after BA too

johnwlambert:
That's cool to see it improves the front-end results. It's a bit hard to tell the back-end influence without a side by side comparison of metrics or something. Could you show the before & after of BA metrics on each of the datasets?

If BA isn't improved, I would probably suggest we hold off for merging until we can figure out why.

travisdriver:
@ayushbaid What's the status on this? I can help complete any TODOs

gtsfm/two_view_estimator.py
```diff
@@ -104,17 +169,31 @@ def create_computation_graph(
...
 
+        i2Ri1_graph, i2Ui1_graph = dask.delayed(self.bundle_adjust, nout=2)(
+            keypoints_i1_graph,
+            keypoints_i2_graph,
+            v_corr_idxs_graph,
+            camera_intrinsics_i1_graph,
+            camera_intrinsics_i2_graph,
+            i2Ri1_pre_ba_graph,
+            i2Ui1_pre_ba_graph,
+        )
```
travisdriver, line 195: Should we be filtering the post BA results here?

tests/test_two_view_estimator.py
travisdriver: Comments for each code block would also be useful.

travisdriver:
The `TwoViewEstimator` currently has its own `DataAssociation` and `BundleAdjustmentOptimizer` objects. There are two options:

1. Keep this format and add these as attributes and add to the config, or
2. use the already instantiated `MultiViewOptimizer`.

What do you guys think?

travisdriver:
The addition of 2-view BA results in a **huge** improvement in the results for the Dawn @ 4 Vesta sequences (see #348 and #295 for visualizations of previous results).

# RC3 (deep_front_end,  2011205_rc3, 20,  png,  wget, astronet, 1024, true)
Screenshot from 2021-11-02 11-42-33

# OpNav022 (deep_front_end,  2011212_opnav_022, 15,  png,  wget, astronet, 1024, true)
Screenshot from 2021-11-02 13-34-13

johnwlambert:
@akshay-krishnan for some reason this 1dsfm test is failing:
```
_____________ TestTranslationAveraging1DSFM.test_computation_graph _____________

self = <test_averaging_1dsfm.TestTranslationAveraging1DSFM testMethod=test_computation_graph>

    def test_computation_graph(self):
        """Test the dask computation graph execution using a valid collection of relative unit-translations."""
    
        """Test a simple case with 8 camera poses.
    
        The camera poses are arranged on the circle and point towards the center
        of the circle. The poses of 8 cameras are obtained from SFMdata and the
        unit translations directions between some camera pairs are computed from their global translations.
    
        This test is copied from GTSAM's TranslationAveragingExample.
        """
    
        fx, fy, s, u0, v0 = 50.0, 50.0, 0.0, 50.0, 50.0
        expected_wTi_list = SFMdata.createPoses(Cal3_S2(fx, fy, s, u0, v0))
    
        wRi_list = [x.rotation() for x in expected_wTi_list]
    
        # create relative translation directions between a pose index and the
        # next two poses
        i2Ui1_dict = {}
        for i1 in range(len(expected_wTi_list) - 1):
            for i2 in range(i1 + 1, min(len(expected_wTi_list), i1 + 3)):
                # create relative translations using global R and T.
                i2Ui1_dict[(i1, i2)] = Unit3(expected_wTi_list[i2].between(expected_wTi_list[i1]).translation())
    
        # use the `run` API to get expected results, ignore the metrics
        wti_expected, _ = self.obj.run(len(wRi_list), i2Ui1_dict, wRi_list)
    
        # form computation graph and execute
        i2Ui1_graph = dask.delayed(i2Ui1_dict)
        wRi_graph = dask.delayed(wRi_list)
        computation_graph = self.obj.create_computation_graph(len(wRi_list), i2Ui1_graph, wRi_graph)
        with dask.config.set(scheduler="single-threaded"):
            wti_computed, _ = dask.compute(computation_graph)[0]
        wTi_computed = [Pose3(wRi, wti) for wRi, wti in zip(wRi_list, wti_computed)]
        wTi_expected = [Pose3(wRi, wti) for wRi, wti in zip(wRi_list, wti_expected)]
>       self.assertTrue(
            geometry_comparisons.compare_global_poses(
                wTi_computed, wTi_expected, RELATIVE_ERROR_THRESHOLD, ABSOLUTE_ERROR_THRESHOLD
            )
        )
E       AssertionError: False is not true

tests/averaging/translation/test_averaging_1dsfm.py:143: AssertionError
```
any idea what's going on there?

johnwlambert:
@travisdriver this is what I'm seeing when I run:
```bash
python visualization/view_scene.py
--output_dir /Users/johnlam/Downloads/results-deep_front_end-2011205_rc3-20-png-wget-astronet-1024-true.zip/results/ba_output
--max_range 10000 --frustum_ray_len 1000 
```

Screen Shot 2021-11-02 at 8 17 01 PM

Were you seeing a lot of randomness between runs?

travisdriver:
@johnwlambert Actually I ran it with a 15 frame lookahead. I'll update the comment.

travisdriver:
> @akshay-krishnan for some reason this 1dsfm test is failing:
> 
> ```
> _____________ TestTranslationAveraging1DSFM.test_computation_graph _____________
> 
> self = <test_averaging_1dsfm.TestTranslationAveraging1DSFM testMethod=test_computation_graph>
> 
>     def test_computation_graph(self):
>         """Test the dask computation graph execution using a valid collection of relative unit-translations."""
>     
>         """Test a simple case with 8 camera poses.
>     
>         The camera poses are arranged on the circle and point towards the center
>         of the circle. The poses of 8 cameras are obtained from SFMdata and the
>         unit translations directions between some camera pairs are computed from their global translations.
>     
>         This test is copied from GTSAM's TranslationAveragingExample.
>         """
>     
>         fx, fy, s, u0, v0 = 50.0, 50.0, 0.0, 50.0, 50.0
>         expected_wTi_list = SFMdata.createPoses(Cal3_S2(fx, fy, s, u0, v0))
>     
>         wRi_list = [x.rotation() for x in expected_wTi_list]
>     
>         # create relative translation directions between a pose index and the
>         # next two poses
>         i2Ui1_dict = {}
>         for i1 in range(len(expected_wTi_list) - 1):
>             for i2 in range(i1 + 1, min(len(expected_wTi_list), i1 + 3)):
>                 # create relative translations using global R and T.
>                 i2Ui1_dict[(i1, i2)] = Unit3(expected_wTi_list[i2].between(expected_wTi_list[i1]).translation())
>     
>         # use the `run` API to get expected results, ignore the metrics
>         wti_expected, _ = self.obj.run(len(wRi_list), i2Ui1_dict, wRi_list)
>     
>         # form computation graph and execute
>         i2Ui1_graph = dask.delayed(i2Ui1_dict)
>         wRi_graph = dask.delayed(wRi_list)
>         computation_graph = self.obj.create_computation_graph(len(wRi_list), i2Ui1_graph, wRi_graph)
>         with dask.config.set(scheduler="single-threaded"):
>             wti_computed, _ = dask.compute(computation_graph)[0]
>         wTi_computed = [Pose3(wRi, wti) for wRi, wti in zip(wRi_list, wti_computed)]
>         wTi_expected = [Pose3(wRi, wti) for wRi, wti in zip(wRi_list, wti_expected)]
> >       self.assertTrue(
>             geometry_comparisons.compare_global_poses(
>                 wTi_computed, wTi_expected, RELATIVE_ERROR_THRESHOLD, ABSOLUTE_ERROR_THRESHOLD
>             )
>         )
> E       AssertionError: False is not true
> 
> tests/averaging/translation/test_averaging_1dsfm.py:143: AssertionError
> ```
> 
> any idea what's going on there?

It was failing on the CI for some other PRs. It's passing locally for me.

johnwlambert:
@travisdriver can you also add a result showing how much longer GTSFM takes with this change? Is it 2x longer, or 10x longer, etc?

travisdriver:
> @travisdriver can you also add a result showing how much longer GTSFM takes with this change? Is it 2x longer, or 10x longer, etc?

Benchmarking _with_ 2-view BA took 6h 6m 16s, and 1h 16m 57s _without_ (the times for the individual benchmarks can be seen in the CI). So, maybe this should be turned off by default? Although the performance increase is worth the extra computations in my opinion.

dellaert:
This is great, let’s discuss during our meeting

travisdriver:
Ideas to make things faster:
- Use mass triangulation instead of triangulating tracks individually
- Tune optimization parameters

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -171,16 +178,20 @@ def run(
             track = initial_data.get_track(j)
             initial_values.insert(P(j), track.point3())
 
+        # Configure optimizer.
+        params = gtsam.LevenbergMarquardtParams()
+        if self._max_iterations:
+            params.setMaxIterations(self._max_iterations)
+        params.setVerbosityLM("ERROR")
```
ayushbaid, line 185: Thanks Travis for this.

tests/test_two_view_estimator.py
```diff
@@ -18,52 +18,49 @@
...
+        # Compute pose errors.
+        rotation_angular_error = comp_utils.compute_relative_rotation_angle(i2Ri1_optimized, i2Ei1.rotation())
+        translation_angular_error = comp_utils.compute_relative_unit_translation_angle(
             i2Ui1_optimized, i2Ei1.direction()
         )
-
-        self.assertLessEqual(rotation_angular_deviation, 1)
-        self.assertLessEqual(translation_angular_deviation, 1)
+        self.assertLessEqual(rotation_angular_error, 1)
+        self.assertLessEqual(translation_angular_error, 1)
```
ayushbaid: I think we should replace this with a mock test?

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Any change in performance?

ayushbaid:
Re: the impact on runtime: With BA capped at 10 iters, the majority of the runtime is coming from the data association step.

ayushbaid:
Even on door-12 dataset, the improvements with bundle-adjustment after verification are clear:

## Verifier Summary Pre Ba 2view Report
image

## Verifier Summary Post Ba 2view Report
image

ayushbaid:
Opened GTSAM issue https://github.com/borglab/gtsam/issues/919 to request for a batch triangulation API.

gtsfm/bundle/bundle_adjustment.py
travisdriver: None that I saw.

gtsfm/scene_optimizer.py
```diff
@@ -160,8 +164,8 @@ def create_computation_graph(
             i2Ri1_graph_dict[(i1, i2)] = i2Ri1
             i2Ui1_graph_dict[(i1, i2)] = i2Ui1
             v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
-            two_view_reports_dict[(i1, i2)] = two_view_report
-            two_view_reports_pp_dict[(i1, i2)] = two_view_report_pp
+            for token, report in two_view_reports.items():
+                two_view_reports_dict[token][(i1, i2)] = report
```
travisdriver, line 168: I do kind of like this "token" idea, but I think it would be cleaner if it was _inside_ the `TwoViewEstimationReport` (or maybe another class that manages all of the reports). Maybe this is ok for as a temporary solution.

gtsfm/scene_optimizer.py
```diff
@@ -400,6 +388,8 @@ def save_full_frontend_metrics(
     """
     metrics_list = []
 
+    round_and_save_fn = lambda x: round(x, PRINT_NUM_SIG_FIGS) if x else None
```
travisdriver, line 391: Maybe just `round_fn`? This function doesn't explicitly save anything.

gtsfm/two_view_estimator.py
```diff
@@ -28,7 +33,20 @@
...
+
+
+DATA_ASSOC_REPROJ_ERROR_THRESH = 3
+BA_OUTPUT_REPROJ_ERROR_THRESH = 100
+DATA_ASSOC_2VIEW = DataAssociation(
+    reproj_error_thresh=DATA_ASSOC_REPROJ_ERROR_THRESH, min_track_len=2, mode=TriangulationParam.NO_RANSAC
+)
+BUNDLE_ADJUST_2VIEW = BundleAdjustmentOptimizer(
+    output_reproj_error_thresh=BA_OUTPUT_REPROJ_ERROR_THRESH, robust_measurement_noise=True, max_iterations=10
+)  # we dont care about output error threshold as we do not access the tracks of 2-view BA, which the threshold affects.
```
travisdriver, line 49: Should we ass these to the config?

johnwlambert: yeah, line 42 and 43 should be loaded from the config

gtsfm/scene_optimizer.py
johnwlambert: agreed

gtsfm/two_view_estimator.py
ayushbaid: 43 is inconsequential. I am basically ignoring the optimized points for now. 

Having 3 in a config is not so straightforward. I think once we have a batch triangulation API, we will just use that instead of using the data association module. So i think we can avoid this effort for now.

gtsfm/frontend/inlier_support_processor.py
```diff
@@ -75,7 +78,8 @@ def run(
             i2Ri1 = None
             i2Ui1 = None
             v_corr_idxs = np.array([], dtype=np.uint64)
-            return i2Ri1, i2Ui1, v_corr_idxs, None
+
+            return i2Ri1, i2Ui1, v_corr_idxs, TwoViewEstimationReport(v_corr_idxs=v_corr_idxs, num_inliers_est_model=0)
```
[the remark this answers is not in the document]
ayushbaid: I do not get your point. Can you elaborate.

I saw these errors: 
```
File "/home/runner/work/gtsfm/gtsfm/gtsfm/averaging/rotation/cycle_consistency.py", line 312, in _compute_metrics
    if report.R_error_deg is None or report.U_error_deg is None:
AttributeError: 'NoneType' object has no attribute 'R_error_deg'
```
but maybe that is because i was returning a report with null values instead of returning None itself.

gtsfm/two_view_estimator.py
travisdriver: I think we should at least be able to turn 2-view BA on or off from the config, especially due to the greatly inflated runtimes.

gtsfm/frontend/inlier_support_processor.py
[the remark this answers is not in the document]
johnwlambert: So we check for None here: 
https://github.com/borglab/gtsfm/blob/master/gtsfm/two_view_estimator.py#L256

So we need some signal that we are discarding. Otherwise the statistics will get polluted with zeros for pairs we are not actually using.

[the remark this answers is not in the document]
travisdriver: Why are we setting `num_inliers_est_model` to 0 instead of `num_inliers_est_model`? Even though it doesn't have _enough_ outliers with respect to our threshold, it's kind of misleading to say it didn't find any.

gtsfm/evaluation/metrics.py
```diff
@@ -91,7 +91,7 @@ def __init__(
         if data is not None:
             # Cast to a numpy array
             if not isinstance(data, np.ndarray):
-                data = np.array(data)
+                data = np.array(data, dtype=np.float32)
```
ayushbaid, line 94: this is to cast `None` to `np.Nan`

gtsfm/frontend/inlier_support_processor.py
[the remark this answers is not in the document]
ayushbaid: Ok I will revert to None.

gtsfm/scene_optimizer.py
ayushbaid: Yeah in the long term I think we want to restructure `TwoViewReport`, so we can handle it there

gtsfm/utils/metrics.py
```diff
@@ -422,3 +422,19 @@ def save_metrics_as_json(metrics_groups: Delayed, output_dir: str) -> None:
     """
     for metrics_group in metrics_groups:
         metrics_group.save_to_json(os.path.join(output_dir, metrics_group.name + ".json"))
+
+
+def get_stats_for_sfmdata(gtsfm_data: GtsfmData, suffix: str) -> List[GtsfmMetric]:
```
ayushbaid, line 427: moved this function from scene_optimizer

gtsfm/two_view_estimator.py
ayushbaid: We dont have that much config anymore. Let me know if you still think we need the configs.

ayushbaid:
The triangulation steps in both data association and bundle adjustment modules consumed a lot of compute. I am skipping calling the triangulation (for reprojection error) now from two-view plate. DA is super fast now,  and bundle adjustment is the major contributor of compute. We do not need the batch triangulation API.

gtsfm/utils/metrics.py
travisdriver: Is it only used in this file?

gtsfm/two_view_estimator.py
```diff
@@ -91,12 +220,11 @@ def create_computation_graph(
             Computed relative rotation wrapped as Delayed.
             Computed relative translation direction wrapped as Delayed.
             Indices of verified correspondences wrapped as Delayed.
-            Two view report w/ verifier metrics wrapped as Delayed.
-            Two view report w/ post-processor metrics wrapped as Delayed.
+            Two-view reports at different stages, as a dictionary.
```
travisdriver, line 223: Maybe elaborate on what the different stages are.

gtsfm/two_view_estimator.py
```diff
@@ -55,6 +76,114 @@ def __init__(
...
+        # Perform data association to construct 2-view BA input.
+        start_time = timeit.default_timer()
+        triangulated_tracks: List[SfmTrack] = cls.triangulate_two_view_correspondences(
+            camera_i1=camera_i1,
+            camera_i2=camera_i2,
+            keypoints_i1=keypoints_i1,
+            keypoints_i2=keypoints_i2,
+            corr_idxs=verified_corr_idxs,
+        )
+        logger.info("Performed DA in %.6f seconds.", timeit.default_timer() - start_time)
```
travisdriver, line 168: Should remove this before merging.

travisdriver:
> The triangulation steps in both data association and bundle adjustment modules consumed a lot of compute. I am skipping calling the triangulation (for reprojection error) now from two-view plate. DA is super fast now, and bundle adjustment is the major contributor of compute. We do not need the batch triangulation API.

Can you elaborate a little more? What do you mean you're skipping triangulation?

gtsfm/two_view_estimator.py
travisdriver: I still think we should be able to turn two-view BA on or off from the config at least, and also maybe change the max iterations.

dellaert:
> > The triangulation steps in both data association and bundle adjustment modules consumed a lot of compute. I am skipping calling the triangulation (for reprojection error) now from two-view plate. DA is super fast now, and bundle adjustment is the major contributor of compute. We do not need the batch triangulation API.
> 
> Can you elaborate a little more? What do you mean you're skipping triangulation?

I'm curious, too. This is not a decision to be made in isolation, but should be a group decision based on metrics across different datasets. And, our plan was to resolve the computational issue with batch triangulation, as I remember it.

gtsfm/two_view_estimator.py
dellaert: Every blue box should have a a config, right? That are sections in some main config file?

ayushbaid:
> > > The triangulation steps in both data association and bundle adjustment modules consumed a lot of compute. I am skipping calling the triangulation (for reprojection error) now from two-view plate. DA is super fast now, and bundle adjustment is the major contributor of compute. We do not need the batch triangulation API.
> > 
> > Can you elaborate a little more? What do you mean you're skipping triangulation?
> 
> I'm curious, too. This is not a decision to be made in isolation, but should be a group decision based on metrics across different datasets. And, our plan was to resolve the computational issue with batch triangulation, as I remember it.

Sorry @dellaert and @travisdriver. I have removed reprojections which we were using for metrics/statistics. Not triangulation. Triangulation with the current API using a python loop and a batch triangulation API did not produce a significant difference.

dellaert:
Ah, ok. Could we still land that GTSAM PR and use it, though? It is much cleaner than doing a loop in python and is generally useful.

travisdriver:
I was doing some testing and it looks like 10 steps may not be enough for certain datasets. Therefore, I think it's definitely worth exposing the two-view BA module to the config.

### 2011205_rc3 w/ `max_interations=10`
Screenshot from 2021-11-16 13-31-47

### 2011205_rc3 w/ `max_interations=20`
Screenshot from 2021-11-16 13-19-00

travisdriver:
@ayushbaid What's the status of this?

gtsfm/utils/metrics.py
ayushbaid: Yep. But I am moving it because it is not a core scene-optimizer functionality (which is to call different modules)

gtsfm/two_view_estimator.py
ayushbaid: I prefer leaving it at debug level.

gtsfm/configs/deep_front_end.yaml
```diff
@@ -14,6 +14,7 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
+    bundle_adjust_2view: True
```
travisdriver, line 17: Is it possible to also expose the number of BA iterations?

ayushbaid: I have removed the number of iterations as not changing the value does not lead to a significant time difference.

travisdriver: Do you mean to say that reducing the number of iterations did not result in a significant time difference? If so, could you support this with some runtime values? 

From what I remember, reducing the number of iterations noticeably reduced runtimes.

ayushbaid: This is from the door dataset on my local machine.

2-view BA with no max iters:
- Average: 1.72
- Median: 1.24


2-view BA with 20 max iters:
- Average: 1.20
- Median: 1.14

From the CIs
- [Without max iters](https://github.com/borglab/gtsfm/actions/runs/1511043825): 2 hrs 27 mins
- [With 20 max iters](https://github.com/borglab/gtsfm/actions/runs/1510032128): 2 hours 10 mins

I feel the difference is not prohibitive to have one more hyperparam.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -191,34 +210,30 @@ def run(
...
+        ba_metrics = GtsfmMetricsGroup(
+            name=METRICS_GROUP, metrics=metrics_utils.get_stats_for_sfmdata(unfiltered_data, suffix="_unfiltered")
+        )
 
         if cameras_gt is not None:
             poses_gt = [cam.pose() for cam in cameras_gt]
 
             # align the sparse multi-view estimate after BA to the ground truth pose graph.
-            filtered_result = filtered_result.align_via_Sim3_to_poses(wTi_list_ref=poses_gt)
+            filtered_result = filtered_data.align_via_Sim3_to_poses(wTi_list_ref=poses_gt)
```
akshay-krishnan, line 236: please consider renaming filtered_result to filtered_data_aligned and use the same in line 251.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -233,15 +248,14 @@ def get_metrics_from_sfm_data(sfm_data: GtsfmData, suffix: str) -> List[GtsfmMet
                 metric_name = "Filtered tracks triangulated with GT cams: {}".format(exit_code.name)
                 ba_metrics.add_metric(GtsfmMetric(name=metric_name, data=count))
 
-        ba_metrics.add_metrics(get_metrics_from_sfm_data(filtered_result, suffix="_filtered"))
+        ba_metrics.add_metrics(metrics_utils.get_stats_for_sfmdata(filtered_data, suffix="_filtered"))
```
akshay-krishnan, line 251: use filtered_data_aligned? even if it does not make a difference...

gtsfm/scene_optimizer.py
```diff
@@ -172,29 +176,11 @@ def create_computation_graph(
...
                         keypoints_graph_list[i2],
                         v_corr_idxs,
-                        two_view_report=two_view_report,
+                        two_view_report=two_view_reports[PRE_BA_REPORT_TAG],
                         file_path=os.path.join(PLOT_CORRESPONDENCE_PATH, f"{i1}_{i2}.jpg"),
                     )
                 )
 
-        # Persist all front-end metrics and their summaries.
-        auxiliary_graph_list.append(
```
akshay-krishnan, line 181: why is this removed?

gtsfm/two_view_estimator.py
```diff
@@ -47,14 +68,124 @@ def __init__(
...
                 (not during estimation).
         """
         self._matcher = matcher
         self._verifier = verifier
         self.processor = inlier_support_processor
+        self._bundle_adjust_2view = bundle_adjust_2view
         self._corr_metric_dist_threshold = eval_threshold_px
 
+    @classmethod
+    def triangulate_two_view_correspondences(
```
akshay-krishnan, line 82: I think it might be more useful to have a triangulation util with variable number of cameras, can be used in DA as well.

gtsfm/two_view_estimator.py
```diff
@@ -136,27 +301,46 @@ def create_computation_graph(
...
+            post_isp_i2Ui1,
+            post_isp_v_corr_idxs,
+            post_isp_report,
+        ) = self.processor.create_computation_graph(post_ba_i2Ri1, post_ba_i2Ui1, post_ba_v_corr_idxs, post_ba_report)
+
+        return (
+            post_isp_i2Ri1,
+            post_isp_i2Ui1,
+            post_isp_v_corr_idxs,
+            {
```
akshay-krishnan, line 338: nit create this dict outside and write a one-liner return?

akshay-krishnan:
regarding my comments above, I realized I had some comments long ago that I never "submitted", they may be outdated now, in which case feel free to ignore.

gtsfm/configs/deep_front_end.yaml
ayushbaid: Average statistics can be skewed by some outliers, which can be caused by some random fluctuations. The median on door-12 on my machine is just off by 8%. The CI run is off by 13%, and that includes a pretty big dataset.

I agree that we will save some compute by having a cap on the number of iterations, but I have two concerns about having a cap:
1. When will we use it? In the CI we would want to achieve the best performance for now. In the future we might revisit this problem when we have a dataset which is huge relative to skydio-501. For the local runs, I think we will either use BA or not use BA. I don't see a reason when we would want to fiddle with the hyperparam to achieve ~10% 2-view BA compute savings.
2. What is the best cap on the number of iterations?


@dellaert @johnwlambert @akshay-krishnan for your opinions too.

gtsfm/two_view_estimator.py
```diff
@@ -28,7 +33,20 @@
...
+POST_CYCLE_CONSISTENT_REPORT_TAG = "POST_CYCLE_CONSISTENT_2VIEW_REPORT"
+
+
+DATA_ASSOC_REPROJ_ERROR_THRESH = 3
+BA_OUTPUT_REPROJ_ERROR_THRESH = 100
+DATA_ASSOC_2VIEW = DataAssociation(
+    reproj_error_thresh=DATA_ASSOC_REPROJ_ERROR_THRESH, min_track_len=2, mode=TriangulationParam.NO_RANSAC
+)
+BUNDLE_ADJUST_2VIEW = BundleAdjustmentOptimizer(
+    output_reproj_error_thresh=BA_OUTPUT_REPROJ_ERROR_THRESH, robust_measurement_noise=True, max_iterations=10
```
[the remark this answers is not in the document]
ayushbaid: not a problem anymore.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -207,7 +218,7 @@ def get_metrics_from_sfm_data(sfm_data: GtsfmData, suffix: str) -> List[GtsfmMet
             return metrics
 
         ba_metrics = GtsfmMetricsGroup(
-            name="bundle_adjustment_metrics", metrics=get_metrics_from_sfm_data(optimized_data, suffix="_unfiltered")
+            name=METRICS_GROUP, metrics=get_metrics_from_sfm_data(optimized_data, suffix="_unfiltered")
```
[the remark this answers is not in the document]
ayushbaid: not applicable anymore.

gtsfm/two_view_estimator.py
```diff
@@ -55,6 +73,63 @@ def __init__(
         self.processor = inlier_support_processor
         self._corr_metric_dist_threshold = eval_threshold_px
 
+    @classmethod
+    def bundle_adjust(
```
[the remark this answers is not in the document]
ayushbaid: It does not use any argument from a particular instantiation and hence I feel `@classmethod` is reasonable. Any other suggestions?

gtsfm/scene_optimizer.py
ayushbaid: Its done later for all the different reports.

gtsfm/two_view_estimator.py
ayushbaid: I think having a GTSAM API as Frank suggested is the best way forward. In that case we don't need a util function right now.

gtsfm/two_view_estimator.py
travisdriver: @ayushbaid

travisdriver (approved):
A couple of small outstanding comments, but everything else looks good to me!

@ayushbaid Would you mind posting the full dashboard file here for future reference?

ayushbaid:
Results w.r.t master:

After rotation cycle consistency check:
cycle_consistent


After averaging:
averaging

Final:
ba

akshay-krishnan (approved):

dellaert:
Yay!

gtsfm/frontend/inlier_support_processor.py
```diff
@@ -75,7 +78,8 @@ def run(
             i2Ri1 = None
             i2Ui1 = None
             v_corr_idxs = np.array([], dtype=np.uint64)
-            return i2Ri1, i2Ui1, v_corr_idxs, None
+
```
johnwlambert, line 81: @ayushbaid @travisdriver @akshay-krishnan This computation here will give us the wrong numbers about inlier stats. If for 3 pairs, we had # inliers as 20, 20, 5, and if we threw away the 3rd, then we should have avg # inliers = 20.

But here it would be return as 15, which isn't correct.

PR #230, opened 2021-08-13
Adding benchmarks in CI

ayushbaid:
The raw metrics are uploaded as a zip file under the "Benchmark GTSFM on select datasets" action. Akshay's HTML reports will be added to this zip too.

This generated artifacts still require a manual diff for comparison, and automating that step is out-of-scope for this PR.


.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: conda-incubator/setup-miniconda@v2
+        with:
+          miniconda-version: "latest"
+          activate-environment: gtsfm-v1
+          environment-file: environment_linux.yml
+          python-version: 3.8
+      - name: Environment setup
+        run: |
+          bash .github/scripts/setup.sh
+          pip install hydra-core==1.1.0
```
johnwlambert, line 25: Thanks for the PR, @ayushbaid. Why do we need the separate pip install here? it should be in the conda env, right?

.github/scripts/setup.sh
```diff
@@ -0,0 +1,28 @@
...
+
+##########################################################
+# GTSFM dependencies (including GTSAM) previously installed using conda
+##########################################################
+
+echo "Running .github/scripts/setup.sh..."
+conda init
+conda info --envs
+
+wget -O 2021_03_12_gtsam_python38_wheel.zip --no-check-certificate "https://drive.google.com/uc?export=download&id=1JvsLYngMzSqvmD6RwEgEb_HXXHfw_EJ2"
```
johnwlambert, line 11: this seems redundant with the code in `.github/scripts/python.sh`. Can we put it in just one place?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,35 @@
+name: Benchmark GTSFM on select datasets
+on: [pull_request]
+jobs:
+  door-dataset-sift:
+    name: Run GTSFM on the door dataset using SIFT front-end
```
johnwlambert, line 5: looks good. can we run the skydio-8 here also, and skydio-32? i guess we can put those in G Drive or upload as files to a separate Github location (not checked into this repo, though)

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+      - name: Environment setup
+        run: |
+          bash .github/scripts/setup.sh
+          conda info
+      - name: Download Skydio-crane-8 dataset
+        run: |
+          wget --save-cookies cookies.txt 'https://docs.google.com/uc?export=download&id=1mmM1p_NpL7-pnf3iHWeWVKpsm1pcBoD5' -O- \
+            | sed -rn 's/.*confirm=([0-9A-Za-z_]+).*/\1/p' > confirm.txt
+          wget --load-cookies cookies.txt -O skydio-8.zip \
+            'https://docs.google.com/uc?export=download&id=1mmM1p_NpL7-pnf3iHWeWVKpsm1pcBoD5&confirm='$(<confirm.txt)
```
johnwlambert, line 41: Very nice, thanks for adding these Ayush. Can you make the URLs stand-alone bash variables, like SKYDIO8_GDRIVE_URL=“blah”?

johnwlambert: Should we define a matrix and then use a for loop over the URLs?

.github/workflows/test-python.yml
```diff
@@ -20,12 +23,20 @@ jobs:
...
-    - name: Python tests and checks
-      shell: bash -l {0}
+    - name: Environment setup
       run: |
+        bash .github/scripts/setup.sh
         pip install pytest-cov
-        conda info --envs
-        bash .github/scripts/python.sh
+        pip install flake8
+        pip install hydra-core==1.1.0
```
johnwlambert, line 31: should already be in conda here, as well, right?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+          python-version: 3.8
+          use-only-tar-bz2: true # IMPORTANT: This needs to be set for caching to work properly!
+      - name: Environment setup
+        run: |
+          bash .github/scripts/setup.sh
+          conda info
+      - name: Download Skydio-crane-8 dataset
+        run: |
+          wget --save-cookies cookies.txt 'https://docs.google.com/uc?export=download&id=1mmM1p_NpL7-pnf3iHWeWVKpsm1pcBoD5' -O- \
+            | sed -rn 's/.*confirm=([0-9A-Za-z_]+).*/\1/p' > confirm.txt
```
johnwlambert, line 39: can we add a comment to explain what this does : - ) ?

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+        uses: TonyBogdanov/zip@1.0
+        with:
+          args: unzip -qq skydio-8.zip 
+      - name: Decompress Skydio-crane-32 dataset
+        uses: TonyBogdanov/zip@1.0
+        with:
+          args: unzip -qq skydio-32.zip -d skydio-32
+      - name: Execute door dataset
+        run: |
+          python gtsfm/runner/run_scene_optimizer.py --config_name sift_front_end.yaml
```
johnwlambert, line 58: maybe we can explicitly pass the door dataset path here via CLI for clarity

gtsfm/utils/metrics.py
```diff
@@ -128,7 +128,7 @@ def compute_translation_angle_metric(
     for (i1, i2) in i2Ui1_dict:
         i2Ui1 = i2Ui1_dict[(i1, i2)]
         angles.append(comp_utils.compute_translation_to_direction_angle(i2Ui1, wTi_list[i2], wTi_list[i1]))
-    return GtsfmMetric("translation_angle_deg", angles)
+    return GtsfmMetric("translation_angle_deg", np.array(angles, dtype=np.float))
```
johnwlambert, line 131: why do we need the cast here? GtsfmMetric should allow accepting List[float]

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,90 @@
...
+          wget --save-cookies cookies.txt 'https://docs.google.com/uc?export=download&id=1mmM1p_NpL7-pnf3iHWeWVKpsm1pcBoD5' -O- \
+            | sed -rn 's/.*confirm=([0-9A-Za-z_]+).*/\1/p' > confirm.txt
+          wget --load-cookies cookies.txt -O skydio-8.zip \
+            'https://docs.google.com/uc?export=download&id=1mmM1p_NpL7-pnf3iHWeWVKpsm1pcBoD5&confirm='$(<confirm.txt)
+      - name: Download Skydio-crane-32 dataset
+        run: |
+          wget --save-cookies cookies.txt 'https://docs.google.com/uc?export=download&id=1BQ6jp0DD3D9yhTnrDoEddzlMYT0RRH68' -O- \
+            | sed -rn 's/.*confirm=([0-9A-Za-z_]+).*/\1/p' > confirm.txt
+          wget --load-cookies cookies.txt -O skydio-32.zip \
+            'https://docs.google.com/uc?export=download&id=1BQ6jp0DD3D9yhTnrDoEddzlMYT0RRH68&confirm='$(<confirm.txt)
```
johnwlambert, line 47: prefer named bash variable for URL here also

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,90 @@
+name: Benchmark GTSFM on select datasets
+on: [pull_request]
+jobs:
+  door-dataset-sift:
+    name: Run GTSFM on the select dataset using SIFT front-end
```
johnwlambert, line 5: typo "on selected datasets"

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,90 @@
+name: Benchmark GTSFM on select datasets
+on: [pull_request]
+jobs:
+  door-dataset-sift:
```
johnwlambert, line 4: needs new name since not only door dataset anymore?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -73,18 +71,17 @@ def run(
...
                 )
+                num_valid_measurements += 1
 
         # sample indices to be used as projection directions
-        num_measurements = len(i2Ui1_dict)
         indices = np.random.choice(
-            num_measurements,
-            min(self._max_1dsfm_projection_directions, num_measurements),
-            replace=False,
+            num_valid_measurements, min(self._max_1dsfm_projection_directions, num_valid_measurements), replace=False,
```
johnwlambert, line 84: i think a cleaner fix is
```python
        # sample indices to be used as projection directions
        # note: if some unit translations are None, then we can only use valid values
        num_measurements = len(w_i2Ui1_measurements)
        indices = np.random.choice(
            num_measurements,
            size=min(self._max_1dsfm_projection_directions, num_measurements),
            replace=False,
        )
```

johnwlambert: do you know why the unit translations end up being None here? does MFAS mark them as outliers or something?

.github/workflows/benchmark.yml
johnwlambert: marking as resolved, thanks for adding this

.github/workflows/benchmark.yml
johnwlambert: https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows#using-a-build-matrix

.github/workflows/benchmark.yml
ayushbaid: Removed it. For some reason it was not able to find hydra and this line made it work. But now we have it working without any issues.

.github/scripts/setup.sh
ayushbaid: This file is now removed.

.github/workflows/benchmark.yml
ayushbaid: I want the file names to be linked to the file ids. Cannot find a way to add them to a matrix.

.github/workflows/benchmark.yml
ayushbaid: This cookie is for the prompt about the virus scan not being performed. I'll add a comment

gtsfm/utils/metrics.py
ayushbaid: For the nan values. Its actually List[Optional[float]].

.github/workflows/benchmark.yml
ayushbaid: I am using select as an adjective :P

.github/workflows/benchmark.yml
ayushbaid: changed

gtsfm/averaging/translation/averaging_1dsfm.py
ayushbaid: changed the logic. I do not think we need the comments though.

No its because rotation averaging might return some None's. Or the input relative directions might itself be None.

ayushbaid:
Merging this PR with the failing unit tests, as this PR is essential to merge other PRs. Will follow up with another review to fix the unit tests

.github/workflows/test-python.yml
```diff
@@ -8,24 +8,43 @@ jobs:
...
 
     env:
       PYTHON_VERSION: 3.8
 
     steps:
     - uses: actions/checkout@v2
+    - name: Cache conda env
+      uses: actions/cache@v2
+      env:
+        # Increase this value to reset cache if environment_linux.yml has not changed
```
johnwlambert, line 23: @ayushbaid should this say "if environment_linux.yml has changed" or "has not changed"?

ayushbaid: If the file has changed, the hash value will change and there will be a cache miss. This comment is to invalidate the cache even when the file is exactly the same.

gtsfm/runner/run_scene_optimizer.py
```diff
@@ -57,24 +56,17 @@ def run_scene_optimizer(args) -> None:
...
-        type=int,
-        default=1,
-        help="Number of threads per each worker",
+        "--threads_per_worker", type=int, default=1, help="Number of threads per each worker",
     )
     parser.add_argument(
         "--config_name",
         type=str,
-        default="deep_front_end.yaml",
+        default="sift_front_end.yaml",
```
johnwlambert, line 67: i think we should revert this one line back to `deep_front_end.yaml`

PR #233, opened 2021-08-16
add LORANSAC to deep_front_end.yaml

johnwlambert:
Uses python wheels for `pycolmap` binary (without compiling COLMAP).

Wheels for Linux have been uploaded to pypi.


gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,146 @@
...
+
+DEFAULT_RANSAC_SUCCESS_PROB = 0.99999
+DEFAULT_RANSAC_MAX_ITERS = 20000
+
+logger = logger_utils.get_logger()
+
+
+class LoRansac(VerifierBase):
+    def __init__(
+        self, use_intrinsics_in_verification: bool, estimation_threshold_px: float, min_allowed_inlier_ratio_est_model: float
```
akshay-krishnan, line 37: rename min_allowed_inlier_ratio_est_model to min_inlier_ratio_est_model?

akshay-krishnan (changes requested):
Have we evaluated this change with the new CI and plots?

gtsfm/frontend/verifier/loransac.py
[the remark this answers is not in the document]
johnwlambert: Thanks for the review, @akshay-krishnan.

So `min_allowed_inlier_ratio_est_model: float` is what we are using in the config and in other classes, so the name would have to match the config to be automatically constructed by hydra. So I think we should keep it as is.

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,146 @@
...
+            camera_intrinsics_i2: intrinsics for image #i2.
+
+        Returns:
+            Estimated rotation i2Ri1, or None if it cannot be estimated.
+            Estimated unit translation i2Ui1, or None if it cannot be estimated.
+            Indices of verified correspondences, of shape (N, 2) with N <= N3. These are subset of match_indices.
+            Inlier ratio of w.r.t. the estimated model, i.e. the #final RANSAC inliers/ #putatives.
+        """
+        if match_indices.shape[0] < self._min_matches:
+            logger.info('[LORANSAC] Not enough correspondences for verification.')
```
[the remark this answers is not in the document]
johnwlambert: I guess I didn't want to make it an exception because this is how we treat it in other classes.

Getting no matches is possible perhaps on a very difficult scene. Implausible, but still possible i think.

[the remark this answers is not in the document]
akshay-krishnan: Indeed, but that should be handled by the caller. As far this class is concerned, it will fail if there are not enough matches. 
Its already failing, and the better way to do that in python is to raise exceptions rather than returning some dummy result. This way the failure has better "visibility". 
I understand that is not how things are done in other classes, but that should not be a reason to improve (at least in future code). Maybe at some point we can change existing code as well.

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,146 @@
...
+        i2Ui1 = result_dict["tvec"]
+        num_inliers = result_dict["num_inliers"]
+
+        inlier_ratio_est_model = num_inliers / match_indices.shape[0]
+        if inlier_ratio_est_model >= self._min_allowed_inlier_ratio_est_model:
+            i2Ri1 = Rot3(Rotation.from_quat([qx,qy,qz,qw]).as_matrix())
+            i2Ui1 = Unit3(i2Ui1)
+            inlier_mask = result_dict["inliers"]
+            v_corr_idxs = match_indices[inlier_mask]
+
```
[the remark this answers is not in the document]
johnwlambert: sure, it's removed now

gtsfm/frontend/verifier/loransac.py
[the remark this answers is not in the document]
akshay-krishnan: Okay, but the _allowed_ is redundant and  we should try to have shorter names. As a convention we could use min_X or max_X for any threshold of X. Maybe another variable to add to the list of things we can have better names for :)

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,146 @@
...
+import pycolmap
+from gtsam import Cal3Bundler, Rot3, Unit3
+from scipy.spatial.transform import Rotation
+
+
+import gtsfm.utils.features as feature_utils
+import gtsfm.utils.logger as logger_utils
+import gtsfm.utils.verification as verification_utils
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.frontend.verifier.verifier_base import VerifierBase, NUM_MATCHES_REQ_E_MATRIX, NUM_MATCHES_REQ_F_MATRIX
```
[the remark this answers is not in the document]
akshay-krishnan: I think conventionally we only import classes directly from modules. Functions, constants etc. are not directly imported. The reader does not have to refer to the import statement when reading to understand the source of the constant, and its more readable. 
Cant find anything on PEP, unfortunately.

gtsfm/frontend/verifier/loransac.py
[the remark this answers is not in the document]
akshay-krishnan: Yes! Or, you could have the if statement in the calling function. 

if match_indices.shape[0] < threshold:
  log error 
else:
  call function. 
  
As far as this function is concerned, it is an error and we should treat it as one.

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,194 @@
...
+            min_inlier_ratio = 0.01
+            min_num_trials = 1000
+            max_num_trials = 100000
+            confidence = 0.9999
+
+        (See https://github.com/mihaidusmanu/pycolmap/blob/master/essential_matrix.cc#L98)
+
+        Args:
+            use_intrinsics_in_verification: Flag to perform keypoint normalization and compute the essential matrix
+                instead of fundamental matrix. This should be preferred when the exact intrinsics are known as opposed
```
akshay-krishnan, line 52: is this indentation added by subblack?

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            result_dict = pycolmap.fundamental_matrix_estimation(
+                uv_i1_, uv_i2_, max_error_px=self._estimation_threshold_px
+            )
+
+        success = result_dict["success"]
+        if not success:
+            matrix_type = "Essential" if self.use_intrinsics_in_verification else "Fundamental"
+            logger.info(f"[LORANSAC] {matrix_type} matrix estimation unsuccessful.")
+            return self._failure_result
```
akshay-krishnan, line 160: how would the caller know if this method returned None because there were not enough matches or because the matches were very bad? It might want to do different things for the two cases. 
this is related to my previous comment about exceptions.

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,194 @@
...
+
+        num_inliers = result_dict["num_inliers"]
+        inlier_ratio_est_model = num_inliers / match_indices.shape[0]
+
+        # no need to extract the relative pose if we have insufficient inliers.
+        if inlier_ratio_est_model < self._min_allowed_inlier_ratio_est_model:
+            i2Ri1 = None
+            i2Ui1 = None
+            v_corr_idxs = np.array([])
```
akshay-krishnan, line 169: why is this empty? maybe I am missing something, but dont we have any inliers? (maybe less but is it actually zero?)

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,194 @@
...
+            return i2Ri1, i2Ui1, v_corr_idxs, inlier_ratio_est_model
+
+        inlier_mask = result_dict["inliers"]
+        v_corr_idxs = match_indices[inlier_mask]
+        if self._use_intrinsics_in_verification:
+            # case where E-matrix was estimated
+            # See https://github.com/colmap/colmap/blob/dev/src/base/pose.h#L72 for quaternion ordering
```
akshay-krishnan, line 176: Im not sure how useful a link to a colmap line is.. what if they change some code? 
It might be better to explain briefly what youre referring to.

akshay-krishnan (changes requested):

gtsfm/frontend/verifier/loransac.py
johnwlambert: no, it was just something i added. but i see indentation here too: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

gtsfm/frontend/verifier/loransac.py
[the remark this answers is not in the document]
johnwlambert: I guess one thing that is maybe different here is that the constants aren't part of a class, though, right? They are just defined in the file

gtsfm/frontend/verifier/loransac.py
johnwlambert: so i guess my thinking here was that if this threshold is not met, then all of the matches are surely garbage. and thus no inliers can be trusted. so, effectively, it's zero

gtsfm/frontend/verifier/loransac.py
johnwlambert: good point. i added a bit more text to make it clearer

gtsfm/frontend/verifier/loransac.py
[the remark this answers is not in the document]
johnwlambert: i suppose in a future PR we could make them part of the VerifierBase class, though

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
+"""
+Wrapper about COLMAP's LORANSAC Essential matrix estimation, using pycolmap's pybind API.
+
+LORANSAC paper:
+ftp://cmp.felk.cvut.cz/pub/cmp/articles/matas/chum-dagm03.pdf
+
+On Linux, a python wheel is available:
+(add URL)
```
ayushbaid, line 8: is it available now?

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
...
+        self._estimation_threshold_px = estimation_threshold_px
+        self._min_allowed_inlier_ratio_est_model = min_allowed_inlier_ratio_est_model
+        self._min_matches = (
+            verifier_base.NUM_MATCHES_REQ_E_MATRIX
+            if self._use_intrinsics_in_verification
+            else verifier_base.NUM_MATCHES_REQ_F_MATRIX
+        )
+
+        # for failure, i2Ri1 = None, and i2Ui1 = None, and no verified correspondences, and inlier_ratio_est_model = 0
+        self._failure_result = (None, None, np.array([], dtype=np.uint64), 0.0)
```
ayushbaid, line 71: global constant?

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
...
+        Args:
+            uv_i1: array of shape (N3,2) representing coordinates of 2d points in image 1.
+            uv_i2: array of shape (N3,2) representing coordinates of 2d points in image 2.
+            camera_intrinsics_i1: intrinsics for image #i1.
+            camera_intrinsics_i2: intrinsics for image #i2.
+
+        Returns:
+            dictionary containing result status code, estimated relative pose (R,t), and inlier mask.
+        """
+        focal_length = camera_intrinsics_i1.fx()
```
ayushbaid, line 91: lets move this to a helper function in the same class. Saves duplication of code.

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
...
+        cx, cy = camera_intrinsics_i1.px(), camera_intrinsics_i1.py()
+
+        # TODO: use more accurate proxy?
+        width = int(cx * 2)
+        height = int(cy * 2)
+
+        camera_dict1 = {
+            "model": "SIMPLE_PINHOLE",
```
ayushbaid, line 99: why do we use simple pinhole camera?

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
...
+        self._min_matches = (
+            verifier_base.NUM_MATCHES_REQ_E_MATRIX
+            if self._use_intrinsics_in_verification
+            else verifier_base.NUM_MATCHES_REQ_F_MATRIX
+        )
+
+        # for failure, i2Ri1 = None, and i2Ui1 = None, and no verified correspondences, and inlier_ratio_est_model = 0
+        self._failure_result = (None, None, np.array([], dtype=np.uint64), 0.0)
+
+    def estimate_essential_matrix(
```
ayushbaid, line 73: private fn?

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
...
+        """
+        if match_indices.shape[0] < self._min_matches:
+            logger.info("[LORANSAC] Not enough correspondences for verification.")
+            return self._failure_result
+
+        uv_i1 = keypoints_i1.coordinates
+        uv_i2 = keypoints_i2.coordinates
+
+        uv_i1_ = uv_i1[match_indices[:, 0]]
+        uv_i2_ = uv_i2[match_indices[:, 1]]
```
ayushbaid, line 150: do it in single step? We will lose some readability but having very similar names would create more confusion?

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,197 @@
...
+        if inlier_ratio_est_model < self._min_allowed_inlier_ratio_est_model:
+            i2Ri1 = None
+            i2Ui1 = None
+            v_corr_idxs = np.array([])
+            return i2Ri1, i2Ui1, v_corr_idxs, inlier_ratio_est_model
+
+        inlier_mask = result_dict["inliers"]
+        v_corr_idxs = match_indices[inlier_mask]
+        if self._use_intrinsics_in_verification:
+            # case where E-matrix was estimated
```
ayushbaid, line 178: this comment is reduntant given the if condition just above it.

gtsfm/frontend/verifier/loransac.py
johnwlambert: good point. that will be cleaner

gtsfm/frontend/verifier/loransac.py
johnwlambert: indeed it is. just updated the file and docstring

gtsfm/frontend/verifier/loransac.py
johnwlambert: good point, updated

gtsfm/frontend/verifier/loransac.py
johnwlambert: agreed, but i thought it added a bit of clarity, so I prefer to keep it

gtsfm/frontend/verifier/loransac.py
johnwlambert: So both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length

johnwlambert: both give identical results in E matrix estimation ( i checked)

gtsfm/frontend/verifier/loransac.py
johnwlambert: i cleaned it up -- let me know if it's clearer now

gtsfm/frontend/verifier/loransac.py
akshay-krishnan: Shouldn't the decision of whether to trust the inliers or not be made outside LORANSAC? I think being functional goes beyond just having non-mutable members. The loransac class has a certain function and its scope should be limited to that. One would expect it to return whatever inliers it found. And that is how pycolmap designed their loransac API as well.

akshay-krishnan: I think the inliers are good inliers with respect to the model estimated. It's possible the model is garbage, but that is for the caller to handle and handling it here would be beyond loransac functionality.

gtsfm/frontend/verifier/loransac.py
akshay-krishnan: Its nice to have, but it's hard to maintain consistency across codebase, so I would not prefer it. Upto you.

ayushbaid (approved):

gtsfm/frontend/verifier/loransac.py
johnwlambert: @akshay-krishnan I see what you mean. LORANSAC should just handle the estimation,and a high-level controller should analyze the results with higher-order logic. However, the tricky thing is that if we can't trust the result bc of low inlier ratios, we want to save time and not convert the F matrix to E matrix, for example.

But the verifier API returns an E matrix. So it isn't super clear how to abstract this away at a higher level, since we would need to go back down to the lower level afterwards.

johnwlambert: @akshay-krishnan gentle reminder about this comment : - )

akshay-krishnan: I am not sure what you mean, what goes wrong if we return the correspondences here? I think that makes more sense API-wise. Like R, t was not estimated because we dont have enough inliers, but hey, if you'd like to know what the correspondences were, here you go.

gtsfm/frontend/verifier/loransac.py
```diff
@@ -0,0 +1,196 @@
...
+
+        Returns:
+            Estimated rotation i2Ri1, or None if it cannot be estimated.
+            Estimated unit translation i2Ui1, or None if it cannot be estimated.
+            Indices of verified correspondences, of shape (N, 2) with N <= N3. These are subset of match_indices.
+            Inlier ratio of w.r.t. the estimated model, i.e. the #final RANSAC inliers/ #putatives.
+        """
+        if match_indices.shape[0] < self._min_matches:
+            logger.info("[LORANSAC] Not enough correspondences for verification.")
+            return self._failure_result
```
akshay-krishnan, line 146: I still think an exception needs to be raised here, because it is incorrect usage.

gtsfm/frontend/verifier/loransac.py
johnwlambert: Gotcha. I guess I see it as wasted compute to estimate R,t in that case, since we already know that the correspondences will be a waste with very high confidence. Maybe a minor thing, but over 10k or 100k image pairs maybe it would start to add up.

akshay-krishnan: Why do we estimate R,t if the inlier ratio is less?

johnwlambert: It's actually the other way around -- currently I'm not estimating R,t if the inlier ratio is less than the required threshold

akshay-krishnan (approved):
My only remaining comments are about 2 things: 
- The way dummy values are returned from the verifier (this makes it harder to track failures). 
- What errors should be ignored by the verifier and what should not. I think it might be better to have the TwoViewEstimator (or wherever verifier is used) handle cases where the verifier cannot work, instead of doing it in the verifier. This might be a larger change than what I thought, so feel free to add a TODO/issue (if you agree that is indeed better).

gtsfm/frontend/verifier/loransac.py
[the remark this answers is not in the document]
johnwlambert: I added this to https://github.com/borglab/gtsfm/issues/283

johnwlambert:
> My only remaining comments are about 2 things:
> 
> * The way dummy values are returned from the verifier (this makes it harder to track failures).
> * What errors should be ignored by the verifier and what should not. I think it might be better to have the TwoViewEstimator (or wherever verifier is used) handle cases where the verifier cannot work, instead of doing it in the verifier. This might be a larger change than what I thought, so feel free to add a TODO/issue (if you agree that is indeed better).

Thanks @akshay-krishnan. I added these comments to https://github.com/borglab/gtsfm/issues/283.

PR #237, opened 2021-08-18
adding benchmarks using deep-front-end

ayushbaid:
Adding new CI to benchmark using the superpoint-superglue based frontend


.github/workflows/benchmark_deep_frontend.yml
```diff
@@ -0,0 +1,96 @@
+name: Benchmark GTSFM on select datasets using deep front-end
+on: [pull_request]
+jobs:
+  benchmark-deep:
+    name: Run GTSFM
```
johnwlambert, line 5: thanks for the PR, @ayushbaid. Could we prevent any copy-paste by looping over the config name in the workflow matrix?

johnwlambert: i think this job is already 3 benchmark jobs right? i would probably prefer no copy-paste to make maintenance easier

ayushbaid:
> @ayushbaid thanks for the updates. Looks very nice. I think a few CI things are failing, though

Yeah my idea of configuring them to run one after the other doesn't work. Let me revert to the simpler version.

johnwlambert (approved):

johnwlambert (approved):

PR #239, opened 2021-08-25
Switching from conda to mamba

ayushbaid:
Switch should allow super-quick setup in the CI jobs.


johnwlambert:
Thanks for the PR, @ayushbaid. Could ou explain a little bit more about the advantages/speedup here, and maybe explain a bit about mamba vs. conda?

ayushbaid:
Do not see any improvements on the CI right now.

PR #240, opened 2021-08-30
Removing numjs dependency to allow "npm install" to function correctly.

adisingh50:
Neha was having trouble building all the dependencies for the React Visualization Tool on her machine. After close investigation into the build failure, I determined that the `numjs` library was causing the issue. Based on this [thread](https://stackoverflow.com/questions/62119181/how-to-install-numjs), it seems that `numjs` is no longer supported for the most recent version of node, and the creator of the package has not fixed the issue yet. 

After this PR, Neha should be able to properly `npm install` the React App on her machine. Some other notes:

- After removing `numjs`, I needed to rebuild the entire dependency tree, which is why `package-lock.json` has been updated (this is responsible for the 24,000+ added lines. But it's nothing significant)
- Needed to remove the `AllFrustums.js` component which was initially responsible for rendering all the camera frustums. It heavily depended on `numjs`, which is currently unusable as explained earlier.


dellaert (approved):
Will trust you on this. Going fwd, @adisingh50 should probably review Neha/Pratyusha PRs in future?

adisingh50:
Yeah, I can definitely review PRs as they continue to make changes to the React Application.

johnwlambert:
Thanks for the PR, @adisingh50. If we want to keep the frustum plotting functionality for now, would an easier fix be to fix the node version to the previous release? and then we can update it to latest when the numjs author fixes things on their end?

johnwlambert:
It's possible that numjs will no longer be maintained (last commit was March 2020) -- I wonder if there are other alternatives we should consider? Which functionality are we using from numjs?

adisingh50:
So to plot the camera frustums, I reimplemented the `SE3` and `ViewFrustum` classes in JS to calculate the pose (in world coordinates) for each camera frustum given all the intrinsic and extrinsic data from `images.txt` and `cameras.txt`. Both of these ported classes heavily relied on `numjs`.

adisingh50:
I will try to revert the nodejs version to v13 (newest version is v16). The comment on [this last commit ](https://github.com/nicolaspanel/numjs/commit/1bd9f59f8eefb78fe33e8eb51964ed03e43187da)back in May 2020 says it should work with node v13.

PR #242, opened 2021-08-31
Remove underconstrained TA square unit test

akshay-krishnan:
This unit test expects translation averaging to recover the translations for 4 points on a circle with only 2 edges for each point.
```
x ----- x
|       |
x ----- x
```
This is an underconstrained problem. Starting with the relative translations, we can only recover a rectangle, not a square. Deprecating this test as there is already another test for a square where each point has 3 edges. 

We could revisit this test if we fix translation averaging to explicitly fail for such degenerate cases someday.


johnwlambert (approved):

johnwlambert:
thank you Akshay

PR #253, opened 2021-09-02
use updated gtsam wheel for linux

johnwlambert:


travisdriver:
I think merging this broke the CI.

johnwlambert:
Hi @travisdriver, the unit test that failed in your CI run is a flaky one, so feel free to ignore it. 

The other CI runs today are passing even with the latest CI (w/ updated GTSAM wheel and master pulled in), so this shouldn't be an issue (https://github.com/borglab/gtsfm/actions)

travisdriver:
Is there a way to make it not flaky? :)

PR #257, opened 2021-09-02
compare previous LM result with new GNC results

johnwlambert:
GNC doesn't seem to converge for the Door dataset currently.


johnwlambert:
Closing for now, since GNC seems to be more appropriate for the averaging stages due to high non-convexity in BA. Will need to expose more verbosity in GTSAM itself to debug.

PR #258, opened 2021-09-02
Astronet loader

travisdriver:


travisdriver:
2011212_opnav_022-gtsfm-scene-02092021

Current results on `2011205_opnav_022` segment of the Dawn mission to 4 Vesta. Plenty of room for improvement :)

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,242 @@
...
+
+from gtsam import Cal3Bundler, Pose3, Rot3, Point3, SfmTrack
+import trimesh
+
+import gtsfm.utils.io as io_utils
+import gtsfm.utils.logger as logger_utils
+from gtsfm.common.image import Image
+from gtsfm.loader.loader_base import LoaderBase
+
+from gtsfm.utils.read_write_model import read_model
```
johnwlambert, line 18: thanks for the PR, Travis. Very cool.

Usually we only import classes from modules, otherwise we just import the module by its name, e.g. `import gtsfm.utils.read_write_model as read_write_model`

Maybe a better name for that module is `mesh_utils` or `model_utils` or `model_io` or `tri_model_io`?

johnwlambert: Oh nevermind, I see that is the original COLMAP name for it? if so, we should put it in the `third_party` folder. What functionality do we need from it that we don't already have in `gtsfm.utils.io`?

tests/loader/test_astronet_loader.py
```diff
@@ -0,0 +1,164 @@
...
+        self.assertTrue(isinstance(fetched_pose, Pose3))
+
+    def test_get_camera_invalid_index(self):
+        """Tests that the camera pose is None, because it is missing on disk."""
+        with self.assertRaises(IndexError):
+            fetched_pose = self.loader.get_camera_pose(25)
+            self.assertIsNone(fetched_pose)
+
+    def test_trimesh_sfmtracks_align(self):
+        """Tests that the ground truth scene mesh aligns with sfmtracks"""
```
johnwlambert, line 132: could you explain a bit more in the docstring about what it means to align a mesh with an SfmTrack?

tests/loader/test_astronet_loader.py
```diff
@@ -0,0 +1,164 @@
...
+        assert self.loader._num_imgs == 4
+        assert len(self.loader._image_paths) == 4
+
+    def test_sfmtracks(self) -> None:
+        """Ensure we have one sfmtrack per mesh vertex"""
+        # there are 4 images and 1465 tracks in 2011212_opnav_022
+        assert self.loader.num_sfmtracks == 1465
+        assert len(self.loader._sfmtracks) == 1465
+
+    def test_get_image_valid_index(self):
```
johnwlambert, line 61: missing return type

tests/loader/test_astronet_loader.py
```diff
@@ -0,0 +1,164 @@
...
+import trimesh
+from gtsam import Pose3, PinholeCameraCal3Bundler
+
+from gtsfm.common.image import Image
+from gtsfm.loader.astronet_loader import AstroNetLoader
+import gtsfm.utils.io as io_utils
+
+TEST_DATA_ROOT = Path(__file__).resolve().parent.parent / "data"
+TEST_POINT3D_IDS = [7, 2560, 3063, 3252, 3424, 3534, 3653, 3786, 3920, 4062, 4420, 4698, 3311, 3963, 325]
+TEST_SFMTRACKS_INDICES = [0, 100, 200, 300, 400, 500, 600, 700, 800, 900, 1000, 1100, 1200, 1300, 1400]
```
johnwlambert, line 19: could you add an inline comment explaining where these come from and how they were selected?

gtsfm/runner/run_scene_optimizer_astronet.py
```diff
@@ -0,0 +1,104 @@
...
+from gtsfm.common.gtsfm_data import GtsfmData
+from gtsfm.loader.astronet_loader import AstroNetLoader
+from gtsfm.scene_optimizer import SceneOptimizer
+
+DATA_ROOT = Path(__file__).resolve().parent.parent.parent / "tests" / "data"
+
+logger = logger_utils.get_logger()
+
+
+def run_scene_optimizer(args) -> None:
```
johnwlambert, line 23: missing type hint `args: argparse.Namespace`

gtsfm/loader/astronet_loader.py
travisdriver: `gtsfm.utils.io` doesn't have any of the binary read/write. I can add it thought if you want.

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,242 @@
...
+             ├── cameras.bin: camera calibrations (see https://colmap.github.io/format.html#cameras-txt)
+             ├── images.bin: 3D poses and 2D tracks (see https://colmap.github.io/format.html#images-txt)
+             └── points3D.bin: 3D tracks (see https://colmap.github.io/format.html#points3d-txt)
+
+
+        Args:
+            data_dir: path to directory containing the COLMAP-formatted data: cameras.bin, images.bin, and points3D.bin
+            gt_scene_mesh_path (optional): path to file of target small body surface mesh.
+                Note: vertex size mismath observed when reading in from OBJ format. Prefer PLY.
+            use_gt_intrinsics: whether to use ground truth intrinsics. If calibration is
```
johnwlambert, line 56: do the astronet files have EXIF?

travisdriver: No they do not.

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,242 @@
...
+            img_fpath = os.path.join(data_dir, "images", img_fname)
+            if not Path(img_fpath).exists():
+                raise FileNotFoundError(f"Could not locate image at {img_fpath}.")
+            self._image_paths.append(img_fpath)
+
+        self._num_imgs = len(self._image_paths)
+        logger.info("AstroNet loader found and loaded %d images and %d tracks.", self._num_imgs, self.num_sfmtracks)
+
+    @staticmethod
+    def colmap2gtsfm(
```
johnwlambert, line 109: this seems like this is a more general purpose function that should go into GtsfmData or something, like 

```python
class GtsfmData:
     @classmethod
     def from_colmap_model(cls, colmap_model_dir: str) -> "GtsfmData":
```
would that make sense?

environment_linux.yml
```diff
@@ -33,12 +33,11 @@ dependencies:
   - cudatoolkit=10.2
   - opencv=4.5.0 # preferring conda-forge's distribution as it provides the latest distribution
   - pytorch
+  - embree
```
johnwlambert, line 36: do we need embree and trimesh for this PR?

akshay-krishnan:
Please provide a PR description. I'll just let John and Ayush review this, as they are familiar with the loaders. :)

johnwlambert:
Is the figure you posted
```
Current results on 2011205_opnav_022 segment of the Dawn mission to 4 Vesta.
```
from the deep front end?

gtsfm/loader/astronet_loader.py
travisdriver: Yeah that's a good idea; I'll do that.

environment_linux.yml
travisdriver: Preferably

> Is the figure you posted
> 
> ```
> Current results on 2011205_opnav_022 segment of the Dawn mission to 4 Vesta.
> ```
> 
> from the deep front end?

Yes it is

travisdriver:
> Is the figure you posted
> 
> ```
> Current results on 2011205_opnav_022 segment of the Dawn mission to 4 Vesta.
> ```
> 
> from the deep front end?

Yes it is

environment_linux.yml
travisdriver: I can comment it out since it's not really being used yet, but could I leave it commented out so I don't have to rewrite it later?

tests/loader/test_astronet_loader.py
travisdriver: Do all methods need a `-> None`?

johnwlambert: thats right

gtsfm/loader/astronet_loader.py
johnwlambert: got it. how are intrinsics estimated if `use_gt_intrinsics` is set to false then?

environment_linux.yml
johnwlambert: so i think we should only check in these dependencies after evaluating the specific use case and tradeoffs with other libraries.

do they have python wheels in conda for all platforms for python 3.8?

gtsfm/loader/astronet_loader.py
johnwlambert: which data is stored in binary format?

johnwlambert:
> > Is the figure you posted
> > ```
> > Current results on 2011205_opnav_022 segment of the Dawn mission to 4 Vesta.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > from the deep front end?
> 
> Yes it is

cool, indeed lots of room for improvement :-)

gtsfm/loader/astronet_loader.py
travisdriver: I guess I could have it require that ground truth intrinsics are provided.

Is the case for the intrinsics being `None` is handled anywhere? The output of `get_intrinsics_from_exif()` is `Optional[Cal3Bundler]`, but it seems the ground truth intrinsics (or a good estimate) are required.

environment_linux.yml
travisdriver: We don't _need_ embree with trimesh, but conducting ray tracing without it is orders of magnitude slower (a couple minutes as opposed to hundredths of a second). 

I've looked into this a lot actually, as it's kind of the backbone of my data generation scheme, and `trimesh` with `embree` are definitely the best for performing ray tracing in python.

`trimesh` by itself is also a really great library on it's own for any tasks involving meshes. It also has a lot of functionality with `o3d` for plotting. 

I can of course forego the ray tracing and ground truth mesh and just use the depth maps if you'd prefer, but it's useful to have the actual ground truth mesh for plotting and evaluation purposes.

Let me know what you think.

dellaert: No great downside, right?

gtsfm/loader/astronet_loader.py
travisdriver: Everything except for the images

environment_linux.yml
johnwlambert: Cool, can we just double check to make sure that wheels are available for both packages in conda for python 3.8 for both mac and linux?

tests/loader/test_astronet_loader.py
travisdriver: Changed this unit test to not use the ground truth mesh (yet) and updated docstring.

tests/loader/test_astronet_loader.py
travisdriver: Addressed.

tests/loader/test_astronet_loader.py
travisdriver: Explanation added.

gtsfm/runner/run_scene_optimizer_astronet.py
travisdriver: Moved arg parsing inside `run_scene_optimizer`.

environment_linux.yml
travisdriver: @dellaert Only upsides :)

However, I've removed all the dependencies on `trimesh` and `embree` as @johnwlambert suggested, as the ground truth mesh isn't being utilized yet in this PR.

dellaert:
So, if it's faster, as Travis says, why would we not depend on trimesh? What's the downside?

travisdriver:
> So, if it's faster, as Travis says, why would we not depend on trimesh? What's the downside?

There's no downside, but the main reason to add trimesh as a dependency is to perform ray tracing for front-end metrics which I have not yet implemented. Therefore, I think it makes sense to delay adding trimesh to a later PR.

However, I can keep it in this PR if you'd like.

gtsfm/loader/astronet_loader.py
travisdriver: Moved this file to `thirdparty.colmap.scripts.python.read_wrtie_model`.

gtsfm/loader/astronet_loader.py
travisdriver: @johnwlambert Let me know how you think this should be handled

gtsfm/loader/astronet_loader.py
travisdriver: @johnwlambert I think this may be better for another PR, as I think the `GtsfmData` calss needs to be restructured slightly for the loaders to be able to utilize it (e.g., storing image names for reading).

travisdriver: Moved this to an issue, as I think it's out of the scope of this PR.

environment_linux.yml
```diff
@@ -38,7 +38,4 @@ dependencies:
   - plotly=4.14.3
   - tabulate
   - pip:
-    - pydegensac
-
-
-
```
ayushbaid, line 44: Unrelated change

gtsfm/configs/deep_front_end.yaml
```diff
@@ -10,7 +10,6 @@ SceneOptimizer:
     detector_descriptor:
       _target_: gtsfm.frontend.detector_descriptor.superpoint.SuperPointDetectorDescriptor
 
-
```
ayushbaid, line 13: unrelated change

environment_linux.yml
travisdriver: Causes a mypy error... but I can add it back

gtsfm/configs/deep_front_end.yaml
travisdriver: Causes a mypy error... but I can add it back

gtsfm/loader/astronet_loader.py
ayushbaid: If we do not have exif, we can approximate the intrinsics from image size? We might have the code somewhere.

I do not think we should make intrinsics as compulsory.

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,232 @@
...
+
+
+        Args:
+            data_dir: path to directory containing the COLMAP-formatted data: cameras.bin, images.bin, and points3D.bin
+            gt_scene_mesh_path (optional): path to file of target small body surface mesh.
+                Note: vertex size mismath observed when reading in from OBJ format. Prefer PLY.
+            use_gt_intrinsics: whether to use ground truth intrinsics. If calibration is
+               not found on disk, then use_gt_intrinsics will be set to false automatically.
+            use_gt_extrinsics (optional): whether to use ground truth extrinsics
+            use_gt_sfmtracks (optional): whether to use ground truth tracks
```
ayushbaid, line 57: why introduce this argument. `LoaderBase` does not have it.

gtsfm/runner/run_scene_optimizer_astronet.py
```diff
@@ -0,0 +1,102 @@
...
+from gtsfm.common.gtsfm_data import GtsfmData
+from gtsfm.loader.astronet_loader import AstroNetLoader
+from gtsfm.scene_optimizer import SceneOptimizer
+
+DATA_ROOT = Path(__file__).resolve().parent.parent.parent / "tests" / "data"
+
+logger = logger_utils.get_logger()
+
+
+def run_scene_optimizer() -> None:
```
ayushbaid, line 23: can we follow the same arrangement of code as other runners?

gtsfm/loader/astronet_loader.py
travisdriver: The main issue would be approximating the focal length, especially for this case as the focal length for camera onboard spacecraft are usually very large (10,000 - 100,000 pixels).

gtsfm/utils/io.py
```diff
@@ -55,7 +55,7 @@ def load_image(img_path: str) -> Image:
         exif_data = parsed_data
 
     img_fname = Path(img_path).name
-    return Image(value_array=np.asarray(original_image), exif_data=exif_data, file_name=img_fname)
+    return Image(value_array=np.asarray(original_image.convert("RGB")), exif_data=exif_data, file_name=img_fname)
```
ayushbaid, line 58: why is this needed? Will this cause any regression elsewhere?

gtsfm/loader/astronet_loader.py
travisdriver: I think it provides a more complete ground truth solution. Also, with respect to issue [#263](https://github.com/borglab/gtsfm/issues/263), `GtsfmData` stores the tracks as well so it would make sense to have this as an option.

gtsfm/runner/run_scene_optimizer_astronet.py
travisdriver: The other runners cause a mypy error:

```
Redefining name 'args' from outer scope (line 90)pylint(redefined-outer-name)
```

gtsfm/utils/io.py
travisdriver: If the source images are grayscale (as is that case for the AstroNet images) it will convert them to RGB. Otherwise, nothing changes.

gtsfm/loader/astronet_loader.py
travisdriver: I don't see that issue with making it required just for this specific loader, as the ground truth intrinsics are always available.

travisdriver: Made calibrations required for now. Opened handling of no intrinsics as an issue [#265](https://github.com/borglab/gtsfm/issues/265).

dask-report.html
```diff
@@ -0,0 +1,88 @@
+
+
+
```
johnwlambert, line 3: thanks for the updates, @travisdriver. Looks like the `dask-report.html` file got checked in. Would you mind deleting it? Maybe we need to add it to the .gitignore?

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,235 @@
...
+    """Simple loader class that reads an AstroNet data segment.
+
+    Refs:
+    - https://github.com/travisdriver/astronet
+    """
+
+    def __init__(
+        self,
+        data_dir: str,
+        use_gt_intrinsics: bool = True,
```
johnwlambert, line 36: I think we aren't handling the case when `use_gt_intrinsics` would be false in the class. Would we expect a silent failure, or what sort of error would be raised?

My sense is that if the AstroNet data will always have intrinsics, we can just note this in the docstring, remove this arg, and throw a `RuntimeError` if the intrinsics files are not found on disk in the constructor.

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,235 @@
...
+        Returns:
+            cameras_gtsfm: list of N camera calibrations corresponding to the N images in images_gtsfm
+            images_gtsfm: list of N camera poses when each image was taken
+            img_fnames: file names of images in images_gtsfm
+            sfmtracks_gtsfm: tracks of points in points3D
+        """
+        cameras_gtsfm, images_gtsfm, img_fnames, sfmtracks_gtsfm = None, None, None, None
+
+        # Note: Assumes input cameras use `PINHOLE` model
+        if len(images) > 0 and len(cameras) > 0:
```
johnwlambert, line 125: if len(images) == 0 or len(cameras) == 0 what is the expected behavior? should we throw an error?

i think we could avoid the nested logic by checking for that and throwing an exception right away

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,235 @@
...
+
+
+        Args:
+            data_dir: path to directory containing the COLMAP-formatted data: cameras.bin, images.bin, and points3D.bin
+            gt_scene_mesh_path (optional): path to file of target small body surface mesh.
+                Note: vertex size mismath observed when reading in from OBJ format. Prefer PLY.
+            use_gt_intrinsics: whether to use ground truth intrinsics. If calibration is
+               not found on disk, then use_gt_intrinsics will be set to false automatically.
+            use_gt_extrinsics (optional): whether to use ground truth extrinsics.
+            use_gt_sfmtracks (optional): whether to use ground truth tracks.
```
johnwlambert, line 57: Hi Travis, could you explain a little more about what the gt SfmTracks will be used for? How could they be accessed in the code if they are not in the BaseLoader? Maybe this is a design decision where we would need to  consider updating the BaseLoader and all loaders.

gtsfm/loader/astronet_loader.py
travisdriver: Yeah sorry, I thought `use_gt_intrinsics` was required by the `BaseLoader`. I'll remove it and throw a runtime error.

gtsfm/loader/astronet_loader.py
travisdriver: Yes, I think throwing a runtime error if any of the Dicts are empty is a good idea.

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,507 @@
...
+                    tvec=tvec,
+                    camera_id=camera_id,
+                    name=image_name,
+                    xys=xys,
+                    point3D_ids=point3D_ids,
+                )
+    return images
+
+
+def read_images_binary(path_to_model_file):
```
johnwlambert, line 215: i am a bit worried checking this 3rd party in since I don't think it conforms to our python coding bar:
(1) no docstrings
(2) no type hints
(3) no explanation of input or return args.

So in https://github.com/borglab/gtsfm/blob/master/gtsfm/utils/io.py#L234, we have:
```python
def read_images_txt(fpath: str) -> Tuple[Optional[List[Pose3]], Optional[List[str]]]:
    """Read camera poses and image file names from a COLMAP-format images.txt file.
    Reference: https://colmap.github.io/format.html#images-txt
        "The coordinates of the projection/camera center are given by -R^t * T, where
        R^t is the inverse/transpose of the 3x3 rotation matrix composed from the
        quaternion and T is the translation vector. The local camera coordinate system
        of an image is defined in a way that the X axis points to the right, the Y axis
        to the bottom, and the Z axis to the front as seen from the image."
    Args:
        fpath: path to images.txt file
    Returns:
        wTi_list: list of camera poses for each image, or None if file path invalid
        img_fnames: name of image file, for each image, or None if file path invalid
    """
    if not Path(fpath).exists():
        logger.info("%s does not exist", fpath)
        return None, None

    with open(fpath, "r") as f:
        lines = f.readlines()

    wTi_list = []
    img_fnames = []
    # ignore first 4 lines of text -- they contain a description of the file format
    # and a record of the number of reconstructed images.
    for line in lines[4::2]:
        i, qw, qx, qy, qz, tx, ty, tz, i, img_fname = line.split()
        # Colmap provides extrinsics, so must invert
        iRw = Rot3(float(qw), float(qx), float(qy), float(qz))
        wTi = Pose3(iRw, np.array([tx, ty, tz], dtype=np.float64)).inverse()
        wTi_list.append(wTi)
        img_fnames.append(img_fname)

    return wTi_list, img_fnames
```
The only difference there is that we don't return the point3D_ids. But I would prefer we update our internal code to return those, otherwise I think we will have two redundant functions in the codebase

johnwlambert: Another option would be to put in a PR to COLMAP to bring the 3rd party code up to our standard. That's what we did for PatchmatchNet. But I don't think it resolves the redundancy issue (we would need to maintain two copies of this function)

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,507 @@
...
+                tvec=tvec,
+                camera_id=camera_id,
+                name=image_name,
+                xys=xys,
+                point3D_ids=point3D_ids,
+            )
+    return images
+
+
+def write_images_text(images, path):
```
johnwlambert, line 251: can be compared with 
```python
def write_images(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> None:
    """Writes the image data file in the COLMAP format.
    Reference: https://colmap.github.io/format.html#images-txt
    Note: the "Number of images" saved to the .txt file is not the number of images
    fed to the SfM algorithm, but rather the number of localized camera poses/images,
    which COLMAP refers to as the "reconstructed cameras".
    Args:
        gtsfm_data: scene data to write.
        images: list of all images for this scene, in order of image index.
        save_dir: folder to put the images.txt file in.
    """
    os.makedirs(save_dir, exist_ok=True)

    num_imgs = gtsfm_data.number_images()
    # TODO: compute this (from keypoint data? or from track data?)
    mean_obs_per_img = 0

    file_path = os.path.join(save_dir, "images.txt")
    with open(file_path, "w") as f:
        f.write("# Image list with two lines of data per image:\n")
        f.write("#   IMAGE_ID, QW, QX, QY, QZ, TX, TY, TZ, CAMERA_ID, NAME\n")
        f.write("#   POINTS2D[] as (X, Y, POINT3D_ID)\n")
        f.write(f"# Number of images: {num_imgs}, mean observations per image: {mean_obs_per_img}\n")

        for i in gtsfm_data.get_valid_camera_indices():
            img_fname = images[i].file_name
            camera = gtsfm_data.get_camera(i)
            # COLMAP exports camera extrinsics (cTw), not the poses (wTc), so must invert
            iTw = camera.pose().inverse()
            iRw_quaternion = iTw.rotation().quaternion()
            itw = iTw.translation()
            tx, ty, tz = itw
            qw, qx, qy, qz = iRw_quaternion

            f.write(f"{i} {qw} {qx} {qy} {qz} {tx} {ty} {tz} {i} {img_fname}\n")
            # TODO: write out the points2d
            f.write("TODO\n")
```

https://github.com/borglab/gtsfm/blob/master/gtsfm/utils/io.py#L273


We actually have a longstanding issue to add some more functionality to this file here: https://github.com/borglab/gtsfm/issues/130

gtsfm/loader/astronet_loader.py
travisdriver: Moved this to issue [#263](https://github.com/borglab/gtsfm/issues/263).

Ground truth tracks could be used for computing frontend metrics when the ground truth mesh isn't available.

Also [here](https://github.com/borglab/gtsfm/pull/258#discussion_r702464001).

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: Again, all of the data is in binary. I can add these functions to `io.py` if you'd like. 

However, all of these functions cater to a third-party data format. Unless we're planning on making this our own data format as well, maybe they should be in thirdparty? Or possibly part of the specific loader(s)?

dask-report.html
travisdriver: Sorry will delete it. Yes, let's add it to the .gitignore.

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: @johnwlambert What do you think.

I think having each loader handle reading of their respective data formats makes sense. However, the COLMAP format is pretty widespread, so it could be useful having them in io.py as well.

johnwlambert: Hi Travis, I couldn't find any binary-encoded data in the COLMAP files. It looks like they parse some of the plain text that way, but it could be done in either way.

I just don't think the COLMAP python code is high-enough quality to check into our repo. I might be ok with the 3rd party route if we can PR into COLMAP and lift the quality of that file to add docstrings, type hints, etc. Otherwise I think its fairly unreadable. But we have different data structures than them, so it might make sense just to add the few missing lines to our files and use our own converter.

travisdriver: The AstroNet files use the _COLMAP_ binary format. You can find them in `tests/data/test_2011212_opnav_022`.

Yes, I also don't think the code is high enough quality to check into the repo. I was wondering whether you suggest I add the (properly documented) functions to the individual loader or `io.py`.

gtsfm/loader/astronet_loader.py
travisdriver: Done

gtsfm/loader/astronet_loader.py
travisdriver: Done

thirdparty/colmap/scripts/python/read_write_model.py
johnwlambert: I see, got it, didn't realize you were using the explicit .bin format. In that case, I would suggest adding a PR to COLMAP to clean it up per our [contributing guidelines](https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md), and then we could add it directly to the `thirdparty` dir. That's what we did for PatchmatchNet

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+        self._max_frame_lookahead = max_frame_lookahead
+
+        # Use COLMAP model reader to load data and convert to GTSfM format.
+        if Path(data_dir).exists():
+            _cameras, _images, _points3d = read_model(path=data_dir, ext=".bin")
+            self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
+                _cameras, _images, _points3d, load_sfmtracks=use_gt_sfmtracks
+            )
+
+        # Camera intrinsics are currently required due to absense of EXIF data and diffculty in approximating focal
```
johnwlambert, line 76: nit: small typo on "absence"

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+            FileNotFoundError if image path does not exist.
+            RuntimeError if ground truth camera calibrations not provided.
+        """
+        self._use_gt_extrinsics = use_gt_extrinsics
+        self._use_gt_sfmtracks = use_gt_sfmtracks
+        self._max_frame_lookahead = max_frame_lookahead
+
+        # Use COLMAP model reader to load data and convert to GTSfM format.
+        if Path(data_dir).exists():
+            _cameras, _images, _points3d = read_model(path=data_dir, ext=".bin")
```
johnwlambert, line 71: usually we only use the leading underscore prefix for class attributes, not for temporary variables

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+        Raises:
+            FileNotFoundError if image path does not exist.
+            RuntimeError if ground truth camera calibrations not provided.
+        """
+        self._use_gt_extrinsics = use_gt_extrinsics
+        self._use_gt_sfmtracks = use_gt_sfmtracks
+        self._max_frame_lookahead = max_frame_lookahead
+
+        # Use COLMAP model reader to load data and convert to GTSfM format.
+        if Path(data_dir).exists():
```
johnwlambert, line 70: what's the expected behavior if `data_dir` does not exist?

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+             └── points3D.bin: 3D tracks (see https://colmap.github.io/format.html#points3d-txt)
+
+
+        Args:
+            data_dir: path to directory containing the COLMAP-formatted data: cameras.bin, images.bin, and points3D.bin
+            gt_scene_mesh_path (optional): path to file of target small body surface mesh.
+                Note: vertex size mismath observed when reading in from OBJ format. Prefer PLY.
+            use_gt_intrinsics: whether to use ground truth intrinsics. If calibration is
+               not found on disk, then use_gt_intrinsics will be set to false automatically.
+            use_gt_extrinsics (optional): whether to use ground truth extrinsics.
```
johnwlambert, line 55: maybe we could explain how we will use the extrinsics in the docstring. e.g. we would use them as the pseudo-ground truth

gtsfm/loader/astronet_loader.py
johnwlambert: I don't think this function belongs in this file -- I'd prefer we avoid the tech debt and move it elsewhere before we check it in.

gtsfm/utils/io.py
johnwlambert: is the conversion expensive? or a no-op? maybe we should have a check to only do this if the image is grayscale?

travisdriver: Good idea. Only converting if it's not already RGB is cleaner.

gtsfm/loader/astronet_loader.py
travisdriver: We spoke offline and decided having the sfmtracks could be useful (e.g., qualitative comparison between reconstructed and ground truth scene).

gtsfm/loader/astronet_loader.py
travisdriver: I think every loader should have a `to_gtsfmdata()` method that converts the respective data formats to `GtsfmData`. I think putting all of these data conversion methods into the core code for every dataset we use will just clutter things. I think the loader should be more self contained, so that if someone wants to implement GTSfM on their data, they just need to fill in the loader.

Maybe the `LoaderBase` should have an abstract method. Maybe it would look something like this:

```python
@abc.abstractmethod
def to_gtsfmdata(self, **kwargs: Dict[str, Any]) -> GtsfmData:
    """Converts individual data formats to `GtsfmData`."""
```

I think leaving this here untill we address [#263](https://github.com/borglab/gtsfm/issues/263) makes sense.

gtsfm/loader/astronet_loader.py
travisdriver: Done

gtsfm/loader/astronet_loader.py
travisdriver: Added `FileNotFound` exception

gtsfm/loader/astronet_loader.py
travisdriver: Done

gtsfm/loader/astronet_loader.py
travisdriver: Corrected

johnwlambert:
Thanks for all the updates, Travis. Can we put the Astronet test data under an astronet subdir?

instead of `tests/data/test_2011212_opnav_022/images/00000004.png`
i would prefer `tests/data/astronet/test_2011212_opnav_022/images/00000004.png`
so people can know what dataset opnav refers to

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,579 @@
...
+        images = read_images_text(os.path.join(path, "images" + ext))
+        points3D = read_points3D_text(os.path.join(path, "points3D") + ext)
+    else:
+        cameras = read_cameras_binary(os.path.join(path, "cameras" + ext))
+        images = read_images_binary(os.path.join(path, "images" + ext))
+        points3D = read_points3D_binary(os.path.join(path, "points3D") + ext)
+    return cameras, images, points3D
+
+
+def write_model(cameras, images, points3D, path, ext=".bin"):
```
johnwlambert, line 502: Thanks for adding this Travis. Much clearer to read this file now. 

I think this function is missing type hints.

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,579 @@
...
+        and os.path.isfile(os.path.join(path, "images" + ext))
+        and os.path.isfile(os.path.join(path, "points3D" + ext))
+    ):
+        print("Detected model format: '" + ext + "'")
+        return True
+
+    return False
+
+
+def read_model(path, ext=""):
```
johnwlambert, line 479: same here -- think we're missing type hints.

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,579 @@
...
+        points3D = read_points3D_text(os.path.join(path, "points3D") + ext)
+    else:
+        cameras = read_cameras_binary(os.path.join(path, "cameras" + ext))
+        images = read_images_binary(os.path.join(path, "images" + ext))
+        points3D = read_points3D_binary(os.path.join(path, "points3D") + ext)
+    return cameras, images, points3D
+
+
+def write_model(cameras, images, points3D, path, ext=".bin"):
+    """Write model data."""
```
johnwlambert, line 503: maybe we could say something a bit more descriptive like

`Write 3d model data to 3 text or binary files`

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,579 @@
...
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+#
+# Author: Johannes L. Schoenberger (jsch-at-demuc-dot-de)
+
+import os
+import collections
+from typing import Dict, Tuple, Any, BinaryIO
+import struct
+import argparse
```
johnwlambert, line 36: we usually put all the `import` lines before all the `from` lines, within a group.

thirdparty/colmap/scripts/python/read_write_model.py
```diff
@@ -0,0 +1,579 @@
...
+
+import numpy as np
+
+
+CameraModel = collections.namedtuple("CameraModel", ["model_id", "model_name", "num_params"])
```
johnwlambert, line 41: Looks very nice, could you add it as PR to COLMAP also?

travisdriver: I had to remove some empty return statements and return statements in some of the writers because the newly added type hints causeed mypy errors. I'm not sure if they will cause issues with COLMAP. You can test it and submit a PR if you'd like.

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: Added reference to output format.

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: fixed

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: fixed

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: fixed

thirdparty/colmap/scripts/python/read_write_model.py
travisdriver: Redocumented all of the functions. However, I still recommend that we move the modules in `thirdparty/` to [submodules](https://git-scm.com/book/en/v2/Git-Tools-Submodules), as they provide an upstream and will likely lead to less licencing issues.

gtsfm/utils/io.py
travisdriver: Done

johnwlambert:
Looks quite good. Thanks Travis. @ayushbaid could you take another look when you get a chance since you are the most familiar with the `BaseLoader` class design?

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+from typing import Optional, Dict, Tuple, List
+
+from gtsam import Cal3Bundler, Pose3, Rot3, Point3, SfmTrack
+
+import gtsfm.utils.io as io_utils
+import gtsfm.utils.logger as logger_utils
+from gtsfm.common.image import Image
+from gtsfm.loader.loader_base import LoaderBase
+
+from thirdparty.colmap.scripts.python.read_write_model import read_model
```
johnwlambert, line 17: nit: typically we only import classes from modules, not individual function names. This is per the Google style guide: https://google.github.io/styleguide/pyguide.html#22-imports.

So I would prefer:
```python
import thirdparty.colmap.scripts.python.read_write_model as colmap_io
```
or
```python
import thirdparty.colmap.scripts.python.read_write_model as read_write_models
```

tests/loader/test_astronet_loader.py
```diff
@@ -0,0 +1,183 @@
...
+
+import unittest
+from pathlib import Path
+
+import numpy as np
+from gtsam import Pose3, PinholeCameraCal3Bundler
+
+from gtsfm.common.image import Image
+from gtsfm.loader.astronet_loader import AstroNetLoader
+from thirdparty.colmap.scripts.python.read_write_model import read_model
```
johnwlambert, line 14: same note here about importing an individual function vs. Google style

gtsfm/loader/astronet_loader.py
akshay-krishnan: +1

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+from thirdparty.colmap.scripts.python.read_write_model import Camera as ColmapCamera
+from thirdparty.colmap.scripts.python.read_write_model import Image as ColmapImage
+from thirdparty.colmap.scripts.python.read_write_model import Point3D as ColmapPoint3D
+
+
+logger = logger_utils.get_logger()
+
+
+class AstroNetLoader(LoaderBase):
+    """Simple loader class that reads an AstroNet data segment.
```
akshay-krishnan, line 27: nit: `Simple` is unnecessary. Might be useful to explain a bit about AstroNet or the data format being read.

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+from thirdparty.colmap.scripts.python.read_write_model import read_model
+from thirdparty.colmap.scripts.python.read_write_model import Camera as ColmapCamera
+from thirdparty.colmap.scripts.python.read_write_model import Image as ColmapImage
+from thirdparty.colmap.scripts.python.read_write_model import Point3D as ColmapPoint3D
+
+
+logger = logger_utils.get_logger()
+
+
+class AstroNetLoader(LoaderBase):
```
akshay-krishnan, line 26: Should the class be called `AstronetLoader`, considering the context here? The file is also called astronet_loader.

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+            RuntimeError if ground truth camera calibrations not provided.
+        """
+        self._use_gt_extrinsics = use_gt_extrinsics
+        self._use_gt_sfmtracks = use_gt_sfmtracks
+        self._max_frame_lookahead = max_frame_lookahead
+
+        # Use COLMAP model reader to load data and convert to GTSfM format.
+        if Path(data_dir).exists():
+            cameras, images, points3d = read_model(path=data_dir, ext=".bin")
+            self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
```
akshay-krishnan, line 70: Are calibrations intrinsics or extrinsics? Maybe use a better variable name?

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+        # Camera intrinsics are currently required due to absence of EXIF data and diffculty in approximating focal
+        # length (usually 10000 to 100000 pixels).
+        if self._calibrations is None:
+            raise RuntimeError("Camera intrinsics cannot be None.")
+
+        if self._wTi_list is None:
+            self._use_gt_extrinsics = False
+
+        if self._sfmtracks is None:
+            self._use_gt_sfmtracks = False
```
akshay-krishnan, line 85: This is weird from a design perspective. if the user specifies use_gt_extrinsics or use_gt_sfmtracks, and they are not found, I think its better to throw an error than to silently force them to `False`. Maybe they had the wrong path or something.

akshay-krishnan: To do that without even logging a warning would be very user-unfriendly :)

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+        Args:
+            index: the index to fetch.
+
+        Returns:
+            pose for the given camera.
+        """
+        if index < 0 or index >= len(self):
+            raise IndexError(f"Image index {index} is invalid")
+
+        if not self._use_gt_extrinsics:
```
akshay-krishnan, line 197: check if _wTi_list is None too?

gtsfm/utils/io.py
```diff
@@ -55,6 +55,7 @@ def load_image(img_path: str) -> Image:
         exif_data = parsed_data
 
     img_fname = Path(img_path).name
+    original_image = original_image.convert("RGB") if original_image.mode != "RGB" else original_image
```
akshay-krishnan, line 58: looks like something that should be documented in the function docstring.

tests/loader/test_astronet_loader.py
akshay-krishnan: If these have a 1:1 correspondences, it might be better to use a list of tuples instead of 2 lists. Ive found that to cause fewer bugs.

tests/loader/test_astronet_loader.py
```diff
@@ -0,0 +1,183 @@
...
+            data_dir,
+            use_gt_extrinsics=True,
+            use_gt_sfmtracks=True,
+            max_frame_lookahead=2,
+        )
+
+    def test_constructor_set_properties(self) -> None:
+        """Ensure that constructor sets class properties correctly."""
+        # assert self.loader._gt_scene_trimesh is not None
+        # assert self.loader._gt_scene_trimesh.vertices;shape[0] == 5002
```
akshay-krishnan, line 50: remove if unnecessary?

tests/loader/test_astronet_loader.py
```diff
@@ -0,0 +1,183 @@
...
+        assert self.loader._use_gt_extrinsics
+        assert self.loader._use_gt_sfmtracks
+        assert self.loader._max_frame_lookahead == 2
+
+    def test_len(self) -> None:
+        """Ensure we have one calibration per image/frame."""
+        # there are 4 images and 1465 tracks in 2011212_opnav_022
+        assert len(self.loader) == 4
+        assert len(self.loader._calibrations) == 4
```
akshay-krishnan, line 59: Consider renaming these to intrinsics. Not sure if we have a convention for this.

akshay-krishnan (changes requested):

gtsfm/loader/astronet_loader.py
travisdriver: I agree; I was following convention of the other loaders. Should I change or keep the same as others?

tests/loader/test_astronet_loader.py
travisdriver: I agree; I was following convention of the other loaders. Should I change or keep the same as others?

gtsfm/loader/astronet_loader.py
travisdriver: Calibrations are intrinsics.

gtsfm/loader/astronet_loader.py
travisdriver: My dataset is called `AstroNet`,  so yeah I think it should be `AstronetLoader` instead.

gtsfm/loader/astronet_loader.py
travisdriver: Agreed. The data format is explained in the `__init__` docstring

tests/loader/test_astronet_loader.py
akshay-krishnan: I think _intrinsincs or _camera_intrinsics is better. We even have a method called get_camera_intrinsics. WDYT @ayushbaid or @johnwlambert ?

gtsfm/loader/astronet_loader.py
akshay-krishnan: I think its better to raise exceptions if its None. The pipeline will crash, but that is probably the better action here, instead of waiting for the results only to realize we did not use ground truth. 
If its an easy change, could you do the same for the other loaders? Or maybe add a TODO? Asking for @johnwlambert 's or @ayushbaid 's opinion here.

gtsfm/loader/astronet_loader.py
akshay-krishnan: same as the other comment, maybe change to _intrinsics?

tests/loader/test_astronet_loader.py
travisdriver: I do think the _calibrations aligns better with GTSAMs naming convention (e.g., Cal3Bundler). I think keeping _calibrations is fine for now.

gtsfm/loader/astronet_loader.py
johnwlambert: Akshay did you mean disambiguating from a CNN architecture/network name here?

gtsfm/loader/astronet_loader.py
```diff
@@ -0,0 +1,231 @@
...
+        self._max_frame_lookahead = max_frame_lookahead
+
+        # Use COLMAP model reader to load data and convert to GTSfM format.
+        if Path(data_dir).exists():
+            cameras, images, points3d = read_model(path=data_dir, ext=".bin")
+            self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
+                cameras, images, points3d, load_sfmtracks=use_gt_sfmtracks
+            )
+        else:
+            raise FileNotFoundError("No data found at %s." % data_dir)
```
johnwlambert, line 74: prefer to raise error immediately, and not nest (from [contributing](https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md#python-style))

So I would prefer:
```python
if not Path(data_dir).exists():
    raise FileNotFoundError("No data found at %s." % data_dir)

cameras, images, points3d = read_model(path=data_dir, ext=".bin")
self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
```

gtsfm/loader/astronet_loader.py
travisdriver: Maybe I could raise a warning instead? The extrinsics and sfmtracks aren't required to reconstruct the scene and are used strictly for computing metrics, so I could see a case for either.

johnwlambert: I think it depends on how general purpose we expect the loader to be.

For the Olsson loader, if no GT is found, we still proceed, since we assume it's ok for the user to just give us a folder of images without GT, since that's a valid situation.

For the Astronet data, if we assume GT is always readily available, then we should probably raise an error. But if downloading the GT might be cumbersome, and someone just wants to try, then maybe no error, but a helpful warning message would be appreciated.

gtsfm/loader/astronet_loader.py
johnwlambert: +1

gtsfm/utils/io.py
johnwlambert: +1

tests/loader/test_astronet_loader.py
johnwlambert: I don't have a strong opinion about this one. Yeah I believe @ayushbaid originally named this one with the GTSAM principle in mind?

travisdriver: However, in the other loaders, the `_use_gt_intrinsics` is used to specify whether to set `_calibrations`. So I do think a naming convention should be decided upon. I'll move this to another issue.

gtsfm/loader/astronet_loader.py
travisdriver: If `_wTi_list` is None then `_use_gt_extrinsics` is set to False in __init__.

gtsfm/loader/astronet_loader.py
travisdriver: Actually, I think raising an exception makes more sense, because if we just ignore that the inputs are missing when they're requested by the user, it kind of defeats the purpose of the `_use_gt_intrinsics`, `_use_gt_extrinsics`, and `_use_gt_sfmtrakcs` variables.

tests/loader/test_astronet_loader.py
travisdriver: They're used individually as well, so I thought it would be clearer to split them up.

gtsfm/loader/astronet_loader.py
travisdriver: Agreed. Fixed

tests/loader/test_astronet_loader.py
travisdriver: fixed

gtsfm/loader/astronet_loader.py
travisdriver: fixed

gtsfm/loader/astronet_loader.py
travisdriver: Keeping with GTSAM naming convention.

gtsfm/loader/astronet_loader.py
travisdriver: Added `RuntimeError` if ground truth requested but not found.

gtsfm/loader/astronet_loader.py
travisdriver: Added check. Should also be added to the other loaders

gtsfm/utils/io.py
travisdriver: added

tests/loader/test_astronet_loader.py
travisdriver: I will be uncommenting this (and some other tests) in the next PR when I add trimesh. It would be useful to be able to keep them here so I don't have to rewrite them.

gtsfm/loader/astronet_loader.py
travisdriver: agreed. fixed

akshay-krishnan (approved):

gtsfm/loader/astronet_loader.py
akshay-krishnan: Yeah I just meant using a small 'n' instead of 'N' because Astronet is one thing as far as the loader is concerned. Its resolved now.

PR #260, opened 2021-09-05
Fresh branch off master -- increase min track length up to 3

johnwlambert:
Increasing min track length from 2 up to 3.

Freshly branched off master for cleaner git history. We'll check pose errors after BA vs. before.


travisdriver (commented):
Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.

johnwlambert:
> Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.

Thanks. The only place I think it would be hard-coded would be here, when we sample 2 views in RANSAC. https://github.com/borglab/gtsfm/blob/master/gtsfm/data_association/point3d_initializer.py#L27. Not sure if that counts, though. @akshay-krishnan is that what you meant?

johnwlambert:
@dellaert Before we merge this change (requiring length-3 tracks as minimum length), it looks like the change hurts all metrics:

After BA, w/ the change rotation angular errors slightly decrease (8.8 -> 8.0 median) but mean increases (9.0 to 9.4 mean)

- After BA, w/ the change, translation angular errors increase from 7.3 -> 8.4 (median), 8.8 -> 13.7 (mean)
- After BA, w/ the change, we get about 1/20th of the number of tracks -> going from 2236 tracks to 118 tracks
- After BA, w/ the change, translation distance errors increase from 0.43 -> 0.93 (median),  0.52 to 1.15 (mean)

I propose we hold off merging it until we understand why this leads to higher pose errors. Otherwise I think it is a step back.

cc @travisdriver @akshay-krishnan @ayushbaid

travisdriver:
> @dellaert Before we merge this change (requiring length-3 tracks as minimum length), it looks like the change hurts all metrics:
> 
> * After BA, w/ the change, translation angular errors **increase** from 7.3 -> 8.4 (median), 8.8 -> 13.7 (mean)
> * After BA, w/ the change, translation distance errors **increase** from 0.43 -> 0.93 (median),  0.52 to 1.15 (mean)
> * After BA, w/ the change rotation angular errors slightly decrease (8.8 -> 8.0 median) but mean **increases** (9.0 to 9.4 mean)
> * After BA, w/ the change, we get about 1/20th of the number of tracks -> going from 2236 tracks to 118 tracks
> 
> I propose we hold off merging it until we understand why this leads to higher pose errors. Otherwise I think it is a step backwards.
> 
> cc @travisdriver @akshay-krishnan @ayushbaid

Which benchmark is this with respect to? There are 6 different benchmarks, but you only refer to one.

johnwlambert:
> > @dellaert Before we merge this change (requiring length-3 tracks as minimum length), it looks like the change hurts all metrics:
> > 
> > * After BA, w/ the change, translation angular errors **increase** from 7.3 -> 8.4 (median), 8.8 -> 13.7 (mean)
> > * After BA, w/ the change, translation distance errors **increase** from 0.43 -> 0.93 (median),  0.52 to 1.15 (mean)
> > * After BA, w/ the change rotation angular errors slightly decrease (8.8 -> 8.0 median) but mean **increases** (9.0 to 9.4 mean)
> > * After BA, w/ the change, we get about 1/20th of the number of tracks -> going from 2236 tracks to 118 tracks
> > 
> > I propose we hold off merging it until we understand why this leads to higher pose errors. Otherwise I think it is a step backwards.
> > cc @travisdriver @akshay-krishnan @ayushbaid
> 
> Which benchmark is this with respect to? There are 6 different benchmarks, but you only refer to one.

Sorry -- meant on skydio32, which I think is the main one that is not fully solved. (w/ deep front end because the sift front end does much worse)

johnwlambert:
On Door 12, I see 5123 tracks before change, 4072 tracks after change.
I see translation error after BA stay the same: 0.02 to 0.02 (mean), but already solved, since so close to zero
I see translation angular errors after BA increase slightly (0.65 to 0.59), basically solved
Rotation angular errors after BA stay the same (0.03 to 0.03), basically solved
Mean track length does increase with the change (6.85 to 8.11).

travisdriver:
I'm not sure it's valid to draw any conclusions from a single datapoint. What about the other benchmarks? If you think the other benchmarks are not sufficient for evaluating the performance of gtsfm, then maybe we should be using other datasets.

In general, I think we need to graduate from the single (and even double) digit length image sequences. I also think we need to inject more variability into the datasets, as there are only two different scenes being used for benchmarking. Maybe we could also add some argoverse and AstroNet datasets for benchmarking.

johnwlambert:
On Skydio-8, I see rotation error after BA decreases with the change, from 0.56 to 0.51 (mean), 0.46 -> 0.43 (median)
After the change, translation angular error decreases from 132 to 112 (mean), 3.39 -> 3.12 (median)

travisdriver:
> @dellaert Before we merge this change (requiring length-3 tracks as minimum length), it looks like the change hurts all metrics on Skydio-32:
> 
> * After BA, w/ the change, translation angular errors **increase** from 7.3 -> 8.4 (median), 8.8 -> 13.7 (mean)
> * After BA, w/ the change, translation distance errors **increase** from 0.43 -> 0.93 (median),  0.52 to 1.15 (mean)
> * After BA, w/ the change rotation angular errors slightly decrease (8.8 -> 8.0 median) but mean **increases** (9.0 to 9.4 mean)
> * After BA, w/ the change, we get about 1/20th of the number of tracks -> going from 2236 tracks to 118 tracks
> 
> I propose we hold off merging it until we understand why this leads to higher pose errors. Otherwise I think it is a step backwards.
> 
> cc @travisdriver @akshay-krishnan @ayushbaid

What about before BA?

akshay-krishnan:
> > Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.
> 
> Thanks. The only place I think it would be hard-coded would be here, when we sample 2 views in RANSAC. https://github.com/borglab/gtsfm/blob/master/gtsfm/data_association/point3d_initializer.py#L27. Not sure if that counts, though. @akshay-krishnan is that what you meant?

Here are are a couple (not sure if there are more): 
- https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L186
- https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L303

akshay-krishnan:
> @dellaert Before we merge this change (requiring length-3 tracks as minimum length), it looks like the change hurts all metrics on Skydio-32:
> 
> * After BA, w/ the change, translation angular errors **increase** from 7.3 -> 8.4 (median), 8.8 -> 13.7 (mean)
> * After BA, w/ the change, translation distance errors **increase** from 0.43 -> 0.93 (median),  0.52 to 1.15 (mean)
> * After BA, w/ the change rotation angular errors slightly decrease (8.8 -> 8.0 median) but mean **increases** (9.0 to 9.4 mean)
> * After BA, w/ the change, we get about 1/20th of the number of tracks -> going from 2236 tracks to 118 tracks
> 
> I propose we hold off merging it until we understand why this leads to higher pose errors. Otherwise I think it is a step backwards.
> 
> cc @travisdriver @akshay-krishnan @ayushbaid

I don't think it is a step backwards, rather the metrics might now be a better signal of the final reconstruction quality.  But yes, I agree that more insights are required before concluding this guess is actually true.

travisdriver:
> > > Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.
> > 
> > 
> > Thanks. The only place I think it would be hard-coded would be here, when we sample 2 views in RANSAC. https://github.com/borglab/gtsfm/blob/master/gtsfm/data_association/point3d_initializer.py#L27. Not sure if that counts, though. @akshay-krishnan is that what you meant?
> 
> Here are are a couple (not sure if there are more):
> 
> * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L186
> * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L303

Thank you for finding these. They should be corrected now.

johnwlambert:
> > > Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.
> > 
> > 
> > Thanks. The only place I think it would be hard-coded would be here, when we sample 2 views in RANSAC. https://github.com/borglab/gtsfm/blob/master/gtsfm/data_association/point3d_initializer.py#L27. Not sure if that counts, though. @akshay-krishnan is that what you meant?
> 
> Here are are a couple (not sure if there are more):
> 
> * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L186
> * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L303

I see those length-2 exceptions as definitions of types of exceptions. The min track length will be checked elsewhere, and will trump all. So I think we should keep the length-2 exceptions as is

travisdriver:
Yes, it looks like the track length is checked [here](https://github.com/borglab/gtsfm/blob/2072a9893d878aa99dfce5c66ab361991034bef4/gtsfm/data_association/data_assoc.py#L118), so the stuff hardcoded in point3d_initiializer doesn't affect the accepted tracks.

Changed it back.

travisdriver (commented):
Everything looks good to me. The change improved performance for a majority of the benchmarks. 

We should merge it before we have to start another fresh branch :)

travisdriver (approved):

akshay-krishnan:
> > > > Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.
> > > 
> > > 
> > > Thanks. The only place I think it would be hard-coded would be here, when we sample 2 views in RANSAC. https://github.com/borglab/gtsfm/blob/master/gtsfm/data_association/point3d_initializer.py#L27. Not sure if that counts, though. @akshay-krishnan is that what you meant?
> > 
> > 
> > Here are are a couple (not sure if there are more):
> > 
> > * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L186
> > * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L303
> 
> I see those length-2 exceptions as definitions of types of exceptions. The min track length will be checked elsewhere, and will trump all. So I think we should keep the length-2 exceptions as is

True, but why are these not defined in terms of min_track_length?

travisdriver:
> > > > > Looks good to me. However, @akshay-krishnan mentioned that the track length is hard coded in places, but I can't find these instances anywhere.
> > > > 
> > > > 
> > > > Thanks. The only place I think it would be hard-coded would be here, when we sample 2 views in RANSAC. https://github.com/borglab/gtsfm/blob/master/gtsfm/data_association/point3d_initializer.py#L27. Not sure if that counts, though. @akshay-krishnan is that what you meant?
> > > 
> > > 
> > > Here are are a couple (not sure if there are more):
> > > 
> > > * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L186
> > > * https://github.com/borglab/gtsfm/blob/61189858e11ba8b4bbb8a57301b8df2c70beaa40/gtsfm/data_association/point3d_initializer.py#L303
> > 
> > 
> > I see those length-2 exceptions as definitions of types of exceptions. The min track length will be checked elsewhere, and will trump all. So I think we should keep the length-2 exceptions as is
> 
> True, but why are these not defined in terms of min_track_length?

I was going to say that the point3d_initializer is just concerned with whether or not the track can be triangulated, but it also handles some outlier (i.e., filtering by rejection reprojection error).

Therefore, I propose we either add another check (and exit code) for the min_track_length within point3d_initializer, or move all of the outlier rejection out of the point3d_initializer and into the DataAssociation module.

I personally think having the outlier rejection outside of the point3d initializer makes the most sense, and that having all of the outlier rejection policies in one place is cleaner.

travisdriver:
Unfortunately, I think we need to figure out why so many tracks are being filtered after BA before merging this PR:

Metric name | Value
-- | --
number_cameras | 32
number_tracks_unfiltered | 2374
mean_3d_track_lengths_unfiltered | 3.86689
mean_reprojection_errors_unfiltered_px | 47.3911
mean_rotation_error_angle_deg | 7.69377
mean_translation_error_distance | 0.756135
mean_translation_angle_error_deg | 11.2585
number_tracks_filtered | 113
mean_3d_track_lengths_filtered | 3.11504
mean_reprojection_errors_filtered_px | 1.6091

min-track-length-3-skidio32

ayushbaid:
Do we want to keep this open? @johnwlambert @travisdriver

travisdriver:
> Do we want to keep this open? @johnwlambert @travisdriver

I'll update and get the benchmark dashboard to see where we're at. It may be the case that we need to increase recall before making this change is effective, and in that case I'll move this to an issue.

johnwlambert:
superseded by https://github.com/borglab/gtsfm/pull/747

PR #262, opened 2021-09-05
Adding number of cameras to metrics

ayushbaid:
This PR adds the metrics of number of cameras estimated in the averaging, data association, and bundle adjustment module.

### DIFF
--- a/gtsfm/bundle/bundle_adjustment.py
+++ b/gtsfm/bundle/bundle_adjustment.py
@@ -131,6 +131,7 @@ class BundleAdjustmentOptimizer(NamedTuple):
         def get_metrics_from_sfm_data(sfm_data: GtsfmData, suffix: str) -> List[GtsfmMetric]:
             """Helper to get bundle adjustment metrics from a GtsfmData object with a suffix for metric names."""
             metrics = []
+            metrics.append(GtsfmMetric(name="number_cameras", data=sfm_data.number_images()))
             metrics.append(GtsfmMetric("number_tracks" + suffix, sfm_data.number_tracks()))
             metrics.append(
                 GtsfmMetric(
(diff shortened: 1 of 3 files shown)


gtsfm/utils/metrics.py
```diff
@@ -176,6 +176,8 @@ def compute_averaging_metrics(
     gt_wRi_list, gt_wti_list = get_rotations_translations_from_poses(gt_wTi_list)
 
     metrics = []
+    metrics.append(GtsfmMetric(name="num_rotations_computed", data=len(wRi_list)))
```
johnwlambert, line 179: Thanks for the PR Ayush.

should it instead be the number of not None entries? Can you double check it works as expected on some data?

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -131,6 +131,7 @@ def run(
         def get_metrics_from_sfm_data(sfm_data: GtsfmData, suffix: str) -> List[GtsfmMetric]:
             """Helper to get bundle adjustment metrics from a GtsfmData object with a suffix for metric names."""
             metrics = []
```
johnwlambert, line 133: I think number_images() is now kept fixed since the loader. I think we want the length of get valid camera indices?

gtsfm/utils/metrics.py
ayushbaid: Thanks John. Fixed it now

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Yep. corrected it.

travisdriver:
Looks good to me.

travisdriver (approved):
Looks good to me.

PR #264, opened 2021-09-05
Reverting NodeJS to Latest Version

adisingh50:
Neha, Pratyusha, and I were able to figure out an alternate solution to the `numjs` and `sharp` dependency issue by fixing an issue with the conda libtool. They are both able to properly run the React Application on their machines now.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -24,7 +24,7 @@ dependencies:
   - matplotlib==3.4.2
   - networkx
   - numpy
-  - nodejs==13.13.0 # will be updated to version 16.16.1 when numjs issue #81 is fixed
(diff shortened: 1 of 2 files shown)


ayushbaid:
LGTM!

Does the existing test sufficiently cover this setup issue? I mean the unit tests were all passing and we were not able to start the react application. It is not straightforward, but is there any way we check it automatically?

ayushbaid (approved):

johnwlambert:
Thanks for the PR @adisingh50. Could you explain the updates need to sharp/libtool to fix the issue? Are there extra instructions we should add to the readme in another PR?

PR #268, opened 2021-09-06
increasing max keypoints to 8192

ayushbaid:
On comparing GTSFM's SIFT with Colmap's, we notice that the maximum number of keypoints is 8192 (compared to our 5000). This is a significant difference, and having more keypoints can improve the performance of our systems.

### DIFF
--- a/gtsfm/frontend/detector/detector_base.py
+++ b/gtsfm/frontend/detector/detector_base.py
@@ -10,16 +10,17 @@ from dask.delayed import Delayed
 from gtsfm.common.image import Image
 from gtsfm.common.keypoints import Keypoints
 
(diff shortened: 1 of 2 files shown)


gtsfm/frontend/detector/detector_base.py
```diff
@@ -10,16 +10,17 @@
 from gtsfm.common.image import Image
 from gtsfm.common.keypoints import Keypoints
 
+DEFAULT_MAX_KEYPOINTS = 8192
```
johnwlambert, line 13: looks good. maybe you can add a URL with line number referencing where COLMAP sets this?

PR #269, opened 2021-09-06
Use native unzip commands in actions

ayushbaid:
We rely on a third party action to unzip datasets, which is flaky. 

This PR switches to directly using `unzip` on the archives.


ayushbaid:
Unzip successful in the benchmarks. Can be merged.

johnwlambert (approved):

johnwlambert:
thanks Ayush, hopefully will be less buggy now

PR #272, opened 2021-09-07
Add references to other SfM libraries in various modules

johnwlambert:


gtsfm/data_association/point3d_initializer.py
```diff
@@ -59,6 +59,8 @@ class Point3dInitializer(NamedTuple):
     """Class to initialize landmark points via triangulation w/ or w/o RANSAC inlier/outlier selection.
 
     Note: We currently limit the size of each sample to 2 camera views in our RANSAC scheme.
+    Comparable to OpenSfM's `TrackTriangulator` class:
+        https://github.com/mapillary/OpenSfM/blob/master/opensfm/reconstruction.py#L755
```
akshay-krishnan, line 63: Did we actually refer to OpenSfM's implementation when implementing this? Because every SfM library has a triangulator class, and we should reference (only) those that we actually referred to.

akshay-krishnan (approved):

gtsfm/data_association/point3d_initializer.py
johnwlambert: Thanks @akshay-krishnan. We did not refer to this class when implementing it (I only recently found it), but I think it's a very useful reference to have closeby. The implementation is basically identical to ours.

ayushbaid (approved):
LGTM!

Were these the only places where references were missing? Or will be adding them as we go.

travisdriver:
@johnwlambert Is this good to merge now?

PR #273, opened 2021-09-07
try benchmarking on 281 image dataset (Olsson Palace of Fine Arts)

johnwlambert:
Parallelize the CI to run 7 benchmarks at once, instead of 3 successively. Speeds up CI by X percent.

Also, add a new benchmark for 281 images, although with only a 20-frame lookahead for now (to be extended later).

### DIFF
--- a/.github/scripts/execute_single_benchmark.sh
+++ b/.github/scripts/execute_single_benchmark.sh
@@ -1,0 +1,75 @@
+#!/bin/bash
+
+
+CONFIG_NAME=$1
+DATASET_NAME=$2
+MAX_FRAME_LOOKAHEAD=$3
+IMAGE_EXTENSION=$4
+DATASET_SRC=$5
+LOADER_NAME=$6
+
+
+echo "Config: ${CONFIG_NAME}, Dataset: ${DATASET_NAME}, Download Source: ${DATASET_SRC}, Loader: ${LOADER_NAME}"
+
+# Prepare the download URLs.
+if [ "$DATASET_NAME" == "skydio-8" ]; then
+  export GDRIVE_FILEID='1mmM1p_NpL7-pnf3iHWeWVKpsm1pcBoD5'
+elif [ "$DATASET_NAME" == "skydio-32" ]; then
+  export GDRIVE_FILEID='1BQ6jp0DD3D9yhTnrDoEddzlMYT0RRH68'
+elif [ "$DATASET_NAME" == "palace-fine-arts-281" ]; then
+  WGET_URL1=http://vision.maths.lth.se/calledataset/fine_arts_palace/fine_arts_palace.zip
+  WGET_URL2=http://vision.maths.lth.se/calledataset/fine_arts_palace/data.mat
+  echo $WGET_URL1
+  echo $WGET_URL2
+fi
+
+# Download the data.
+if [ "$DATASET_SRC" == "gdrive" ]; then
+  echo "Downloading ${DATASET_NAME} from GDRIVE"
+  export GDRIVE_URL='https://docs.google.com/uc?export=download&id='$GDRIVE_FILEID
+  wget --save-cookies cookies.txt $GDRIVE_URL -O- | sed -rn 's/.*confirm=([0-9A-Za-z_]+).*/\1/p' > confirm.txt
+  wget --load-cookies cookies.txt -O ${DATASET_NAME}.zip $GDRIVE_URL'&confirm='$(<confirm.txt)
+
+elif [ "$DATASET_SRC" == "wget" ]; then
+  echo "Downloading ${DATASET_NAME} with WGET"
+  wget $WGET_URL1
+  wget $WGET_URL2
+fi
+
+# Extract the data, configure arguments for runner.
+if [ "$DATASET_NAME" == "door-12" ]; then
+  DATASET_ROOT=tests/data/set1_lund_door
+
+elif [ "$DATASET_NAME" == "skydio-8" ]; then
+  unzip -qq skydio-8.zip
+  IMAGES_DIR=skydio_crane_mast_8imgs_with_exif/images
+  COLMAP_FILES_DIRPATH=skydio_crane_mast_8imgs_with_exif/crane_mast_8imgs_colmap_output
+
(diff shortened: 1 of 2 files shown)


travisdriver:
How many images are in this dataset?

johnwlambert:
> How many images are in this dataset?

@travisdriver  there are 281 in this dataset. But the baselines are tiny and the dataset is pretty easy. But the amount of images is a big stress test.

from https://www1.maths.lth.se/matematiklth/personal/calle/dataset/dataset.html

travisdriver:
> > How many images are in this dataset?
> 
> @travisdriver there are 281 in this dataset. But the baselines are tiny and the dataset is pretty easy. But the amount of images is a big stress test.
> 
> from https://www1.maths.lth.se/matematiklth/personal/calle/dataset/dataset.html

Sounds good; the 16 image dataset that was posted on Slack had the same name so wasn't sure. Think this is a great benchmark to have.

johnwlambert:
cc'ing @adisingh50 on the Palace-of-Fine-Arts-281 dataset

akshay-krishnan:
Can we remove the benchmark for Door if its not giving us useful signals anymore?

.github/workflows/benchmark.yml
```diff
@@ -38,64 +46,83 @@ jobs:
...
+              --config_name ${CONFIG_NAME}.yaml \
+              --image_extension $IMAGE_EXTENSION
+
+          elif [ "$LOADER_NAME" == "colmap-loader" ]; then
+              python gtsfm/runner/run_scene_optimizer_colmaploader.py \
+              --images_dir ${IMAGES_DIR} \
+              --colmap_files_dirpath $COLMAP_FILES_DIRPATH \
+              --max_frame_lookahead $MAX_FRAME_LOOKAHEAD \
+              --config_name ${CONFIG_NAME}.yaml
+          fi
```
akshay-krishnan, line 121: Could you put this into a shell script and just call it here?

akshay-krishnan: All the commands for that action I mean

johnwlambert: Good call @akshay-krishnan. It's updated now.

.github/scripts/execute_single_benchmark.sh
```diff
@@ -0,0 +1,75 @@
...
+  --config_name ${CONFIG_NAME}.yaml \
+  --image_extension $IMAGE_EXTENSION
+
+elif [ "$LOADER_NAME" == "colmap-loader" ]; then
+  python gtsfm/runner/run_scene_optimizer_colmaploader.py \
+  --images_dir ${IMAGES_DIR} \
+  --colmap_files_dirpath $COLMAP_FILES_DIRPATH \
+  --max_frame_lookahead $MAX_FRAME_LOOKAHEAD \
+  --config_name ${CONFIG_NAME}.yaml
+fi
```
akshay-krishnan, line 75: We should have default else for all these elifs which prints an error and returns without executing the rest

johnwlambert: so for door, we don't need to download it, or unzip it, so i think we want to fall through all branches on some of the datasets

akshay-krishnan (approved):

ayushbaid:
I do like the fact that we can run all the benchmarks in parallel but I think we might have overdone it. 

We could have just had 6/7 different bash scripts, one for each benchmark. And from the yml file call the bash script. The bash script downloads the dataset, unzips it, and runs gtsfm for it. The name of the bash script can go in the config.

I feel we have way too many variables which we will change ocassionaly, and hance we can hardcode them in the bash script. No need to have all the names in the zip files too.

For example: a config entry can just be 'palace-of-arts': the bash script will be 'palace-of-arts.sh', and the output zip will be 'palace-of-arts.zip'.

johnwlambert:
> I do like the fact that we can run all the benchmarks in parallel but I think we might have overdone it.
> 
> We could have just had 6/7 different bash scripts, one for each benchmark. And from the yml file call the bash script. The bash script downloads the dataset, unzips it, and runs gtsfm for it. The name of the bash script can go in the config.
> 
> I feel we have way too many variables which we will change ocassionaly, and hance we can hardcode them in the bash script. No need to have all the names in the zip files too.
> 
> For example: a config entry can just be 'palace-of-arts-sift': the bash script will be 'palace-of-arts-sift.sh', and the output zip will be 'palace-of-arts-sift.zip'.

Thanks for the review @ayushbaid. I was trying to avoid having a separate "sift" and "deep" bash script though, since that code would all be copy-pasted. Same for copy-pasting the GDrive download code from one to another. I couldn't figure out how to have a double-for loop in the Github Actions matrix (I don't think it's supported).

How about we make one bash script for gdrive, and then separate one for the other olsson loader ones? I'll make a separate PR for that. I guess in the interest of cutting down the CI time from 5 to 3 hours for all our other PRs, I'll merge this now, and submit the changes to in the next PR.

ayushbaid:
> > I do like the fact that we can run all the benchmarks in parallel but I think we might have overdone it.
> > 
> > We could have just had 6/7 different bash scripts, one for each benchmark. And from the yml file call the bash script. The bash script downloads the dataset, unzips it, and runs gtsfm for it. The name of the bash script can go in the config.
> > 
> > I feel we have way too many variables which we will change ocassionaly, and hance we can hardcode them in the bash script. No need to have all the names in the zip files too.
> > 
> > For example: a config entry can just be 'palace-of-arts-sift': the bash script will be 'palace-of-arts-sift.sh', and the output zip will be 'palace-of-arts-sift.zip'.
> 
> Thanks for the review @ayushbaid. I was trying to avoid having a separate "sift" and "deep" bash script though, since that code would all be copy-pasted. Same for copy-pasting the GDrive download code from one to another. I couldn't figure out how to have a double-for loop in the Github Actions matrix (I don't think it's supported).
> 
> How about we make one bash script for gdrive, and then separate one for the other olsson loader ones? I'll make a separate PR for that. I guess in the interest of cutting down the CI time from 5 to 3 hours for all our other PRs, I'll merge this now, and submit the changes to in the next PR.

I am ok with the current change too! Just one suggestion to improve the names of the artifacts. No need to change it to different scripts.

PR #279, opened 2021-09-07
Add open3d to conda environment

johnwlambert:
Used for visualization.


tests/visualization/test_open3d_vis_utils.py
```diff
@@ -0,0 +1,35 @@
...
+import numpy as np
+import open3d
+
+import visualization.open3d_vis_utils as open3d_vis_utils
+
+
+def test_create_colored_spheres_open3d() -> None:
+    """Try creating spheres centered at randomly generated 3d locations (disclaimer: slow)."""
+    num_points = 1000
+    point_cloud = np.random.randn(num_points, 3)
```
akshay-krishnan, line 16: we should set the seed before using random in tests for reproducibility.

tests/visualization/test_open3d_vis_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    spheres = open3d_vis_utils.create_colored_spheres_open3d(point_cloud, rgb, sphere_radius)
+
+    assert isinstance(spheres, list)
+    assert all([isinstance(sphere, open3d.geometry.TriangleMesh) for sphere in spheres])
+
+
+def test_create_colored_point_cloud_open3d() -> None:
+    """Try creating points at randomly generated 3d locations (should be fast)."""
+    num_points = 10000
+    point_cloud = np.random.randn(num_points, 3)
```
akshay-krishnan, line 30: maybe a testcase class is better here so you can have the point cloud as a member shared across tests.

tests/visualization/test_open3d_vis_utils.py
```diff
@@ -0,0 +1,35 @@
...
+    # high bound is exclusive
+    rgb = np.random.randint(low=0, high=256, size=(num_points, 3))
+
+    pcd = open3d_vis_utils.create_colored_point_cloud_open3d(point_cloud, rgb)
```
akshay-krishnan, line 34: also this seems like a super loose test.

akshay-krishnan (changes requested):

tests/visualization/test_open3d_vis_utils.py
johnwlambert: Thanks for the review @akshay-krishnan. Agreed its a super loose test. I mostly wanted to make sure the imports worked in the env

tests/visualization/test_open3d_vis_utils.py
johnwlambert: That's fair. I used a different sized point cloud for each though, since 10k is too many for the sphere rendering (way too slow) but no problem for the point rendering

travisdriver: You could just create one point cloud and subsample as needed.

tests/visualization/test_open3d_vis_utils.py
```diff
@@ -0,0 +1,35 @@
+"""
+Unit tests for Open3d visualizaation code.
```
travisdriver, line 2: Make this inline with `"""`.

travisdriver:
Could we also add mayavi to the environment since it's also reference in `view_scene.py`?

tests/visualization/test_open3d_vis_utils.py
johnwlambert: I added a random seed. Although these viz tests should work for any amount of randomness : - )

tests/visualization/test_open3d_vis_utils.py
johnwlambert: thanks, updated

tests/visualization/test_open3d_vis_utils.py
```diff
@@ -0,0 +1,36 @@
+"""Unit tests for Open3d visualizaation code.
```
akshay-krishnan, line 1: typo in visualization.

akshay-krishnan (approved):

travisdriver:
It seems that the wheels for mayavi for python 3.8 are broken due to a circular dependency with VTK (at least for Linux). Could we just remove mayavi? I think open3d looks better, anyways, and especially since mayavi isn't included in the environment currently.

tests/visualization/test_open3d_vis_utils.py
johnwlambert: That's true. Not sure if it makes a big difference though, i think it's ok.

johnwlambert:
> It seems that the wheels for mayavi for python 3.8 are broken due to a circular dependency with VTK (at least for Linux). Could we just remove mayavi? I think open3d looks better, anyways, and especially since mayavi isn't included in the environment currently.

@travisdriver Sure, I'm fine with removing mayavi. The functionality is about the same, although mayavi can draw spheres faster than Open3d. Could you open an issue on the mayavi github page about the circular dependency you saw? I think it's a bug they just introduced, I can't seem to load it anymore either now.

tests/visualization/test_open3d_vis_utils.py
johnwlambert: thanks, fixed now

Issue #244, opened 2021-09-01
Improving data association 


akshay-krishnan:
Data association is the association of features into tracks before bundle adjustment. Currently this is done with Union-find. 

We currently observe the following problems with data association: 
- Too many small tracks
- Wrong correspondences may cause two different tracks to be merged into one track. 
- Once a track is triangulated, if the mean reprojection error is high, all observations are discarded. 
- The reprojection error threshold used to make this work is too high. 
- We dont refine tracking with better poses after bundle adjustment. 

Creating a high level issue to address these problems. Feel free to add to the list any discrepancies you observe and add yourself to assignees when contributing. :)

akshay-krishnan:
Initial directions to investigate: 

- Force tracks to have minimum length 3 ()
- Remove only outliers from the track after triangulation. 
- Try creating more tracks from outliers and merging them after bundle adjustment.

travisdriver:
[#281](https://github.com/borglab/gtsfm/pull/281) to address 3.

travisdriver:
- [ ] Use RANSAC _by default_ in triangulation
- [ ] Reduce reprojection error threshold <100 in triangulation

PR #281, opened 2021-09-08
Reject individual tracks by reprojection error

travisdriver:
Rejects individual tracks based on reprojection error as opposed to rejecting the entire frame if one measurement exceeds to reprojection error threshold as per [#244](https://github.com/borglab/gtsfm/issues/244).


johnwlambert:
Thanks for the PR, @travisdriver. How does this change impact the metrics on our different benchmarks?

travisdriver:
I'll just summarize the change in metrics for skydio-32 since they all seem to follow the same pattern. Seems as though almost all the metrics are improved, most notably in translation and rotation error, except for a marginal increase in reprojection error after filtering. The small increase in filtered reprojection error is not present in the other datasets.

# Bundle Adjustment Metrics

## Before

Metric name | Value
-- | --
number_cameras | 32
number_tracks_unfiltered | 5036
mean_3d_track_lengths_unfiltered | 2.66044
mean_reprojection_errors_unfiltered_px | 7.0951
mean_rotation_error_angle_deg | 9.41532
mean_translation_error_distance | 0.596332
mean_translation_angle_error_deg | 9.66996
number_tracks_filtered | 2211
mean_3d_track_lengths_filtered | 2.10312
mean_reprojection_errors_filtered_px | 1.28554

## After

Metric name | Value
-- | --
number_cameras | 32
number_tracks_unfiltered | 5634
mean_3d_track_lengths_unfiltered | 2.74938
mean_reprojection_errors_unfiltered_px | 8.27895
mean_rotation_error_angle_deg | 8.16042
mean_translation_error_distance | 0.501854
mean_translation_angle_error_deg | 8.5187
number_tracks_filtered | 2221
mean_3d_track_lengths_filtered | 2.15038
mean_reprojection_errors_filtered_px | 1.30445

dellaert:
Awesome!Q

gtsfm/data_association/point3d_initializer.py
```diff
@@ -210,9 +210,9 @@ def triangulate(self, track_2d: SfmTrack2d) -> Tuple[Optional[SfmTrack], Optiona
             self.track_camera_dict, triangulated_pt, inlier_track.measurements
         )
 
-        # Check that all the measurements have reprojection error < threshold.
-        # TODO(johnwlambert): compare with approach where we only throw away the outlier measurements.
-        if not np.all(reproj_errors < self.reproj_error_thresh):
+        # Filter measurements by the reprojection error.
+        inlier_track = inlier_track.select_subset(np.where(reproj_errors.flatten() < self.reproj_error_thresh)[0])
+        if inlier_track.number_measurements() < 2:
```
akshay-krishnan, line 215: Should this not be the min track length parameter? Why 2?

travisdriver: Min track length is checked outside of the point3d_initializer. See [#260](https://github.com/borglab/gtsfm/pull/260).

travisdriver: A check for min track length can be added to the point3d_initializer, but I think that's more in the scope of the other PR.

akshay-krishnan: Yeah I think we need to pass that parameter here.

akshay-krishnan (approved):

tests/data_association/test_point3d_initializer.py
```diff
@@ -147,10 +147,8 @@ def testSimpleTriangulationWithOneMeasurement(self):
         self.assertTrue(self.__runWithOneMeasurement(self.simple_triangulation_initializer))
 
     def testSimpleTriangulationWithOutlierMeasurements(self):
-
         sfm_track, _, _ = self.simple_triangulation_initializer.triangulate(SfmTrack2d(get_track_with_one_outlier()))
-
-        self.assertIsNone(sfm_track)
+        assert sfm_track.number_measurements() == 7
```
johnwlambert, line 151: thanks for the PR, Travis. Could we add an inline comment explaining why we expect 7 tracks here?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -210,9 +210,9 @@ def triangulate(self, track_2d: SfmTrack2d) -> Tuple[Optional[SfmTrack], Optiona
             self.track_camera_dict, triangulated_pt, inlier_track.measurements
         )
 
-        # Check that all the measurements have reprojection error < threshold.
-        # TODO(johnwlambert): compare with approach where we only throw away the outlier measurements.
-        if not np.all(reproj_errors < self.reproj_error_thresh):
+        # Filter measurements by the reprojection error.
```
johnwlambert, line 213: maybe these lines below are more readable on 2 lines? like 
```python
valid_reproj_error_idxs = np.where(reproj_errors.flatten() < self.reproj_error_thresh)[0]
inlier_track = inlier_track.select_subset(valid_reproj_error_idxs)
```
maybe there's a better name for this. like instead of `valid_reproj_error_idxs`, something like `idxs_reproj_err_satisfied`?

tests/data_association/test_point3d_initializer.py
travisdriver: Done

gtsfm/data_association/point3d_initializer.py
travisdriver: Done

johnwlambert:
@travisdriver I took another look at the metrics for this PR:

skydio - 8

Bundle Adjustment Metrics

| Metric name	| Value Before  | Value After
| number_cameras |	8 | 8 | 
| number_tracks_unfiltered |	1704  | 2024 | 
| mean_3d_track_lengths_unfiltered |	2.59683 | 2.53211 | 
| mean_reprojection_errors_unfiltered_px |	3.39898 | 5.1279 | 
| mean_rotation_error_angle_deg |	0.63194 | 0.876624 | 
| mean_translation_error_distance |	2.15029 | 2.12615 | 
| mean_translation_angle_error_deg |	4.53636 | 5.70282 | 
| number_tracks_filtered |	1162 | 1036 | 
| mean_3d_track_lengths_filtered |	2.28055 | 2.2249 | 
| mean_reprojection_errors_filtered_px |	0.983117 | 1.06239 | 


skydio - 32

Bundle Adjustment Metrics

| Metric name |	Value Before | Value After |  
| number_cameras |	32 | 32 | 
| number_tracks_unfiltered |	5724 | 5621 | 
| mean_3d_track_lengths_unfiltered |	2.67313 | 2.77282 | 
| mean_reprojection_errors_unfiltered_px |	10.9595 | 73.9129 | 
| mean_rotation_error_angle_deg |	7.326 | 7.79943 | 
| mean_translation_error_distance |	0.651919| 0.578229 | 
| mean_translation_angle_error_deg |	11.4042 | 10.424 | 
| number_tracks_filtered |	1256 | 1113 | 
| mean_3d_track_lengths_filtered |	2.08758 | 2.07637 | 
| mean_reprojection_errors_filtered_px | 1.39266  | 1.45703 | 



I wonder if we should re-triangulate again after removing the outliers? Is there any way you could visualize what the outlier patches look like for the elements that are removed from tracks, for this PR?

johnwlambert:
It seems like 
mean_reprojection_errors_unfiltered_px  goes from 10.9595 -> 73.9129 which is confusing to me.

travisdriver:
> It seems like
> mean_reprojection_errors_unfiltered_px goes from 10.9595 -> 73.9129 which is confusing to me.

I did not see this when I originally posted metrics for this PR, which makes me think it was something else that was merged into master in between...

johnwlambert:
Hmm could be, let me check. 

I think it would be a good idea to look at the individual track patches regardless though : - ) i think that's the only way to make sense of how the inlier/outlier classification is going.

travisdriver:
You can check the commit before I merged in master this morning and the mean reprojection error was not 73.

PR #288, opened 2021-09-11
Using robust Huber loss in Translation Averaging

akshay-krishnan:
The metrics on skydio-32 show a clear improvement: 

Before the change:

|Metric | Value|
|---|---|
|mean_translation_angle_error_deg | 16.2499|
|mean_translation_error_distance | 0.98776 |

After the change:

|Metric | Value|
|---|---|
|mean_translation_angle_error_deg | 10.3671|
|mean_translation_error_distance | 0.546729 |

There are also small improvements after bundle adjustment. The improvement is similar on skydio-8, not too significant on door. 
I see from the plots that the inlier errors significantly decrease, but large outliers are not quite affected, which is expected.  

[CI run for plots](https://github.com/borglab/gtsfm/actions/runs/1223467854)

### DIFF
--- a/gtsfm/averaging/translation/averaging_1dsfm.py
+++ b/gtsfm/averaging/translation/averaging_1dsfm.py
@@ -48,6 +48,7 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
         i2Ui1_dict: Dict[Tuple[int, int], Optional[Unit3]],
         wRi_list: List[Optional[Rot3]],
         scale_factor: float = 1.0,
+        robust_measurement_noise: bool = True,
         gt_wTi_list: Optional[List[Optional[Pose3]]] = None,
     ) -> Tuple[List[Optional[Point3]], Optional[GtsfmMetricsGroup]]:
         """Run the translation averaging.
@@ -57,6 +58,7 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
             i2Ui1_dict: relative unit-translation as dictionary (i1, i2): i2Ui1
             wRi_list: global rotations for each camera pose in the world coordinates.
             scale_factor: non-negative global scaling factor.
+            robust_measurement_noise: Whether to use Huber noise model for the measurements, defaults to true.
             gt_wTi_list: ground truth poses for computing metrics.
 
         Returns:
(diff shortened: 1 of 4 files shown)


johnwlambert:
Looks awesome @akshay-krishnan. Can we make it a default flag set to true to use_robust in the translation averaging? That way in ablation tables we can just flip the switch

akshay-krishnan:
> Looks awesome @akshay-krishnan. Can we make it a default flag in the config set to true to use_robust in the translation averaging? That way in ablation tables we can just flip the switch

Yeah, but we would have change the base class API right? And this is something implementation-specific. What if there is a translation averaging implementation which does not need that flag?

johnwlambert:
> > Looks awesome @akshay-krishnan. Can we make it a default flag in the config set to true to use_robust in the translation averaging? That way in ablation tables we can just flip the switch
> 
> Yeah, but we would have change the base class API right? And this is something implementation-specific. What if there is a translation averaging implementation which does not need that flag?
> Do we have any implementation-specific flags in other classes?

That's true. But I still think it's a good idea. Could be an optional flag that defaults to False.
I like how Ayush did it here: https://github.com/borglab/gtsfm/pull/70/files

Maybe for some implementations, making it robust is tricky, so we would keep default false.

akshay-krishnan:
> > > Looks awesome @akshay-krishnan. Can we make it a default flag in the config set to true to use_robust in the translation averaging? That way in ablation tables we can just flip the switch
> > 
> > 
> > Yeah, but we would have change the base class API right? And this is something implementation-specific. What if there is a translation averaging implementation which does not need that flag?
> > Do we have any implementation-specific flags in other classes?
> 
> That's true. But I still think it's a good idea. Could be an optional flag that defaults to False.
> I like how Ayush did it here: https://github.com/borglab/gtsfm/pull/70/files
> 
> Maybe for some implementations, making it robust is tricky, so we would keep default false.

Not exactly the same case, since there is no bundle adjustment base class. But sure, I can add a flag. 
What I mean is, the flag might be meaningless for some other implementation, it's a base class argument that is implementation-specific, which is not very clean API-wise.

akshay-krishnan:
Done, and that will also be my first commit using github.dev :)

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -64,7 +67,12 @@ def run(
                 or ill-constrained system).
             A GtsfmMetricsGroup of 1DSfM metrics.
         """
-        noise_model = gtsam.noiseModel.Isotropic.Sigma(NOISE_MODEL_DIMENSION, NOISE_MODEL_SIGMA)
+        isotropic_noise_model = gtsam.noiseModel.Isotropic.Sigma(NOISE_MODEL_DIMENSION, NOISE_MODEL_SIGMA)
```
ayushbaid, line 70: just rename the first variable as noise_model, and overwrite it if the flag is true?

ayushbaid (approved):
LGTM except one minor comment.

ayushbaid:
Redacting approval because John's comment needs to be addressed.

akshay-krishnan:
> should this line be
> 
> ```python
> noise_model = gtsam.noiseModel.Robust.Create(huber_loss, isotropic_noise_model)
> ```
> 
> ?

Good catch, too bad my first github.dev commit was buggy :/

johnwlambert:
Looks really good Akshay. Can we make it an argument of the constructor instead though? And not extra arg to run? That way we can flip switch in config. Can we put it True in the config too?

dellaert (approved):
This is awesome!

johnwlambert (approved):
Nice work, thanks Akshay

PR #292, opened 2021-09-13
Enable RANSAC in triangulation

travisdriver:
Enabled RANSAC in the triangulation module.


travisdriver:
The results on Skydio-32 look promising (will post other results once the benchmarking finishes):

# Before

Metric name | Value
-- | --
number_cameras | 32
number_tracks_unfiltered | 5724
mean_3d_track_lengths_unfiltered | 2.67313
mean_reprojection_errors_unfiltered_px | 10.9595
mean_rotation_error_angle_deg | 7.326
mean_translation_error_distance | 0.651919
mean_translation_angle_error_deg | 11.4042
number_tracks_filtered | 1256
mean_3d_track_lengths_filtered | 2.08758
mean_reprojection_errors_filtered_px | 1.39266

# After

Metric name | Value
-- | --
number_cameras | 32
number_tracks_unfiltered | 3482
mean_3d_track_lengths_unfiltered | 2.04451
mean_reprojection_errors_unfiltered_px | 0.669101
mean_rotation_error_angle_deg | 6.7561
mean_translation_error_distance | 0.920434
mean_translation_angle_error_deg | 12.0292
number_tracks_filtered | 3445
mean_3d_track_lengths_filtered | 2.04325
mean_reprojection_errors_filtered_px | 0.641061

johnwlambert:
Thanks for the PR, Travis. Cool to see. 

I'm curious about why the average camera distance from GT goes from 0.65 to 0.9. Could you try sharing what the visualizations look like before and after the change?

johnwlambert:
Travis, is there any way you could share some examples of how RANSAC is doing? Like for some visualizations of tracks (using our DA track viz script that dumps vertically stacked patches), could we see which ones RANSAC is labeling as outliers? Do the outliers look correct in practice? I think having GT metrics here on 3d GT will help too : - )

travisdriver:
> Travis, is there any way you could share some examples of how RANSAC is doing? Like for some visualizations of tracks (using our DA track viz script that dumps vertically stacked patches), could we see which ones RANSAC is labeling as outliers? Do the outliers look correct in practice? I think having GT metrics here on 3d GT will help too : - )

Which GT metrics are you referring to?

johnwlambert:
> > Travis, is there any way you could share some examples of how RANSAC is doing? Like for some visualizations of tracks (using our DA track viz script that dumps vertically stacked patches), could we see which ones RANSAC is labeling as outliers? Do the outliers look correct in practice? I think having GT metrics here on 3d GT will help too : - )
> 
> Which GT metrics are you referring to?

Oh I meant the track precision/recall from the 3d model -- I think we might need a bit more DA metrics before merging this DA PR

travisdriver:
@johnwlambert Once you land #385, feel free to close this and open a fresh branch with the same changes.

johnwlambert:
How does the dashboard look for this PR Travis? Could you share the file if you have created it?

travisdriver:
Here are the results for 100 RANSAC iterations with a 5 pixel threshold.

I think door-12 may be so "converged" that small perturbations in the results make it look like it's doing really bad in the dashboard. For example, the ~1000% increase in translation error is a result of it going from 0.002 to 0.02. Nevertheless, the results are still not great.

I still think there's something peculiar going on. I would have at least expected the number of unfiltered tracks to increase (since were no longer rejecting entire tracks), but this is not the case. Also, the mean track length seems to have decreased substantially in most cases, suggesting that RANSAC is rejecting a significant amount of track correspondences (maybe we need to increase the pixel threshold).

I will investigate and debug. 

image

travisdriver:
Actually I guess I would expect the number of unfiltered tracks to decrease since the default reprojection error threshold is 100 pixels on `master`...

travisdriver:
Here are the results with a RANSAC threshold of 100 pixels. 

My hypothesis is that the poses from the averaging module are not accurate enough to perform RANSAC at more reasonable thresholds (this is supported by the huge performance increase afforded by #225).

newplot (1)

johnwlambert:
Hi @travisdriver, could you post the dashboard to show the improvement on AstroNet with these changes?

travisdriver:
newplot (7)

The decrease in performance for **Skydio-32** is likely just noise, as it has already converged, i.e., the decrease in the metrics is on the order of .1 and .01.

**Skydio-8** performed poorly on the CI but looked perfect on my machine; this is likely a symptom of #323. Here is the final reconstruction from running Skydio-8 locally:

Screenshot from 2021-12-15 00-02-29

### Skydio-8 metrics **w/ RANSAC** from local run
Screenshot from 2021-12-15 09-58-14

### Skydio-8 metrics **w/o RANSAC** from local run
Screenshot from 2021-12-15 10-01-44

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,39 @@ class TriangulationParam(Enum):
...
+    """Options for triangulation RANSAC solver."""
+
+    reproj_error_threshold: float
+    min_track_len: int
+    mode: TriangulationParam
+
+    # RANSAC parameters
+    min_inlier_ratio: float = 0.5
+    confidence: float = 0.9999
+    dyn_num_hypotheses_multiplier: float = 3.0
```
johnwlambert, line 69: Hi @travisdriver, thanks for the PR. Could you add docstrings explaining each of these params, e.g. `dyn_num_hypotheses_multiplier`?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,39 @@ class TriangulationParam(Enum):
...
+    def num_ransac_hypotheses(self) -> int:
+        """Compute maximum number of hypotheses."""
+        self.__check_ransac_params()
+        dyn_num_hypotheses = int(
+            np.log(1 - self.confidence)
+            / np.log(1 - (1 - self.min_inlier_ratio) ** NUM_SAMPLES_PER_RANSAC_HYPOTHESIS)
+            * self.dyn_num_hypotheses_multiplier
+        )
+        return min(self.max_num_hypotheses, dyn_num_hypotheses)
+        # return self.max_num_hypotheses
```
johnwlambert, line 89: commented out code -- can it be removed?

gtsfm/data_association/data_assoc.py
```diff
@@ -91,13 +92,12 @@ def run(
...
-        point3d_initializer = Point3dInitializer(
-            cameras,
-            self.mode,
-            self.reproj_error_thresh,
-            self.num_ransac_hypotheses,
-        )
+        # Initialize 3D landmark for each track
+        point3d_initializer = Point3dInitializer(cameras, self.triangulation_options)
+        #     self.mode,
+        #     self.reproj_error_thresh,
```
johnwlambert, line 98: commented-out code -- can it be removed?

gtsfm/configs/deep_front_end_astronet.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+    data_association_module:
+      _target_: gtsfm.data_association.data_assoc.DataAssociation
+      triangulation_options:
+        _target_: gtsfm.data_association.point3d_initializer.TriangulationOptions
+        reproj_error_threshold: 100
+        min_track_len: 2
+        mode:
+          _target_: gtsfm.data_association.point3d_initializer.TriangulationParam
+          value: 1 # 0 corresponds to NO_RANSAC
```
johnwlambert, line 55: could we add here what a value of `1` corresponds to, instead of what `0` corresponds to?

gtsfm/configs/deep_front_end_astronet.yaml
```diff
@@ -0,0 +1,63 @@
...
+        max_num_hypotheses: 100
+      save_track_patches_viz: False
```
johnwlambert, line 57: @travisdriver does the RANSAC triangulation setting hurt the other benchmarks (besides AstroNet)? If so, do we know why?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,39 @@ class TriangulationParam(Enum):
...
+
+    def __check_ransac_params(self) -> None:
+        """Check that input value are valid"""
+        assert self.reproj_error_threshold > 0
+        assert 0 < self.min_inlier_ratio < 1
+        assert 0 < self.confidence < 1
+        assert self.dyn_num_hypotheses_multiplier > 0
+        assert self.max_num_hypotheses > 0
+
+    def num_ransac_hypotheses(self) -> int:
```
johnwlambert, line 80: @travisdriver could we add a unit test on this?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,39 @@ class TriangulationParam(Enum):
...
 
+class TriangulationOptions(NamedTuple):
+    """Options for triangulation RANSAC solver."""
+
+    reproj_error_threshold: float
+    min_track_len: int
+    mode: TriangulationParam
+
+    # RANSAC parameters
+    min_inlier_ratio: float = 0.5
```
johnwlambert, line 67: I'm more accustomed to "inlier_ratio" rather than "min_inlier_ratio" when computing the number of RANSAC hypotheses -- what does `min_inlier_ratio` mean here?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -85,14 +120,15 @@ def execute_ransac_variant(self, track_2d: SfmTrack2d) -> np.ndarray:
             best_inliers: boolean array of length N. Indices of measurements
                are set to true if they correspond to the best RANSAC hypothesis
         """
-        if self.num_ransac_hypotheses is None:
-            raise ValueError("RANSAC triangulation requested but number of hypothesis is None.")
+        # if self.num_ransac_hypotheses is None:
```
johnwlambert, line 123: just curious -- why do we not need this check anymore?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -85,14 +120,15 @@ def execute_ransac_variant(self, track_2d: SfmTrack2d) -> np.ndarray:
...
+        #     raise ValueError("RANSAC triangulation requested but number of hypothesis is None.")
 
         # Generate all possible matches
         measurement_pairs = generate_measurement_pairs(track_2d)
 
         # limit the number of samples to the number of available pairs
-        num_hypotheses = min(self.num_ransac_hypotheses, len(measurement_pairs))
+        num_hypotheses = min(self.options.num_ransac_hypotheses(), len(measurement_pairs))
+        logger.debug(num_hypotheses)
```
johnwlambert, line 131: Would be helpful to make this log message more descriptive, e.g. 
```python
logger.debug("Num RANSAC hypotheses for triangulation: %d", num_hypotheses)
```

gtsfm/configs/deep_front_end.yaml
```diff
@@ -46,12 +46,14 @@ SceneOptimizer:
...
-        _target_: gtsfm.data_association.data_assoc.TriangulationParam
-        value: 0 # 0 corresponds to NO_RANSAC
-      num_ransac_hypotheses: 20
+      triangulation_options:
+        _target_: gtsfm.data_association.point3d_initializer.TriangulationOptions
+        reproj_error_threshold: 10
+        min_track_len: 2
+        mode:
+          _target_: gtsfm.data_association.point3d_initializer.TriangulationParam
+          value: 1 # 0 corresponds to NO_RANSAC
```
johnwlambert, line 55: same here -- would be nice to explain what 1 corresponds to. If we are using RANSAC in this config, what do we need the new `deep_front_end_astronet` config for?

gtsfm/configs/deep_front_end_astronet.yaml
travisdriver: The RANSAC triangulation doesn't seem to be hurting any of the relevant benchmarks: see [this comment](https://github.com/borglab/gtsfm/pull/292#issuecomment-999275501). 

It significantly improves AstroNet with a more appropriate threshold of 100 pixels (since the averaged poses are still not very good). It's all a function of the accuracy of the averaged poses.

gtsfm/data_association/point3d_initializer.py
travisdriver: This was just for debugging purposes; I will remove it.

gtsfm/configs/deep_front_end.yaml
travisdriver: AstroNet performs better with a higher RANSAC threshold (100 pixels as opposed to 10) due to the quality of the averaged poses.

gtsfm/data_association/point3d_initializer.py
travisdriver: Because the checks are performed by `__check_ransac_params()` when `num_ransac_hypotheses()` is called.

gtsfm/data_association/point3d_initializer.py
travisdriver: Yes, the name is actually misleading (I was following COLMAP's implementation). It's actually the assumed outlier ratio, i.e., a higher value will result in more iterations in order to achieve the defined confidence.

I will change the name and add a description to the docstring.

johnwlambert: Thanks

gtsfm/configs/deep_front_end_astronet.yaml
```diff
@@ -0,0 +1,63 @@
+SceneOptimizer:
```
johnwlambert, line 1: maybe we could add a comment to the top of this yaml file, explaining that the only change w.r.t. `deep_front_end.yaml` is the RANSAC triangulation reproj threshold (10x looser)

travisdriver: added

gtsfm/configs/deep_front_end_astronet.yaml
travisdriver: added

gtsfm/data_association/data_assoc.py
travisdriver: removed

gtsfm/data_association/point3d_initializer.py
travisdriver: added

gtsfm/data_association/point3d_initializer.py
travisdriver: added

gtsfm/data_association/point3d_initializer.py
travisdriver: removed

gtsfm/data_association/point3d_initializer.py
travisdriver: removed

johnwlambert:
Thanks for the PR, Travis. I'm taking a look at the report now:

[visual_comparison_dashboard 3.html.zip](https://github.com/borglab/gtsfm/files/7810261/visual_comparison_dashboard.3.html.zip)

Just trying to figure out which changes are sort of random changes vs. deterministic ones. I think you have a better sense from me from having run the CI a few times on this branch.

Maybe if we fix the determinism bug first it will be a bit clearer. What are your thoughts?

johnwlambert:
Hi @travisdriver the PR looks good. Do you have a sense for the new runtime? I'm curious if it's 1% slower, 10% slower, or 100%+ slower with the new RANSAC loop.

travisdriver:
> Hi @travisdriver the PR looks good. Do you have a sense for the new runtime? I'm curious if it's 1% slower, 10% slower, or 100%+ slower with the new RANSAC loop.

I'm not sure what you're referring to; there's no increase in runtime according to the CI.

johnwlambert:
> > Hi @travisdriver the PR looks good. Do you have a sense for the new runtime? I'm curious if it's 1% slower, 10% slower, or 100%+ slower with the new RANSAC loop.
> 
> I'm not sure what you're referring to; there's no increase in runtime according to the CI.

I was thinking since we were adding a new inner Python for-loop for our 2-view geometry estimation, so that could slow things down considerably. But I guess since we limit to just 100 hypotheses the effect on runtime is neglible?

travisdriver:
> > > Hi @travisdriver the PR looks good. Do you have a sense for the new runtime? I'm curious if it's 1% slower, 10% slower, or 100%+ slower with the new RANSAC loop.
> > 
> > 
> > I'm not sure what you're referring to; there's no increase in runtime according to the CI.
> 
> I was thinking since we were adding a new inner Python for-loop for our 2-view geometry estimation, so that could slow things down considerably. But I guess since we limit to just 100 hypotheses the effect on runtime is neglible?

Yes, triangulation is a relatively inexpensive operation.

travisdriver:
@johnwlambert What else needs to be done in order to merge this PR?

johnwlambert:
Hi Travis, thanks for your patience here. I'm a bit confused about the performance on skydio-8 and skydio-32. I triggered the CI one more time to see if those regressions were random. @akshay-krishnan @ayushbaid could you all take a look? I'm fine with landing the new config just for astronet right away, though. 
[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/7893074/visual_comparison_dashboard.html.zip)

I saw that the runtime went from 58 min to 2 hours 20 min just for one benchmark (deep front end RC3). I'm running again to see if that's repeatable.
Screen Shot 2022-01-18 at 6 41 26 PM

johnwlambert:
I checked the dashboard for the second run, and still seeing some confusing things. Hopefully if we land the repeatability stuff first, it will be more clear for us
Screen Shot 2022-01-19 at 9 34 39 AM

travisdriver:
> I saw that the runtime went from 58 min to 2 hours 20 min just for one benchmark (deep front end RC3). I'm running again to see if that's repeatable.

This is from the most recent commit on master; the benchmark runtime fluctuates a lot for some reason, but there are plenty of instances other than this branch where is takes >2 hours.

Screenshot from 2022-01-19 09-34-40

johnwlambert:
> > I saw that the runtime went from 58 min to 2 hours 20 min just for one benchmark (deep front end RC3). I'm running again to see if that's repeatable.
> 
> This is from the most recent commit on master; the benchmark runtime fluctuates a lot for some reason, but there are plenty of instances other than this branch where is takes >2 hours.
> 
> Screenshot from 2022-01-19 09-34-40

I see, cool. Didn't realize runtime was fluctuating that much too. I wonder if the hardware specs are different between runs on the CI

travisdriver:
> I checked the dashboard for the second run, and still seeing some confusing things. Hopefully if we land the repeatability stuff first, it will be more clear for us 

I don't think an increase in angular error from .3 to .8 should completely hold the PR, especially when there are significant improvements across the board. The repeatability issues effect all PRs, but it seems to only be holding up this one...



> I checked the dashboard for the second run, and still seeing some confusing things. Hopefully if we land the repeatability stuff first, it will be more clear for us Screen Shot 2022-01-19 at 9 34 39 AM

Thanks for looking at the benchmarks. What's confusing from the output? A .5 degree increase in the mean rotation angle error seems insignificant compared to the other improvements.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,54 @@ class TriangulationParam(Enum):
...
+        assert self.dyn_num_hypotheses_multiplier > 0
+        assert 0 <= self.min_num_hypotheses < self.max_num_hypotheses
+
+    def num_ransac_hypotheses(self) -> int:
+        """Compute maximum number of hypotheses."""
+        self.__check_ransac_params()
+        dyn_num_hypotheses = int(
+            np.log(1 - self.confidence)
+            / np.log(1 - self.min_inlier_ratio ** NUM_SAMPLES_PER_RANSAC_HYPOTHESIS)
+            * self.dyn_num_hypotheses_multiplier
```
johnwlambert, line 100: Hi Travis, could you add a reference for this dynamic multiplier? If COLMAP is using it, do you mind adding an inline comment w/ URL to the COLMAP source where they do it?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,54 @@ class TriangulationParam(Enum):
...
+        mode: triangulation mode, which dictates whether or not to use robust estimation.
+        min_inlier_ratio: a priori assumed minimum probability that a point is an inlier.
+        confidence: desired confidence that at least one hypothesis is outlier free.
+        dyn_num_hypotheses_multiplier: multiplication factor for dynamically computed hyptheses based on confidence.
+        min_num_hypotheses: minimum number of hypotheses.
+        mas_num_hypotheses: masimum number of hypotheses.
+    """
+
+    reproj_error_threshold: float
+    mode: TriangulationParam
```
akshay-krishnan, line 77: I feel like TriangulationParam is not a good name for this enum. It means the same thing as `TriangulationOptions` but the enum is about what type of ransac to run. Maybe RansacMode or something is better.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,54 @@ class TriangulationParam(Enum):
...
+
+    # RANSAC parameters
+    min_inlier_ratio: float = 0.1
```
akshay-krishnan, line 80: isnt that a really small ratio? do we have a min_num_inliers too? just to be sure this does not select tracks with one or two inliers?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,54 @@ class TriangulationParam(Enum):
...
+
+    def num_ransac_hypotheses(self) -> int:
+        """Compute maximum number of hypotheses."""
+        self.__check_ransac_params()
+        dyn_num_hypotheses = int(
+            np.log(1 - self.confidence)
+            / np.log(1 - self.min_inlier_ratio ** NUM_SAMPLES_PER_RANSAC_HYPOTHESIS)
+            * self.dyn_num_hypotheses_multiplier
+        )
+        num_hypotheses = min(self.max_num_hypotheses, dyn_num_hypotheses)
```
akshay-krishnan, line 102: `num_hypotheses = max(min(self.max_num_hypotheses, dyn_num_hypotheses), self.min_num_hypothesis)` ?

gtsfm/data_association/point3d_initializer.py
akshay-krishnan, line 100: this works, but just for super clear readability could you add an additional pair of parenthesis? (A / B) * C instead of A / B * C

gtsfm/data_association/point3d_initializer.py
johnwlambert: I'm ok with it currently. I see what you mean Akshay, but I think `RansacMode.NO_RANSAC` would be sort of weird.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,54 @@ class TriangulationParam(Enum):
...
+      - http://www.cse.psu.edu/~rtc12/CSE486/lecture15.pdf
+
+    Args:
+        reproj_error_threshold: the maximum reprojection error allowed.
+        mode: triangulation mode, which dictates whether or not to use robust estimation.
+        min_inlier_ratio: a priori assumed minimum probability that a point is an inlier.
+        confidence: desired confidence that at least one hypothesis is outlier free.
+        dyn_num_hypotheses_multiplier: multiplication factor for dynamically computed hyptheses based on confidence.
+        min_num_hypotheses: minimum number of hypotheses.
+        mas_num_hypotheses: masimum number of hypotheses.
```
johnwlambert, line 73: smal typo here: "max" and "maximum"

gtsfm/data_association/point3d_initializer.py
johnwlambert: Maybe `SamplingMode` is better?

gtsfm/data_association/point3d_initializer.py
johnwlambert: marking as resolved

gtsfm/data_association/point3d_initializer.py
travisdriver: I like `SamplingMode`

gtsfm/data_association/point3d_initializer.py
travisdriver: COLMAP uses it, but all it does it multiply the number of dynamic hypotheses by a constant so I'm indifferent about whether it warrants a reference

gtsfm/data_association/point3d_initializer.py
travisdriver: Yes, a smaller ratio will result in a larger number of dynamic hypotheses.

Keep in mind that the measurements are the cameras (with the corresponding keypoint), so a track with one inlier would result in a underdetermined problem (which we check for and triangulation would fail anyways). Two inliers is allowed, but tracks are discarded according to `min_track_length`.

gtsfm/data_association/point3d_initializer.py
travisdriver: This is better thank you

gtsfm/data_association/point3d_initializer.py
johnwlambert: or `TriangulationSamplingMode`?

gtsfm/data_association/point3d_initializer.py
johnwlambert: i updated it, marking as resolved

gtsfm/data_association/point3d_initializer.py
johnwlambert: i updated it, marking as resolved

gtsfm/data_association/point3d_initializer.py
johnwlambert: i updated it, marking as resolved

gtsfm/configs/deep_front_end_astronet.yaml
```diff
@@ -0,0 +1,68 @@
+# Relaxes triangulation RANSAC threshold to 100 (10x threshold of deep_front_end.yaml)
```
ayushbaid, line 1: Can we name it something which mentions this attribute value being high instead of tagging it with a database?

gtsfm/configs/scene_optimizer_unit_test_config.yaml
```diff
@@ -44,12 +43,14 @@ SceneOptimizer:
...
-        _target_: gtsfm.data_association.data_assoc.TriangulationParam
-        value: 0 # 0 corresponds to NO_RANSAC
-      num_ransac_hypotheses: 20
+      triangulation_options:
+        _target_: gtsfm.data_association.point3d_initializer.TriangulationOptions
+        reproj_error_threshold: 10
+        mode:
+          _target_: gtsfm.data_association.point3d_initializer.TriangulationSamplingMode
+          value: 0 # 0 corresponds to NO_RANSAC
+        max_num_hypotheses: 20
```
ayushbaid, line 53: Why change to ransac in the unit test?

akshay-krishnan (approved):

gtsfm/data_association/point3d_initializer.py
akshay-krishnan: Okay, I wonder if a higher min inlier ratio produces better hypothesis? It would definitely save some compute.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -55,6 +56,53 @@ class TriangulationParam(Enum):
...
+    Reference: https://github.com/colmap/colmap/blob/dev/src/optim/ransac.h
+
+    See the following slides for a derivation of the #req'd samples: http://www.cse.psu.edu/~rtc12/CSE486/lecture15.pdf
+
+    Args:
+        reproj_error_threshold: the maximum reprojection error allowed.
+        mode: triangulation mode, which dictates whether or not to use robust estimation.
+        min_inlier_ratio: a priori assumed minimum probability that a point is an inlier.
+        confidence: desired confidence that at least one hypothesis is outlier free.
+        dyn_num_hypotheses_multiplier: multiplication factor for dynamically computed hyptheses based on confidence.
```
ayushbaid, line 72: Can you elaborate what this is in the PR comments?

Also good to have an example computation in the docstring for the number of hypothesis.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -70,9 +118,7 @@ class Point3dInitializer(NamedTuple):
     """
 
     track_camera_dict: Dict[int, PinholeCameraCal3Bundler]
-    mode: TriangulationParam
-    reproj_error_thresh: float
-    num_ransac_hypotheses: Optional[int] = None
+    options: TriangulationOptions
```
ayushbaid, line 121: can we rename to triangulation_options to prevent ambiguity?

ayushbaid (approved):
LGTM except few minor comments.

Do you want to wait for the repeatability issues in 1dsfm to land before using it in astronet, or do you want to start using it regardless?

gtsfm/configs/deep_front_end_astronet.yaml
johnwlambert: I agree, I think this would improve clarity

gtsfm/data_association/point3d_initializer.py
johnwlambert: I feel like maybe that would be too wordy/long. Just my two cents.

gtsfm/configs/scene_optimizer_unit_test_config.yaml
johnwlambert: I didn't quite follow here Ayush -- I believe we were using NO_RANSAC before, and the test is still using NO_RANSAC now.

gtsfm/data_association/point3d_initializer.py
johnwlambert: thanks, marking as resolved

gtsfm/data_association/point3d_initializer.py
johnwlambert: @travisdriver we could add the following math:
```python
The RANSAC module defaults to 2749 iterations, computed as 
np.log(1-0.9999) / np.log( 1 - 0.1 **2) * 3 = 2749.3
```

johnwlambert: I added this to the docstring

gtsfm/data_association/point3d_initializer.py
johnwlambert: I updated this, marking as resolved

johnwlambert (approved):
LGTM

gtsfm/data_association/point3d_initializer.py
travisdriver: I think just `options` is sufficient since there aren't any similar attributes.

gtsfm/configs/deep_front_end_astronet.yaml
travisdriver: AstroNet is the only dataset that uses the config by default, so it seems fitting to name it accordingly.

PR #295, opened 2021-09-14
AstroNet mesh frontend metrics

travisdriver:
This PR adds a front-end evaluation based on a provided ground truth surface mesh of the scene. High-fidelity ground truth shape models are available for small bodies explored by past missions, allowing for highly accurate evaluation of the correspondences computes by the front-end. 

Through testing, the current Sampson distance (SD) method seems to consistently over-estimate the precision of the front end, at least for the AstroNet examples I have run. For example, on a 14-image sequence from Dawn @ 4 Vesta, the _Sampson distance method_ estimated an overall precision of **.9998** while the _mesh-based method_ computed a precision of **.9388**. 

Moreover, I ran GTSfM using the ground truth information to verify correspondences using both the SD and mesh-based methods as a sanity check. When correspondences were verfied using the ground truth mesh, the final reconstruction was perfect. However, the result using the SD method was far from perfect.

I still think the Sampson distance is a useful metric for when a ground truth mesh is not present, but I think the mesh-based front-end evaluation provides a more accurate estimate of the precision

# Mesh-based method

mean_inlier_ratio_wrt_gt_model: 0.9998

### Metrics for pair (0, 8)
```
(0, 8)
"rotation_angular_error": 15.69,
"translation_angular_error": 8.21,
"num_inliers_gt_model": 376,
 "inlier_ratio_gt_model": 0.82,
"inlier_avg_reproj_error_gt_model": 1.86,
 "outlier_avg_reproj_error_gt_model": 7.34,
"inlier_ratio_est_model": 0.78,
 "num_inliers_est_model": 459

```

### Final reconstruction with ground truth verified correspondences

Screenshot from 2021-10-09 11-00-06

# Sampson distance method

mean_inlier_ratio_wrt_gt_model: 0.9388

### Metrics for pair (0, 8)
```
(0, 8)
"rotation_angular_error": 15.69,
"translation_angular_error": 8.21,
"num_inliers_gt_model": 459,
"inlier_ratio_gt_model": 1.0,
"inlier_avg_reproj_error_gt_model": 0.63,
"outlier_avg_reproj_error_gt_model": NaN,
"inlier_ratio_est_model": 0.78,
"num_inliers_est_model": 459

```

### Final reconstruction with ground truth verified correspondences

Screenshot from 2021-10-09 10-44-36


gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
...
+
+    Args:
+        keypoints_i1: N keypoints in image i1.
+        keypoints_i2: N corresponding keypoints in image i2.
+        intrinsics_i1: intrinsics for i1.
+        intrinsics_i2: intrinsics for i2.
+        gt_wTi1: ground truth pose of the world frame relative to i1.
+        gt_wTi2: ground truth pose of the world frame relative to i2.
+        gt_scene_mesh: ground truth triangular surface mesh of the scene in the world frame.
+        dist_threshold: max acceptable distance for a correct correspondence.
```
johnwlambert, line 144: this `dist_threshold` is a distance from what object to what object? is it a reprojection error?

gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
...
+def mesh_inlier_correspondences(
+    keypoints_i1: Keypoints,
+    keypoints_i2: Keypoints,
+    gt_camera_i1: PinholeCameraCal3Bundler,
+    gt_camera_i2: PinholeCameraCal3Bundler,
+    gt_scene_mesh: trimesh.Trimesh,
+    dist_threshold: float,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Compute inlier correspondences using the ground truth triangular surface mesh of the scene.
+
```
johnwlambert, line 135: i think a high-level description of the algorithm (at least 2-3 sentences) is needed in the docstring.

gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
...
+        )
+        drc_i2[corr_idx, :] = (
+            gt_camera_i2.BackprojectFromCamera(keypoints_i2.coordinates[corr_idx], depth=1.0)
+            - gt_camera_i2.translation()
+        )
+
+    # Perform ray tracing.
+    src = np.vstack((src_i1, src_i2))
+    drc = np.vstack((drc_i1, drc_i2))
+    logger.info("Computing ray intersections...")
```
johnwlambert, line 177: Nit: this logging could get expensive for 50K+ image pairs, like skydio-501 w/ 150 frame lookahead

visualization/view_scene.py
```diff
@@ -62,9 +62,9 @@ def view_scene(args: argparse.Namespace) -> None:
     # expression below is equivalent to applying zcwTw.transformFrom() to each world point
     point_cloud -= mean_pt
 
-    is_nearby = np.linalg.norm(point_cloud, axis=1) < args.max_range
-    point_cloud = point_cloud[is_nearby]
-    rgb = rgb[is_nearby]
+    # is_nearby = np.linalg.norm(point_cloud, axis=1) < args.max_range
```
johnwlambert, line 65: why the change here?

gtsfm/scene_optimizer.py
```diff
@@ -145,14 +146,24 @@ def create_computation_graph(
...
 
+            # Store results.
             i2Ri1_graph_dict[(i1, i2)] = i2Ri1
             i2Ui1_graph_dict[(i1, i2)] = i2Ui1
             v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
-
             two_view_reports_dict[(i1, i2)] = two_view_report
 
+            # # Use ground truth relative poses.
+            # from gtsam import Unit3
```
johnwlambert, line 161: looks like this should be removed before we check it in

gtsfm/scene_optimizer.py
```diff
@@ -103,6 +104,8 @@ def create_computation_graph(
         camera_intrinsics_graph: List[Delayed],
         image_shape_graph: List[Delayed],
         gt_cameras_graph: Optional[List[Delayed]] = None,
+        gt_pose_graph: Optional[List[Delayed]] = None,
+        gt_scene_mesh: Optional[trimesh.Trimesh] = None,
```
johnwlambert, line 108: this is a big change to the SceneOptimizer. I really would like to see some examples of when Sampson fails and when we need this, before we check it in for all datasets (since we only have a mesh for 1 of 9 benchmark runs in the CI currently)

gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
         raise ValueError("Keypoints must have same counts")
 
     if len(keypoints_i1) == 0:
-        return 0
+        return None
+
+    # Compute ground truth correspondences.
+    if None not in [gt_scene_mesh, gt_wTi1, gt_wTi2]:
```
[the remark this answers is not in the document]
travisdriver: Not sure if I agree that what you suggested is cleaner.

There are different prints in the terminal based on which on is run. I'm not really sure it makes sense to put it in the config as it's not configurable. You either had the required ground truth information information or the metrics cannot be computed.

gtsfm/utils/metrics.py
travisdriver: the same as `count_correct_correspondences`

gtsfm/scene_optimizer.py
travisdriver: There are examples on the last weeks presentation. I can add more benchmarks if you'd like.

I'm open to suggestions on how to run this evaluation without having top pass anything to `SceneOptimizer`

gtsfm/utils/metrics.py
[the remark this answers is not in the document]
johnwlambert: Yeah maybe that syntax fix is not cleaner actually.

@akshay-krishnan and @ayushbaid what are your thoughts on how to configure the eval options?

gtsfm/scene_optimizer.py
johnwlambert: Cool, I guess we didn't get to discuss those yet. Could you add them to the PR description?

travisdriver: Added

gtsfm/scene_optimizer.py
```diff
@@ -145,14 +146,24 @@ def create_computation_graph(
...
-
             two_view_reports_dict[(i1, i2)] = two_view_report
 
+            # # Use ground truth relative poses.
+            # from gtsam import Unit3
+
+            # i2Ri1_graph_dict[(i1, i2)] = gt_i2Ti1.rotation()
+            # i2Ui1_graph_dict[(i1, i2)] = dask.delayed(lambda x: Unit3(np.array([x.x(), x.y(), x.z()])))(gt_i2Ti1)
+
+            # Visualize verified two-view correspondnces.
```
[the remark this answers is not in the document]
travisdriver: fixed

gtsfm/utils/metrics.py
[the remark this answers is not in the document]
travisdriver: bump @akshay-krishnan @ayushbaid

gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
...
+            - gt_camera_i2.translation()
+        )
+
+    # Perform ray tracing.
+    src = np.vstack((src_i1, src_i2))
+    drc = np.vstack((drc_i1, drc_i2))
+    logger.info("Computing ray intersections...")
+    start_time = timeit.default_timer()
+    loc, idr, _ = gt_scene_mesh.ray.intersects_location(src, drc, multiple_hits=False)
+    logger.info(f"Cast {2 * n_corrs} rays in {timeit.default_timer() - start_time} seconds.")
```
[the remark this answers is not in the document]
travisdriver: Seems to take about ~0.13 seconds per image pair on average. What do you mean 60K+ times

gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
...
+    """
+    if len(keypoints_i1) != len(keypoints_i2):
+        raise ValueError("Keypoints must have same counts")
+
+    n_corrs = len(keypoints_i1)
+    is_inlier = np.zeros(n_corrs, dtype=bool)
+    src_i1 = np.repeat(np.reshape(gt_camera_i1.translation(), (-1, 3)), n_corrs, axis=0)
+    src_i2 = np.repeat(np.reshape(gt_camera_i2.translation(), (-1, 3)), n_corrs, axis=0)
+
+    # Compute ketpoint rays.
```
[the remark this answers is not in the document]
travisdriver: fixed

visualization/view_scene.py
travisdriver: fixed

gtsfm/utils/metrics.py
```diff
@@ -46,7 +51,7 @@ def count_correct_correspondences(
         intrinsics_i1: intrinsics for i1.
         intrinsics_i2: intrinsics for i2.
         i2Ti1: relative pose
-        epipolar_dist_threshold: max acceptable distance for a correct correspondence.
+        dist_threshold: max acceptable distance for a correct correspondence.
```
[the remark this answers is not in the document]
travisdriver: Added

gtsfm/scene_optimizer.py
```diff
@@ -145,14 +146,24 @@ def create_computation_graph(
                 image_shape_graph[i1],
                 image_shape_graph[i2],
                 gt_i2Ti1,
+                gt_pose_i1,
```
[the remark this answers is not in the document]
travisdriver: fixed

gtsfm/utils/metrics.py
```diff
@@ -59,15 +64,145 @@ def count_correct_correspondences(
...
+    idr_i1 = idr[idr < n_corrs]
+    loc_i1 = loc[idr < n_corrs]
+    idr_i2 = idr[idr >= n_corrs] - n_corrs
+    loc_i2 = loc[idr >= n_corrs]
+    idr, i1_idx, i2_idx = np.intersect1d(idr_i1, idr_i2, return_indices=True)
+
+    # Forward project intersections into other image to compute error.
+    reproj_err = np.array([np.nan] * len(keypoints_i1))
+    for i in range(len(idr)):
+        xy_i1 = keypoints_i1.coordinates[idr[i]]
```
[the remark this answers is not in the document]
travisdriver: fixed

ayushbaid (commented):
I personally disagree with the API design of having both GT poses and GT mesh, and having if-else statements in the code. 

Having a mesh is a specific feature of the astronet dataset, and some other dataset in the future will have a different way of representing a ground truth.

What we essentially want is a better way of classifying the correspondences, right? The current method of using GT poses in the sampson distance is not the correct way as Travis's experiments show us.

I think the loader should have an API `evaluate_correspondences`, which takes in the verified correspondences and marks then as correct/incorrect. The benefit of this method is that we have abstracted out the implementation details of an evaluation function from the scene optimizer and all other plates and that leads to better maintainability as we do not need to add a new argument every time a new 3D representation is encountered. The downside is that we will need to think about how we can wrap up the function as a delayed and pass it around for the two-view optimizer to use and generate the metric.

gtsfm/utils/metrics.py
```diff
@@ -60,16 +66,147 @@ def count_correct_correspondences(
...
+    Returns:
+        keypoint_ind: (M,) array of keypoint indices whose corresponding ray intersected the ground truth mesh.
+        intersections_locations: (M, 3), array of ray intersection locations.
+    """
+    num_kpts = len(keypoints)
+    src = np.repeat(gt_camera.pose().translation().reshape((-1, 3)), num_kpts, axis=0)  # At_i1A
+    drc = np.asarray([gt_camera.backproject(keypoints.coordinates[i], depth=1.0) - src[i, :] for i in range(num_kpts)])
+    start_time = timeit.default_timer()
+    intersections, keypoint_ind, _ = gt_scene_mesh.ray.intersects_location(src, drc, multiple_hits=False)
+    logger.info(f"Case {num_kpts} rays in {timeit.default_timer() - start_time} seconds.")
```
johnwlambert, line 207: Still a bit hesitant about logging this per image pair, since we could be printing out 100,000s of lines with this call here on large datasets.

gtsfm/two_view_estimator.py
```diff
@@ -218,16 +240,83 @@ def check_for_degeneracy(
...
+    keypoints_i1: Keypoints,
+    keypoints_i2: Keypoints,
+    matches_i1i2: np.ndarray,
+    intrinsics_i1: Cal3Bundler,
+    intrinsics_i2: Cal3Bundler,
+    dist_threshold: float,
+    gt_wTi1: Pose3,
+    gt_wTi2: Pose3,
+    gt_scene_mesh: trimesh.Trimesh,
+) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray, float]:
```
johnwlambert, line 253: Should this be a child class of `BaseVerifier`?

gtsfm/two_view_estimator.py
```diff
@@ -150,36 +158,50 @@ def create_computation_graph(
             camera_intrinsics_i1_graph,
             camera_intrinsics_i2_graph,
         )
+        # gt_verifier_result = dask.delayed(ground_truth_verifier)(
+        #     keypoints_i1_graph,
```
johnwlambert, line 162: commented out code we probably want to remove before checking in

gtsfm/scene_optimizer.py
johnwlambert: Could we add an explanation of what the technical specification of this mesh is? e.g. Is it a triangle mesh, or is it a quad mesh? Need the face vertices be arranged in CCW or CW order, etc?

gtsfm/two_view_estimator.py
travisdriver: I can definitely add it, as it's useful for debugging. What do you think?

gtsfm/scene_optimizer.py
travisdriver: I believe I mention that it's a triangular surface mesh in multiple places, and I think the fact that the library is called `trimesh` gives it away. 

There are many different mesh formats that `trimesh` can handle, so I don't think it would be useful to include specific formatting details. If your mesh was generated properly for the given format (e.g., PLY, OBJ, STL), `trimesh` can handle it.

gtsfm/scene_optimizer.py
```diff
@@ -396,6 +404,17 @@ def save_full_frontend_metrics(
                 "inlier_ratio_gt_model": round(report.inlier_ratio_gt_model, PRINT_NUM_SIG_FIGS)
                 if report.inlier_ratio_gt_model
                 else None,
+                "inlier_avg_reproj_error_gt_model": round(
```
akshay-krishnan, line 407: should we populate these on datasets that do not have the mesh?

gtsfm/two_view_estimator.py
```diff
@@ -65,6 +68,9 @@ class TwoViewEstimationReport:
     U_error_deg: Optional[float] = None
     i2Ri1: Optional[Rot3] = None
     i2Ui1: Optional[Unit3] = None
+    reproj_error_gt_model: Optional[np.ndarray] = None
```
akshay-krishnan, line 71: we could use some explanation of these metrics here. maybe a one liner for each. is gt_model == gt_mesh?

gtsfm/two_view_estimator.py
```diff
@@ -106,7 +112,9 @@ def create_computation_graph(
         camera_intrinsics_i2_graph: Delayed,
         im_shape_i1_graph: Delayed,
         im_shape_i2_graph: Delayed,
-        i2Ti1_expected_graph: Optional[Delayed] = None,
+        wTi1_expected_graph: Optional[Delayed] = None,
```
akshay-krishnan, line 115: For consistency purposes, can we call these gt_wTi1_graph instead of wTi1_expected_graph?

akshay-krishnan: same with mesh

gtsfm/two_view_estimator.py
akshay-krishnan: +1, even if its not a derived verifier (which sounds like a good idea), it would be nice to have it outside this file.

gtsfm/two_view_estimator.py
```diff
@@ -218,16 +240,83 @@ def check_for_degeneracy(
         return i2Ri1, i2Ui1, v_corr_idxs, two_view_report
 
 
+def ground_truth_verifier(
+    keypoints_i1: Keypoints,
+    keypoints_i2: Keypoints,
+    matches_i1i2: np.ndarray,
+    intrinsics_i1: Cal3Bundler,
+    intrinsics_i2: Cal3Bundler,
```
akshay-krishnan, line 248: are these GT intrinsics?

gtsfm/two_view_estimator.py
```diff
@@ -218,16 +240,83 @@ def check_for_degeneracy(
...
+    dist_threshold: float,
+    gt_wTi1: Pose3,
+    gt_wTi2: Pose3,
+    gt_scene_mesh: trimesh.Trimesh,
+) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray, float]:
+    """Verify correspondences and compute relative pose using ground truth data."""
```
akshay-krishnan, line 254: add Args and return sections in docstring

gtsfm/two_view_estimator.py
```diff
@@ -247,9 +339,11 @@ def compute_correspondence_metrics(
     corr_idxs_i1i2: np.ndarray,
     intrinsics_i1: Cal3Bundler,
     intrinsics_i2: Cal3Bundler,
-    i2Ti1: Pose3,
     epipolar_distance_threshold: float,
-) -> Tuple[int, float, Optional[np.ndarray]]:
+    wTi1: Optional[Pose3] = None,
```
akshay-krishnan, line 343: I think we dont need this function anymore.

gtsfm/utils/metrics.py
akshay-krishnan: we should start using logging levels instead of unnecessary arguments: https://docs.python.org/3/library/logging.html#logging.Logger.log

gtsfm/two_view_estimator.py
```diff
@@ -218,16 +240,83 @@ def check_for_degeneracy(
...
+    )
+    inlier_idxs = np.where(inlier_mask.ravel() == 1)[0]
+    v_corr_idxs = matches_i1i2[inlier_idxs]
+    logger.info("computed ground truth correspondences")
+
+    if v_corr_idxs.shape[0] < 15:
+        return None, None, np.array([], dtype=np.uint64), 0.0
+
+    # Compute essential matrix.
+    i2Ei1, _ = cv2.findEssentialMat(
```
johnwlambert, line 275: I think we can cut this down by about 20 lines if we just call `Ransac.verify()` here:
https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/verifier/ransac.py#L62

johnwlambert:
I retrieved the meshed result for skydio-32 for COLMAP, but it doesn't look too great on the crane mast itself

. Maybe the mesh eval approach works best with convex bodies.

@codyly are these the sort of COLMAP results you were seeing after dense fusion from COLMAP?
Screen Shot 2021-10-21 at 3 50 02 PM
Screen Shot 2021-10-21 at 3 45 33 PM
Screen Shot 2021-10-21 at 3 44 45 PM

akshay-krishnan:
> > > > I personally disagree with the API design of having both GT poses and GT mesh, and having if-else statements in the code.
> > > > Having a mesh is a specific feature of the astronet dataset, and some other dataset in the future will have a different way of representing a ground truth.
> > > > What we essentially want is a better way of classifying the correspondences, right? The current method of using GT poses in the sampson distance is not the correct way as Travis's experiments show us.
> > > > I think the loader should have an API `evaluate_correspondences`, which takes in the verified correspondences and marks then as correct/incorrect. The benefit of this method is that we have abstracted out the implementation details of an evaluation function from the scene optimizer and all other plates and that leads to better maintainability as we do not need to add a new argument every time a new 3D representation is encountered. The downside is that we will need to think about how we can wrap up the function as a delayed and pass it around for the two-view optimizer to use and generate the metric.
> > > 
> > > 
> > > I'm not sure if we have enough different dataset-specific metrics _yet_ to warrant moving all of this to the individual loaders, but this could definitely be something that could be of use later.
> > > I think maybe the functionality of `TwoViewEstimationReport` should be expanded. In my opinion, we should have a separate class (I use a class called `Evaluator` in my other work) that collects all the outputs from the different modules of the SfM pipeline and then compiles the reports at the end.
> > > It should not be the job of the `SceneOptimizer` or `TwoViewEstimator` to compute metrics or carry around variables that are only used for evaluation purposes. These classes should only take inputs that are required to perform SfM. For example, the `SceneOptimizer` should only take in the ground truth poses if it's actually using them in the optimization. I think this would greatly simplify the API of `SceneOptimizer`.
> > > I think this would be a lot more clean workflow wise, i.e., perform SfM, then evaluate the result instead of perform intermediate SfM task -> compile metrics -> perform next intermediate SfM task -> compile more metrics -> ...
> > > I also think that this is slightly out of the scope of this specific PR, and that adding a single _optional_ variable is sufficient for now until we decide on a new evaluation implementation.
> > 
> > 
> > I think if we delay any evaluation until after the entire process is complete, we would need to keep in memory every single intermediate variable from GTSFM. Which could be expensive, and messy in the code. @travisdriver could you think of a way around that?
> > @ayushbaid I do like the idea of making it loader-specific. Maybe the base loader uses Sampson, but AstroNet can override that with its own implementation. If we pass in the loader to SceneOptimizer, this may be better from at least one other perspective -- we may not need to keep 500 images in memory at once, for a 500-image dataset, which seems very inefficient. (At least I think that is what is happening currently). Instead, we could make individual calls for images when we need them from the loader object.
> > Before we change the SceneOptimizer inputs for just one dataset, I think we should try this approach on another dataset as well. Travis, could you try retrieving the Delaunayed dense mesh from COLMAP for skydio-32, and visualizing the classified correspondences using your code on that dataset? Maybe the mesh based eval is a more general solution to look at. Or maybe there is a bug in Sampson (from the figure I showed in the slides today).
> 
> Well seeing as how the `TwoViewEstimationReport` stores the intermediate values already, I don't see how it would be much different from the current implementation. I was simply meaning we remove the reports from _inside_ the `SceneOptimizer`, and have a separate class the queries the different modules and organizes the reports.
> 
> The loader idea does seem more useful the more I think about it. Maybe the loader could even maintain its own optimization policy for datasets with measurements other than just images (e.g., GPS)? I would be slightly apprehensive about passing the loader object to _every single_ node, but I recently learned about the [scattering functionality in Dask](https://ericmjl.github.io/blog/2018/10/26/more-dask-pre-scattering-data/) that significantly reduces the overhead of transferring data between nodes. Think it's definitely something we could try.
> 
> @johnwlambert With respect to trying it on a COLMAP mesh. If you send me the mesh, I'll see what I can do. I do think investigating the inconsistencies with the Sampson distance metric would be useful either way. I think SD is more general than the mesh-based method since it requires far less information.

Chiming in, I agree with @ayushbaid  this is a problem. But I dont think the solution is having the evaluation logic in the loader. The solution has to be scalable to all modules. Why would we expect a loader class to produce evaluation metrics for all modules? 

Agree with @travisdriver , this is an existing problem and is not in the scope of PR, maybe we should open an issue and discuss there? 

We want to have the logic of computing metrics within the individual modules. But I think we must separate the metrics computation from the result computation. This means that we need more than just the single `run()` method for each module. And probably also have a container to store ground truth: poses, meshes etc.

gtsfm/utils/metrics.py
travisdriver: fixed

johnwlambert:
@travisdriver thanks for adding the unit tests.

It looks like we have a merge conflict from the visualization files here. Could you take a look when you get a chance?

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+from gtsam import PinholeCameraCal3Bundler, Cal3Bundler
+from gtsfm.common.keypoints import Keypoints
+
+import gtsfm.utils.metrics as metric_utils
+
+
+class TestMetricUtils(unittest.TestCase):
+    """Class containing all unit tests for metric utils."""
+
+    def test_mesh_inlier_correspondences(self):
```
johnwlambert, line 19: nit: would be nice to add a return type here for consistency
```python
def test_mesh_inlier_correspondences(self) -> None:
```

gtsfm/utils/metrics.py
```diff
@@ -55,21 +62,152 @@ def count_correct_correspondences(
...
+            is_inlier[keypoint_ind[i]] = max(err_i2i1, err_i1i2) < dist_threshold
+            reproj_err[keypoint_ind[i]] = max(err_i2i1, err_i1i2)
+        else:
+            is_inlier[keypoint_ind[i]] = False
+            reproj_err[keypoint_ind[i]] = np.nan
+
+    return is_inlier, reproj_err
+
+
+def compute_keypoint_intersections(
```
johnwlambert, line 188: @akshay-krishnan @ayushbaid do you think we should name this file `raycasting.py` or `mesh_intersection.py` and put it into the utils? Not sure if `metrics` is quite the right place for it.

gtsfm/utils/metrics.py
```diff
@@ -55,21 +62,152 @@ def count_correct_correspondences(
...
+    keypoints: Keypoints, gt_camera: PinholeCameraCal3Bundler, gt_scene_mesh: trimesh.Trimesh, verbose: bool = False
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Computes intersections between ground truth surface mesh and rays originating from image keypoints.
```
johnwlambert, line 191: do we not need Embree anymore? or does `trimesh` use it by default, so it doesn't need to be installed here?

It would be nice in this docstring to explain what library is performing the raycasting routine.

gtsfm/utils/metrics.py
```diff
@@ -55,21 +62,152 @@ def count_correct_correspondences(
...
+
+    # Forward project intersections into other image to compute error.
+    reproj_err = np.array([np.nan] * len(keypoints_i1))
+    for i in range(len(keypoint_ind)):
+        uv_i1 = keypoints_i1.coordinates[keypoint_ind[i]]
+        uv_i2 = keypoints_i2.coordinates[keypoint_ind[i]]
+        uv_i2i1, success_flag_i1 = gt_camera_i1.projectSafe(intersections_i2[i2_idx[i]])
+        uv_i1i2, success_flag_i2 = gt_camera_i2.projectSafe(intersections_i1[i1_idx[i]])
+        if success_flag_i1 and success_flag_i2:
+            err_i2i1 = np.linalg.norm(uv_i1.flatten() - uv_i2i1.flatten())
```
johnwlambert, line 177: why do we need all the flatten() calls here? Does `projectSafe()` not return a (2,) size array?

gtsfm/utils/metrics.py
```diff
@@ -55,21 +62,152 @@ def count_correct_correspondences(
...
+    keypoints_i1: Keypoints,
+    keypoints_i2: Keypoints,
+    gt_camera_i1: PinholeCameraCal3Bundler,
+    gt_camera_i2: PinholeCameraCal3Bundler,
+    gt_scene_mesh: trimesh.Trimesh,
+    dist_threshold: float,
+) -> Tuple[np.ndarray, np.ndarray]:
+    """Compute inlier correspondences using the ground truth triangular surface mesh of the scene. First, rays are
+    back-projected at each keypoint in the images and intersections between these rays and the ground truth mesh are
+    recorded. Next, given a match, the the mesh intersections corresponding to each keypoint is forward-projected into
```
johnwlambert, line 140: nit: typo here. `the the mesh intersections`

johnwlambert: nit: "are forward-projected ..."

gtsfm/utils/metrics.py
```diff
@@ -55,21 +62,152 @@ def count_correct_correspondences(
...
+    the other image and the reprojection error is computed to decide whether the match is an inlier.
+
+    Args:
+        keypoints_i1: N keypoints in image i1.
+        keypoints_i2: N corresponding keypoints in image i2.
+        gt_camera_i1: ground truth camera for image i1, i.e., wTi1 and intrinsics.
+        gt_camera_i1: ground truth camera for image i2, i.e., wTi2 and intrinsics.
+        gt_scene_mesh: ground truth triangular surface mesh of the scene in the world frame.
+        dist_threshold: max acceptable distance (in pixels) between image coordinates of ground truth landmark and
```
johnwlambert, line 149: i think it would be clearer to call it reprojection error in the description:
```python
dist_threshold: max acceptable reprojection error (in pixels) between image coordinates of ground truth landmark and keypoint.
```

gtsfm/utils/metrics.py
travisdriver: IMO, I don't think it's worth it to make a new file for one function.

gtsfm/utils/metrics.py
travisdriver: It can be run with or without Embree. However, it's orders of magnitude faster with it.

johnwlambert: I see. Are we running it with Embree enabled? It would be nice to include this info in the docstring.

gtsfm/utils/metrics.py
travisdriver: fixed

gtsfm/utils/metrics.py
travisdriver: fixed

gtsfm/utils/metrics.py
travisdriver: fixed

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+class TestMetricUtils(unittest.TestCase):
+    """Class containing all unit tests for metric utils."""
+
+    def test_mesh_inlier_correspondences(self) -> None:
+        """Tests `compute_keypoint_intersections()` function."""
+        # Create cube mesh with side length one centered at origin.
+        box = trimesh.primitives.Box()
+
+        # Create arrangement of two camera pointing at center of one of the cube's faces.
```
johnwlambert, line 24: nit: type on "two cameras"

johnwlambert: "pointing at the center"

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        calibration = Cal3Bundler(10, 0, 0, 1, 1)
```
johnwlambert, line 25: nit: prefer to use named args so the reader can know what 10,, 0, 0, 1, 1 correspond to

johnwlambert: even
```python
        fx, k1, k2, u0, v0 = 10, 0, 0, 1, 1
        calibration = Cal3Bundler(fx, k1, k2, u0, v0)
```
would be preferable IMHO

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        target_pos = [0.5, 0, 0]
+        cam_i1 = PinholeCameraCal3Bundler().Lookat([2, 1, 0], target_pos, [0, -1, 0], calibration)
+        cam_i2 = PinholeCameraCal3Bundler().Lookat([2, -1, 0], target_pos, [0, -1, 0], calibration)
```
johnwlambert, line 28: prefer a named arg for (0, -1, 0) and for [2, -1, 0] to help the reader understand what it represents

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        keypoints_i1 = Keypoints(coordinates=np.array([[1, 1]]).astype(np.float32))
+        keypoints_i2 = Keypoints(coordinates=np.array([[1, 1]]).astype(np.float32))
+
+        # Project keypoint at center of each simulated image and record intersection.
+        is_inlier, reproj_err = metric_utils.mesh_inlier_correspondences(
+            keypoints_i1, keypoints_i2, cam_i1, cam_i2, box, 0.1
```
johnwlambert, line 34: nit: prefer named args here for at least the last two args for improved readability. Named args for all args would also be fine. That way the reader can know right away what 0.1 stands for

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        assert np.count_nonzero(is_inlier) == 1
+        assert reproj_err[0] < 1e-4
+
+    def test_compute_keypoint_intersections(self) -> None:
+        """Tests `compute_keypoint_intersections()` function."""
+        # Create cube mesh with side length one centered at origin.
+        box = trimesh.primitives.Box()
+
+        # Create arrangement of 4 cameras in x-y plane pointing at the cube.
+        calibration = Cal3Bundler(10, 0, 0, 1, 1)
```
johnwlambert, line 45: same note here about disambiguating args to Cal3Bundler

gtsfm/utils/metrics.py
johnwlambert: marking as resolved, thanks

gtsfm/utils/metrics.py
johnwlambert: marking as resolved, thanks

tests/utils/test_metric_utils.py
johnwlambert: marking as resolved, thanks

gtsfm/common/two_view_estimation_report.py
```diff
@@ -46,3 +45,6 @@ class TwoViewEstimationReport:
     U_error_deg: Optional[float] = None
     i2Ri1: Optional[Rot3] = None
     i2Ui1: Optional[Unit3] = None
+    reproj_error_gt_model: Optional[np.ndarray] = None
```
johnwlambert, line 48: since we starting to get a lot of attributes here, I wonder if we should make this hierarchical. I propose that we make a report like
```python
@dataclass
class MeshReprojectionReport
    reproj_error_gt_model: float
    inlier_avg_reproj_error_gt_model: float
    outlier_avg_reproj_error_gt_model: float
```
Then 
```python
@dataclass(frozen=False)
class TwoViewEstimationReport:
    mesh_reprojection_report: Optional[MeshReprojectionReport] = None
```
What are your thoughts about that? I think it keeps things cleaner for attributes that we usually won't have.

We need docstrings explaining what each of those quantities mean here for the constructor, though.

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        )
+        assert np.count_nonzero(is_inlier) == 1
+        assert reproj_err[0] < 1e-4
+
+    def test_compute_keypoint_intersections(self) -> None:
+        """Tests `compute_keypoint_intersections()` function."""
+        # Create cube mesh with side length one centered at origin.
+        box = trimesh.primitives.Box()
+
+        # Create arrangement of 4 cameras in x-y plane pointing at the cube.
```
johnwlambert, line 44: should this be "x-z" plane instead?

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+        # Project keypoint at center of each simulated image and record intersection.
+        is_inlier, reproj_err = metric_utils.mesh_inlier_correspondences(
+            keypoints_i1, keypoints_i2, cam_i1, cam_i2, box, 0.1
+        )
+        assert np.count_nonzero(is_inlier) == 1
+        assert reproj_err[0] < 1e-4
+
+    def test_compute_keypoint_intersections(self) -> None:
+        """Tests `compute_keypoint_intersections()` function."""
```
johnwlambert, line 40: I think we could add a few more sentences to the docstring to make the reader have to do no work whatsoever to understand.

```
"""Tests the `compute_keypoint_intersections()` function.

We arrange four cameras in the x-z plane around a cube centered at the origin.
These cameras are placed at (2,0,0), (-2,0,0), (0,0,2) and (0,0,-2). We project a single
3d point located at origin into each camera and ... expect ... blah blah blah ...
"""

File: tests/utils/test_metric_utils.py [hunk 38]
```diff
@@ -0,0 +1,61 @@
...
+from gtsfm.common.keypoints import Keypoints
+
+import gtsfm.utils.metrics as metric_utils
+
+
+class TestMetricUtils(unittest.TestCase):
+    """Class containing all unit tests for metric utils."""
+
+    def test_mesh_inlier_correspondences(self) -> None:
+        """Tests `compute_keypoint_intersections()` function."""
```
johnwlambert, line 20: same note here as for the function below. it would be nice to have 2-3 sentences explaining the setup so the reader can be extremely lazy and gauge correctness in a single glance at the code after reading the docstring.

tests/utils/test_metric_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        box = trimesh.primitives.Box()
+
+        # Create arrangement of 4 cameras in x-y plane pointing at the cube.
+        calibration = Cal3Bundler(10, 0, 0, 1, 1)
+        cam_pos = [[2, 0, 0], [-2, 0, 0], [0, 0, 2], [0, 0, -2]]
+        target_pos = [0, 0, 0]
+        cams = [PinholeCameraCal3Bundler().Lookat(c, target_pos, [0, -1, 0], calibration) for c in cam_pos]
+
+        # Project keypoint at center of each simulated image and record intersection.
+        kpt = Keypoints(coordinates=np.array([[1, 1]]).astype(np.float32))
```
johnwlambert, line 51: might be nice to say in the docstring that each image is of size 2x2.

gtsfm/common/two_view_estimation_report.py
travisdriver: The reprojection error is defined for the Sampson distance metric as well.

gtsfm/utils/metrics.py
travisdriver: This is defined when the mesh is read in with the `use_embree` option. So, this function just goes with whatever the loader defines.

For the `AstronetLoader`, Embree is used.

tests/utils/test_metric_utils.py
travisdriver: I think the comments paint a very clear picture of what's going on.

tests/utils/test_metric_utils.py
travisdriver: Comments are very clear already.

tests/utils/test_metric_utils.py
travisdriver: It's not really an important detail.

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
travisdriver: fixed

tests/utils/test_metric_utils.py
johnwlambert: I like the inline comments. I think a few more sentences would help make it easier for us and others to maintain this code in the future since we're getting into the 100s of unit tests now that we need to maintain.

travisdriver: What else needs to be added that's not already in the comments? I'll add it.

akshay-krishnan (approved):

ayushbaid (commented):
Looks good overall. My main concern is that why we compute reprojection errors only on points which are inliers from epipolar constraint.

gtsfm/utils/metrics.py
```diff
@@ -30,23 +32,28 @@
...
     """Checks the correspondences for epipolar distances and counts ones which are below the threshold.
 
     Args:
         keypoints_i1: keypoints in image i1.
         keypoints_i2: corr. keypoints in image i2.
         intrinsics_i1: intrinsics for i1.
         intrinsics_i2: intrinsics for i2.
-        i2Ti1: relative pose
-        epipolar_dist_threshold: max acceptable distance for a correct correspondence.
+        dist_threshold: max acceptable distance for a correct correspondence.
```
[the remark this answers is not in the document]
travisdriver: They can be manipulated in the config based on whichever method the user would like to use.

gtsfm/two_view_estimator.py
```diff
@@ -152,23 +154,37 @@ def create_computation_graph(
...
 ) -> TwoViewEstimationReport:
     """Wrapper around class constructor for Dask."""
+    # Compute ground truth metrics.
+    if v_corr_idxs_inlier_mask_gt is not None and reproj_error_gt_model is not None:
+        num_inliers_gt_model = np.count_nonzero(v_corr_idxs_inlier_mask_gt)
+        inlier_ratio_gt_model = (
+            np.count_nonzero(v_corr_idxs_inlier_mask_gt) / v_corr_idxs.shape[0] if len(v_corr_idxs) > 0 else 0.0
+        )
+        inlier_avg_reproj_error_gt_model = np.mean(reproj_error_gt_model[v_corr_idxs_inlier_mask_gt])
+        outlier_avg_reproj_error_gt_model = np.nanmean(
```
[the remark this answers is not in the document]
travisdriver: Gives a measure of how _good_ the inliers are and how _bad_ the outliers are.

gtsfm/two_view_estimator.py
```diff
@@ -152,23 +154,37 @@ def create_computation_graph(
...
+    v_corr_idxs_inlier_mask_gt: Optional[np.ndarray] = None,
+    reproj_error_gt_model: Optional[np.ndarray] = None,
 ) -> TwoViewEstimationReport:
     """Wrapper around class constructor for Dask."""
+    # Compute ground truth metrics.
+    if v_corr_idxs_inlier_mask_gt is not None and reproj_error_gt_model is not None:
+        num_inliers_gt_model = np.count_nonzero(v_corr_idxs_inlier_mask_gt)
+        inlier_ratio_gt_model = (
+            np.count_nonzero(v_corr_idxs_inlier_mask_gt) / v_corr_idxs.shape[0] if len(v_corr_idxs) > 0 else 0.0
+        )
```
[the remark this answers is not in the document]
travisdriver: Yeah I'm getting a lot of mypy errors for a lot of variables because many of them are labeled as `Optional`. I think this should be part of a larger effort where we reassess whether certain variables are, in fact, optional.

gtsfm/loader/astronet_loader.py
```diff
@@ -78,6 +82,19 @@ def __init__(
             cameras, images, points3d, load_sfmtracks=use_gt_sfmtracks
         )
 
+        # Read in scene mesh as Trimesh object
+        if gt_scene_mesh_path is not None:
+            if not Path(gt_scene_mesh_path).exists():
+                raise FileNotFoundError(f"No mesh found at {gt_scene_mesh_path}")
+            self.gt_scene_trimesh = trimesh.load(gt_scene_mesh_path, process=False, maintain_order=True)
```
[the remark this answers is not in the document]
travisdriver: Other functions access it.

gtsfm/loader/astronet_loader.py
```diff
@@ -49,6 +51,8 @@ def __init__(
 
         Args:
             data_dir: path to directory containing the COLMAP-formatted data: cameras.bin, images.bin, and points3D.bin
+            gt_scene_mesh_path (optional): path to file of target small body surface mesh.
+                Note: vertex size mismath observed when reading in from OBJ format. Prefer PLY.
```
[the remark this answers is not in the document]
travisdriver: I guess I meant it's optional in the sense that the pipeline does not require it for reconstruction (as is the case with the other ground truth variables).

gtsfm/runner/run_scene_optimizer_astronet.py
```diff
@@ -34,17 +37,46 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
+    def run(self) -> None:
+        """Run Structure-from-Motion (SfM) pipeline."""
+        # Create dask client.
+        cluster = LocalCluster(
+            n_workers=self.parsed_args.num_workers, threads_per_worker=self.parsed_args.threads_per_worker
+        )
+
+        with Client(cluster) as client, performance_report(filename="dask-report.html"):
+            # Scatter surface mesh across all nodes to preserve computation time and memory.
+            gt_scene_trimesh_future = client.scatter(self.loader.gt_scene_trimesh, broadcast=True)
```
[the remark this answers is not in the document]
travisdriver: To be honest, the main reason I did it is because Dask told me to, but I can definitely say it saves a significant amount of computation time (although I don't have any exact numbers). 

The documentation is a little spotty, but from my understanding, it preallocates memory for the object on each worker and then passes around the pointer to the worker threads.

It definitely may be useful for other large objects in the pipeline.

gtsfm/scene_optimizer.py
```diff
@@ -103,6 +104,7 @@ def create_computation_graph(
         camera_intrinsics_graph: List[Delayed],
         image_shape_graph: List[Delayed],
         gt_cameras_graph: Optional[List[Delayed]] = None,
+        gt_scene_mesh: Optional[trimesh.Trimesh] = None,
```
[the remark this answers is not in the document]
travisdriver: This is a good point I'll change it

gtsfm/utils/metrics.py
```diff
@@ -30,23 +32,28 @@
 logger = logger_utils.get_logger()
 
 
-def count_correct_correspondences(
+def compute_correspondence_metrics(
```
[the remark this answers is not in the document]
travisdriver: The reprojection error is computed for every for _all_ keypoints, not just the inliers. The `is_inlier` variable is a boolean array specifying whether each match is an inlier, and the `reproj_error` variable is the same length. These are then used to compute `inlier_avg_reproj_error_gt_model` and `outlier_avg_reproj_error_gt_model`.

travisdriver:
> Looks good overall. My main concern is that why we compute reprojection errors only on points which are inliers from epipolar constraint.

The reprojection errors are computed for all keypoints, not just inliers.

gtsfm/common/two_view_estimation_report.py
```diff
@@ -46,3 +45,6 @@ class TwoViewEstimationReport:
...
+    inlier_avg_reproj_error_gt_model: Optional[float] = None
+    outlier_avg_reproj_error_gt_model: Optional[float] = None
```
johnwlambert, line 50: We still need to add definitions for each of these new variables to the class in the docstring.

tests/utils/test_metric_utils.py
johnwlambert: Would be great if you could add a few sentences to the docstring giving a high-level description of the test.

You can use my summary below if you think it's accurate : - )

```python
"""Tests the `compute_keypoint_intersections()` function. We arrange four cameras in the x-z plane around a cube centered at the origin. These cameras are placed at (2,0,0), (-2,0,0), (0,0,2) and (0,0,-2). We project a single 3d point located at the origin into each camera. Since the cube has unit length on each dimension, we expect a keypoint located at the center of each image to be found at the boundary of the cube -- 0.5 meters from the origin for each side on the z-x plane.
```

gtsfm/common/two_view_estimation_report.py
travisdriver: added

gtsfm/scene_optimizer.py
travisdriver: Isn't it projecting a point onto a line?

gtsfm/two_view_estimator.py
```diff
@@ -152,23 +154,37 @@ def create_computation_graph(
...
+        inlier_avg_reproj_error_gt_model = np.mean(reproj_error_gt_model[v_corr_idxs_inlier_mask_gt])
+        outlier_avg_reproj_error_gt_model = np.nanmean(
+            reproj_error_gt_model[np.logical_not(v_corr_idxs_inlier_mask_gt)]
+        )
+    else:
+        num_inliers_gt_model = 0
```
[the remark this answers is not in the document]
travisdriver: It was returning 0 before. Should I change it to `Nan`?

gtsfm/common/two_view_estimation_report.py
```diff
@@ -34,6 +33,9 @@ class TwoViewEstimationReport:
         U_error_deg: relative translation error w.r.t. GT. Only defined if GT poses provided.
         i2Ri1: relative rotation.
         i2Ui1: relative translation direction.
+        reproj_error_gt_model: reprojection error betwen correspondences w.r.t. GT.
```
johnwlambert, line 36: thanks for adding these. small typo here on "between" I might say 
`reprojection errors between correspondences w.r.t. GT.` to indicate its plural

gtsfm/scene_optimizer.py
```diff
@@ -153,15 +151,25 @@ def create_computation_graph(
                 camera_intrinsics_graph[i2],
                 image_shape_graph[i1],
                 image_shape_graph[i2],
-                gt_i2Ti1,
+                gt_wTi1,
+                gt_wTi2,
+                gt_scene_mesh,
             )
 
+            # Store results.
```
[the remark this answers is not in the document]
travisdriver: Trying to avoid floating code blocks

tests/utils/test_metric_utils.py
travisdriver: reworded and added

gtsfm/common/two_view_estimation_report.py
travisdriver: fixed

gtsfm/scene_optimizer.py
travisdriver: removed

gtsfm/scene_optimizer.py
```diff
@@ -415,6 +417,17 @@ def save_full_frontend_metrics(
                 "inlier_ratio_gt_model": round(report.inlier_ratio_gt_model, PRINT_NUM_SIG_FIGS)
                 if report.inlier_ratio_gt_model
                 else None,
+                "inlier_avg_reproj_error_gt_model": round(
+                    np.nanmean(report.reproj_error_gt_model[report.v_corr_idxs_inlier_mask_gt]), PRINT_NUM_SIG_FIGS
+                )
+                if report.reproj_error_gt_model is not None and report.v_corr_idxs_inlier_mask_gt is not None
+                else None,
+                "outlier_avg_reproj_error_gt_model": round(
+                    np.nanmean(report.reproj_error_gt_model[np.logical_not(report.v_corr_idxs_inlier_mask_gt)]),
```
johnwlambert, line 426: @travisdriver can't we just use `report.inlier_avg_reproj_error_gt_model` here? why do we need to compute the mean over again? I think we already compute it in the TwoViewEstimator

https://github.com/borglab/gtsfm/pull/295/files?authenticity_token=sUeFac35fdQHIM%2FWNeNuKnBbZw%2Bq8phgw%2B4iKHfmflK7LqGME7HzIYxnQRv3OJqxvyVxjM1akbCEnMutH8PLxg%3D%3D&file-filters%5B%5D=.py#diff-1f5b63f35ed4d88709ce154ae9808f38a1a7178fb5f74211b1053ae5a8c37894R177

gtsfm/scene_optimizer.py
johnwlambert: It's related to it. It's a first order approximation derived using Lagrangian multipliers. I wouldn't call it reprojection error, since it's not a projection of a 3d point into the image, but maybe we can get away with calling it this.

gtsfm/two_view_estimator.py
```diff
@@ -278,8 +255,10 @@ def aggregate_frontend_metrics(
     for report in two_view_reports_dict.values():
         if report is None:
             continue
-        rot3_angular_errors.append(report.R_error_deg)
-        trans_angular_errors.append(report.U_error_deg)
+        if report.R_error_deg is not None:
+            rot3_angular_errors.append(report.R_error_deg)
```
johnwlambert, line 259: @travisdriver we shouldn't need this `is not None` check here, since we are just filling the array with None, and then casting it to float which converts them to Nan, and then they are ignored in the `np.nanmean` call.

johnwlambert: But if keeps the types simpler, maybe it's ok.

travisdriver: Was trying to fix some of the mypy errors

gtsfm/scene_optimizer.py
travisdriver: You're right. I think it's cleaner to just compute it here when everything else is computed. I'll remove it here

gtsfm/two_view_estimator.py
johnwlambert: Got it, sure.

gtsfm/utils/metrics.py
```diff
@@ -55,21 +62,152 @@ def count_correct_correspondences(
...
+        gt_camera_i2 = PinholeCameraCal3Bundler(gt_wTi2, intrinsics_i2)
+        is_inlier, reproj_error = mesh_inlier_correspondences(
+            matched_keypoints_i1,
+            matched_keypoints_i2,
+            gt_camera_i1,
+            gt_camera_i2,
+            gt_scene_mesh,
+            dist_threshold,
+        )
+    elif gt_wTi1 is not None and gt_wTi2 is not None:
```
johnwlambert, line 82: i don't think we need this check here to see if the ground truth poses are provided since we already check it in the TwoViewEstimator:  

https://github.com/borglab/gtsfm/pull/295/files?authenticity_token=QXZc5PBgvc%2BslrvBTHXtF%2BPMIrnyqmt7Pn2kCL1%2B2atLH3gBLigzOifRNQyOrhmMLLI0OpUCYqt5D02N1VtsPw%3D%3D&file-filters%5B%5D=.py#diff-1f5b63f35ed4d88709ce154ae9808f38a1a7178fb5f74211b1053ae5a8c37894R120

johnwlambert: i think the double nesting here makes it harder to read, so i think a single nesting would be better, just checking to see if the mesh is there. or you can exit immediately with None,None.

We discuss this a but in [Contributing.md](https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md#python-style), but returning early is always preferred to nesting.

gtsfm/two_view_estimator.py
```diff
@@ -152,23 +154,37 @@ def create_computation_graph(
...
+    R_error_deg: Optional[float] = None,
+    U_error_deg: Optional[float] = None,
+    v_corr_idxs_inlier_mask_gt: Optional[np.ndarray] = None,
+    reproj_error_gt_model: Optional[np.ndarray] = None,
 ) -> TwoViewEstimationReport:
     """Wrapper around class constructor for Dask."""
+    # Compute ground truth metrics.
+    if v_corr_idxs_inlier_mask_gt is not None and reproj_error_gt_model is not None:
+        num_inliers_gt_model = np.count_nonzero(v_corr_idxs_inlier_mask_gt)
+        inlier_ratio_gt_model = (
```
johnwlambert, line 174: @travisdriver  this is actually a different definition of the metric than the one we have been using. Before we were using the # of putatives as the denominator

https://github.com/borglab/gtsfm/pull/295/files#diff-1f5b63f35ed4d88709ce154ae9808f38a1a7178fb5f74211b1053ae5a8c37894L223

The way we had it before follows Heinly12eccv:
 https://www.cs.unc.edu/~jheinly/publications/eccv2012-heinly.pdf

If we try to compute `inlier_ratio_gt_model` here, we don't have access to the # of putatives (`corr_idxs` instead of `v_corr_idxs`)

johnwlambert: Maybe that metric should be moved to the matcher, actually, since it's a matcher-based metric, and has nothing to do with the verifier.

johnwlambert: It looks like the fraction here is (#verified w.r.t. GT model) / (# verified w.r.t. estimated model), which I don't think is what we want for inlier ratio, since it could be greater than 1.

johnwlambert: Actually, it looks like we were computing it this way before, since we were passing in `v_corr_idxs` to `compute_correspondence_metrics()`, not `corr_idxs`.

johnwlambert: So it looks like the `inlier_ratio_gt_model` here is actually 

                                          # verified correspondences that were right w.r.t. GT model
inlier_ratio_gt_model = ------------
                                           # verrified correspondences

johnwlambert: @akshay-krishnan I think this explains the bug we saw in https://github.com/borglab/gtsfm/issues/306, actually

PR #296, opened 2021-09-14
Create controller to gradually relax problem strictness

johnwlambert:
Tries first with 200 required inliers and 0.6 require inlier ratio, and decreases this until we get to our current default.

Empirically requiring at least 3 * number of images in the dataset for # of backend measurements, to accept result.

### DIFF
--- a/gtsfm/averaging/rotation/cycle_consistency.py
+++ b/gtsfm/averaging/rotation/cycle_consistency.py
@@ -98,7 +98,7 @@ def compute_cycle_error(
     i2Ri1_dict: Dict[Tuple[int, int], Rot3],
     cycle_nodes: Tuple[int, int, int],
     two_view_reports_dict: Dict[Tuple[int, int], TwoViewEstimationReport],
-    verbose: bool = True,
+    verbose: bool = False,
 ) -> Tuple[float, Optional[float], Optional[float]]:
     """Compute the cycle error by the magnitude of the axis-angle rotation after composing 3 rotations.
 
--- a/gtsfm/runner/run_scene_optimizer_colmaploader.py
+++ b/gtsfm/runner/run_scene_optimizer_colmaploader.py
@@ -5,6 +5,7 @@ import hydra
 from dask.distributed import Client, LocalCluster, performance_report
 from hydra.utils import instantiate
 
+import gtsfm.utils.io as io_utils
 import gtsfm.utils.logger as logger_utils
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.loader.colmap_loader import ColmapLoader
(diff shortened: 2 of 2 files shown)


gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -98,7 +98,7 @@ def compute_cycle_error(
     i2Ri1_dict: Dict[Tuple[int, int], Rot3],
     cycle_nodes: Tuple[int, int, int],
     two_view_reports_dict: Dict[Tuple[int, int], TwoViewEstimationReport],
-    verbose: bool = True,
+    verbose: bool = False,
```
johnwlambert, line 101: was way too verbose

travisdriver:
Is this mainly for analysis to decide which number of inliers and inlier ratio combination is best?

johnwlambert:
> Is this mainly for analysis to decide which number of inliers and inlier ratio combination is best?

@travisdriver I think we need a controller like this in our main system, because no set of hyperparameters will generalize to all scenes, based on the width of baselines and # of total images. COLMAP does the same thing here:
https://github.com/colmap/colmap/blob/dev/src/controllers/incremental_mapper.cc#L322

gtsfm/averaging/rotation/cycle_consistency.py
akshay-krishnan: Agreed

gtsfm/runner/run_scene_optimizer_colmaploader.py
```diff
@@ -13,37 +14,81 @@
 logger = logger_utils.get_logger()
 
 
+# fmt: off
+# Successive relaxation threshold pairs -- from strictest to loosest
+# `inlier ratio` is the minimum allowed inlier ratio w.r.t. the estimated model
+NUM_INLIERS_THRESHOLDS      =  [200, 175, 150, 125, 100, 75,  50,   25, 15]
+MIN_INLIER_RATIOS_THRESHOLDS = [0.6, 0.5, 0.4, 0.3, 0.2, 0.1, 0.1, 0.1, 0.1]
```
akshay-krishnan, line 21: What is the reasoning behind this? It does not make sense for me to use 15 inliers with a 0.1 ratio. From an estimation perspective I think inlier ratio is expected to be constant or increase as we decrease the number of inliers. 
Happy to look at some literature if this choice is based on something.

akshay-krishnan (commented):
From a design perspective I think this is better done in TwoViewEstimator.

gtsfm/runner/run_scene_optimizer_colmaploader.py
johnwlambert: @akshay-krishnan thanks for taking a look. 15 inliers with a 0.1 ratio is our current default. So I just relaxed it all the way back to where we are currently (which can be bad). But this makes a huge difference in performance.

I'll post the full metrics, but in the other PR here (which I will close), I found that for skydio-32, we can now get to:
- 6600 tracks after filtering after BA (from 612)
- 2.2 deg translation error after BA (from 9.0)
- 0.97 deg rotation error after BA (from 7.9)
- 0.12 translation distance error after BA (from 0.53)
- 2.91 avg track length unfiltered (from 2.76)
- 3.8 mean unfiltered reproj error (from 2829)

akshay-krishnan:
Other than that, I like the idea of using multiple thresholds :)

johnwlambert:
> From a design perspective I think this is better done in TwoViewEstimator.

Yeah we want to avoid re-computing the front-end more than once. I think we should put it after the cycle consistency though, since i'm using as a proxy "number of backend measurements coming out of cycle consistency" for "is the problem solvable"

gtsfm/runner/run_scene_optimizer_colmaploader.py
akshay-krishnan: Thanks @johnwlambert I think we should look beyond statistics here. It is not a deep model that estimates the F/E/H, and whatever hyperparameter should have some reasoning. 
I agree the current setting is not good, but as far as I can see setting a moderate number of inliers (like 50) with a high inlier ratio (0.6 or more) has not been tried in the other PR.

akshay-krishnan: And the fact that we are seeing so much improvement is some justification towards not using the 15, 0.1 setting we currently use.

akshay-krishnan: Basically I agree with most of your points, except that we should use inlier ratios as low as 0.1. As you noted, this may not give us a good model, so should be avoided.

johnwlambert: @akshay-krishnan I'm sure the thresholds can be tuned further. But I think this can help us get unblocked in terms of accuracy, and then we can try to tune these thresholds to work on many more datasets.

To put this in the two-view estimator, we are going to need to restructure the scene optimizer, two view estimator, and verifiers. I'd prefer to do that in a different PR so we could discuss the design in some more detail.

akshay-krishnan: Sure, it just becomes harder to do later as more PRs get merged. 
The thresholds can be tuned further, but why are we using a 0.1 inlier ratio if we know it doesn’t work well?

johnwlambert:
effect on Notre-Dame-20 dataset, on BA metrics:

|Metric name|	Value Before | Value After |
| :-: | :-: | :-: |
|number_cameras |	19 | 17 |
|number_tracks_unfiltered |	2316 | 2557 |
|mean_3d_track_lengths_unfiltered |	2.93566 | 3.04537 |
|mean_reprojection_errors_unfiltered_px |	1.01345 | 191.751 |
|mean_rotation_error_angle_deg |	0.982867 | 0.505581 |
|mean_translation_error_distance |	0.607116 | 0.147453 |
|mean_translation_angle_error_deg |	7.62274 | 5.06972 |
|number_tracks_filtered |	2128 | 2214 |
|mean_3d_track_lengths_filtered |	2.6532 | 2.80804 |
|mean_reprojection_errors_filtered_px |	0.697504 | 0.878593  |

camera dist to GT cameras goes from 0.6 -> 0.15
Rot error 0.98 -> 0.5 deg.

dellaert:
> Tries first with 200 required inliers and 0.6 require inlier ratio, and decreases this until we get to our current default.
> 
> Empirically requiring at least 3 * number of images in the dataset for # of backend measurements, to accept result.

Could you elaborate in The PR comment when and where this controller is run? is it it’s own process in the “graph” ? Maybe paste an image of the revised graph.

travisdriver:
> effect on Notre-Dame-20 dataset, on BA metrics:
> 
> Metric name	Value Before	Value After
> number_cameras	19	17
> number_tracks_unfiltered	2316	2557
> mean_3d_track_lengths_unfiltered	2.93566	3.04537
> mean_reprojection_errors_unfiltered_px	1.01345	191.751
> mean_rotation_error_angle_deg	0.982867	0.505581
> mean_translation_error_distance	0.607116	0.147453
> mean_translation_angle_error_deg	7.62274	5.06972
> number_tracks_filtered	2128	2214
> mean_3d_track_lengths_filtered	2.6532	2.80804
> mean_reprojection_errors_filtered_px	0.697504	0.878593
> camera dist to GT cameras goes from 0.6 -> 0.15
> Rot error 0.98 -> 0.5 deg.

The value for `mean_reprojection_errors_unfiltered_px` goes up significantly (1.01345 -> 191.751), any idea why this happens?

dellaert:
Ok…

PR #221, opened 2021-07-29
Aligning Point Cloud and Camera Frustums

adisingh50:
It's been a while since my last commit, but I realized my last PR was left in an unfinished state. Thus, I decided to make an entirely new PR which is up to date with all the new codebases changes made over the summer. Once this is merged, I'll remove the old, lingering PRs.

In this PR, I have modified `io.py` and added a new file `ellipsoid.py` which contains all the helper functions necessary to align the point cloud and camera frustums to the x,y,z axes using SVD. I've also added `test_ellipsoid_utils.py` to perform unit testing on the helper functions in `ellipsoid.py`.

After Point Cloud Alignment:
image

PR #297, opened 2021-09-15
Fix COLMAP extrinsics reading bug in ReactJS

johnwlambert:
COLMAP stores extrinsics cTw not world poses wTc
- Read incorrectly in React
- Read correctly in Python: https://github.com/borglab/gtsfm/blob/master/gtsfm/utils/io.py#L268

We should add a `.inverse()` method to our SE(3) class in ReactJS to make this a trivial one-line fix:
```python
def inverse(self) -> "SE3":
    return SE3(rotation=self.rotation.T, translation=self.rotation.T.dot(-self.translation))
```


johnwlambert:
@PratyushaMaiti @nehau9 here's the fix I had in mind for this part

johnwlambert:
Fixed by Adi's PR in https://github.com/borglab/gtsfm/pull/221

PR #303, opened 2021-09-16
add num_inliers and inlier_ratio distribution metrics to report

johnwlambert:
These are helpful for tuning thresholds

### DIFF
--- a/gtsfm/configs/deep_front_end.yaml
+++ b/gtsfm/configs/deep_front_end.yaml
@@ -12,7 +12,7 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    eval_threshold_px: 3 # in px
+    eval_threshold_px: 4 # in px
     min_num_inliers_acceptance: 15
 
     matcher:
--- a/gtsfm/configs/scene_optimizer_unit_test_config.yaml
+++ b/gtsfm/configs/scene_optimizer_unit_test_config.yaml
@@ -13,7 +13,7 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    eval_threshold_px: 3 # in px
+    eval_threshold_px: 0.5 # in px
     min_num_inliers_acceptance: 15
     
     matcher:
--- a/gtsfm/two_view_estimator.py
+++ b/gtsfm/two_view_estimator.py
@@ -316,14 +316,23 @@ def aggregate_frontend_metrics(
     """
     num_image_pairs = len(two_view_reports_dict.keys())
 
-    inlier_ratio_gt_model_all_pairs = [report.inlier_ratio_gt_model for report in two_view_reports_dict.values()]
-    inlier_ratio_est_model_all_pairs = [report.inlier_ratio_est_model for report in two_view_reports_dict.values()]
-    num_inliers_gt_model_all_pairs = [report.num_inliers_est_model for report in two_view_reports_dict.values()]
-    num_inliers_est_model_all_pairs = [report.num_inliers_gt_model for report in two_view_reports_dict.values()]
+    rot3_angular_errors = []
+    trans_angular_errors = []
 
-    # all rotational errors in degrees
-    rot3_angular_errors = [report.R_error_deg for report in two_view_reports_dict.values()]
-    trans_angular_errors = [report.U_error_deg for report in two_view_reports_dict.values()]
+    inlier_ratio_gt_model_all_pairs = []
+    inlier_ratio_est_model_all_pairs = []
+    num_inliers_gt_model_all_pairs = []
+    num_inliers_est_model_all_pairs = []
+    # populate the distributions -- all rotational errors in degrees
+    for report in two_view_reports_dict.values():
+        rot3_angular_errors.append(report.R_error_deg)
+        trans_angular_errors.append(report.U_error_deg)
+
+        inlier_ratio_gt_model_all_pairs.append(report.inlier_ratio_gt_model)
+        inlier_ratio_est_model_all_pairs.append(report.inlier_ratio_est_model)
+        num_inliers_gt_model_all_pairs.append(report.num_inliers_est_model)
+        num_inliers_est_model_all_pairs.append(report.num_inliers_gt_model)
+
     rot3_angular_errors = np.array(rot3_angular_errors, dtype=float)
     trans_angular_errors = np.array(trans_angular_errors, dtype=float)
     # count number of rot3 errors which are not None. Should be same in rot3/unit3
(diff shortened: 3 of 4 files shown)


johnwlambert:
Screen Shot 2021-09-16 at 3 32 23 PM

skydio-32, deep front end. justifies why 0.6 is a good inlier ratio for this dataset

johnwlambert:
Screen Shot 2021-09-16 at 3 34 30 PM

door-12

johnwlambert:
Screen Shot 2021-09-16 at 3 35 30 PM

door-12, sift

johnwlambert:
skydio-8, deep front end, before cycle consistency

Screen Shot 2021-09-16 at 3 38 13 PM

after cycle consistency
Screen Shot 2021-09-16 at 3 38 39 PM

akshay-krishnan:
I dont quite understand how the inlier ratio with respect to GT model is greater than that with respect to estimated model. Isnt the estimated model supposed to be the one with the most inliers in our data?

akshay-krishnan:
Im hoping this also convinced you that a min inlier ratio of 0.1 is unacceptable.. :)

akshay-krishnan:
Do we need all 4 of these metrics? They seem to be quite correlated. At some point we have to care about the report being succinct and not confusing with a ton of metrics.

gtsfm/two_view_estimator.py
```diff
@@ -316,6 +316,11 @@ def aggregate_frontend_metrics(
     """
     num_image_pairs = len(two_view_reports_dict.keys())
 
+    inlier_ratio_gt_model_all_pairs = [report.inlier_ratio_gt_model for report in two_view_reports_dict.values()]
+    inlier_ratio_est_model_all_pairs = [report.inlier_ratio_est_model for report in two_view_reports_dict.values()]
+    num_inliers_gt_model_all_pairs = [report.num_inliers_est_model for report in two_view_reports_dict.values()]
+    num_inliers_est_model_all_pairs = [report.num_inliers_gt_model for report in two_view_reports_dict.values()]
```
akshay-krishnan, line 322: I think it's better to iterate over two_view_reports_dict once instead of four times. Not a bottleneck, but seems a bit odd.

johnwlambert: thanks for taking a look @akshay-krishnan.

I agree it's a bit weird, but it was 4 lines instead of 9, so i sort of like it. But I'm good with whatever you prefer here

```python
inlier_ratio_gt_model_all_pairs = []
inlier_ratio_est_model_all_pairs = []
num_inliers_gt_model_all_pairs = []
num_inliers_est_model_all_pairs = []
for report in two_view_reports_dict.values()
    inlier_ratio_gt_model_all_pairs.append(report.inlier_ratio_gt_model)
    inlier_ratio_est_model_all_pairs.append(report.inlier_ratio_est_model)
    num_inliers_gt_model_all_pairs.append(report.num_inliers_est_model)
    num_inliers_est_model_all_pairs.append(report.num_inliers_gt_model)
```
vs.

```python
    inlier_ratio_gt_model_all_pairs = [report.inlier_ratio_gt_model for report in two_view_reports_dict.values()]
    inlier_ratio_est_model_all_pairs = [report.inlier_ratio_est_model for report in two_view_reports_dict.values()]
    num_inliers_gt_model_all_pairs = [report.num_inliers_est_model for report in two_view_reports_dict.values()]
    num_inliers_est_model_all_pairs = [report.num_inliers_gt_model for report in two_view_reports_dict.values()]
```

johnwlambert:
> I dont quite understand how the inlier ratio with respect to GT model is greater than that with respect to estimated model. Isnt the estimated model supposed to be the one with the most inliers in our data?

It is weird that the GT model inlier ratio is higher than the est. model inlier ratio. In seems to show that the ground truth explains the matches better than what we estimated.

 I think there are two plausible explanations:
(1) we didn't run enough RANSAC iterations to find the best model
(2) thresholds arent strict enough

More unlikely explanations are that:
(3) GT is wrong
(4) we are computing our statistics wrong

gtsfm/two_view_estimator.py
johnwlambert: I agree we are getting a lot of metrics now, but I think these metrics are some of the most important for further improvement, so I vote we include them

akshay-krishnan: > I agree it's a bit weird, but it was 4 lines instead of 9, so i sort of like it. But I'm good with whatever you prefer here

Isnt it better to have a 4x computational improvement than a 2x code length improvement? In my opinion, it is not  even a comparison we should be making.

akshay-krishnan:
> > I dont quite understand how the inlier ratio with respect to GT model is greater than that with respect to estimated model. Isnt the estimated model supposed to be the one with the most inliers in our data?
> 
> It is weird that the GT model inlier ratio is higher than the est. model inlier ratio. In seems to show that the ground truth explains the matches better than what we estimated.
> 
> I think there are two plausible explanations:
> (1) we didn't run enough RANSAC iterations to find the best model
> (2) thresholds arent strict enough
> 
> More unlikely explanations are that:
> (3) GT is wrong
> (4) we are computing our statistics wrong

Following up, 
(1) How many do we run? We can theoretically quantify how many iterations are to be run to find a good model with with at least x inlier ratio with a likelihood of p. Then in a few cases, we may get suboptimal models, but we seem to be dealing with many cases of this happening. 
(2) The threshold to define an inlier? Are we not using the same thresholds for inliers in the metrics as we do in RANSAC? 

(3) Doesnt look like it, because it seems to have higher inlier ratio. 
(4) We have unit tests, right?

gtsfm/two_view_estimator.py
johnwlambert: @akshay-krishnan that's a good point. I updated it, should be cleaner now.

johnwlambert: turns out the weird phenomenon was due to a simple bug:
```
    num_inliers_gt_model_all_pairs = [report.num_inliers_est_model for report in two_view_reports_dict.values()]
    num_inliers_est_model_all_pairs = [report.num_inliers_gt_model for report in two_view_reports_dict.values()]
```
incorrect assignment

gtsfm/configs/deep_front_end.yaml
```diff
@@ -12,7 +12,7 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    eval_threshold_px: 3 # in px
+    eval_threshold_px: 4 # in px
```
johnwlambert, line 15: setting eval and estimation thresholds to be the same, for now

gtsfm/configs/scene_optimizer_unit_test_config.yaml
```diff
@@ -13,7 +13,7 @@ SceneOptimizer:
 
   two_view_estimator:
     _target_: gtsfm.two_view_estimator.TwoViewEstimator
-    eval_threshold_px: 3 # in px
+    eval_threshold_px: 0.5 # in px
```
johnwlambert, line 16: est threshold is 0.5 for this unit test config

PR #308, opened 2021-09-19
Adding retries for wget

ayushbaid:
Current benchmarks in CI are flaky because of download failures. These retries might help.

Reference: https://github.com/borglab/gtsam-manylinux-build/blob/master/build-macos-new.sh#L35

### DIFF
--- a/.github/scripts/execute_single_benchmark.sh
+++ b/.github/scripts/execute_single_benchmark.sh
@@ -14,7 +14,7 @@ function retry {
   shift
 
   local count=0
-  until "$@"; do
+  until "$@"; do # actual command execution happening here, and will continue till signal 0 (success).
     exit=$?
     wait=$((2 ** $count))
     count=$(($count + 1))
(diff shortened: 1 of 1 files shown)


.github/scripts/execute_single_benchmark.sh
```diff
@@ -9,6 +9,26 @@ DATASET_SRC=$5
 LOADER_NAME=$6
 MAX_RESOLUTION=$7
 
+function retry {
+  local retries=$1
+  shift
+
+  local count=0
+  until "$@"; do
```
akshay-krishnan, line 17: I think this line is executing the actual command and if that command succeeds it gets the 0 signal.

akshay-krishnan: Maybe you want to document that.

akshay-krishnan (approved):

akshay-krishnan:
The unit tests seem to pass, so maybe this fixes it?

akshay-krishnan:
Also we should not block PRs by waiting 5h for the deep front-end. Maybe that can be changed in a different PR.

.github/scripts/execute_single_benchmark.sh
ayushbaid: Added

travisdriver (approved):
LGTM

PR #311, opened 2021-09-20
Save another full front-end report after cycle-consistency checks

johnwlambert:
This report tells us which front-end error slips through cycle-consistency, and we can use it for more analysis about noise that is sneaking into the backend.

### DIFF
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -160,7 +160,11 @@ class SceneOptimizer:
                 )
 
         # persist all front-end metrics and its summary
-        auxiliary_graph_list.append(dask.delayed(save_full_frontend_metrics)(two_view_reports_dict, image_graph))
+        auxiliary_graph_list.append(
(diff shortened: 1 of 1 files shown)


gtsfm/scene_optimizer.py
```diff
@@ -313,13 +324,14 @@ def save_metrics_reports(metrics_graph_list: Delayed) -> List[Delayed]:
 
 
 def save_full_frontend_metrics(
-    two_view_report_dict: Dict[Tuple[int, int], TwoViewEstimationReport], images: List[Image]
+    two_view_report_dict: Dict[Tuple[int, int], TwoViewEstimationReport], images: List[Image], save_fname: str
```
akshay-krishnan, line 327: super nit: the function already has _save_ in it, so "filename" is sufficient for the argument name.

akshay-krishnan (approved):

gtsfm/scene_optimizer.py
johnwlambert: good point, updated now

PR #207, opened 2021-07-02
Add baseline check for triangulation

ayushbaid:

PR #317, opened 2021-09-24
Add computation of triangulation angle between two rays

johnwlambert:
Was previously implemented by @ayushbaid, but closed in https://github.com/borglab/gtsfm/pull/207/files

Now useful for MVS initialization for @codyly


gtsfm/utils/triangulation.py
```diff
@@ -0,0 +1,42 @@
...
+
+Authors: Ayush Baid
+"""
+import numpy as np
+from gtsam import PinholeCameraCal3Bundler, Unit3
+
+import gtsfm.utils.geometry_comparisons as geometry_utils
+
+
+def calculate_triangulation_angle_in_degrees(
```
akshay-krishnan, line 11: I think this could be in geometry_utils too, unless you think we would have more functions here. A util file just for one function would be odd.

akshay-krishnan (approved):

codyly:
I have already included this function in `gtsfm/densify/mvs_utils.py` and a unit test in `test/densify/test_mvs_utils.py` (implemented by Ayush in the previous PR). I guess these files should also be updated after this PR merged.

codyly (approved):

johnwlambert:
Closing, as relevant code was included in https://github.com/borglab/gtsfm/pull/426

PR #65, opened 2020-12-28
switching to GeneralSFMFactor2 and adding option for shared intrinsics

ayushbaid:
This PR depends on https://github.com/borglab/gtsam/pull/644

PR #319, opened 2021-09-27
adding docstring and reverting unnessary changes

ayushbaid:
Improving dosctring and removing unneccesary changes

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -12,12 +12,12 @@ jobs:
           [
             "sift_front_end,  door-12,               12,  JPG,  test_data,  olsson-loader,  1296, true",
             "deep_front_end,  door-12,               12,  JPG,  test_data,  olsson-loader,  1296, true",
-            "sift_front_end,  skydio-8,              8,   jpg,  gdrive ,    colmap-loader,  760, true",
-            "deep_front_end,  skydio-8,              8,   jpg,  gdrive,     colmap-loader,  760, true",
-            "sift_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true",
-            "deep_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true",
-            "sift_front_end,  palace-fine-arts-281,  25,  jpg,  wget,       olsson-loader,  320, true",
-            "deep_front_end,  notre-dame-20,         20,  jpg,  gdrive,     colmap-loader,  760, false",
+            "sift_front_end,  skydio-8,              8,   jpg,  gdrive ,    colmap-loader,  760,  true",
+            "deep_front_end,  skydio-8,              8,   jpg,  gdrive,     colmap-loader,  760,  true",
+            "sift_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760,  true",
+            "deep_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760,  true",
+            "sift_front_end,  palace-fine-arts-281,  25,  jpg,  wget,       olsson-loader,  320,  true",
+            "deep_front_end,  notre-dame-20,         20,  jpg,  gdrive,     colmap-loader,  760,  false",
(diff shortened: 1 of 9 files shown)


gtsfm/bundle/bundle_adjustment.py
```diff
@@ -96,7 +96,7 @@ def run(
             logger.error(
                 "Bundle adjustment aborting, optimization cannot be performed without any tracks or any cameras."
             )
-            return initial_data, GtsfmMetricsGroup()
+            return initial_data
```
ayushbaid, line 99: Added in #65, but not present in master before that

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -178,7 +178,7 @@ def run(
         except Exception:
             logger.exception("LM Optimization failed")
             # as we did not perform the bundle adjustment, we skip computing the total reprojection error
-            return GtsfmData(initial_data.number_images()), GtsfmMetricsGroup()
+            return GtsfmData(initial_data.number_images())
```
ayushbaid, line 181: Added in #65, but not present in master before that

gtsfm/bundle/bundle_adjustment.py
johnwlambert, line 99: thanks for the PR @ayushbaid. Don't we need to return two things though?

.github/workflows/benchmark.yml
```diff
@@ -12,12 +12,12 @@ jobs:
...
-            "sift_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true",
-            "deep_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true",
-            "sift_front_end,  palace-fine-arts-281,  25,  jpg,  wget,       olsson-loader,  320, true",
-            "deep_front_end,  notre-dame-20,         20,  jpg,  gdrive,     colmap-loader,  760, false",
+            "sift_front_end,  skydio-8,              8,   jpg,  gdrive ,    colmap-loader,  760,  true",
+            "deep_front_end,  skydio-8,              8,   jpg,  gdrive,     colmap-loader,  760,  true",
+            "sift_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760,  true",
+            "deep_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760,  true",
+            "sift_front_end,  palace-fine-arts-281,  25,  jpg,  wget,       olsson-loader,  320,  true",
+            "deep_front_end,  notre-dame-20,         20,  jpg,  gdrive,     colmap-loader,  760,  false",
```
johnwlambert, line 20: extra space needed on the last line as well, and first two lines, since comma followed by two spaces is the delimiter?

gtsfm/bundle/bundle_adjustment.py
ayushbaid: We do, but we did not have consensus on what to return. `GtsfmMetricsGroup` has args in the constructor which are compulsory. Either way this will break when we encounter this error.

I did not want this change to go in with my shared intrinsics PR.

.github/workflows/benchmark.yml
ayushbaid: Oh ok. I thought it was just for prettifying.

johnwlambert:
Hi @ayushbaid, thanks for this PR. Looks like the CI is failing due to 
```
gtsfm/runner/run_scene_optimizer_astronet.py:5:1: F401 'argparse.Namespace' imported but unused
```
Could you update that one line?

.github/workflows/benchmark.yml
johnwlambert: could you add the extra space for sed to split it?

johnwlambert: @ayushbaid gentle reminder about this comment

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Thanks @ayushbaid. I think we should just return `None` if we don't know what to return for the second arg. But currently this change will not pass `mypy`

PR #321, opened 2021-09-29
Using delayed superpoint model for inference

ayushbaid:
WIP

### DIFF
--- a/gtsfm/frontend/detector_descriptor/superpoint.py
+++ b/gtsfm/frontend/detector_descriptor/superpoint.py
@@ -10,10 +10,12 @@ References:
 Authors: Ayush Baid
 """
 from pathlib import Path
-from typing import Tuple, Union
+from typing import Any, Dict, Optional, Tuple, Union
 
+import dask
 import numpy as np
 import torch
+from dask.delayed import Delayed
 
 import gtsfm.utils.images as image_utils
 from gtsfm.common.image import Image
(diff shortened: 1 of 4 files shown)


gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -73,3 +75,14 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
         keypoints = Keypoints(feature_points, responses=scores, scales=np.ones(scores.shape))
 
         return keypoints, descriptors
+
+    def get_delayed_extra_args(self) -> Delayed:
```
johnwlambert, line 79: thanks for the PR, @ayushbaid. Could we add a docstring here?

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -39,19 +41,19 @@ def __init__(self, use_cuda: bool = True, weights_path: Union[Path, str] = MODEL
...
-        device = torch.device("cuda" if self._use_cuda else "cpu")
-        model = SuperPoint(self._config).to(device)
-        model.eval()
+        model = extra_args["model"]
         # TODO: fix inference issue #110
 
         image_tensor = torch.from_numpy(
             np.expand_dims(image_utils.rgb_to_gray_cv(image).value_array.astype(np.float32) / 255.0, (0, 1))
-        ).to(device)
+        ).to("cpu")
```
johnwlambert, line 56: why always to CPU? won't this lead to performance degredation?

johnwlambert:
@ayushbaid Thanks for working on this.

Could you add an analysis of the execution time before and after on your machine, with this change? Could you add a few more details to the PR description too, with a bit more context about the rationale behind the changes?

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -73,3 +75,14 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
         keypoints = Keypoints(feature_points, responses=scores, scales=np.ones(scores.shape))
 
         return keypoints, descriptors
+
+    def get_delayed_extra_args(self) -> Delayed:
+        if self._extra_args is None:
+            model = SuperPoint(self._config).cpu()
```
adisingh50, line 81: Same question as John. Why force CPU instead of checking if `torch.cuda_is_available()`?

ayushbaid: Because we are using the delayed version of the model. We cannot map the model to a GPU and pass it around to different workers. If GPU processing is to be done, we need to do the mapping on the worker code,  i.e. in the `detect_and_describe_function`.

ayushbaid:
Update: I am still trying to find the memory leak in superglue. I want to fix inference for both the networks as superglue is the primary consumer of compute.

PR #327, opened 2021-10-01
Fixing Camera Frustum Rendering

adisingh50:
Before Fix:
image

After Fix:
image

Both Visualiztions in React Three Fiber

### DIFF
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -230,9 +230,9 @@ class SceneOptimizer:
         auxiliary_graph_list.extend(save_metrics_reports(metrics_graph_list))
 
         # # Modify BA input and BA output to have point clouds and frustums aligned with x,y,z axes.
-        # ba_input_graph, ba_output_graph, gt_pose_graph = dask.delayed(align_estimated_gtsfm_data, nout=3)(
-        #     ba_input_graph, ba_output_graph, gt_pose_graph
-        # )
+        ba_input_graph, ba_output_graph, gt_pose_graph = dask.delayed(align_estimated_gtsfm_data, nout=3)(
+            ba_input_graph, ba_output_graph, gt_pose_graph
+        )
 
         if self._save_3d_viz:
             auxiliary_graph_list.extend(save_visualizations(ba_input_graph, ba_output_graph, gt_pose_graph))
--- a/tests/utils/test_ellipsoid_utils.py
+++ b/tests/utils/test_ellipsoid_utils.py
@@ -3,6 +3,9 @@
 Authors: Adi Singh
 """
 import unittest
+from gtsfm.common.gtsfm_data import GtsfmData
+from gtsam import Cal3Bundler, PinholeCameraCal3Bundler, Pose3, Rot3, SfmTrack, Similarity3
+from gtsfm.utils.ellipsoid import get_ortho_axis_alignment_transform
 
 import numpy as np
 import numpy.testing as npt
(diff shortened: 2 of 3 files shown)


tests/utils/test_ellipsoid_utils.py
```diff
@@ -13,6 +16,93 @@
...
+
+        # Add 6 tracks to sample_data
+        track1 = SfmTrack(np.array([1, 1, 0]))
+        track2 = SfmTrack(np.array([-1, 1, 0]))
+        track3 = SfmTrack(np.array([-2, 2, 0]))
+        track4 = SfmTrack(np.array([-1, -1, 0]))
+        track5 = SfmTrack(np.array([1, -1, 0]))
+        track6 = SfmTrack(np.array([2, -2, 0]))
+
+        sample_data.add_track(track1)
```
johnwlambert, line 59: looks great. can we do it in a for-loop to prevent any copy-paste? Like I'm thinking
```python
points3d = np.array(
    [
          [1, 1, 0],
          [-1, 1, 0],
         [-2, 2, 0],
         [-1, -1, 0],
        [1, -1, 0],
        [2, -2, 0]
    ]
)
for pt_3d in points_3d:
      sample_data.add_track(SfmTrack(pt_3d))
```
    
```

File: tests/utils/test_ellipsoid_utils.py [hunk 2]
```diff
@@ -13,6 +16,93 @@
...
+                [-np.sqrt(2), 0, 0],
+                [-2 * np.sqrt(2), 0, 0],
+            ]
+        )
+        npt.assert_almost_equal(computed_3d_points, expected_3d_points, decimal=6)
+
+        # Verify correct camera poses.
+        computed_camera_poses = sample_data.get_camera_poses()
+
+        computed_cam0_pose = computed_camera_poses[0]
```
johnwlambert, line 88: maybe down here we can make a bit more concise doing something like
```python
expected_wTi_list = [
     Pose3(walignedTw.rotation(), np.array([np.sqrt(2), 0, 2]),
     Pose3(walignedTw.rotation(), np.array([0, -np.sqrt(2), 2])),
     Pose3(walignedTw.rotation(), np.array([-np.sqrt(2), 0, 2]))
]
wTi_list = sample_data.get_camera_poses():
for wTi, wTi_expected in zip(wTi_list, expected_wTi_list):
     assert wTi == wTi_expected
```

tests/utils/test_ellipsoid_utils.py
adisingh50: Yup. This looks much cleaner.

tests/utils/test_ellipsoid_utils.py
adisingh50: Ok, comparing the Pose3 objects themselves between computed and expected looks a lot better than individually comparing each camera's translation and rotation separately. thanks

gtsfm/scene_optimizer.py
```diff
@@ -231,15 +231,16 @@ def _filter_dict_keys(dict: Dict[Any, Any], ref_dict: Dict[Any, Any]) -> Dict[An
         # Save metrics to JSON and generate HTML report.
         auxiliary_graph_list.extend(save_metrics_reports(metrics_graph_list))
 
-        # # Modify BA input and BA output to have point clouds and frustums aligned with x,y,z axes.
-        # ba_input_graph, ba_output_graph, gt_pose_graph = dask.delayed(align_estimated_gtsfm_data, nout=3)(
-        #     ba_input_graph, ba_output_graph, gt_pose_graph
-        # )
+        # Modify BA input, BA output, and GT poses to have point clouds and frustums aligned with x,y,z axes.
+        gt_poses_graph = (
```
adisingh50, line 235: needed to move `gt_poses_graph` up in the code so that the function call to `align_estimated_gtsfm_data` has access to it

tests/utils/test_ellipsoid_utils.py
```diff
@@ -13,6 +16,88 @@
...
+        # fmt: off
+        points3d = np.array(
+            [
+                [1, 1, 0],
+                [-1, 1, 0],
+                [-2, 2, 0],
+                [-1, -1, 0],
+                [1, -1, 0],
+                [2, -2, 0],
+                [5, 5, 0]
```
adisingh50, line 62: added this last point as an outlier point to be filtered out inside the ellipsoid utils wrapper function

adisingh50 (commented):
Did some debugging with PDB to determine the difference between 

- Approach 1: first filtering then centering
- Approach 2: first centering then filtering

Here's what I found when I tested out both approaches on the `test_get_ortho_axis_alignment_transform` unit test function in `test_ellipsoid_utils.py`. Approach 1 passes the unit test and approach 2 failes the unit test.

When first filtering and then centering:
`wuprightRw`: 
` [  -sqrt(2)/2  sqrt(2)/2    0  ]`
` [  -sqrt(2)/2  -sqrt(2)/2   0 ]`
` [  0               0                1  ]`

`mean`: `[0, 0, 0]`

When first centering and then filtering:
`wuprightRw`: 
` [  -sqrt(2)/2  sqrt(2)/2    0  ]`
` [  sqrt(2)/2  sqrt(2)/2   0 ]`
` [  0               0                1  ]`

`mean`: `[0.714, 0.714, 0]`


So clearly, the order of operations matters in the final computation of `walignedTw`. Approach 2 has the wrong translation primarily because it calculates the mean based off the raw point cloud (the point (5,5,0) skews the mean). Whereas approach 1 properly filters out the (5,5,0) outlier to begin with, and then computes the mean on the proper collection of points. Approach 2 yields the point cloud visualized under the caption "Before Fix" and approach 1 yields the point cloud visualized under the caption "After Fix".

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,203 @@
...
+
+        sample_data = GtsfmData(number_images=12)
+
+        # Read all 12 camera poses from data.mat file which contains extrinsics for all cameras in door12 dataset.
+        data = loadmat("data.mat")
+        M_list = [data["P"][0][i] for i in range(12)]
+        K = M_list[0][:3, :3]
+        Kinv = np.linalg.inv(K)
+        iTw_list = [Kinv @ M_list[i] for i in range(12)]
+        wTi_list = [Pose3(Rot3(iTw[:3, :3]), iTw[:, 3]).inverse() for iTw in iTw_list]
```
johnwlambert, line 118: Hi @adisingh50, thanks for the PR. It looks like some of this I/O code here is borrowed from the OlssonLoader. Can we just instantiate the OlssonLoader on the path to the Door test data (like we do in other unit tests), and use that object to get the poses you need?

adisingh50: Yes, instantiating the OlssonLoader would certainly be cleaner and may fix the unit test error I was originally getting from directly reading from `data.mat`

johnwlambert: let's follow the example here: https://github.com/borglab/gtsfm/blob/master/tests/test_scene_optimizer.py#L27

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,206 @@
...
+        M = camera_translations.shape[0]
+        N = points_3d.shape[0]
+
+        relative_distances = np.zeros((M, N))
+
+        for camera_index in range(0, M):
+            for point_index in range(0, N):
+                camera_center = camera_translations[camera_index, :]
+                point = points_3d[point_index, :]
+
```
johnwlambert, line 200: nit: prefer no extra new line between 199 and 201

tests/utils/test_ellipsoid_utils.py
johnwlambert: marking as resolved, thanks

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,206 @@
...
+        wTi_list = self.loader._wTi_list
+
+        # Add 12 camera frustums to sample_data.
+        default_intrinsics = Cal3Bundler(fx=100, k1=0, k2=0, u0=0, v0=0)
+        for idx, pose in enumerate(wTi_list):
+            camera = PinholeCameraCal3Bundler(pose, default_intrinsics)
+            sample_data.add_camera(idx, camera)
+
+        # fmt: off
+        points_3d = np.array(
```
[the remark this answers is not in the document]
adisingh50: sure

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,206 @@
...
+
+        c = point at (xi,yi,0) with a camera frustum at (xi,yi,2)
+        o = point at (xi,yi,0)
+        """
+
+        sample_data = GtsfmData(number_images=3)
+        default_intrinsics = Cal3Bundler(fx=100, k1=0, k2=0, u0=0, v0=0)
+
+        # Add 3 camera frustums to sample_data (looking down at z=0 plane)
+        camera0 = PinholeCameraCal3Bundler(Pose3(Rot3(np.eye(3)), np.array([-1, 1, 2])), default_intrinsics)
```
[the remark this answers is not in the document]
johnwlambert: marking as resolved, thanks

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,206 @@
...
+        computed_wTi_list = sample_data.get_camera_poses()
+        for wTi_computed, wTi_expected in zip(computed_wTi_list, expected_wTi_list):
+            assert wTi_computed.equals(wTi_expected, tol=1e-9)
+
+    def test_point_cloud_cameras_locked(self) -> None:
+        """Tests the get_ortho_axis_alignment_transform() function with a GtsfmData object containing 11 point cloud
+        points and 12 camera frustums from the door-12 dataset. Determines if the points are properly "locked" in
+        with one another before and after the alignment transformation is applied.
+        """
+
```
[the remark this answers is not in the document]
johnwlambert: resolved, thanks

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,203 @@
...
+        final_relative_distances = self.compute_relative_distances(
+            transformed_camera_translations, transformed_points_3d
+        )
+
+        npt.assert_almost_equal(final_relative_distances, initial_relative_distances, decimal=6)
+
+    def compute_relative_distances(
+        self, camera_translations: np.ndarray, points_3d: np.ndarray
+    ) -> Dict[Tuple[int, int], float]:
```
johnwlambert, line 171: return type should be `np.ndarray` I believe Adi

adisingh50: oops. I forgot to update that. Thanks for catching it

tests/utils/test_ellipsoid_utils.py
```diff
@@ -2,17 +2,201 @@
...
+        walignedTw = Similarity3(R=walignedTw.rotation(), t=walignedTw.translation(), s=1.0)
+        sample_data = sample_data.apply_Sim3(walignedTw)
+
+        # Aggregate the final, transformed points
+        num_tracks = sample_data.number_tracks()
+        transformed_points_3d = [np.array(sample_data.get_track(i).point3()) for i in range(num_tracks)]
+        transformed_points_3d = np.array(transformed_points_3d)
+        transformed_camera_translations = np.array([pose.translation() for pose in sample_data.get_camera_poses()])
+
+        final_relative_distances = self.compute_relative_distances(
```
akshay-krishnan, line 163: why only distances? do the points have a rigid transform relative to the cameras?

PR #330, opened 2021-10-02
add coordinate frame plotting with open3d

johnwlambert:
Very useful for debugging, so I added it to the Open3d visualization utilities.

### DIFF
--- a/visualization/open3d_vis_utils.py
+++ b/visualization/open3d_vis_utils.py
@@ -1,5 +1,5 @@
 """
-Utilities for rendering camera frustums and 3d point clouds using the Open3d library.
+Utilities for rendering camera frustums, 3d point clouds, and coordinate frames using the Open3d library.
 
 Authors: John Lambert
 """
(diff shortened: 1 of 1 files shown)


visualization/open3d_vis_utils.py
```diff
@@ -67,7 +67,7 @@ def create_colored_spheres_open3d(
 
 
 def create_all_frustums_open3d(wTi_list: List[Pose3], calibrations: List[Cal3Bundler]) -> List[open3d.geometry.LineSet]:
-    """Render camera frustums as collections of line segments, using Open3d.
+    """Render camera frustums as collections of line segments, using Open3d. Colored red-to-green by trajectory order.
```
akshay-krishnan, line 70: What is trajectory order for unordered collections?

akshay-krishnan (approved):

visualization/open3d_vis_utils.py
johnwlambert: Trajectory order for unordered collections would just be the order by which the images are numbered, when we read the from disk.

johnwlambert: I added a comment to explain this a bit more @akshay-krishnan

PR #331, opened 2021-10-02
Add separate file for plane utils

johnwlambert:
A utility to sample 3d points on a 3d plane is useful for many unit tests in various files, so making it its own utility.


ayushbaid (approved):
LGTM. Please merge after tests are successful

akshay-krishnan:
I think plane utils is super specific for a utils file. Do you expect more functions to go into it? We should not have a utils file for just one function

gtsfm/utils/plane.py
```diff
@@ -0,0 +1,41 @@
...
+"""
+from typing import Tuple
+
+import numpy as np
+
+
+def sample_points_on_plane(
+    plane_coefficients: Tuple[float, float, float, float],
+    range_x_coordinate: Tuple[float, float],
+    range_y_coordinate: Tuple[float, float],
```
akshay-krishnan, line 13: range_x and range_y are shorter names without much loss of readability.

gtsfm/utils/plane.py
```diff
@@ -0,0 +1,41 @@
...
+    num_points: int,
+) -> np.ndarray:
+    """Sample random points on a 3D plane ax + by + cz + d = 0.
+
+    Args:
+        plane_coefficients: coefficients (a,b,c,d) of the plane equation.
+        range_x_coordinate: desired range of the x coordinates of samples.
+        range_y_coordinate: desired range of the y coordinates of samples.
```
akshay-krishnan, line 21: This is kind of weird, how is this used? I'd imagine a use for limiting the range within the plane's frame of reference, but in the global frame, not so much. The points could be very far apart, depending on the slope of the plane.

johnwlambert:
> I think plane utils is super specific for a utils file. Do you expect more functions to go into it? We should not have a utils file for just one function

Thanks for taking a look @akshay-krishnan. Wasn’t sure where else to put it though, since it doesn’t belong to any other module. We might add some more stuff to it later I suppose. Not sure. Was hoping to use this in some new unit tests Im making for homography.

akshay-krishnan (changes requested):
I think we should have more generic name for the utils file, so that we could put more functions here. Maybe geometry_utils, but that is a little too generic.

gtsfm/utils/plane.py
johnwlambert: good point, i'll update that

gtsfm/utils/plane.py
johnwlambert: so the main way we're using it is by assuming that z is the only truly free variable, and x and y should be sampled within some range.

gtsfm/utils/plane.py
johnwlambert: updated now

johnwlambert:
> I think we should have more generic name for the utils file, so that we could put more functions here. Maybe geometry_utils, but that is a little too generic.

True. I just haven't been able to think of something better. How about we stick with `utils/plane.py` for now, and then re-organize later when necessary? I think technically almost all of the util files could be considered geometry-related, so I agree that's a bit too generic.

PR #333, opened 2021-10-04
Add script to collate and merge gtsfm reports

johnwlambert:
Allows side by side comparisons of tables only (not plots).

### DIFF
--- a/gtsfm/evaluation/merge_reports.py
+++ b/gtsfm/evaluation/merge_reports.py
@@ -9,12 +9,12 @@ from typing import Any, Dict, List, Tuple
 
 from tabulate import tabulate
 
-GTSFM_REPORT_TABLES = Dict[str, Dict[str, Any]]
+SINGLE_REPORT_TABLES = Dict[str, Dict[str, Any]]
 # contains info about each metric, from two separate reports
-GTSFM_MERGED_REPORT_TABLES = Dict[str, List[Tuple[str, Any, Any]]]
+MERGED_REPORT_TABLES = Dict[str, List[Tuple[str, Any, Any]]]
 
 
-def extract_tables_from_report(report_fpath: str) -> GTSFM_REPORT_TABLES:
+def extract_tables_from_report(report_fpath: str) -> SINGLE_REPORT_TABLES:
     """Given an HTML text file containing HTML table info (and Plotly rendering code), rip out the table information
     for each table.
 
@@ -61,7 +61,7 @@ def extract_tables_from_report(report_fpath: str) -> GTSFM_REPORT_TABLES:
     return table_dict
 
 
-def merge_tables(table_dict1: GTSFM_REPORT_TABLES, table_dict2: GTSFM_REPORT_TABLES) -> GTSFM_MERGED_REPORT_TABLES:
+def merge_tables(table_dict1: SINGLE_REPORT_TABLES, table_dict2: SINGLE_REPORT_TABLES) -> MERGED_REPORT_TABLES:
     """Combine 2-column tables from single reports, to 3-column tables, with columns corresponding to each report."""
 
     new_table_dict = defaultdict(list)
@@ -74,7 +74,7 @@ def merge_tables(table_dict1: GTSFM_REPORT_TABLES, table_dict2: GTSFM_REPORT_TAB
     return new_table_dict
 
 
-def print_tables_tabulate(merged_table_dict: GTSFM_MERGED_REPORT_TABLES, table_format: str) -> None:
+def print_tables_tabulate(merged_table_dict: MERGED_REPORT_TABLES, table_format: str) -> None:
     """Dump to stdout Markdown-formatted tables."""
     for tab_name, tab_metrics in merged_table_dict.items():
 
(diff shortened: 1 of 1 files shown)


gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,99 @@
...
+
+    return new_table_dict
+
+
+def main() -> None:
+    """ """
+
+    report_fpath1 = "/Users/jlambert/Downloads/prev_master/metrics-deep_front_end____skydio-32_______________32____jpg____gdrive_______colmap-loader____760___true.zip/gtsfm_metrics_report.html"
+
+    report_fpath2 = "/Users/jlambert/Downloads/rcc_median/metrics-deep_front_end____skydio-32_______________32____jpg____gdrive_______colmap-loader____760___true.zip/gtsfm_metrics_report.html"
```
akshay-krishnan, line 83: Remove these, make them command line args?

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,99 @@
...
+
+    """ """  # TODO(johnwlambert): remove spaces from artifact file names
```
akshay-krishnan, line 85: Remove comment if this is done

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,99 @@
...
+    return table_dict
+
+
+def print_tables_tabulate(merged_table_dict: GTSFM_MERGED_REPORT_TABLES) -> None:
+    """Dump to stdout Markdown-formatted tables."""
+    for tab_name, tab_metrics in merged_table_dict.items():
+
+        headers = [f"{tab_name}", "Report 1", "Report 2"]
+        table = tab_metrics
+        print(tabulate(table, headers, tablefmt="github"))
```
akshay-krishnan, line 60: Might be nice to have argument to control whether we need a html or github.

akshay-krishnan (changes requested):

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,99 @@
...
+        # add newline between tables
+        print()
+
+
+def merge_tables(table_dict1: GTSFM_REPORT_TABLES, table_dict2: GTSFM_REPORT_TABLES) -> GTSFM_MERGED_REPORT_TABLES:
+    """Combine 2-column tables from single reports, to 3-column tables, with columns corresponding to each report."""
```
womackj1, line 66: Note: I may need to generalize this code to also combine 3-column tables to 5-column tables for elements with COLMAP comparison.

gtsfm/evaluation/merge_reports.py
johnwlambert: thanks, fixed now

gtsfm/evaluation/merge_reports.py
johnwlambert: thanks, fixed now, works without having to modify the file paths in any way

gtsfm/evaluation/merge_reports.py
johnwlambert: good idea, i added a command line option for that

gtsfm/evaluation/merge_reports.py
johnwlambert: Sure, feel free to add a separate PR into this branch with the changes you propose. Although for Colmap, we don't have an HTML report, so maybe a separate script is needed altogether to get the info

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,107 @@
...
+Authors: John Lambert
+"""
+
+import argparse
+from collections import defaultdict
+from typing import Any, Dict, List, Tuple
+
+from tabulate import tabulate
+
+GTSFM_REPORT_TABLES = Dict[str, Dict[str, Any]]
```
ayushbaid, line 12: I'd prefer the name explicitly convey that is is a table coming from a single report.

Also, think we can drop the GTSFM prefix

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,107 @@
...
+# contains info about each metric, from two separate reports
+GTSFM_MERGED_REPORT_TABLES = Dict[str, List[Tuple[str, Any, Any]]]
```
ayushbaid, line 14: Why does the type of the value differ? I think the only difference should be going from 2-tuple to 3-tuple

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,107 @@
...
+
+
+def extract_tables_from_report(report_fpath: str) -> GTSFM_REPORT_TABLES:
```
ayushbaid, line 17: Have you looked into third-party libraries to do this?

Example: pandas does this out-of-the-box. https://pbpython.com/pandas-html-table.html. We can also merge dataframes pretty easily.

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,107 @@
...
+
+    for tab_name, tab_metrics1 in table_dict1.items():
+        for tab_name1, tab_val1 in tab_metrics1.items():
+            tab_val2 = table_dict2[tab_name].get(tab_name1, None)
+            new_table_dict[tab_name] += [(tab_name1, tab_val1, tab_val2)]
+
+    return new_table_dict
+
+
+def print_tables_tabulate(merged_table_dict: GTSFM_MERGED_REPORT_TABLES, table_format: str) -> None:
```
ayushbaid, line 77: Why not output a html table itself? I would prefer exactly the same structure of the diff-report as the original report, but with more columns and added information in the plots. 

Printing as markdown tables is not a scalable solution as we cannot do the plots in the same way.

gtsfm/evaluation/merge_reports.py
```diff
@@ -0,0 +1,107 @@
...
+    table_dict1 = extract_tables_from_report(report_fpath1)
+    table_dict2 = extract_tables_from_report(report_fpath2)
+    merged_table_dict = merge_tables(table_dict1, table_dict2)
+    print_tables_tabulate(merged_table_dict, output_format)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--report_fpath1", required=True, help="Path to previous report.")
+    parser.add_argument("--report_fpath2", required=True, help="Path to new report.")
```
ayushbaid, line 99: would prefer `report1_fpath` and `report2_fpath`

ayushbaid (changes requested):

gtsfm/evaluation/merge_reports.py
johnwlambert: @ayushbaid sure, we can do that -- dump an HTML file you mean to disk, instead of just printing to stdout?

ayushbaid (approved):

PR #334, opened 2021-10-04
create artifacts for plots and output data in CI benchmark

johnwlambert:
Currently only the result metrics are uploaded after a CI benchmark run. It would be nice to be able to inspect the actual output poses and point cloud, by downloading the artifacts.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -72,23 +72,11 @@ jobs:
             $SHARE_INTRINSICS
 
 
-      - name: Archive dataset metrics
-        uses: actions/upload-artifact@v2
-        with:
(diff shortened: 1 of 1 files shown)


akshay-krishnan:
Would we not want them to be a part of the same zip file? That might be easier to manage, and would also not lead to a lot of artifacts.

johnwlambert:
> Would we not want them to be a part of the same zip file? That might be easier to manage, and would also not lead to a lot of artifacts.

Good point @akshay-krishnan. Let me make that update

johnwlambert:
@akshay-krishnan I put everything into a single zip file per (dataset, front-end) pair. Much cleaner.

akshay-krishnan (approved):

PR #335, opened 2021-10-05
Update cycle consistency verbosity to be false by default

akshay-krishnan:
--verbose flags are generally set to `false` by default. It needs to be true only when debugging, and then it has to be manually specified, potentially via command line args. 

The problem with setting this to true is that generates a lot of text making it hard to find other log statements. It gets worse when the dataset is huge, the logs for the palace of fine arts 281 image dataset generated by the CI is ~65 MB of text.


ayushbaid (approved):
LGTM.

akshay-krishnan:
Thanks @ayushbaid

PR #336, opened 2021-10-05
Prevents other configs from being canceled if one of them fails in the benchmark

akshay-krishnan:
Some configs fail due to flaky wget or unzip, and we dont have to cancel other configs in the benchmark. 
GitHub documentation [here](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast)


ayushbaid (approved):
Thanks for this!!

johnwlambert:
thanks!

PR #338, opened 2021-10-06
Use scipy's axis-angle computation for rel. error computation until we patch GTSAM

johnwlambert:
Use scipy's axis-angle computation for rel. error computation until we patch GTSAM

See https://github.com/borglab/gtsam/issues/886 for more details.

Unblocks us for now in GTSFM.

### DIFF
--- a/gtsfm/utils/geometry_comparisons.py
+++ b/gtsfm/utils/geometry_comparisons.py
@@ -275,7 +275,11 @@ def compute_relative_rotation_angle(R_1: Optional[Rot3], R_2: Optional[Rot3]) ->
         return None
 
     relative_rot = R_1.between(R_2)
-    relative_rot_angle_rad = relative_rot.axisAngle()[1]
+    # TODO(johnwlambert): we are using scipy.spatial.transform as a hotfix until GTSAM axisAngle() is patched.
+    # See https://github.com/borglab/gtsam/issues/886
+    scaled_axis = Rotation.from_matrix(relative_rot.matrix()).as_rotvec()
+    relative_rot_angle_rad = np.linalg.norm(scaled_axis)
+    axis = scaled_axis / relative_rot_angle_rad
     relative_rot_angle_deg = np.rad2deg(relative_rot_angle_rad)
     return relative_rot_angle_deg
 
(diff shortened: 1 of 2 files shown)


tests/utils/test_geometry_comparisons.py
```diff
@@ -221,6 +222,41 @@ def test_compute_relative_rotation_angle(self):
 
         np.testing.assert_allclose(computed_deg, expected_deg, rtol=1e-3, atol=1e-3)
 
+
+    def test_compute_relative_rotation_angle(self) -> None:
```
ayushbaid, line 226: How does this test pass when GTSAM's comparisons are buggy?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -275,7 +275,11 @@ def compute_relative_rotation_angle(R_1: Optional[Rot3], R_2: Optional[Rot3]) ->
         return None
 
     relative_rot = R_1.between(R_2)
-    relative_rot_angle_rad = relative_rot.axisAngle()[1]
+    # TODO(johnwlambert): we are using scipy.spatial.transform as a hotfix until GTSAM axisAngle() is patched.
+    # See https://github.com/borglab/gtsam/issues/886
+    scaled_axis = Rotation.from_matrix(relative_rot.matrix()).as_rotvec()
+    relative_rot_angle_rad = np.linalg.norm(scaled_axis)
+    axis = scaled_axis / relative_rot_angle_rad
```
ayushbaid, line 282: Do we need to cast to scipy rotation? Can we get the rotation vector from gtsam?

tests/utils/test_geometry_comparisons.py
johnwlambert: @ayushbaid So the GTSAM axisAngle *usually* works. Just not for a few edge cases. Over 50,000 image pairs, this can be 100+ completely erroneous values, making our plots look like garbage. Scipy seems to handle these edge cases properly currently. So this unit test here is checking some generic cases, but can’t sample the edge cases. So it passes.

gtsfm/utils/geometry_comparisons.py
johnwlambert: @ayushbaid hmm this is what we were doing before, right? But it was buggy.

ayushbaid (approved):

ProfFan:
@johnwlambert  Please test if the fix works and maybe revert this?

PR #340, opened 2021-10-06
Learning-based outlier rejection

johnwlambert:
Training script for outlier rejection

## Linear Classifier
Linear classifier learned weights: 
```
Outlier (Class 0) weights:  [ 0.47  0.64  0.7   0.23  0.2   0.02 -0.07  0.15 -0.13 -0.08 -0.18]
Inlier (Class 1) weights:  [-0.45 -0.25 -0.69 -0.28  0.03  0.25 -0.17  0.08 -0.06  0.1  -0.19]
```
sorted edge cycle errors [0 - 7] indices, then weight for (inlier_ratio_est_model, num_inliers_est_model, num_participating_cycles)

- more inliers -> tends to class 1 (and less tends to class 0)
- higher cycle errors -> class 0 (And lower cycle errors -> class 1)

inlier/outlier threshold defined at 5 degrees currently.

## 2-layer net
On skydio-501, with 26000 examples split 80/20 on train/val, on val we get
```
F1:0.88 Prec:0.96 Rec:0.82
```
vs `median`: R precision 0.95, R recall 0.51

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -19,6 +19,7 @@ jobs:
             "sift_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true",
             "deep_front_end,  skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true",
             "sift_front_end,  palace-fine-arts-281,  25,  jpg,  wget,       olsson-loader,  320, true",
+            "deep_front_end,  palace-fine-arts-281,  5,   jpg,  wget,       olsson-loader,  760, true",
             "deep_front_end,  notre-dame-20,         20,  jpg,  gdrive,     colmap-loader,  760, false",
             "sift_front_end,  2011205_rc3,           65,  png,  wget,       astronet,       1024, true",
           ]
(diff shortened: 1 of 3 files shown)


johnwlambert:
> Cool results. Can you explain what's going on here a little more in-depth?

Sure. We're looking for ways to reduce the number of outliers that go into Shonan and 1dsfm. But it's quite trickly to reason about the outliers globally. We get increased precision from cycle consistency, but I don't think it's enough. We're using a heuristic on a feature currently (median of cycle errors), but it seems the most informative feature are the # of inliers, # of cycles an edge appears in, and inlier ratio w.r.t. estimated model (plus the cycle errors).

So we can learn the weights of a 1 layer net, or simple 2-layer MLP, and try to predict inlier/outlier.

So far, this seems to be better than just a median (F1 can be much higher). But I still need to measure generalization across datasets.

I don't think an F1 of 88% is high enough, though. I think using the actual point correspondences to reason about 3 views at once (i.e. trifocal tensor) is a stronger way.

johnwlambert:
> Is this ready for review? I saw many methods in your script that needed documentation or some cleanup and thought you would update PR further.

@akshay-krishnan thanks for asking -- not ready for review. Although the performance is better for Skydio-501, I don't think it's a general enough solution. I think we will get more performance with more generalization if we rely on geometry-based point and pose filtering like epipolar point transfer or Trifocal tensor point transfer. I'm going to go ahead and close for now.

PR #343, opened 2021-10-10
Color inliers and outliers (w.r.t. GT poses) green/red, and make CI upload these images as artifacts.

johnwlambert:
Color inliers and outliers (w.r.t. GT poses) green/red, and make CI upload these images as artifacts.

Currently CI does not generate these figures as artifacts.
Currently, locally, even when GT pose is available, we use random colors for each correspondence, which makes it hard to see when Sampson error classifies correspondences correctly.

This provides important info bout where we expect to see outlier 3d points in free space after BA. This "per-correspondence ground truth" also will allow us to start using this information to analyze the purity of our tracks in DA.

Examples from skydio-8, deep front end:

### DIFF
--- a/gtsfm/configs/deep_front_end.yaml
+++ b/gtsfm/configs/deep_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: True
+  save_two_view_correspondences_viz: False
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 
--- a/gtsfm/configs/sift_front_end.yaml
+++ b/gtsfm/configs/sift_front_end.yaml
@@ -1,7 +1,7 @@
 SceneOptimizer:
   _target_: gtsfm.scene_optimizer.SceneOptimizer
   save_gtsfm_data: True
-  save_two_view_correspondences_viz: True
+  save_two_view_correspondences_viz: False
   save_3d_viz: True
   pose_angular_error_thresh: 5 # degrees
 


gtsfm/utils/metrics.py
```diff
@@ -53,6 +53,8 @@ def count_correct_correspondences(
 
     Returns:
         Number of correspondences which are correct.
+        Mask of which verified correspondences are classified as correct under Sampson error
+            (using GT epipolar geometry).
```
akshay-krishnan, line 57: I dont think this is a good API, its like returning a mask and a non-zero count of the mask. The user could just count the nonzeros outside.

gtsfm/two_view_estimator.py
```diff
@@ -271,7 +279,7 @@ def compute_correspondence_metrics(
         epipolar_distance_threshold,
     )
     inlier_ratio_gt_model = num_inliers_gt_model / corr_idxs_i1i2.shape[0]
-    return num_inliers_gt_model, inlier_ratio_gt_model
+    return num_inliers_gt_model, inlier_ratio_gt_model, v_corr_idxs_inlier_mask_gt
```
akshay-krishnan, line 282: even here, ideally we only need to return the mask, num inliers are inlier ratio can be computed outside.

akshay-krishnan (changes requested):

akshay-krishnan:
> The main issue I see is that we turn on the visualization by default, and upload it to CI. I think we should avoid this.

+1, I'll add the reasons: 
- increases CI time, 
- waste of resources (both space and compute) as this runs for all commits on all PRs and we dont have many PRs on the frontend. 
- its not what the CI is meant for, are there any uses to the visualization other than debugging?

johnwlambert:
@ayushbaid @akshay-krishnan thanks for taking a look. I agree that most users wouldnt want the "save 2-view viz" setting turned on locally, since it blows up the disk. But in the CI, it's free. Maybe it should be a CLI arg we can turn on in the CI then?

I think these plots that the CI generates are quite informative. For example, the matches here look great, but Sampson at 4 px for 760p says they're wrong. So this threshold must be bad:


On the other hand, we get a lot of visibility into why this dataset is hard (object symmetry):

travisdriver:
The coloring of inliers and outliers is also in [#295](https://github.com/borglab/gtsfm/pull/295)

ayushbaid (approved):

PR #344, opened 2021-10-10
Use robust shonan

johnwlambert:


gtsfm/averaging/rotation/shonan.py
```diff
@@ -37,7 +37,7 @@ def __init__(self) -> None:
         """
         Note: `p_min` and `p_max` describe the minimum and maximum relaxation rank.
         """
-        self._p_min = 5
+        self._p_min = 30
```
ayushbaid, line 40: Why this change?

travisdriver:
What about the other datasets?

gtsfm/averaging/rotation/shonan.py
johnwlambert: so when we use a robust cost like Huber, we don't climb the Riemannian staircase from SO(5) to SO(30). Instead, we choose a single relaxation rank p and solve for the SO(p) matrices. So pmin must equal pmax

dellaert: Hmmm. Let’s discuss tomorrow

travisdriver:
> Seems like the 2-view BA helps on this dataset. For example, that (0,4) image pair gets to
> 
> ```
>     {
>         "i1": 0,
>         "i2": 4,
>         "i1_filename": "00000000.png",
>         "i2_filename": "00000004.png",
>         "rotation_angular_error": 3.5,
>         "translation_angular_error": 1.76,
>         "num_inliers_gt_model": 812,
>         "inlier_ratio_gt_model": 1.0,
>         "inlier_ratio_est_model": 0.92,
>         "num_inliers_est_model": 812
>     },
> ```
> 
> after 2-view BA

This is great. Keypoint measurements are notoriously noisy at the imaging distances inherent to deep space missions, so it makes sense that the 2-view BA helps a lot in this case.

In this case, let's get [#225](https://github.com/borglab/gtsfm/pull/225) merged as soon as possible.

johnwlambert:
For reference, Skydio-8 is unaffected by robust shonan (vs. traditional Shonan)

PR #346, opened 2021-10-12
Comparing metrics across pipelines side-by-side in columns of metrics report

womackj1:
This adds a new table to the GTSfM metrics report. The rows of the table is information that is shared across common SfM pipelines (like Number of 3D Points, Mean Observations Per Image, Number of Cameras), although currently only GTSfM and COLMAP is supported.


gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+import os
+import json
+from typing import Dict
+
+from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
+
+
+
+def compare_metrics(txt_metric_paths: Dict[str, str], json_path: str) -> None:
+    """
```
johnwlambert, line 14: missing 1-line docstring/explanation

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,60 @@
...
+        # Add 3D point information
+        fpath = os.path.join(txt_metric_paths[pipeline_name], "points3D.txt")
+        with open(fpath, "r") as f:
+            data = f.readlines()
+            points = data[3:]
+            track_lengths = []
+            for point in points:
+                track_lengths.append(len(point.split()[8:]) / 2)
+
+        gtsfm_metrics.append(GtsfmMetric('Number of 3D Points ' + pipeline_name,
```
johnwlambert, line 30: can we reformat this file with `python black`?

akshay-krishnan:
Could you share an example report?

gtsfm/evaluation/compare_metrics.py
womackj1: Added.

gtsfm/evaluation/compare_metrics.py
womackj1: Done

johnwlambert:
> > Could you share an example report?
> 
> Sure thing, here's a screenshot of the modified portion of the report.
> 
> Screen Shot 2021-10-11 at 10 48 17 PM

thanks @womackj1. Can you make the bin edges in the histogram for track lengths be integer valued? this is what we do in gtsfm/data_association/data_association.py

they should be counts of length-2 tracks, length-3 tracks, etc

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,81 @@
+"""Script to log metrics for comparison across SfM pipelines (GTSfM, COLMAP, etc.) as a JSON.
+
+Authors: Jon Womack
+"""
+import os
+import json
+from typing import Dict
```
akshay-krishnan, line 7: I dont think that is the right order, @johnwlambert ?

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,81 @@
...
+    gtsfm_metrics = []
+    for pipeline_name in txt_metric_paths.keys():
+        # Add 3D point information
+        fpath = os.path.join(txt_metric_paths[pipeline_name], "points3D.txt")
+        with open(fpath, "r") as f:
+            data = f.readlines()
+            points = data[3:]
+            track_lengths = []
+            for point in points:
+                track_lengths.append(len(point.split()[8:]) / 2)
```
akshay-krishnan, line 27: its not clear to the reader how this is parsed, could you explain the format of the file being parsed in a comment?

gtsfm/evaluation/metrics_report.py
```diff
@@ -42,10 +42,24 @@ def create_table_for_scalar_metrics(metrics_dict: Dict[str, Union[float, int]])
     Returns:
         Table with scalar metrics and their values in HTML format.
     """
-    table = {"Metric name": list(metrics_dict.keys()), "Value": list(metrics_dict.values())}
+    if "gtsfm" in list(metrics_dict.keys())[0] or "colmap" in list(metrics_dict.keys())[0]:
```
akshay-krishnan, line 45: this is not a clean way to write this code. Ideally, we'd like to have a function that gets colmap metrics and gtsfm metrics separately, why are the two combined into the same metrics dict when they are from different sources?

womackj1: In generate_metrics_report_html(), the tables and plots are created in a for-loop over each metric group, with get_figures_for_metrics(metrics_group), which doesn't support storing those metrics separately.

Would it be cleaner to write an additional function for getting_figures_for_multiple_metrics([metrics_group, ...]), then add the heading and table outside of the loop where the rest of the metric groups' headings and tables are written?

gtsfm/evaluation/compare_metrics.py
womackj1: Yes, fixed.

gtsfm/evaluation/compare_metrics.py
womackj1: Fixed.

gtsfm/evaluation/compare_metrics.py
womackj1: This reader only parses x, y, z, r, g, and b for each point, not information about each point's corresponding track.
Similarly, the reader for images doesn't parse each images 2D point information i.e. POINTS2D[] as (X, Y, POINT3D_ID).

The script now uses the camera parser (it is compatible as is).

Should I modify the readers to parse this additional information and the functions currently using these readers to disregard the additional information?

womackj1:
> > > Could you share an example report?
> > 
> > 
> > Sure thing, here's a screenshot of the modified portion of the report.
> > Screen Shot 2021-10-11 at 10 48 17 PM
> 
> thanks @womackj1. Can you make the bin edges in the histogram for track lengths be integer valued? this is what we do in gtsfm/data_association/data_association.py
> 
> they should be counts of length-2 tracks, length-3 tracks, etc

Screen Shot 2021-10-12 at 7 50 09 PM

Fixed!

gtsfm/evaluation/compare_metrics.py
johnwlambert: Hi @womackj1, let's go ahead and use the official COLMAP python loader then. @travisdriver already wrote the converter to GtsfmData here:
https://github.com/borglab/gtsfm/blob/master/gtsfm/loader/astronet_loader.py#L76

so I propose we move his `colmap2gtsfm` function to `utils/io.py` and read off the tracks out of the `GtsfmData` using the helper methods we already have

womackj1:
Screen Shot 2021-11-02 at 12 02 14 PM
Screen Shot 2021-11-02 at 12 02 30 PM
The latest HTML report.

womackj1:
> > Screen Shot 2021-11-02 at 12 02 14 PM
> > Screen Shot 2021-11-02 at 12 02 30 PM
> > The latest HTML report.
> 
> Nice @womackj1. Can you restrict the sig figs to what we are using in the left column?

The sig figs in the left column seem to be set automatically in the tabulate method. So I've made the whole numbers integers, and floats has 3 sig figs after decimal.

johnwlambert:
Thanks for the PR, @womackj1. It looks like all the jobs in the CI are failing:

```
[2021-11-02 17:15:06,018 INFO geometry_comparisons.py line 131 5212] Sim(3) Translation `atb`: [tx,ty,tz]=[-0.35  0.1   0.45]
[2021-11-02 17:15:06,019 INFO geometry_comparisons.py line 132 5212] Sim(3) Scale `asb`: 6.18
[2021-11-02 17:15:06,019 INFO geometry_comparisons.py line 140 5212] Pose graph Sim(3) alignment complete.
/home/runner/work/gtsfm/gtsfm/gtsfm/utils/reprojection.py:44: RuntimeWarning: Mean of empty slice
  avg_track_reproj_error = np.nanmean(errors)
/home/runner/work/gtsfm/gtsfm/gtsfm/evaluation/metrics.py:174: RuntimeWarning: All-NaN slice encountered
  "min": np.nanmin(data).tolist(),
/home/runner/work/gtsfm/gtsfm/gtsfm/evaluation/metrics.py:175: RuntimeWarning: All-NaN slice encountered
  "max": np.nanmax(data).tolist(),
/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/numpy/lib/nanfunctions.py:1119: RuntimeWarning: All-NaN slice encountered
  r, k = function_base._ureduce(a, func=_nanmedian, axis=axis, out=out,
/home/runner/work/gtsfm/gtsfm/gtsfm/evaluation/metrics.py:177: RuntimeWarning: Mean of empty slice
  "mean": np.nanmean(data).tolist(),
/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/numpy/lib/nanfunctions.py:1670: RuntimeWarning: Degrees of freedom <= 0 for slice.
  var = nanvar(a, axis=axis, dtype=dtype, out=out, ddof=ddof,
/home/runner/work/gtsfm/gtsfm/gtsfm/common/gtsfm_data.py:318: RuntimeWarning: All-NaN slice encountered
  "Min scene reproj error: %.3f", np.nanmin(scene_reproj_errors) if len(scene_reproj_errors) else np.NaN
/home/runner/work/gtsfm/gtsfm/gtsfm/common/gtsfm_data.py:321: RuntimeWarning: Mean of empty slice
  "Avg scene reproj error: %.3f", np.nanmean(scene_reproj_errors) if len(scene_reproj_errors) else np.NaN
/home/runner/work/gtsfm/gtsfm/gtsfm/common/gtsfm_data.py:327: RuntimeWarning: All-NaN slice encountered
  "Max scene reproj error: %.3f", np.nanmax(scene_reproj_errors) if len(scene_reproj_errors) else np.NaN
distributed.worker - WARNING - Compute Failed
Function:  generate_metrics_report_html
args:      ([<gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f064d28e0>, <gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f04967e50>, <gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f04f2f370>, <gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f054ffc40>, <gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f06193e50>, <gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f05302eb0>, <gtsfm.evaluation.metrics.GtsfmMetricsGroup object at 0x7f7f04c316d0>], '/home/runner/work/gtsfm/gtsfm/result_metrics/gtsfm_metrics_report.html')
kwargs:    {}
Exception: 'TypeError("generate_metrics_report_html() missing 2 required positional arguments: \'colmap_files_dirpath\' and \'metric_paths\'")'

  File "gtsfm/runner/run_scene_optimizer_colmaploader.py", line 45, in <module>
    runner.run()
  File "/home/runner/work/gtsfm/gtsfm/gtsfm/runner/gtsfm_runner_base.py", line 106, in run
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/dask/base.py", line 288, in compute
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/dask/base.py", line 570, in compute
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 2693, in get
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 1969, in gather
    return self.sync(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 865, in sync
    return sync(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/utils.py", line 327, in sync
    raise exc.with_traceback(tb)
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/utils.py", line 310, in f
    result[0] = yield future
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 1834, in _gather
TypeError: generate_metrics_report_html() missing 2 required positional arguments: 'colmap_files_dirpath' and 'metric_paths'
```
Could you take a look when you get a chance?

gtsfm/scene_optimizer.py
```diff
@@ -43,8 +43,12 @@
 
 
 # Paths to Save Output in React Folders.
-REACT_METRICS_PATH = Path(__file__).resolve().parent.parent / "rtf_vis_tool" / "src" / "result_metrics"
-REACT_RESULTS_PATH = Path(__file__).resolve().parent.parent / "rtf_vis_tool" / "public" / "results"
+REACT_METRICS_PATH = (
```
akshay-krishnan, line 46: it looks like you have reformatted this file, is this correct and/or necessary?

gtsfm/utils/io.py
```diff
@@ -23,6 +23,10 @@
 from gtsfm.common.image import Image
 from gtsfm.common.sfm_track import SfmTrack2d
 
+from thirdparty.colmap.scripts.python.read_write_model import Camera as ColmapCamera
+from thirdparty.colmap.scripts.python.read_write_model import Image as ColmapImage
+from thirdparty.colmap.scripts.python.read_write_model import Point3D as ColmapPoint3D
```
akshay-krishnan, line 28: from A import X, Y, Z. You can import them all in the same line.

akshay-krishnan (changes requested):
Could you check the settings of your formatter? Ideally, if there are too many formatting changes to be made, we should move them to another PR. But I think these might be because your settings are not the same as Gtsfm

akshay-krishnan:
There are formatting instructions here: https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md#python-style
You might be missing the 120 character per line setting if you used `black` already

gtsfm/utils/io.py
womackj1: Fixed

womackj1:
> There are formatting instructions here: https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md#python-style You might be missing the 120 character per line setting if you used `black` already

I was missing this setting. All the files I modified have been reformatted with `black` and 120 characters per line.

gtsfm/utils/io.py
```diff
@@ -186,7 +236,13 @@ def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
         cam_params = line.split()
         # Note that u0 is px, and v0 is py
         cam_id, model, img_w, img_h, fx, u0, v0 = cam_params[:7]
-        img_w, img_h, fx, u0, v0 = int(img_w), int(img_h), float(fx), float(u0), float(v0)
+        img_w, img_h, fx, u0, v0 = (
```
akshay-krishnan, line 239: why this change?

gtsfm/utils/io.py
```diff
@@ -291,14 +347,28 @@ def write_images(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> N
 
     num_imgs = gtsfm_data.number_images()
     # TODO: compute this (from keypoint data? or from track data?)
```
akshay-krishnan, line 349: remove the TODO if this is done?

gtsfm/utils/io.py
```diff
@@ -291,14 +347,28 @@ def write_images(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> N
...
+
+    image_id_num_measurements = {}
+    for j in range(gtsfm_data.number_tracks()):
+        track = gtsfm_data.get_track(j)
+        for k in range(track.number_measurements()):
+            image_id, uv_measured = track.measurement(k)
+            if image_id not in image_id_num_measurements:
+                image_id_num_measurements[image_id] = 1
+            else:
+                image_id_num_measurements[image_id] += 1
```
akshay-krishnan, line 359: you can avoid this if-else by using a defaultdict

gtsfm/utils/io.py
```diff
@@ -291,14 +347,28 @@ def write_images(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> N
 
     num_imgs = gtsfm_data.number_images()
     # TODO: compute this (from keypoint data? or from track data?)
-    mean_obs_per_img = 0
+
+    image_id_num_measurements = {}
```
akshay-krishnan, line 351: It might be better to have a method in gtsfmdata to get the number of measurements for an image. @johnwlambert ?

gtsfm/utils/io.py
```diff
@@ -311,8 +381,16 @@ def write_images(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> N
...
 
             f.write(f"{i} {qw} {qx} {qy} {qz} {tx} {ty} {tz} {i} {img_fname}\n")
-            # TODO: write out the points2d
-            f.write("TODO\n")
+
+            # write out points2d
+            for j in range(gtsfm_data.number_tracks()):
+                track = gtsfm_data.get_track(j)
+                for k in range(track.number_measurements()):
+                    # process each measurement
```
akshay-krishnan, line 389: nit: change to "write each measurement" ?

gtsfm/utils/io.py
```diff
@@ -159,6 +165,50 @@ def export_model_as_colmap_text(gtsfm_data: GtsfmData, images: List[Image], save
...
+    Args:
+        cameras: dictionary of COLMAP-formatted Cameras
+        images: dictionary of COLMAP-formatted Images
+        points3D: dictionary of COLMAP-formatted Point3Ds
+        return_tracks (optional): whether or not to return tracks
+    Returns:
+        cameras_gtsfm: list of N camera calibrations corresponding to the N images in images_gtsfm
+        images_gtsfm: list of N camera poses when each image was taken
+        img_fnames: file names of images in images_gtsfm
+        sfmtracks_gtsfm: tracks of points in points3D
```
akshay-krishnan, line 184: just a nit: it seems more meaningful (to me) to return these in this order: filenames, images, cameras, tracks. @johnwlambert ?

gtsfm/scene_optimizer.py
```diff
@@ -117,7 +117,10 @@ def create_computation_graph(
         keypoints_graph_list = []
         descriptors_graph_list = []
         for delayed_image in image_graph:
-            (delayed_dets, delayed_descs) = self.feature_extractor.create_computation_graph(delayed_image)
+            (
```
akshay-krishnan, line 120: it looks like this whole file has been reformatted without the 120 character limit, did you change that already?

akshay-krishnan: I dont see any code changes here, just reformatting that may not be intentional.

gtsfm/loader/astronet_loader.py
```diff
@@ -78,9 +75,12 @@ def __init__(
...
         cameras, images, points3d = colmap_io.read_model(path=data_dir, ext=".bin")
-        self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
-            cameras, images, points3d, load_sfmtracks=use_gt_sfmtracks
-        )
+        (
+            self._calibrations,
+            self._wTi_list,
+            img_fnames,
+            self._sfmtracks,
+        ) = io_utils.colmap2gtsfm(cameras, images, points3d, load_sfmtracks=use_gt_sfmtracks)
```
akshay-krishnan, line 83: probably needs to be reformatted with the 120 character limit? When you do that, remove the parenthesis on the return values before reformatting.

gtsfm/loader/astronet_loader.py
```diff
@@ -116,53 +116,11 @@ def __init__(
...
-        if len(points3D) > 0 and load_sfmtracks:
-            sfmtracks_gtsfm = []
-            for point3D in points3D.values():
-                sfmtrack = SfmTrack(point3D.xyz)
-                for (image_id, point2d_idx) in zip(point3D.image_ids, point3D.point2D_idxs):
-                    sfmtrack.add_measurement(image_id_to_idx[image_id], images[image_id].xys[point2d_idx])
-                sfmtracks_gtsfm.append(sfmtrack)
-
-        return cameras_gtsfm, images_gtsfm, img_fnames, sfmtracks_gtsfm
+        logger.info(
```
akshay-krishnan, line 119: probably needs to be reformatted with the 120 character limit?

gtsfm/evaluation/plot_metrics.py
```diff
@@ -29,21 +30,30 @@ def create_metrics_plots_html(json_path: str, output_dir: str) -> None:
         "rotation_cycle_consistency_metrics.json",
         "averaging_metrics.json",
         "data_association_metrics.json",
-        "bundle_adjustment_metrics.json"
+        "bundle_adjustment_metrics.json",
     ]
+    if colmap_files_dirpath != "":
```
akshay-krishnan, line 35: maybe check if the path exists instead?

gtsfm/evaluation/metrics_report.py
```diff
@@ -42,7 +42,31 @@ def create_table_for_scalar_metrics(metrics_dict: Dict[str, Union[float, int]])
     Returns:
         Table with scalar metrics and their values in HTML format.
     """
-    table = {"Metric name": list(metrics_dict.keys()), "Value": list(metrics_dict.values())}
+    table = {
+        "Metric name": list(metrics_dict.keys()),
+        "Value": list(metrics_dict.values()),
+    }
+    return tabulate(table, headers="keys", tablefmt="html")
```
akshay-krishnan, line 49: probably needs to be reformatted with the 120 character limit?

gtsfm/evaluation/metrics_report.py
```diff
@@ -134,13 +169,67 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
                 raise ValueError(f"Metric {metric_name} does not contain a summary.")
             # Add a scalar metric for mean of 1D distributions.
             scalar_metrics["mean_" + metric_name] = value[metrics.SUMMARY_KEY]["mean"]
+            scalar_metrics["median_" + metric_name] = value[metrics.SUMMARY_KEY]["median"]
```
akshay-krishnan, line 172: This might be too verbose, I suggest we use either median or mean.

gtsfm/evaluation/metrics_report.py
```diff
@@ -84,12 +110,22 @@ def create_plots_for_distributions(metrics_dict: Dict[str, Any]) -> str:
...
             histogram = metric_value[metrics.SUMMARY_KEY]["histogram"]
             fig.add_trace(
-                go.Bar(x=list(histogram.keys()), y=list(histogram.values()), name=metric_name), row=row, col=col
+                go.Bar(
+                    x=list(histogram.keys()),
+                    y=list(histogram.values()),
+                    name=metric_name,
+                ),
+                row=row,
+                col=col,
```
akshay-krishnan, line 119: probably needs to be reformatted with the 120 character limit?

gtsfm/evaluation/metrics_report.py
```diff
@@ -84,12 +110,22 @@ def create_plots_for_distributions(metrics_dict: Dict[str, Any]) -> str:
...
             )
         elif "quartiles" in metric_value[metrics.SUMMARY_KEY]:
             # If all values are available, use them to create box plot.
             if metrics.FULL_DATA_KEY in metric_value:
-                fig.add_trace(go.Box(y=metric_value[metrics.FULL_DATA_KEY], name=metric_name), row=row, col=col)
+                fig.add_trace(
+                    go.Box(y=metric_value[metrics.FULL_DATA_KEY], name=metric_name),
+                    row=row,
+                    col=col,
+                )
```
akshay-krishnan, line 128: probably needs to be reformatted with the 120 character limit?

gtsfm/evaluation/metrics_report.py
```diff
@@ -62,9 +86,11 @@ def create_plots_for_distributions(metrics_dict: Dict[str, Any]) -> str:
     # Separate all the 1D distribution metrics.
     for metric, value in metrics_dict.items():
         if isinstance(value, dict):
-            distribution_metrics.append(metric)
+            all_nan_summary = all(v != v for v in value[metrics.SUMMARY_KEY].values())
```
akshay-krishnan, line 89: is there a better (more readable) way to check for nan?

gtsfm/evaluation/metrics_report.py
```diff
@@ -134,13 +169,67 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+            else:
+                scalar_metrics[metric_name] = value
+        all_scalar_metrics.append(scalar_metrics)
+    table = create_table_for_scalar_metrics_and_compare(all_scalar_metrics)
+
+    # TODO Add plots for COLMAP, not just GTSfM
+    plots_fig = ""
+    for metrics_group in all_metrics_groups:
+        plots_fig += create_plots_for_distributions(metrics_group.get_metrics_as_dict()[metrics_group.name])
+    # plots_fig += create_plots_for_distributions(all_metrics_groups[0].get_metrics_as_dict()[metrics_group.name])
```
akshay-krishnan, line 229: why is this commented out?

gtsfm/evaluation/metrics_report.py
```diff
@@ -191,13 +285,16 @@ def generate_metrics_report_html(metrics_groups: List[GtsfmMetricsGroup], html_p
...
         # Iterate over all metrics groups
-        for metrics_group in metrics_groups:
+        for i, metrics_group in enumerate(metrics_groups):
 
             # Write name of the metric group in human readable form.
             f.write(get_html_metric_heading(metrics_group.name))
 
             # Write plots and tables.
-            table, plots_fig = get_figures_for_metrics(metrics_group)
+            if colmap_files_dirpath == "":
```
akshay-krishnan, line 294: this is probably not a great check, it might be better to have a None default, and check for it instead.

gtsfm/evaluation/metrics_report.py
```diff
@@ -134,13 +169,67 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+                mean_nan = value[metrics.SUMMARY_KEY]["mean"] != value[metrics.SUMMARY_KEY]["mean"]
+                median_nan = value[metrics.SUMMARY_KEY]["median"] != value[metrics.SUMMARY_KEY]["median"]
+                if mean_nan or median_nan:
+                    scalar_metrics["mean_" + metric_name] = ""
+                    scalar_metrics["median_" + metric_name] = ""
+                else:
+                    scalar_metrics["mean_" + metric_name] = value[metrics.SUMMARY_KEY]["mean"]
+                    scalar_metrics["median_" + metric_name] = value[metrics.SUMMARY_KEY]["median"]
+            else:
+                scalar_metrics[metric_name] = value
```
akshay-krishnan, line 221: this does not guarantee that the two dicts (colmap and gtsfm) have the same keys in order. If for some reason they don't, things might fail, or even worse show incorrect results. I think `create_table_for_scalar_metrics_and_compare` must accept a Dict[str, List[Union[float, int]]], and so this function should handle the logic of merging the keys.

gtsfm/evaluation/metrics_report.py
```diff
@@ -134,13 +169,67 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+        metrics_dict = metrics_group.get_metrics_as_dict()[metrics_group.name]
+        # Separate the scalar metrics.
+        for metric_name, value in metrics_dict.items():
+            if isinstance(value, dict):
+                # Metrics with a dict representation must contain a summary.
+                if metrics.SUMMARY_KEY not in value:
+                    raise ValueError(f"Metric {metric_name} does not contain a summary.")
+                # Add a scalar metric for mean of 1D distributions.
+                mean_nan = value[metrics.SUMMARY_KEY]["mean"] != value[metrics.SUMMARY_KEY]["mean"]
+                median_nan = value[metrics.SUMMARY_KEY]["median"] != value[metrics.SUMMARY_KEY]["median"]
```
akshay-krishnan, line 213: do we have median in summary?

akshay-krishnan (changes requested):

gtsfm/evaluation/metrics_report.py
womackj1: np.isnan() is more readable, but requires an import.

gtsfm/evaluation/metrics_report.py
womackj1: Thanks, it is now just median.

gtsfm/evaluation/metrics_report.py
womackj1: Yes - min, max, median, mean, stddev.

gtsfm/evaluation/metrics_report.py
womackj1: Why should `create_table_for_scalar_metrics_and_compare` accept a Dict[str, List[Union[float, int]]] instead of List[Dict[str, Union[float, int]]]? `create_table_for_scalar_metrics`, which operates on a single pipeline's (GTSfM) metrics accepts a Dict[str, Union[float, int]]. Extending this to multiple pipelines for comparison led me to a list of pipeline's metrics. 

Would you be opposed to using an OrderedDict or sorting the dicts by key to ensure order?

gtsfm/evaluation/metrics_report.py
womackj1: Removed.

gtsfm/evaluation/metrics_report.py
womackj1: Done.

gtsfm/evaluation/plot_metrics.py
womackj1: Check added

gtsfm/utils/io.py
womackj1: This is not a TODO I left or completed. I removed it though.

womackj1: (The TODO is indeed complete)

gtsfm/utils/io.py
womackj1: Done

gtsfm/utils/io.py
womackj1: Now a defaultdict

gtsfm/utils/io.py
womackj1: black formatter

gtsfm/scene_optimizer.py
womackj1: I've formatted all of these files with `black --line-length=120 filepath`. I didn't see mention of additional settings in the style guide - is there additional black configuration to get the formatting you're expecting?

gtsfm/evaluation/metrics_report.py
womackj1: Reverted.

gtsfm/evaluation/metrics_report.py
womackj1: Reformatted.

gtsfm/evaluation/metrics_report.py
womackj1: Reverted

gtsfm/loader/astronet_loader.py
womackj1: Reverted.

gtsfm/loader/astronet_loader.py
womackj1: Reformatted

gtsfm/evaluation/metrics_report.py
akshay-krishnan: `Dict[str, List[Union[float, int]]]` because the metrics from multiple pipelines we compare are still the same metrics, that share the same keys. Only the value of the metric differs. 

Yeah I think I would be opposed. Using an OrderedDict may work (it is the reason its currently working because python dicts are ordered by default), but what if one of the pipelines don't output a metric? Then all the other metrics get matched incorrectly. So we need some function that takes these individual dicts and merges them into one dict. The logic for that I think is straightforward.

gtsfm/utils/io.py
akshay-krishnan: remove the parenthesis and retry formatting? does the change persist?

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +149,73 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
 
+def get_figures_for_metrics_and_compare(metrics_group: GtsfmMetricsGroup, metric_path: str) -> Tuple[str, str]:
+    """Gets the tables and plots for individual metrics in a metrics group.
+
+    All scalar metrics are reported in the table.
+    Metrics of 1-D distributions have an entry in the table for the mean,
+    and a histogram or box plot as per their property.
+
+    Args:
+        metrics_group: A GtsfmMetricsGroup for any gtsfm module.
```
akshay-krishnan, line 175: update docstring

gtsfm/evaluation/metrics_report.py
```diff
@@ -179,7 +254,12 @@ def get_html_header() -> str:
...
-def generate_metrics_report_html(metrics_groups: List[GtsfmMetricsGroup], html_path: str) -> None:
+def generate_metrics_report_html(
+    metrics_groups: List[GtsfmMetricsGroup],
+    html_path: str,
+    colmap_files_dirpath: str,
+    metric_paths: List[str],
+) -> None:
     """Generates a report for metrics groups with plots and tables and saves it to HTML.
 
     Args:
```
akshay-krishnan, line 265: docstring needs update

gtsfm/evaluation/plot_metrics.py
```diff
@@ -5,15 +5,18 @@
...
 import gtsfm.utils.logger as logger_utils
 
 logger = logger_utils.get_logger()
 
 
-def create_metrics_plots_html(json_path: str, output_dir: str) -> None:
+def create_metrics_plots_html(json_path: str, colmap_files_dirpath: str, output_dir: str) -> None:
     """Creates a HTML report of metrics from frontend, averaging, data association and bundle adjustment.
 
     Reads the metrics from JSON files in a previous run.
```
akshay-krishnan, line 22: docstring needs update here as well

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,66 @@
+"""Script to log metrics for comparison across SfM pipelines (GTSfM, COLMAP, etc.) as a JSON.
```
akshay-krishnan, line 1: this only contains a function? so we don't actually refer to it as script.. although you could call it that. 
Not sure what it does as well.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,66 @@
...
+import gtsfm.utils.io as io_utils
+from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
+
+import thirdparty.colmap.scripts.python.read_write_model as colmap_io
+
+
+def compare_metrics(
+    txt_metric_paths: Dict[str, str],
+    json_path: str,
+    GTSFM_MODULE_METRICS_FNAMES: List[str],
```
akshay-krishnan, line 17: is the documentation for this function up to date? I see the args section isn't but the docstring does not make a lot of sense either.

akshay-krishnan (changes requested):
Could you make sure all documentation is updated?

gtsfm/evaluation/metrics_report.py
womackj1: I've changed to `Dict[str, List[Union[float, int]]]`. For future reference, the Dicts for other pipelines have the same keys (metrics) as GTSfM and are left empty if the corresponding metric isn't available.

gtsfm/evaluation/metrics_report.py
womackj1: Updated.

gtsfm/evaluation/metrics_report.py
womackj1: Updated

gtsfm/evaluation/plot_metrics.py
womackj1: updated

gtsfm/utils/io.py
womackj1: Change doesn't persist - artifact of 88 line default setting.

gtsfm/evaluation/compare_metrics.py
womackj1: Updated

gtsfm/evaluation/compare_metrics.py
womackj1: Changed script to function and added more description.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+                    image_id_num_measurements[image_id] += 1
+
+        colmap2gtsfm = {
+            "number_cameras": GtsfmMetric("number_cameras", num_cameras),
+            "3d_tracks_length": GtsfmMetric(
+                "3d_tracks_length",
+                track_lengths,
+                plot_type=GtsfmMetric.PlotType.HISTOGRAM,
+            ),
+        }
```
akshay-krishnan, line 56: so what is the plan here? do we currently only support these two metrics from other pipelines? If we need more should we recompute them?

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +150,68 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+    scalar_metrics = defaultdict(list)
+    for metrics_group in all_metrics_groups:
+        metrics_dict = metrics_group.get_metrics_as_dict()[metrics_group.name]
+        # Separate the scalar metrics.
+        for metric_name, value in metrics_dict.items():
+            if isinstance(value, dict):
+                # Metrics with a dict representation must contain a summary.
+                if metrics.SUMMARY_KEY not in value:
+                    raise ValueError(f"Metric {metric_name} does not contain a summary.")
+                # Add a scalar metric for mean of 1D distributions.
```
akshay-krishnan, line 199: update comment for "median" not "mean"

gtsfm/evaluation/metrics_report.py
```diff
@@ -179,25 +250,38 @@ def get_html_header() -> str:
               </head>"""
 
 
-def generate_metrics_report_html(metrics_groups: List[GtsfmMetricsGroup], html_path: str) -> None:
+def generate_metrics_report_html(
+    metrics_groups: List[GtsfmMetricsGroup],
+    html_path: str,
+    colmap_files_dirpath: str,
+    metric_paths: List[str],
```
akshay-krishnan, line 257: arent these colmap_metric_paths ?

gtsfm/evaluation/metrics_report.py
```diff
@@ -179,25 +250,38 @@ def get_html_header() -> str:
...
+) -> None:
     """Generates a report for metrics groups with plots and tables and saves it to HTML.
 
     Args:
         metrics_groups: List of metrics to be reported.
+        html_path: Path where this report is written to (default: output_dir/gtsfm_metrics_report.html).
+        colmap_files_dirpath: Optional; If a path to a directory containing a COLMAP reconstruction
+          (as cameras.txt, images.txt, and points3D.txt) is provided, the COLMAP metrics will also be
+          included in the report.
+        metric_paths: A list of paths to GTSfM metrics.
```
akshay-krishnan, line 267: are these Gtsfm metrics? or colmap?

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+
+        # Create comparable result_metric json for COLMAP
+        for filename in GTSFM_MODULE_METRICS_FNAMES:
+            metrics = []
+            metrics_group = GtsfmMetricsGroup.parse_from_json(os.path.join(json_path, filename))
+            for metric in metrics_group.metrics:
+                # Case 1: mapping from COLMAP to GTSfM is known
+                if metric.name in colmap2gtsfm.keys():
+                    metrics.append(colmap2gtsfm[metric.name])
+                # Case 2: mapping from COLMAP to GTSfM is known
```
akshay-krishnan, line 66: is unknown you mean?

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+        for track in sfmtracks:
+            track_lengths.append(track.number_measurements())
+            for k in range(track.number_measurements()):
+                image_id, uv_measured = track.measurement(k)
+                if image_id not in image_id_num_measurements:
+                    image_id_num_measurements[image_id] = 1
+                else:
+                    image_id_num_measurements[image_id] += 1
+
+        colmap2gtsfm = {
```
akshay-krishnan, line 49: maybe choose a better name? its already a function name.. colmap_metrics might be better.

gtsfm/evaluation/metrics_report.py
```diff
@@ -46,6 +48,27 @@ def create_table_for_scalar_metrics(metrics_dict: Dict[str, Union[float, int]])
     return tabulate(table, headers="keys", tablefmt="html")
 
 
+def create_table_for_scalar_metrics_and_compare(metrics_dict: Dict[str, List[Union[float, int]]]) -> str:
+    """Creates a table in HTML format for scalar metrics from multiple SfM pipelines (GTSfM, COLMAP).
+
```
akshay-krishnan, line 53: missing args section

gtsfm/evaluation/metrics_report.py
```diff
@@ -46,6 +48,27 @@ def create_table_for_scalar_metrics(metrics_dict: Dict[str, Union[float, int]])
...
+    Returns:
+        Table with scalar metrics and their values in HTML format.
+    """
+    for metric_key, metric_values in metrics_dict.items():
+        for metric_indice, metric_value in enumerate(metric_values):
```
akshay-krishnan, line 58: nit: "metric_index" or "metric_idx"

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +150,68 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+    Args:
+        metrics_group: A GtsfmMetricsGroup for any gtsfm module.
+        metric_path: A path to the GTSfM generated metrics (default: gtsfm/result_metrics)
+
+    Returns:
+        A tuple of table and plotly figures as HTML code.
+    """
+    all_metrics_groups = []
+
+    colmap_metric_path = metric_path[: metric_path.rindex("/")] + "/colmap" + metric_path[metric_path.rindex("/") :]
```
akshay-krishnan, line 184: why not pass this directly to this method?

akshay-krishnan: actually you can just pass 2 metrics groups

gtsfm/evaluation/compare_metrics.py
womackj1: Thanks, I've addressed this.

gtsfm/evaluation/compare_metrics.py
womackj1: We currently only support these two metrics as they are the only ones directly accessible from COLMAP output.
The output of COLMAP doesn't have a 1-to-1 mapping with the output of GTSfM, so many of the GTSFM metrics will be blank for COLMAP and others require more rigorous processing of COLMAP's database, output, and/or logs.

gtsfm/evaluation/metrics_report.py
womackj1: Updated!

gtsfm/evaluation/metrics_report.py
womackj1: Yes, addressed.

gtsfm/evaluation/metrics_report.py
womackj1: GTSfM Metrics, which will typically be GTSFM_MODULE_METRICS_FNAMES.

gtsfm/evaluation/compare_metrics.py
womackj1: yes, updated.

gtsfm/evaluation/compare_metrics.py
womackj1: Addressed.

gtsfm/evaluation/metrics_report.py
womackj1: Added.

gtsfm/evaluation/metrics_report.py
womackj1: Addressed.

gtsfm/evaluation/metrics_report.py
womackj1: Addressed.

womackj1:
[gtsfm_metrics_report.pdf](https://github.com/borglab/gtsfm/files/7966951/gtsfm_metrics_report.pdf)
Here is the HTML Report. This PR is ready to go.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+        colmap_metrics = {
+            "number_cameras": GtsfmMetric("number_cameras", num_cameras),
+            "3d_tracks_length": GtsfmMetric(
+                "3d_tracks_length",
+                track_lengths,
+                plot_type=GtsfmMetric.PlotType.HISTOGRAM,
+            ),
+        }
+
+        # Create comparable result_metric json for COLMAP
```
akshay-krishnan, line 58: for colmap? or for other pipelines?

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+        for filename in metric_filenames:
+            metrics = []
+            metrics_group = GtsfmMetricsGroup.parse_from_json(os.path.join(json_path, filename))
+            for metric in metrics_group.metrics:
+                # Case 1: mapping from COLMAP to GTSfM is known
```
akshay-krishnan, line 63: function documentation says "other pipelines", but here you explicitly mention colmap. I think we should be more consistent.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+                    image_id_num_measurements[image_id] += 1
+
+        colmap_metrics = {
+            "number_cameras": GtsfmMetric("number_cameras", num_cameras),
+            "3d_tracks_length": GtsfmMetric(
+                "3d_tracks_length",
+                track_lengths,
+                plot_type=GtsfmMetric.PlotType.HISTOGRAM,
+            ),
+        }
```
akshay-krishnan, line 56: could you separate this logic where we actually compute the metrics from other pipeline data to a different function? This would help add more metrics later if we need to.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+                if metric.name in colmap_metrics.keys():
+                    metrics.append(colmap_metrics[metric.name])
+                # Case 2: mapping from COLMAP to GTSfM is unknown
+                else:
+                    if metric._dim == 1:
+                        # Case 2a: dict summary
+                        metrics.append(GtsfmMetric(metric.name, []))
+                    else:
+                        # Case 2b: scalar metric
+                        metrics.append(GtsfmMetric(metric.name, ""))
```
akshay-krishnan, line 73: why do we need these? I think a better approach is to save only the metrics for which we know the mapping. The empty string can be inserted when we create the HTML. A GtsfmMetric with an empty string does not mean anything, and is a little hacky. :/

gtsfm/evaluation/metrics.py
```diff
@@ -89,9 +89,10 @@ def __init__(
 
         self._name = name
         if data is not None:
-            # Cast to a numpy array
+            # Cast to a numpy array, unless data is an empty string,
+            # which enables blank values in metrics report.
             if not isinstance(data, np.ndarray):
-                data = np.array(data, dtype=np.float32)
```
akshay-krishnan, line 94: as mentioned above, there are some methods in this class that assume a metric is a number, and I think its better to keep it that way. it has also caused some NaN/None bugs and we should not change it unless there are good reasons to.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,76 @@
...
+import os
+from typing import Dict, List
+
+import gtsfm.utils.io as io_utils
+from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
+
+import thirdparty.colmap.scripts.python.read_write_model as colmap_io
+
+
+def compare_metrics(
```
akshay-krishnan, line 17: also the function is not actually comparing, so we need a good name for it. something along "save_metrics_from_colmap_as_gtsfmmetricsgroup" but shorter.

gtsfm/evaluation/metrics_report.py
```diff
@@ -39,13 +41,42 @@ def get_readable_metric_name(metric_name: str) -> str:
...
+
     Returns:
         Table with scalar metrics and their values in HTML format.
     """
     table = {"Metric name": list(metrics_dict.keys()), "Value": list(metrics_dict.values())}
     return tabulate(table, headers="keys", tablefmt="html")
 
 
+def create_table_for_scalar_metrics_and_compare(metrics_dict: Dict[str, List[Union[float, int]]]) -> str:
+    """Creates a table in HTML format for scalar metrics from multiple SfM pipelines (GTSfM, COLMAP).
```
akshay-krishnan, line 56: could you be a little more clear on whether it is specific to colmap or not? The "(GTSFM, COLMAP) " may be confusing.

gtsfm/evaluation/metrics_report.py
```diff
@@ -39,13 +41,42 @@ def get_readable_metric_name(metric_name: str) -> str:
...
+
+    Args:
+        metrics_dict: A dict, where keys are names of metrics and values are
+        the dictionary representation of the metric.
```
akshay-krishnan, line 60: values are not Dict in the type hint for the argument...

gtsfm/evaluation/metrics_report.py
akshay-krishnan: from the implementation, it seems colmap specific.

gtsfm/evaluation/plot_metrics.py
```diff
@@ -30,21 +36,33 @@ def create_metrics_plots_html(json_path: str, output_dir: str) -> None:
         "rotation_averaging_metrics.json",
         "translation_averaging_metrics.json",
         "data_association_metrics.json",
-        "bundle_adjustment_metrics.json"
+        "bundle_adjustment_metrics.json",
     ]
+
+    if Path(colmap_files_dirpath).exists():
```
akshay-krishnan, line 42: what is the behavior of Path() if colmap_files_dirpath is None?

gtsfm/evaluation/metrics_report.py
```diff
@@ -179,25 +253,43 @@ def get_html_header() -> str:
...
+    metrics_groups: List[GtsfmMetricsGroup],
+    html_path: str,
+    colmap_output_dir: str,
+    metric_paths: List[str],
+) -> None:
     """Generates a report for metrics groups with plots and tables and saves it to HTML.
 
     Args:
         metrics_groups: List of metrics to be reported.
+        html_path: Path where this report is written to (default: output_dir/gtsfm_metrics_report.html).
```
akshay-krishnan, line 266: there does not seem to be a default in the function, its just using whatever is passed in.

gtsfm/evaluation/metrics_report.py
```diff
@@ -179,25 +253,43 @@ def get_html_header() -> str:
...
+        colmap_output_dir: Optional; If a path to a directory containing a COLMAP reconstruction
+          (as cameras.txt, images.txt, and points3D.txt) is provided, the COLMAP metrics will also be
+          included in the report.
+        metric_paths: A list of paths to GTSfM metrics.
```
akshay-krishnan, line 270: the function seems to be doing too many things. It is supposed to generate a HTML from known metrics. Instead of passing in the colmap path to it, I recommend passing in colmap GtsfmMetricGroups. 
The computation of these metric groups can be done in another function (which is not called by this).

akshay-krishnan: If youre curious, here are some of the issues I see with this design: 
- it is asking for the GTSfM metric paths although it already has GTSfM metrics. 
- it is assuming the colmap metrics are situated are in a specific location with respect to gtsfm metrics. 
- its job is to generate a html, not to process colmap output :)

gtsfm/evaluation/plot_metrics.py
```diff
@@ -30,21 +36,33 @@ def create_metrics_plots_html(json_path: str, output_dir: str) -> None:
...
 
 
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
     parser.add_argument("--metrics_dir", default="result_metrics", help="Directory containing the metrics json files.")
+    parser.add_argument("--colmap_files_dirpath", default=None, type=str, help="Directory containing COLMAP output .")
     parser.add_argument("--output_dir", default="", help="Directory to save plots to. Same as metrics_dir by default.")
     args = parser.parse_args()
-    create_metrics_plots_html(args.metrics_dir, args.output_dir)
+    create_metrics_plots_html(args.metrics_dir, args.colmap_files_dirpath, args.output_dir)
```
akshay-krishnan, line 68: I think a better flow is like this: 

```
if args.colmap_files_dirpath is not None:
  save_colmap_metrics(args.colmap_files_dirpath, colmap_json_path)  # saves metrics to the json path
create_metrics_plots_html(args.metrics_dir, colmap_json_path, args.output_dir)
```

akshay-krishnan: basically separate metric computation from plotting.

johnwlambert:
Thanks for the updates @womackj1. There are 6 more metrics I would like us to add to the COLMAP column before merging:

number_tracks_filtered
median_3d_track_lengths_filtered
median_reprojection_errors_filtered_px
number_tracks_unfiltered
median_3d_track_lengths_unfiltered
median_reprojection_errors_unfiltered_px

We can get these all from the COLMAP data.

gtsfm/evaluation/compare_metrics.py
womackj1: Agreed, changed.

gtsfm/evaluation/compare_metrics.py
womackj1: Other pipelines (colmap included).  Updated.

gtsfm/evaluation/compare_metrics.py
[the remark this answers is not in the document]
womackj1: Fixed.

gtsfm/evaluation/compare_metrics.py
womackj1: Good point. Separated.

gtsfm/evaluation/compare_metrics.py
womackj1: How about "save_other_pipelines_metrics" (the function isn't limited to colmap)?

gtsfm/evaluation/compare_metrics.py
womackj1: Agreed. The empty string is now inserted while creating the HTML.

gtsfm/evaluation/metrics.py
womackj1: Reverted.

gtsfm/evaluation/metrics_report.py
womackj1: Changed to COLMAP. Note: I plan on extending this to additional pipelines later.

gtsfm/evaluation/metrics_report.py
womackj1: Updated the comment.

gtsfm/evaluation/metrics_report.py
womackj1: The default referenced in this comment is the default argument for the method call in plot_metrics.py. The comment is now removed to avoid confusion.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,80 @@
...
+            track_lengths,
+            plot_type=GtsfmMetric.PlotType.HISTOGRAM,
+        ),
+    }
+    return other_pipeline_metrics
+
+def save_other_pipelines_metrics(
+    txt_metric_paths: Dict[str, str],
+    json_path: str,
+    metric_filenames: List[str],
```
akshay-krishnan, line 53: nit: alternative (hopefully more clear) variable names: 
colmap_format_outputs (these are not metrics)
gtsfm_json_path
gtsfm_metrics_filenames

gtsfm/evaluation/metrics.py
akshay-krishnan: please revert the comment as well :)

akshay-krishnan: actually, all changes to this file? is the other one necessary?

gtsfm/evaluation/metrics_report.py
```diff
@@ -7,8 +7,10 @@
 
 Authors: Akshay Krishnan
```
akshay-krishnan, line 8: please add yourself here :)

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +158,99 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
 
 
+def add_summary(metric_name, metric_value, scalar_metrics):
+    # Metrics with a dict representation must contain a summary.
+    if metrics.SUMMARY_KEY not in metric_value:
+        raise ValueError(f"Metric {metric_name} does not contain a summary.")
+    # Add a scalar metric for median of 1D distributions.
+    median_nan = metric_value[metrics.SUMMARY_KEY]["median"] != metric_value[metrics.SUMMARY_KEY][
+        "median"]
+    if median_nan:
```
akshay-krishnan, line 183: `if np.isnan(metric_value[metrics.SUMMARY_KEY]["median"]): `

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +158,99 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+            scalar_metrics["median_" + metric_name] = value[metrics.SUMMARY_KEY]["median"]
         else:
             scalar_metrics[metric_name] = value
     table = create_table_for_scalar_metrics(scalar_metrics)
     plots_fig = create_plots_for_distributions(metrics_dict)
     return table, plots_fig
 
 
+def add_summary(metric_name, metric_value, scalar_metrics):
+    # Metrics with a dict representation must contain a summary.
```
akshay-krishnan, line 177: missing documentation

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +158,99 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+    # Add a scalar metric for median of 1D distributions.
+    median_nan = metric_value[metrics.SUMMARY_KEY]["median"] != metric_value[metrics.SUMMARY_KEY][
+        "median"]
+    if median_nan:
+        scalar_metrics["median_" + metric_name].append("")
+    else:
+        scalar_metrics["median_" + metric_name].append(metric_value[metrics.SUMMARY_KEY]["median"])
+
+def add_metric(gtsfm_metric_name, gtsfm_metric_value, scalar_metrics, other_pipelines_metric_dicts):
+    """
```
akshay-krishnan, line 189: missing documentation summary.

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +158,99 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+            # Add a scalar metric for median of 1D distributions.
+            scalar_metrics["median_" + metric_name] = value[metrics.SUMMARY_KEY]["median"]
         else:
             scalar_metrics[metric_name] = value
     table = create_table_for_scalar_metrics(scalar_metrics)
     plots_fig = create_plots_for_distributions(metrics_dict)
     return table, plots_fig
 
 
+def add_summary(metric_name, metric_value, scalar_metrics):
```
akshay-krishnan, line 176: regarding both add_metric and add_summary. I dont think we mutate arguments anywhere in gtsfm, and try to be functional despite being in python land. @johnwlambert please correct me if that is not the case.

akshay-krishnan: so I think we dont need this as a separate function, but if you have other ways to make it functional, that also works.

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +158,99 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+        for other_pipeline_metric_value in other_pipelines_metric_values:
+            if isinstance(other_pipeline_metric_value, dict):
+                add_summary(gtsfm_metric_name, other_pipeline_metric_value, scalar_metrics)
+            else:
+                other_pipeline_metric_value = {'summary': {'median': ""}}
+                add_summary(gtsfm_metric_name, other_pipeline_metric_value, scalar_metrics)
+    else:
+        scalar_metrics[gtsfm_metric_name].append(gtsfm_metric_value)
+        for other_pipeline_metric_value in other_pipelines_metric_values:
+            scalar_metrics[gtsfm_metric_name].append(other_pipeline_metric_value)
```
akshay-krishnan, line 215: as my other comment above says, I think we want to be functional here, so could we return a key, value maybe?

gtsfm/evaluation/plot_metrics.py
akshay-krishnan: this probably needs update again?

gtsfm/evaluation/plot_metrics.py
womackj1: The behavior was TypeError. Your new design avoids that though.

gtsfm/evaluation/metrics.py
womackj1: Reverted the comment and other change (was necessary when I was storing empty strings).

gtsfm/evaluation/metrics_report.py
womackj1: This function now receives colmap GtsfmMetricGroups, which are computed in another function.

gtsfm/evaluation/plot_metrics.py
womackj1: Thanks for the suggestion - the new design is better.

gtsfm/evaluation/compare_metrics.py
womackj1: This json_path contains jsons from gtsfm and other pipelines metrics, but I've modified the other variable names.

gtsfm/evaluation/metrics_report.py
womackj1: Done, thanks!

gtsfm/evaluation/metrics_report.py
womackj1: Addressed.

gtsfm/evaluation/metrics_report.py
womackj1: Sure thing, now using key, value.

womackj1:
HTML of report
[gtsfm_metrics_report.zip](https://github.com/borglab/gtsfm/files/7988015/gtsfm_metrics_report.zip)

gtsfm/evaluation/metrics.py
```diff
@@ -167,9 +169,8 @@ def _create_summary(self, data: np.ndarray) -> Dict[str, Any]:
         """
         if data.ndim != 1:
             raise ValueError("Metric must be a 1D distribution to get summary.")
-        if data.size == 0 or np.isnan(data).all():
+        if data.size == 0:  # or np.isnan(data).all():
```
akshay-krishnan, line 172: why is the NaN handling removed?

gtsfm/evaluation/metrics_report.py
```diff
@@ -110,6 +146,110 @@ def create_plots_for_distributions(metrics_dict: Dict[str, Any]) -> str:
...
+        # j = 0
+        # Check if this is a 1D distribution metric and has a summary.
+        if metric_name not in distribution_metrics or metrics.SUMMARY_KEY not in metric_value:
+            continue
+        row = 1
+        col = 1
+        # Histogram metrics are plotted directly from summary.
+        if "histogram" in metric_value[metrics.SUMMARY_KEY]:
+            histogram = metric_value[metrics.SUMMARY_KEY]["histogram"]
+            fig.add_trace(
```
akshay-krishnan, line 193: is there some way we can reduce the amount of code in this function by having another function to create the plots, which gets called with both gtsfm metrics and other pipeline metrics (i.e the plotting code is the same and does not need repetition).

gtsfm/evaluation/metrics_report.py
```diff
@@ -125,22 +265,110 @@ def get_figures_for_metrics(metrics_group: GtsfmMetricsGroup) -> Tuple[str, str]
...
+
+    for gtsfm_metric_name, gtsfm_metric_value in gtsfm_metric_dict.items():
+        other_pipelines_metrics = defaultdict(list)
+        for other_pipeline_metric_dict in other_pipelines_metric_dicts:
+            if gtsfm_metric_name in other_pipeline_metric_dict:
+                other_pipeline_metric_value = other_pipeline_metric_dict[gtsfm_metric_name]
+                other_pipelines_metrics[gtsfm_metric_name].append(other_pipeline_metric_value)
+            else:
+                other_pipelines_metrics[gtsfm_metric_name].append("")
+        if isinstance(gtsfm_metric_value, dict):
```
akshay-krishnan, line 316: I think there is some repeated code below that can be refactored into a reusable function.

akshay-krishnan (commented):
sorry for the late review!

dellaert:
Cooll!

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+metrics from GTSfM.
+
+Authors: Jon Womack
+"""
+import os
+from typing import Dict, List
+
+import numpy as np
+
+
```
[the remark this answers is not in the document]
womackj1: Removed.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+import gtsfm.utils.io as io_utils
+from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
+
+import thirdparty.colmap.scripts.python.read_write_model as colmap_io
+
+
+def compute_metrics_from_txt(cameras, images, points3d):
```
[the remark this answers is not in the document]
womackj1: Of course, added!

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+        points3D: dictionary of COLMAP-formatted Point3Ds
+    Returns:
+        other_pipeline_metrics: A dictionary of metrics from another pipeline that are comparable with GTSfM
+    """
+    cameras, images, image_files, sfmtracks = io_utils.colmap2gtsfm(cameras, images, points3d, load_sfmtracks=True)
+    num_cameras = len(cameras)
+    unfiltered_track_lengths = []
+    image_id_num_measurements = {}
+    for track in sfmtracks:
+        unfiltered_track_lengths.append(int(track.numberMeasurements()))
```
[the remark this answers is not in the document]
womackj1: Good catch, not necessary, removed.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+
+    # Note: IDs begin at 1, so id-1 for indexing list
+    unfiltered_reproj_errors = []
+    filtered_reproj_errors = []
+    filtered_track_lengths = []
+    for point3d_id, point3d in points3d.items():
+        reproj_error = point3d.error
+        unfiltered_reproj_errors.append(reproj_error)
+        # TODO (Jon): make reproj_error_threshold an argument to this method
+        if reproj_error < 3:
```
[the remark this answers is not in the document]
womackj1: Reprojection error threshold is now an argument.

gtsfm/evaluation/compare_metrics.py
```diff
@@ -0,0 +1,116 @@
...
+
+def save_other_pipelines_metrics(
+    colmap_format_outputs: Dict[str, str],
+    json_path: str,
+    gtsfm_metric_filenames: List[str],
+) -> None:
+    """Converts the outputs of other SfM pipelines to GTSfMMetricsGroups saved as json files.
+
+    Creates folders for each additional SfM pipeline that contain GTSfMMetricsGroups (stored as json files)
+    containing the same metrics as GTSFM_MODULE_METRICS_FNAMES. If one of the GTSfM metrics
```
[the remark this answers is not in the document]
womackj1: I added a note to the comment specifying where GTSFM_MODULE_METRICS_FNAMES is defined.

gtsfm/evaluation/metrics.py
womackj1: I removed this while debugging and forgot to add it back in. I reverted it.

gtsfm/evaluation/metrics_report.py
womackj1: Added a function called make_fig() to reuse code.

gtsfm/evaluation/metrics_report.py
womackj1: I refactored the code into the function `add_scalar_metric()`.

gtsfm/evaluation/compare_metrics.py
akshay-krishnan: nit: Please leave a line between the description and args, and between args and returns. 
```
"""blah blah

Args:
  foo

Returns:
  bar
"""
```

gtsfm/utils/io.py
akshay-krishnan: any thoughts on this @womackj1 ?

gtsfm/evaluation/compare_metrics.py
akshay-krishnan: same in other places

akshay-krishnan (approved):

PR #348, opened 2021-10-13
add AstronNet 2011205_rc3 with a 10 frame lookahead and deep front end to the CI

johnwlambert:
- SIFT front-end is still under development.
- Full 65-frame lookahead won't finish in 6 hours of CI budget.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -12,15 +12,17 @@ jobs:
       matrix:
         config_dataset_info:
           [
-            "sift_front_end,  door-12,               12,  JPG,  test_data,  olsson-loader,  1296, true",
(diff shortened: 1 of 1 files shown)


travisdriver:
> @travisdriver the result isn't so good yet. But I think that's ok for the CI benchmark because I think we'll continue to improve. Here's the AstroNet deep-front end result w/ 20 frame lookahead:
> 
> Screen Shot 2021-10-15 at 5 01 53 PM
> 
> I'd probably rather continue iterating on this one rather than add a new set of data just yet. Is this one of the most difficult sequences from AstroNet?

It's one of the easiest sequences I have :/ I have ones that are similar (large overlap between images, same distance from surface, consistent illumination) and with less images that will likely be easier to manage and allow for it to be run with deep front end easier.

PR #349, opened 2021-10-13
Adding conda env file for CPU only machines

ayushbaid:
This PR adds a conda env file for CPU only machines. 

CPU only machines do not need `cudatoolkit` package, and hence replacing it with `cpuonly` saves a lot of space and will make the environment creation faster.

This will be particularly helpful on the CI.

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -52,7 +52,7 @@ jobs:
         with:
           miniconda-version: "latest"
           activate-environment: gtsfm-v1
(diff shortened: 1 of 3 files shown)


ayushbaid:
The new conda environment file speeds up the computation. We go from 15-20 mins on the conda env creation step to ~7 mins. 

This is expected as the `cudatoolkit` package was a huge one.

ayushbaid:
> LGTM, but does the cache still work? is the CI still fast (under 45 min)?

Cache will be regenerated when I run it on master

PR #354, opened 2021-10-17
Added Round Trip Unit Test for Writing and Reading Images

adisingh50:
image

Trying to inspect the read and write functions for `images.txt`. I used a sample pose that was produced after applying an alignment transformation to a camera frustum. Converted the sample pose into extrinsics. And then converted those extrinsics back into a recovered pose. Then checked if the recovered pose matches the original pose. As shown above, the recovered pose doesn't match up with the original pose.

Not sure why since I thought GTSAM only accepted valid Pose objects.


tests/utils/test_io_utils.py
```diff
@@ -91,3 +92,28 @@ def test_read_cameras_txt_nonexistent_file() -> None:
...
+
+def test_round_trip_images_txt() -> None:
+    """Loads a pose from file that was produced after applying an alignment transformation to a camera frustum. Converts
+    sample pose into extrinsics. Then converts these extrinsics back into a pose and verifies it matches
+    with the original."""
+
+    # Setup dummy pinhole camera object
+    default_intrinsics = Cal3Bundler(fx=100, k1=0, k2=0, u0=0, v0=0)
+
+    pose_from_file = np.load("gtsfm_incorrect_alignment_cameras_wTc.npy")
```
johnwlambert, line 105: thanks for the PR, Adi. Could you provide the pose matrix values as floats in the .py file, rather than in a numpy file? Just 16 numbers in the file you have here, right?

tests/utils/test_io_utils.py
```diff
@@ -91,3 +92,34 @@ def test_read_cameras_txt_nonexistent_file() -> None:
...
+    """Starts with a pose that was produced after applying an alignment transformation to a camera frustum. Converts
+    sample pose into extrinsics. Then converts these extrinsics back into a pose and verifies it matches
+    with the original."""
+
+    # Setup dummy pinhole camera object
+    default_intrinsics = Cal3Bundler(fx=100, k1=0, k2=0, u0=0, v0=0)
+
+    original_pose = np.array(
+        [
+            [0.0018451, -0.994551, 0.104234, -0.0121756],
```
johnwlambert, line 107: @adisingh50 I checked the determinant of the rotation portion of this matrix and it seems to be -1.

Where did you get this rotation matrix from? It looks like we are dealing with a reflection, not a rotation.

Likely this is the issue

johnwlambert: Cool. @adisingh50 can you try the following?  Can you check the determinant of V^T that you take out of the SVD, and if it's -1, check which singular value was -1? and then correct that singular value to be +1, and apply it to the corresponding row or column of V? And see if that fixes all our issues?

You can read a bit more about it here under Orthogonal Procrustes:
https://en.wikipedia.org/wiki/Orthogonal_Procrustes_problem#Generalized.2Fconstrained_Procrustes_problems

https://dsp.stackexchange.com/questions/3431/reflection-rotation-matrix

https://towardsdatascience.com/essential-math-for-data-science-visual-introduction-to-singular-value-decomposition-svd-ca549ab7eb9c

adisingh50: Yup. I'll get back to you later today about this.

akshay-krishnan:
Thanks for adding this Adi! Is the test currently broken or is it fixed?

gtsfm/utils/io.py
```diff
@@ -265,15 +265,36 @@ def read_images_txt(fpath: str) -> Tuple[Optional[List[Pose3]], Optional[List[st
...
+        iRw_quaternion: camera rotation extrinsics, array of shape (4,)
+        itw: camera translation extrinsics, array of shape (3,)
+
+    Returns:
+        The pose of the camera in world frame.
+    """
+    qw, qx, qy, qz = iRw_quaternion
+    tx, ty, tz = itw
+
+    # Colmap provides extrinsics, so must invert
```
johnwlambert, line 291: thanks for this PR, Adi. I think this was definitely useful for debugging. I think I found the previous I/O to be more readable though. So not sure I would want to check this in. @ayushbaid @akshay-krishnan what are your thoughts?

akshay-krishnan: This was actually my suggestion to debug the issue that Adi was seeing. This is more unit-testable. This should also not be in io.py, it is a transform.

akshay-krishnan:
Adi, I am not sure where you borrowed the `extrinsics` vs `pose` terminology from, is this common somewhere? To me, they meant the same thing. Quaternions and matrices are the two different representations for both wTc and cTw.

PR #357, opened 2021-10-18
fix the reported number of inputs to Rotation Cycle Consistency module in the HTML report

johnwlambert:
Currently, the reported # of inputs to cycle consistency is incorrect.

For example, for 63K input pairs, and 26K valid pairs, only 26K are truly fed into the cycle consistency check.

However, the metrics suggest that 63K input pairs are going into the cycle consistency check, which is wrong (the correct number is 26K).
Screen Shot 2021-10-18 at 7 21 16 PM
Screen Shot 2021-10-18 at 7 21 25 PM

### DIFF
--- a/gtsfm/averaging/rotation/cycle_consistency.py
+++ b/gtsfm/averaging/rotation/cycle_consistency.py
@@ -385,7 +385,7 @@ def _compute_metrics(
     )
 
     rcc_metrics = [
-        GtsfmMetric("num_total_rcc_input_measurements", n_valid_edges),
+        GtsfmMetric("num_input_measurements", n_valid_edges),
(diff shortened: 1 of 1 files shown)


gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -380,7 +385,7 @@ def _compute_metrics(
     )
 
     rcc_metrics = [
-        GtsfmMetric("num_total_frontend_measurements", num_total_measurements),
+        GtsfmMetric("num_total_rcc_input_measurements", n_valid_edges),
```
akshay-krishnan, line 388: what is rcc?

akshay-krishnan (approved):

PR #358, opened 2021-10-19
Reverse changes made by PR #281

travisdriver:
Reverses changes made by PR #281, as rejecting individual measurements outside the context of RANSAC is not reliable imo.


johnwlambert:
Thanks for the PR @travisdriver. Before we revert this change, could you find any examples of tracks where #281 does the wrong thing? e.g. lets in bad tracks (with patch centers that clearly dont match each other) vs. rejects those bad tracks?

travisdriver:
If we triangulate a track and then find an outlier when reprojecting this point back into the cameras, then we know the point was triangulated with an outlier measurement (or at least a measurement with a relatively significant amount of noise). Moreover, we don't even know whether that specific measurement that failed the reprojection error check is an _actual_ outlier, since the estimated position of the landmark is corrupted (I don't believe we use any sort of robust estimator when triangulating). 

With this in mind, I think it's straightforward to reason how bad tracks could make it through the current outlier rejection process without comparing thousands of track plots.

One thing that could be done is retriangulating after rejecting outliers and checking to see that the reprojection error improves. However, I think RANSAC is the more precise thing to do.

akshay-krishnan:
Should this not be done after RANSAC is enabled? I think it makes sense to do without RANSAC, unless we have solid metrics that suggest otherwise.

travisdriver:
## Skydio-32
Metric name | Before | After
-- | -- | --
number_cameras | 30 | 31
number_tracks_unfiltered | 8027 | 7856
mean_3d_track_lengths_unfiltered | 2.87044 | 2.89473
mean_reprojection_errors_unfiltered_px | 4.00686 | 3.17239
mean_rotation_error_angle_deg | 0.723362 | 0.298187
mean_translation_error_distance | 3.44216 | 0.0271484
mean_translation_angle_error_deg | 3.58754 | 0.529126
Filtered tracks triangulated with GT cams: SUCCESS | 5307 | 5885
Filtered tracks triangulated with GT cams: EXCEEDS_REPROJ_THRESH | 52 | 194
Filtered tracks triangulated with GT cams: CHEIRALITY_FAILURE | 19 | 10
number_tracks_filtered | 5378 | 6089
mean_3d_track_lengths_filtered | 2.44719 | 2.46395
mean_reprojection_errors_filtered_px | 0.800642 | 0.796796

akshay-krishnan:
Has the homography detection on Notre Dame been merged? If not, the homographies could be the cause of the high translation direction error.

akshay-krishnan:
> > Should this not be done after RANSAC is enabled? I think it makes sense to do without RANSAC, unless we have solid metrics that suggest otherwise.
> 
> How does it make sense to do outside of RANSAC?

Data association errors will cause outlier measurements in a track. Assuming these are not too many (because if they are, the triangulated point is just unreliable), if we triangulate naively (without RANSAC), we can use the triangulated point to discard them.

travisdriver:
> > > Should this not be done after RANSAC is enabled? I think it makes sense to do without RANSAC, unless we have solid metrics that suggest otherwise.
> > 
> > 
> > How does it make sense to do outside of RANSAC?
> 
> Data association errors will cause outlier measurements in a track. Assuming these are not too many (because if they are, the triangulated point is just unreliable), if we triangulate naively (without RANSAC), we can use the triangulated point to discard them.

Any sort of data association errors make the triangulated point unreliable and unfit to be used for subsequent filtering (especially since we're not using a robust estimator). If there are enough outliers, then using the landmark to filter the cameras could result in completely bogus associations because the estimated position of the landmark is likely no where near the ground truth.

Moreover, I don't think it's valid to assume there "are not too many" outliers. The metrics have shown that this change consistently increases performance.

johnwlambert:
Thanks for taking a look at the skydio-32 metrics Travis. The latest on the CI should have translation distance after BA of 0.02 (I pulled down this one, for example: https://github.com/borglab/gtsfm/pull/370)

So I think we need to compare with this from master: 

<p style="color: rgb(0, 0, 0); font-style: normal; font-variant-ligatures: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: 2; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: 2; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration-thickness: initial; text-decoration-style: initial; text-decoration-color: initial; font-size: 25px; font-family: Arial;">Bundle Adjustment Metrics</p>

Metric name | Value
-- | --
number_cameras | 31
number_tracks_unfiltered | 8037
mean_3d_track_lengths_unfiltered | 2.92049
mean_reprojection_errors_unfiltered_px | 3.75204
mean_rotation_error_angle_deg | 0.294806
mean_translation_error_distance | 0.0266476
mean_translation_angle_error_deg | 0.521482
Filtered tracks triangulated with GT cams: SUCCESS | 6041
Filtered tracks triangulated with GT cams: EXCEEDS_REPROJ_THRESH | 55
Filtered tracks triangulated with GT cams: CHEIRALITY_FAILURE | 10
number_tracks_filtered | 6106
mean_3d_track_lengths_filtered | 2.46987
mean_reprojection_errors_filtered_px | 0.798481

So I don't think there is any improvement for skydio-32. Let's report for each of the benchmark ones. I guess we really need a way to automated this to see it visually, since it's quite annoying to have to do it manually currently.

akshay-krishnan:
> > > > Should this not be done after RANSAC is enabled? I think it makes sense to do without RANSAC, unless we have solid metrics that suggest otherwise.
> > > 
> > > 
> > > How does it make sense to do outside of RANSAC?
> > 
> > 
> > Data association errors will cause outlier measurements in a track. Assuming these are not too many (because if they are, the triangulated point is just unreliable), if we triangulate naively (without RANSAC), we can use the triangulated point to discard them.
> 
> Any sort of data association errors make the triangulated point unreliable and unfit to be used for subsequent filtering (especially since we're not using a robust estimator). If there are enough outliers (or the magnitude of even a single outlier is large enough), then using the landmark to filter the cameras could result in completely bogus associations because the estimated position of the landmark is likely no where near the ground truth.
> 
> Moreover, I don't think it's valid to assume there "are not too many" outliers. The metrics have shown that this change consistently increases performance.

I see, yeah makes sense. If the triangulated point is reliable, and if there is still is an outlier (which perhaps did not impact triangulation), then it would help to remove it during bundle adjustment. But maybe bundle adjustment is also robust to it, in which case there is no problem.

akshay-krishnan:
Could you pull from master, compare metrics again and let us know if there is a noticeable improvement?

travisdriver:
> > > > > Should this not be done after RANSAC is enabled? I think it makes sense to do without RANSAC, unless we have solid metrics that suggest otherwise.
> > > > 
> > > > 
> > > > How does it make sense to do outside of RANSAC?
> > > 
> > > 
> > > Data association errors will cause outlier measurements in a track. Assuming these are not too many (because if they are, the triangulated point is just unreliable), if we triangulate naively (without RANSAC), we can use the triangulated point to discard them.
> > 
> > 
> > Any sort of data association errors make the triangulated point unreliable and unfit to be used for subsequent filtering (especially since we're not using a robust estimator). If there are enough outliers (or the magnitude of even a single outlier is large enough), then using the landmark to filter the cameras could result in completely bogus associations because the estimated position of the landmark is likely no where near the ground truth.
> > Moreover, I don't think it's valid to assume there "are not too many" outliers. The metrics have shown that this change consistently increases performance.
> 
> I see, yeah makes sense. If the triangulated point is reliable, and if there is still is an outlier (which perhaps did not impact triangulation), then it would help to remove it during bundle adjustment. But maybe bundle adjustment is also robust to it, in which case there is no problem.

If there is an outlier, then it is safe to assume that the triangulated point is _unreliable_ since we are not using a robust estimator.

travisdriver:
> Could you pull from master, compare metrics again and let us know if there is a noticeable improvement?

Results are similar to what I already posted. Seems to be improvements across the board for **Skydio-32**, **notre-dame-20** and **2011212_rc3** are both equally bad in slightly different ways (whereas before there was a pretty significant improvement for notre-dame-20). 

We have already reasoned why the current implementation is incorrect, and I think we've seen enough in the benchmarks thus far to approve this PR. 

Any time the current implementation matches/beats the proposed is because we got lucky, i.e., the corrupted landmark was corrupted in such a way that that it didn't allow for enough poses to be rejected. However, we have reasoned how the corrupted landmark could lead to bogus poses using the current implementation (which I think is what we saw from the previous notre-dame-20 run that I posted that was completely broken).

akshay-krishnan (approved):

johnwlambert:
> Seems to be improvements across the board for **Skydio-32**, **notre-dame-20** and **2011212_rc3** are both equally bad in slightly different ways (whereas before there was a pretty significant improvement for notre-dame-20).

Hi @travisdriver, I didn't quite follow what you meant here:

```
Seems to be improvements across the board for Skydio-32, notre-dame-20 and 2011212_rc3 are both equally bad in slightly different ways (whereas before there was a pretty significant improvement for notre-dame-20).
```

Could you explain that a bit more? What are the differences we see?

johnwlambert:
Here's the diff against Master that I see:

Screen Shot 2021-11-21 at 1 30 27 AM

Huge improvements for Notre Dame # of filtered tracks. Some other strange regressions. I think this needs a bit more analysis about what's going on.

johnwlambert:
visual diff against master (zipped):
[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/7575841/visual_comparison_dashboard.html.zip)

johnwlambert (approved):

PR #359, opened 2021-10-20
Add table for InlierSupportPostprocessor to HTML report

johnwlambert:
As we discussed, we want a table in the report for each blue box in the computational graph. Currently the "PostProcessor" does not have one. This means that we cannot reason about the number of inliers for accepted image pairs, but only about the number of inliers for all image pairs (even if rejected), since we were trying to keep around that info.

This PR converts the old "Frontend Summary" box to a "Verify Report" box, and adds a "Postprocessor" Table.

Screen Shot 2021-10-18 at 7 01 56 PM

### DIFF
--- a/gtsfm/frontend/inlier_support_processor.py
+++ b/gtsfm/frontend/inlier_support_processor.py
@@ -1,4 +1,3 @@
-
 """Post-processor that uses information about RANSAC support (for verified correspondences) to filter out image pairs.
 
 Authors: John Lambert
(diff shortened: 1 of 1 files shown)


akshay-krishnan:
lack of context for me, do the image pairs get rejected in the verifier or in the post processor?

johnwlambert:
> lack of context for me, do the image pairs get rejected in the verifier or in the post processor?

@akshay-krishnan I'm proposing that we reject image pairs in the PostProcessor if they have insufficient support. And I propose we also fit the homography and check the homography support in the PostProcessor.

PR #362, opened 2021-10-22
Adding view graph estimator base class and cycle-consistent-rotations estimator

akshay-krishnan:
A Viewgraph is a  data structure that stores all the two-view relationships between cameras in the scene.

Estimating the ViewGraph can be done trivially by adding all the two-view estimates into a ViewGraph data structure.
The purpose of this PR, however, is to define an API for more sophisticated methods for estimating a ViewGraph 
that include filtering or optimizing the two-view estimates.

In future PRs, we could use this API for cycle consistency or other three-view filtering or optimization methods.


akshay-krishnan:
> It's a little difficult to see how this class fits into the pipeline without a specific use case implemented.

I think changes can always be made later, as long as the class in itself makes sense. I also prefer to keep PRs small. 

But anyway, Ayush started with changing the cycle consistency to use this in the same PR, so now we have to wait to complete this and merge the PR before we can use the class for other algorithms. Which motivates the original intent to only submit the class :)

akshay-krishnan:
Refactor of the existing cycle consistency code to use ViewGraphEstimator should now be complete, please take a look!

gtsfm/view_graph_estimator/view_graph_estimator_base.py
```diff
@@ -0,0 +1,123 @@
...
+
+METRIC_GROUP = "view_graph"
+
+
+class ViewGraphEstimatorBase(metaclass=abc.ABCMeta):
+    """Base class for ViewGraph estimation.
+
+    A ViewGraphEstimator aggregates two-view estimates into a ViewGraph.
+    It could also improve the two-view estimates using filtering or optimization techniques.
+    """
```
travisdriver, line 30: Is this module based off of another implementation you've seen. If so, could you add some references here?

akshay-krishnan: It’s not based on another implementation Ive seen. The term ViewGraph is borrowed/inspired from other papers. Have you seen similar modules? Happy to add references if you’ve seen any.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -347,6 +347,23 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
     return np.linalg.norm(wti1 - wti2)
 
 
+def compute_cyclic_rotation_error(i1Ri0: Rot3, i2Ri1: Rot3, i2Ri0: Rot3) -> float:
+    """Computes the cycle error in degrees after composing the three input rotations.
+
+    The cycle error is given by the angle of inv(i2Ri0) * i2Ri1 * i1Ri0.
```
ayushbaid, line 353: nit: can be a bit more descriptive in describing what the cycle is.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -347,6 +347,23 @@ def compute_points_distance_l2(wti1: Optional[Point3], wti2: Optional[Point3]) -
...
+
+    Args:
+        i1Ri0: Relative rotation of first edge.
+        i2Ri1: Relative rotation of second edge.
+        i2Ri0: Relative rotation of third edge.
```
ayushbaid, line 358: prefer telling edge from camera i1 to camera i2 and so on as we have not defined edges.

gtsfm/scene_optimizer.py
```diff
@@ -202,43 +200,7 @@ def create_computation_graph(
...
-                )
-            )
-            auxiliary_graph_list.append(
-                dask.delayed(save_full_frontend_metrics)(
-                    two_view_reports_dict_cycle_consistent,
-                    image_graph,
-                    filename="cycle_consistent_frontend_full.json",
-                )
-            )
+        # TODO: compute 2-view report post view graph optimization
```
ayushbaid, line 203: can we do it in the same PR? or will it require too much refactoring?

ayushbaid (approved):

gtsfm/scene_optimizer.py
akshay-krishnan: Ah, somehow missed this TODO, will do in same PR

gtsfm/common/view_graph.py
```diff
@@ -0,0 +1,34 @@
...
+    Args:
+        i2Ri1: dictionary of relative rotations between camera pairs.
+        i2Ui1: dictionary of relative unit-translations between camera pairs.
+        calibrations: list of calibrations for each camera.
+        corr_idxs_i1i2: dictionary of correspondence indices between camera pairs.
+    """
+
+    i2Ri1: Dict[Tuple[int, int], Rot3]
+    i2Ui1: Dict[Tuple[int, int], Unit3]
+    calibrations: List[Cal3Bundler]
```
travisdriver, line 25: Unless these are intended to be optional, I think it would be cleaner to represent this data as a `PinholeCameraCal3Bundler`.

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -0,0 +1,237 @@
...
+    Note: all summary statistics will be compared with an allowed upper bound/threshold. If they exceed the
+    upper bound, they will be rejected.
+    """
+
+    MIN_EDGE_ERROR = "MIN_EDGE_ERROR"
+    MEDIAN_EDGE_ERROR = "MEDIAN_EDGE_ERROR"
+
+
+class CycleConsistentRotationViewGraphEstimator(ViewGraphEstimatorBase):
+    """A ViewGraphEstimator that filters two-view edges with high rotation-cycle consistency error.
```
travisdriver, line 46: "A ViewGraphEstimator that filters two-view edges within a triplet with ..."

gtsfm/view_graph_estimator/view_graph_estimator_base.py
travisdriver: https://github.com/AIBluefisher/DAGSfM/blob/dev/src/graph/view_graph.h

gtsfm/view_graph_estimator/view_graph_estimator_base.py
```diff
@@ -0,0 +1,123 @@
...
+
+import gtsfm.utils.geometry_comparisons as comp_utils
+from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.common.view_graph import ViewGraph
+
+METRIC_GROUP = "view_graph"
+
+
+class ViewGraphEstimatorBase(metaclass=abc.ABCMeta):
```
travisdriver, line 25: What's the motivation behind making this an abstract class where filtering techniques are implemented as derived classes as opposed to a class with the filtering techniques as methods?

gtsfm/common/view_graph.py
travisdriver: Why are the calibrations stored as a list while all other information is a Dict? Is it assumed that the camera IDs are contiguous and start at 0?

gtsfm/common/view_graph.py
```diff
@@ -0,0 +1,34 @@
...
+"""
+from dataclasses import dataclass
+from typing import Dict, Tuple, List
+
+import numpy as np
+from gtsam import Rot3, Unit3, Cal3Bundler
+
+
+@dataclass
+class ViewGraph:
```
travisdriver, line 13: I'm slightly apprehensive about calling this a graph without defining the actual structure of the graph a little more rigorously. 

For example, I'm assuming the vertices should be the individual cameras and should store the calibrations and absolute pose information (set to identity if global poses have not been computed). The edges indicate that the cameras observe enough of the same landmarks to compute their relative geometry, and the edges should store this information along with the correspondences. I think representing the edges as a single dictionary with image pair keys would be a lot cleaner, or at least having some (public) master list of image pairs that is then used to access the (private) dictionaries.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
```diff
@@ -0,0 +1,123 @@
...
+            i2Ui1: Dict from (i1, i2) to relative translation direction of i1 with respect to i2.
+            calibrations: list of calibrations for each image.
+            corr_idxs_i1i2: Dict from (i1, i2) to indices of verified correspondences from i1 to i2.
+            keypoints: keypoints for each images.
+
+        Returns:
+            ViewGraph object.
+        """
+
+    def compute_metrics(
```
travisdriver, line 54: I like the idea of having a standard set of metrics for each estimator, but it seems as though CycleConsistentRotationViewGraphEstimator just overwrites this. I think this can be implemented in such a way that it could capture all relevant metrics for any derived class.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
```diff
@@ -0,0 +1,123 @@
...
+
+    @abc.abstractmethod
+    def run(
+        self,
+        i2Ri1: Dict[Tuple[int, int], Rot3],
+        i2Ui1: Dict[Tuple[int, int], Unit3],
+        calibrations: List[Cal3Bundler],
+        corr_idxs_i1i2: Dict[Tuple[int, int], np.ndarray],
+        keypoints: List[Keypoints],
+    ) -> ViewGraph:
```
travisdriver, line 40: I think this should also return the filtered edges and vertices, and then this could be passed to `compute_metrics` to compute the metrics with respect to both the inliers and outliers as it's done in the `CycleConsistentRotationViewGraphEstimator`.

travisdriver: Also, why not have `run` just take a `ViewGraph` as input? The same information that's being passed in is stored in the `Viewgraph` (except keypoints).

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: what is the motivation for a class with filtering techniques as methods? we would not use more than one of them at any time. 
The abstract class is the same design we follow elsewhere in gtsfm. For example, for detection we don't have one class with different methods for SIFT and deep front end. To me it captures the idea of the derived classes implementing different ways to achieve a common goal, using the same API.

gtsfm/common/view_graph.py
akshay-krishnan: > I think representing the edges as a single dictionary with image pair keys would be a lot cleaner

It would be cleaner, yes. But where is this structure for an edge actually used? The outputs of this class are fed to rotation and translation averaging which need dicts of rotations and translations (separately).  I am open to changing APIs for these classes but that needs more discussion and is beyond the scope of this PR. 

> or at least having some (public) master list of image pairs 

Yeah, we can add a member to store valid image pairs.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: I understand its not ideal to recompute them in compute_metrics(), but returning "inlier indices" assumes that the class only does filtering. It is designed to also support approaches that may refine the edges using optimization. 

about your second point, yeah I considered that. It will introduce some overhead in the user of this class to create a ViewGraph, and that seems like something this class should be doing.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: Yeah we can perhaps move those metrics here, and provide a method like _compute_additional_metrics that users can override to add more metrics if they like.

gtsfm/common/view_graph.py
travisdriver: Why not have absolute pose information? I think it would provide a more complete data structure.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
travisdriver: I assuming that we'd only ever want to keep track of a single ViewGraph (or multiple ViewGraphs that have been segmented from some global ViewGraph), as there is a lot of memory overhead as the number of views increase. 

Therefore, I think this would make more sense if the `ViewGraphEstimator.run`  took as input a ViewGraph and then output a filtered ViewGraph. This would make it more obvious how the outputs of the different estimators fit together.

gtsfm/common/view_graph.py
travisdriver: Representing the ViewGraph as an actual graph is a useful data abstraction, especially when we want to start considering [more sophisticated view filtering algorithms](https://arxiv.org/pdf/1902.02256.pdf). I think what you've done so far is great, and I think a very minimal refactoring will make the graph structure more obvious. Here's a quick draft of what I had in mind (somewhat motivated by [DAGSfM](https://github.com/AIBluefisher/DAGSfM/blob/dev/src/graph/view_graph.h)):

```python
class TwoViewData(NamedTuple):
    i2Ti1: Pose3
    corr_idxs_i1i2: np.ndarray
   
class ViewGraph:
    # Put these in an __init__ and initialize from their constituent parts
    self.view_pairs: List[Tuple[int, int]]
    self._views: Dict[int, PinholeCameraCal3Bundler]
    self._view_edges: Dict[Tuple[int, int], TwoViewData]
    
    # Add getters for the attributes of TwoViewData.
    def get_i2Ui1(i1, i2):
        return self._view_edges[(i1, i2)].i2Ti1.translation()

    # etc., etc.
```

travisdriver: > It would be cleaner, yes. But where is this structure for an edge actually used? The outputs of this class are fed to rotation and translation averaging which need dicts of rotations and translations (separately). I am open to changing APIs for these classes but that needs more discussion and is beyond the scope of this PR.

The edge is just holds any relative view information, e.g., relative translation and rotations (which you could combine into a pose or keep separate), two view correspondences, etc.

travisdriver: > It does not store the absolute pose information. I am open to other names if you think it is misleading. Maybe TwoViewGraph. The definition is clear to me: it is a structure that stores all the two-view relationships before averaging (including intrinsic priors). I can define an edge in the documentation if that is what your comment is about.

I think having the absolute information would be useful. Then, we could pass the ViewGraph to averaging, and it would output a new ViewGraph with the absolute pose data filled in.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: Where would you create the ViewGraph that is passed to the ViewGraphEstimator ? It seems odd to have a class for estimating a ViewGraph receive a ViewGraph as input. That API would surely surprise a reader?

travisdriver: According to your class description: "Estimating the ViewGraph can be done trivially by adding all the two-view estimates into a ViewGraph data structure. The purpose of this class, however, is to define an API for more sophisticated methods for estimating a ViewGraph that include filtering or optimizing the two-view estimates."

Creating the ViewGraph is trivial. The real purpose of this class is to filter the ViewGraph. You could change the name if you'd like.

gtsfm/common/view_graph.py
akshay-krishnan: I understood the design you had in mind from your previous comments, but I already mentioned my thoughts on it: 
- We cannot have PinholeCameraCal3Bundler as we do not have the absolute camera pose from two-view information. 
- I like the TwoViewData struct, but to pass inputs to RA and TA we would have to re-extract a Dict[Tuple[int, int], Rot3] and Dict[Tuple[int, int], Unit3], which is unnecessary. I think changes to RA and TA should be done in a different PR. 

The edge is holding two-view information even now. The question is how should this be internally represented. Multiple dicts vs single dict. I would prefer a representation that avoids redundant computation.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: Two reasons why I dont think we should pass in a ViewGraph: 
- We expect a ViewGraphEstimator to compute what is semantically defined as a  “ViewGraph”. These are good edges (say, cycle consistent) for averaging .
- The ViewGraph struct should not be used to store something we don’t semantically call a “ViewGraph”. This will lead to confusion. 
- More overhead (even if its two lines) in MVO

Only reason why we should pass in a ViewGraph: 
- Cleaner API for the ViewGraph class

gtsfm/common/view_graph.py
travisdriver: > We cannot have PinholeCameraCal3Bundler as we do not have the absolute camera pose from two-view information.

I think this class could useful even after RA and TA, so I think at least having the option to populate the absolute poses later would be a good idea.

> I like the TwoViewData struct, but to pass inputs to RA and TA we would have to re-extract a Dict[Tuple[int, int], Rot3] and Dict[Tuple[int, int], Unit3], which is unnecessary. I think changes to RA and TA should be done in a different PR.

Sure, then just forego the `TwoViewData` for now and just have something like:

```python
class ViewGraph:
    # Put these in an __init__ and initialize from their constituent parts
    self.view_pairs: List[Tuple[int, int]]
    self._views: Dict[int, PinholeCameraCal3Bundler]
    self._i1Ri2: Dict[Tuple[int, int], Rot3]
    self._i1ti2: Dict[Tuple[int, int], Union[Unit3, Point3]]
    self._corr_idxs_i1i2: Dict[Tuple[int, int], np.ndarray]
    # optionally add keypoints?
    
    # Add getters for the attributes of TwoViewData.
    def get_i2Ui1(i1, i2):
        return self._view_edges[(i1, i2)].i2Ti1.translation()

    # etc., etc.
```

akshay-krishnan: Sure, although I would stick with Cal3Bundler and introduce PinholeCamera only in a PR where its populated.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: Could you suggest better names if you think we should pass in a ViewGraph? Maybe ViewGraphRefiner.

travisdriver: > We expect a ViewGraphEstimator to compute what is semantically defined as a “ViewGraph”. These are good edges (say, cycle consistent) for averaging.

It would still be computing a ViewGraph, just from another ViewGraph. You could change it to something like `ViewGraphOperator` or something like that idk. 

> The ViewGraph struct should not be used to store something we don’t semantically call a “ViewGraph”. This will lead to confusion.

I'm not following.

akshay-krishnan: Maybe this is also a change to be made when we use the ViewGraph beyond filtering two-view edges.

gtsfm/common/view_graph.py
travisdriver: That works for me. 

Would you mind creating an issue outlining that we'd eventually like to pass a ViewGraph to the averaging module (which would then populate the absolute poses)?

gtsfm/view_graph_estimator/view_graph_estimator_base.py
travisdriver: > I understand its not ideal to recompute them in compute_metrics(), but returning "inlier indices" assumes that the class only does filtering. It is designed to also support approaches that may refine the edges using optimization.

If no edges got filtered, then it simply wouldn't return a filtered graph; just make it Optional.

> about your second point, yeah I considered that. It will introduce some overhead in the user of this class to create a ViewGraph, and that seems like something this class should be doing.

What would the overhead be? You can just create a ViewGraph after running the TwoView Estimator on all the putative image pairs and then all the ViewGraphEstimators can operate off that.

johnwlambert:
@akshay-krishnan looks like there is a merge conflict in `scene_optimizer.py` -- could you take a look when you get a chance?

gtsfm/common/view_graph.py
ayushbaid: View-graph is meant to represent a graph forming out of 2-view relative measurements. We have one for absolute data right now: `GtsfmData`. In the future we can evaluate if we need to represent the relative and absolute measurements, but we do not have any documented use-case/to-do for this right now.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
ayushbaid: Re the abstract class design: this is what we follow everywhere in `GtSFM`. Coding to interface is the way to go, even if we have one implementation right now.

I do not understand the second point of Travis. I think we had this discussion long back on this PR, where I suggested to have `ViewGraph` as input and as output too. But Akshay commented (and I agree with that) that someone has to construct that view-graph data structure too. We cannot do it in two-view plate, so multi-view remains the only choice.

I wrap my head around this by thinking that `ViewGraph` is a graph made of good 2-view relationships (like say rotation-cycle-consistent).

gtsfm/common/view_graph.py
johnwlambert: I also am not seeing the use case currently. If it's compelling, I'm happy to rethink it

gtsfm/common/view_graph.py
ayushbaid: I am not convinced about this design of having an incomplete data structure, and one-by-one modules filling up more information in the same structure.

Are there any advantage of this approach? Having attributes which can be `None` is something I try to avoid.

What we want in this PR is a way to capture two-view relationships (edges) between images. And we want to just keep the edges which are good 2-view relationship. Having absolute poses here just muddies what we introduce as the output to the `ViewGraphEstimator` module. Having the estimator output an object with missing attributes do not make sense to me. 

Also, I agree with Akshay that we should just pass as inputs to a module which is needed for computation. As we follow a functional pattern for gtsfm, passing in objects with `None` attributes and filling it in the module is not something we would do. We would make a copy of the whole data-structure. I find this needless.

I would infact advocate to remove calibrations from the dataclass too, and have just edge information there. No need to put any property of the node in there.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
ayushbaid: If the derived implementation wants to override the method, they can call the super method. We can document it in the docstring. I am against having any *additional* function.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
ayushbaid: > If no edges got filtered, then it simply wouldn't return a filtered graph; it can Optional.

This is not functional style.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
travisdriver: > I do not understand the second point of Travis. I think we had this discussion long back on this PR, where I suggested to have ViewGraph as input and as output too. But Akshay commented (and I agree with that) that someone has to construct that view-graph data structure too. We cannot do it in two-view plate, so multi-view remains the only choice.

Creating the ViewGraph is trivial, just fill it with the output of the two-view estimation loop. It creats all of the require data products anyways: `i2Ri1_graph_dict`, `i2Ui1_graph_dict`, `v_corr_idxs_graph_dict`. Then just simply pass these to the ViewGraph.

gtsfm/common/view_graph.py
travisdriver: > What we want in this PR is a way to capture two-view relationships (edges) between images. And we want to just keep the edges which are good 2-view relationship. Having absolute poses here just muddies what we introduce as the output to the ViewGraphEstimator module. Having the estimator output an object with missing attributes do not make sense to me.

Edges encode correspondences between images, and correspondences can be filtered using the absolute poses, e.g., what we're already doing in data association.

> Also, I agree with Akshay that we should just pass as inputs to a module which is needed for computation. As we follow a functional pattern for gtsfm, passing in objects with None attributes and filling it in the module is not something we would do. We would make a copy of the whole data-structure. I find this needless.

It doesn't have to be None; it could just be identity. Also, `GtsfmData` is "filled in" after initialization, and there are plenty of other examples of classes with Optional attributes.

> I would infact advocate to remove calibrations from the dataclass too, and have just edge information there. No need to put any property of the node in there.

If the nodes are inconsequential, then this isn't really a graph.

akshay-krishnan: > I would infact advocate to remove calibrations from the dataclass too, and have just edge information there. No need to put any property of the node in there.

@ayushbaid I would be okay with removing calibrations for now as RCC does not use it. We can introduce it in the same PR as the fundamental matrix based cycle-consistency if that works well.

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: Yeah but like I mentioned that changes the concept of the estimator from something that creates a ViewGraph to something that refines/changes it. This questions the very meaning of a ViewGraph. Is it just a collection of two-view poses or is it two-view noise-free poses that can be passed to averaging? Throughout the design, we have thought of it as the latter.

akshay-krishnan: I dont feel too strongly about this, but we would have to change the name of the class as well, and establish a meaning for the ViewGraph. open to comments from @ayushbaid and @johnwlambert

gtsfm/view_graph_estimator/view_graph_estimator_base.py
akshay-krishnan: > What would the overhead be? You can just create a ViewGraph after running the TwoView Estimator on all the putative image pairs and then all the ViewGraphEstimators can operate off that.

I am talking about just the two lines. Overlapping discussion in the other thread.

gtsfm/multi_view_optimizer.py
```diff
@@ -50,48 +54,74 @@ def create_computation_graph(
...
+        )
+        view_graph_two_view_report = dask.delayed(filter_dict_keys)(
+            two_view_reports_dict, dask.delayed(lambda x: x.i2Ri1)(view_graph)
+        )
 
-        pruned_i2Ri1_graph = pruned_graph[0]
-        pruned_i2Ui1_graph = pruned_graph[1]
+        # prune the graph to a single connected component.
+        pruned_i2Ri1_graph, pruned_i2Ui1_graph = dask.delayed(graph_utils.prune_to_largest_connected_component, nout=2)(
+            dask.delayed(lambda x: x.i2Ri1)(view_graph), dask.delayed(lambda x: x.i2Ui1)(view_graph)
```
johnwlambert, line 80: @akshay-krishnan what's the advantage of returning a `ViewGraph` type? It looks like we have to use these `dask.delayed()` calls to access its attributes, which seems less clean than just returning 3 return args

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -0,0 +1,237 @@
+"""A ViewGraphEstimator implementation which ensures relative rotations are consistent in the cycles of the graph.
+
```
johnwlambert, line 2: @akshay-krishnan can we add back in these original references?

```
    Based off of:
        https://github.com/sweeneychris/TheiaSfM/blob/master/src/theia/sfm/filter_view_graph_cycles_by_rotation.cc
    See also:
        C. Zach, M. Klopschitz, and M. Pollefeys. Disambiguating visual relations using loop constraints. In CVPR, 2010
        http://people.inf.ethz.ch/pomarc/pubs/ZachCVPR10.pdf
        Enqvist, Olof; Kahl, Fredrik; Olsson, Carl. Non-Sequential Structure from Motion. ICCVW, 2011.
        https://portal.research.lu.se/ws/files/6239297/2255278.pdf
```

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -1,343 +0,0 @@
...
-        if error_aggregate < CYCLE_ERROR_THRESHOLD:
-            cycle_consistent_keys.add((i1, i2))
-            inlier_errors_aggregate.append(error_aggregate)
-            inlier_errors_wrt_gt.append(two_view_reports_dict[(i1, i2)].R_error_deg)
-        else:
-            outlier_errors_aggregate.append(error_aggregate)
-            outlier_errors_wrt_gt.append(two_view_reports_dict[(i1, i2)].R_error_deg)
-
-    if visualize:
-        plt.scatter(
```
johnwlambert, line 229: @akshay-krishnan I've found these scatter plots to be very useful, so I'd like to keep them

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -1,343 +0,0 @@
...
-    logger.info("cycle_consistent_keys: " + str(cycle_consistent_keys))
-
-    i2Ri1_dict_consistent, i2Ui1_dict_consistent, v_corr_idxs_dict_consistent = {}, {}, {}
-    for (i1, i2) in cycle_consistent_keys:
-        i2Ri1_dict_consistent[(i1, i2)] = i2Ri1_dict[(i1, i2)]
-        i2Ui1_dict_consistent[(i1, i2)] = i2Ui1_dict[(i1, i2)]
-        v_corr_idxs_dict_consistent[(i1, i2)] = v_corr_idxs_dict[(i1, i2)]
-
-    n_valid_edges = len([i2Ri1 for (i1, i2), i2Ri1 in i2Ri1_dict.items() if i2Ri1 is not None])
-    logger.info("Found %d consistent rel. rotations from %d original edges.", len(i2Ri1_dict_consistent), n_valid_edges)
```
johnwlambert, line 268: it would be nice to still log this summary info @akshay-krishnan

gtsfm/view_graph_estimator/view_graph_estimator_base.py
travisdriver: I think just having the single line to create the initial ViewGraph would be much less overhead than having to deconstruct it into its constituent parts everytime we want to pass the output to another estimator IMHO

gtsfm/averaging/rotation/cycle_consistency.py
ayushbaid: @johnwlambert Do we want something out of this data in the report too? Things like precision and recall?

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
akshay-krishnan: Added to the class documentation below.

gtsfm/averaging/rotation/cycle_consistency.py
akshay-krishnan: added it.

gtsfm/averaging/rotation/cycle_consistency.py
akshay-krishnan: added the plots and the precision recall metrics.

johnwlambert:
thanks for the updates, @akshay-krishnan. I see some of the jobs are failing with:

```
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/tornado/ioloop.py", line 741, in _run_callback
    ret = callback()
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/tornado/ioloop.py", line 765, in _discard_future_result
    future.result()
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/worker.py", line 1301, in handle_scheduler
    await self.handle_stream(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/core.py", line 569, in handle_stream
    msgs = await comm.read()
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/comm/tcp.py", line 233, in read
    msg = await from_frames(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/comm/utils.py", line 78, in from_frames
    res = _from_frames()
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/comm/utils.py", line 61, in _from_frames
    return protocol.loads(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/core.py", line 111, in loads
    return msgpack.loads(
  File "msgpack/_unpacker.pyx", line 194, in msgpack._cmsgpack.unpackb
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/core.py", line 103, in _decode_default
    return merge_and_deserialize(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/serialize.py", line 488, in merge_and_deserialize
    return deserialize(header, merged_frames, deserializers=deserializers)
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/serialize.py", line 401, in deserialize
    deserialize(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/serialize.py", line 417, in deserialize
    return loads(header, frames)
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/serialize.py", line 96, in pickle_loads
    return pickle.loads(x, buffers=new)
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/protocol/pickle.py", line 75, in loads
    return pickle.loads(x)
AttributeError: 'CycleConsistentRotationViewGraphEstimator' object has no attribute '__filter_with_edges'
```

johnwlambert:
A few last issues I'm seeing:
```
  File "gtsfm/runner/run_scene_optimizer_colmaploader.py", line 45, in <module>
    runner.run()
  File "/home/runner/work/gtsfm/gtsfm/gtsfm/runner/gtsfm_runner_base.py", line 106, in run
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/dask/base.py", line 288, in compute
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/dask/base.py", line 571, in compute
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 2691, in get
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 1946, in gather
    return self.sync(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/utils.py", line 310, in sync
    return sync(
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/utils.py", line 364, in sync
    raise exc.with_traceback(tb)
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/utils.py", line 349, in f
    result[0] = yield future
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/tornado/gen.py", line 762, in run
    value = future.result()
  File "/usr/share/miniconda3/envs/gtsfm-v1/lib/python3.8/site-packages/distributed/client.py", line 1811, in _gather
  File "/home/runner/work/gtsfm/gtsfm/gtsfm/view_graph_estimator/view_graph_estimator_base.py", line 136, in compute_metrics
    R_precision, R_recall = metrics_utils.get_precision_recall_from_errors(
  File "/home/runner/work/gtsfm/gtsfm/gtsfm/utils/metrics.py", line 393, in get_precision_recall_from_errors
    fn = np.sum(np.array(negative_errors) <= max_positive_error)
TypeError: '<=' not supported between instances of 'NoneType' and 'float'
```
and
```
Run flake8 --max-line-length 120 --ignore E201,E202,E203,E231,W291,W293,E303,W391,E402,W503,E731 gtsfm
gtsfm/view_graph_estimator/view_graph_estimator_base.py:10:1: F401 'logging.Logger' imported but unused
gtsfm/view_graph_estimator/view_graph_estimator_base.py:11:1: F401 'typing.Optional' imported but unused
gtsfm/view_graph_estimator/view_graph_estimator_base.py:29:1: E302 expected 2 blank lines, found 1
gtsfm/view_graph_estimator/view_graph_estimator_base.py:67:121: E501 line too long (157 > 120 characters)
gtsfm/view_graph_estimator/view_graph_estimator_base.py:184:121: E501 line too long (122 > 120 characters)
gtsfm/view_graph_estimator/view_graph_estimator_base.py:185:121: E501 line too long (139 > 120 characters)
gtsfm/view_graph_estimator/view_graph_estimator_base.py:188:121: E501 line too long (138 > 120 characters)
gtsfm/view_graph_estimator/view_graph_estimator_base.py:189:121: E501 line too long (128 > 120 characters)
gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py:8:1: F401 'typing.Optional' imported but unused
gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py:18:1: F401 'gtsfm.evaluation.metrics.GtsfmMetricsGroup' imported but unused
```

akshay-krishnan:
Thanks John, I would normally run and make sure there are no errors, but Ive not been able to get the pipeline running on my new env yet. So Im relying on the CI here. Will fix these later today.

johnwlambert:
Sounds good. Thanks!

travisdriver:
Could you post the benchmarks dashboard just to make sure nothing regressed when switching to this new module?

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
johnwlambert: thanks!

tests/averaging/rotation/test_cycle_consistency.py
```diff
@@ -1,165 +0,0 @@
...
-import numpy as np
-import pytest
-from gtsam import Rot3, Unit3
-
-import gtsfm.averaging.rotation.cycle_consistency as cycle_utils
-from gtsfm.evaluation.metrics import GtsfmMetricsGroup
-from gtsfm.two_view_estimator import TwoViewEstimationReport
-
-
-def test_compute_cycle_error_known_GT() -> None:
```
johnwlambert, line 17: making a note for myself that this test was renamed to `test_compute_cyclic_rotation_error()`

tests/view_graph_estimator/test_cycle_consistent_rotation_estimator.py
```diff
@@ -0,0 +1,61 @@
...
+        (0, 2): Unit3(np.array([1, 0, 0])),
+        (2, 3): Unit3(np.array([1, 0, 0])),
+        (3, 4): Unit3(np.array([1, 0, 0])),
+        (2, 4): Unit3(np.array([1, 0, 0])),
+    }
+    # The ViewGraphEstimator assumes these dicts contain the keys corresponding to the the rotation edges.
+    calibrations = {k: Cal3Bundler() for k in range(0, 5)}
+    corr_idxs_i1i2 = {i1i2: np.array([]) for i1i2 in i2Ri1_dict.keys()}
+    keypoints = {k: np.array([]) for k in range(0, 5)}
+    two_view_reports = None
```
johnwlambert, line 54: the CI says that:
```python
>               inlier_errors_wrt_gt.append(two_view_reports_dict[(i1, i2)].R_error_deg)
E               TypeError: 'NoneType' object is not subscriptable
```

Before I was populating this with:
```python
    two_view_reports_dict = {}
    for (i1, i2) in i2Ri1_dict.keys():
        two_view_reports_dict[(i1, i2)] = TwoViewEstimationReport(
            v_corr_idxs=np.array([]),  # dummy array
            num_inliers_est_model=10,  # dummy value
        )
```
which i believe is necessary

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -0,0 +1,231 @@
...
+        self,
+        edge_error_aggregation_criterion: EdgeErrorAggregationCriterion,
+        error_threshold: float = ERROR_THRESHOLD,
+    ) -> None:
+        self._edge_error_aggregation_criterion = edge_error_aggregation_criterion
+        self._error_threshold = error_threshold
+
+    def run(
+        self,
+        i2Ri1: Dict[Tuple[int, int], Rot3],
```
johnwlambert, line 76: @akshay-krishnan I think we should name this arg `i2Ri1_dict`, since `i2Ri1`means a single rotation elsewhere, but here we mean a dict

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -0,0 +1,231 @@
...
+        2. i2Ri1 is not None
+
+        Args:
+            i2Ri1: input dictionary of relative rotations.
+
+        Returns:
+            List of valid edges.
+        """
+        valid_edges = []
+        for (i1, i2), i2Ri1 in i2Ri1.items():
```
johnwlambert, line 206: @akshay-krishnan we need to rename this variable otherwise we are overwriting it in the loop
```python
for (i1, i2), i2Ri1 in i2Ri1.items():
```

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
johnwlambert: I think the current docstring is actually a bit more accurate. We aren't filtering two-view edges that occur within a single triplet, but rather that occur across potentially many triplets.

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -114,11 +116,14 @@ def run(
...
-                gt_error2 = two_view_reports[(i1, i2)].R_error_deg
-                gt_error3 = two_view_reports[(i0, i2)].R_error_deg
-                max_gt_error_in_cycle.append(max(gt_error1, gt_error2, gt_error3))
+            
+            # form 3 edges e_i, e_j, e_k between fully connected subgraph (nodes i0,i1,i2)
+            edges = [(i0, i1), (i1, i2), (i0, i2)]
+            rot_errors = [two_view_reports[e].R_error_deg for e in edges]
+            gt_known = all([err is not None for err in rot_errors])
+            # if ground truth unknown, cannot estimate error w.r.t. GT
+            max_rot_error = float(np.max(rot_errors)) if gt_known else None
```
akshay-krishnan, line 125: nit: maybe just use max() instead of float(np.max()) ?

akshay-krishnan (commented):
Thanks for cleaning it up, John!

johnwlambert:
Here's the latest report:
[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/7883305/visual_comparison_dashboard.html.zip)

johnwlambert:
report vs. the previous commit on master (to show stochasticitiy)
[viewgraphest-report-v2.zip](https://github.com/borglab/gtsfm/files/7883343/viewgraphest-report-v2.zip)

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
johnwlambert: good point -- I'll fix in a follow up PR, just to avoid having to run the CI again now after I've checked it

travisdriver:
> report vs. the previous commit on master (to show stochasticitiy) [viewgraphest-report-v2.zip](https://github.com/borglab/gtsfm/files/7883343/viewgraphest-report-v2.zip)

Can you explain more about what this is showing? The previous dashboard you posted showed some pretty significant regression.

johnwlambert:
> > report vs. the previous commit on master (to show stochasticitiy) [viewgraphest-report-v2.zip](https://github.com/borglab/gtsfm/files/7883343/viewgraphest-report-v2.zip)
> 
> Can you explain more about what this is showing? The previous dashboard you posted showed some pretty significant regression.

Sure, thanks for taking a look Travis. The first diff is w.r.t. c2ff9bee7743efaa0a4af8228dae9a377a7a2df6, and the second diff is w.r.t. 04a83656b645cc46bb22b7769b6ca0237b143a02

Both were changes to envs/ tests (not any core logic), so there is some weird stochasticity going on still.

johnwlambert:
Here's the diff vs. master from an earlier actions run on the ViewGraphEstimator branch. Here the changes seem random for other benchmarks:
[visual_comparison_dashboard 4.html.zip](https://github.com/borglab/gtsfm/files/7883731/visual_comparison_dashboard.4.html.zip)

johnwlambert (approved):

travisdriver:
@johnwlambert The RCC portion of the benchmark dashboard seemed to be broken in all of the benchmarks you posted. I think it would have been appropriate to fix this before merging this branch to assess performance, as this branch reimplements the RCC module. 

I also think some more investigation should have been done before writing off the (rather significant) performance regressions as stochasticity.

newplot (9)

johnwlambert:
> 

Hi Travis, you're right about the report -- unfortunately the dashboard currently cannot handle a diff when the table names are changed. The new PR here has the fix: https://github.com/borglab/gtsfm/pull/422

I looked at each of the cycle consistency stages for the runs where we saw varying performance, and the cycle consistency output was identical / metrics were identical. So I'm fairly confident there. I'll take a look at the CI again when the next round finishes, as well, just to confirm one more time.

PR #363, opened 2021-10-22
Move triplet extraction from cycle consistency to graph utils

akshay-krishnan:
Move triplet extraction from cycle consistency to graph utils, so that it can be re-used in other places. 

Also moved the unit tests.

### DIFF
--- a/tests/utils/test_graph_utils.py
+++ b/tests/utils/test_graph_utils.py
@@ -129,8 +129,8 @@ class TestGraphUtils(unittest.TestCase):
         triplets = graph_utils.extract_cyclic_triplets_from_edges(edges)
         assert isinstance(triplets, list)
         assert len(triplets) == 2
-        assert triplets[0] == (1, 3, 5)
-        assert triplets[1] == (1, 2, 3)
+        assert triplets[0] == (1, 2, 3)
+        assert triplets[1] == (1, 3, 5)
 
     def test_extract_triplets_3(self) -> None:
         """Ensure triplets are recovered accurately via intersection of adjacency lists.


johnwlambert:
Looks like a few tests are failing:
```
gtsfm/averaging/rotation/cycle_consistency.py:13:1: F401 'typing.DefaultDict' imported but unused
gtsfm/averaging/rotation/cycle_consistency.py:13:1: F401 'typing.Set' imported but unused
gtsfm/utils/graph.py:63:54: F821 undefined name 'DefaultDict'
gtsfm/utils/graph.py:63:71: F821 undefined name 'Set'
gtsfm/utils/graph.py:76:16: F821 undefined name 'defaultdict'
gtsfm/utils/graph.py:101:38: F821 undefined name 'i2Ri1_dict'
```

gtsfm/averaging/rotation/cycle_consistency.py
```diff
@@ -248,7 +181,16 @@ def filter_to_cycle_consistent_edges(
 
     per_edge_errors = defaultdict(list)
 
-    triplets = extract_triplets(i2Ri1_dict)
+    # Extract edges that are ordered correctly and have a valid rotation.
+    valid_edges = []
+    for (i1, i2), i2Ri1 in i2Ri1_dict.items():
+        if i2Ri1 is None or i1 > i2:
```
johnwlambert, line 187: i think this should be i1 >= i2

johnwlambert (approved):
LGTM except the failing CI stuff. Feel free to merge after those are fixed.

Can you also confirm the metrics are unchanged?

gtsfm/utils/graph.py
```diff
@@ -58,3 +58,62 @@ def prune_to_largest_connected_component(
...
+    """
+    adj_list = create_adjacency_list(i2Ri1_dict)
+
+    # only want to keep the unique ones
+    triplets = set()
+
+    # find intersections
+    for (a, b) in edges:
+        if a > b:
+            a, b = b, a
```
johnwlambert, line 109: why do we need to enforce this here?

akshay-krishnan: so that the output triplets are ordered correctly even if the input is not? 
I basically did not want to change the input and output of the crux of your implementation. this will ensure that inputs are ordered even if some user does not pass them as ordered.

PR #366, opened 2021-10-25
use economy svd for axis alignment

johnwlambert:


gtsfm/utils/ellipsoid.py
```diff
@@ -31,11 +31,15 @@ def get_ortho_axis_alignment_transform(gtsfm_data: GtsfmData) -> Pose3:
 
     # Filter outlier points, Center point cloud, and obtain alignment rotation.
     points_filtered = remove_outlier_points(point_cloud)
-    points_centered, mean = center_point_cloud(points_filtered)
+    points_centered = center_point_cloud(points_filtered)
     wuprightRw = get_alignment_rotation_matrix_from_svd(points_centered)
 
+    # Calculate translation vector based off rotated point cloud.
+    point_cloud_rotated = (wuprightRw @ point_cloud.T).T
```
johnwlambert, line 38: I think `point_cloud @ wuprightRw.T` is cleaner IMHO

gtsfm/utils/ellipsoid.py
```diff
@@ -31,11 +31,15 @@ def get_ortho_axis_alignment_transform(gtsfm_data: GtsfmData) -> Pose3:
 
     # Filter outlier points, Center point cloud, and obtain alignment rotation.
     points_filtered = remove_outlier_points(point_cloud)
-    points_centered, mean = center_point_cloud(points_filtered)
+    points_centered = center_point_cloud(points_filtered)
     wuprightRw = get_alignment_rotation_matrix_from_svd(points_centered)
 
+    # Calculate translation vector based off rotated point cloud.
```
adisingh50, line 37: I realized the translation vector that I'd been using in the final `walignedTw` pose was incorrect. Previously, I was using the `-1 * mean` of the **original point cloud** and using that as the translation. However, I realized that first rotation is applied, then translation. Thus, the correct translation to use is `-1 * mean` of the **rotated point cloud** (so that after rotation is applied, I can center it).

johnwlambert: nice, can you show what it looks like both ways?

gtsfm/utils/ellipsoid.py
adisingh50: Agreed

gtsfm/utils/ellipsoid.py
adisingh50: With the incorrect translation, the point cloud renders as:
image

With the correct translation vector, the point cloud renders as:
image

gtsfm/utils/ellipsoid.py
```diff
@@ -78,8 +83,9 @@ def remove_outlier_points(point_cloud: np.ndarray) -> np.ndarray:
 
     mags = np.linalg.norm(point_cloud, axis=1)
     cutoff_mag = np.percentile(mags, OUTLIER_DISTANCE_PERCENTILE)
+    inlier_mask = mags < cutoff_mag
```
johnwlambert, line 86: maybe we can use the inlier mask on the line below, also?

adisingh50: yup. forgot to change to update that.

johnwlambert:
LGTM

adisingh50:
Getting ready to merge this PR into master. Some final visualizations:

SIFT Frontend: Door 12
image

Deep Frontend: Door 12
image

SIFT Frontend: Skydio Crane 8
image

Deep Frontend was taking too long to run on Skydio Crane 32 dataset, so I decided to not visualize that one.

PR #373, opened 2021-11-05
remove unnecessary TwoViewReport attributes

johnwlambert:
i2Ri1 and i2Ui1 shouldn't be in this report (I don't remember how they got in here, probably was from some old debugging i did while prototyping)...

### DIFF
--- a/gtsfm/common/two_view_estimation_report.py
+++ b/gtsfm/common/two_view_estimation_report.py
@@ -31,8 +31,6 @@ class TwoViewEstimationReport:
             Sampson error (using GT epipolar geometry).
         R_error_deg: relative pose error w.r.t. GT. Only defined if GT poses provided.
         U_error_deg: relative translation error w.r.t. GT. Only defined if GT poses provided.
-        i2Ri1: relative rotation.
-        i2Ui1: relative translation direction.
         reproj_error_gt_model: reprojection errors between correspondences w.r.t. GT.
         inlier_avg_reproj_error_gt_model: average reprojection error of inliers.
         outlier_avg_reproj_error_gt_model: average reprojection error of outliers.
@@ -46,8 +44,6 @@ class TwoViewEstimationReport:
     v_corr_idxs_inlier_mask_gt: Optional[np.ndarray] = None
     R_error_deg: Optional[float] = None
     U_error_deg: Optional[float] = None
-    i2Ri1: Optional[Rot3] = None
-    i2Ui1: Optional[Unit3] = None
     reproj_error_gt_model: Optional[np.ndarray] = None
     inlier_avg_reproj_error_gt_model: Optional[float] = None
     outlier_avg_reproj_error_gt_model: Optional[float] = None


travisdriver (changes requested):
I agree that these are unnecessary. However, they are still set [here](https://github.com/borglab/gtsfm/blob/b298d6a2bc17e3ad804b672af0959d1b98988056/gtsfm/frontend/inlier_support_processor.py#L93) and possibly other places.

Also, now `two_view_report_pp` is really just an exact copy of `two_view_report` and should be removed.

johnwlambert:
> I agree that these are unnecessary. However, they are still set [here](https://github.com/borglab/gtsfm/blob/b298d6a2bc17e3ad804b672af0959d1b98988056/gtsfm/frontend/inlier_support_processor.py#L93) and possibly other places.
> 
> Also, now `two_view_report_pp` is really just an exact copy of `two_view_report` and should be removed.

Thanks for finding the other reference, I had missed that one. It's removed now.

@dellaert and I pair-programmed the new output of the Postprocessor (InlierSupportProcessor) last week and added this. The main idea behind it is:
(1) We have to be able to know the statistics of the raw matches, vs. the postprocessed matches.
(2) Two different variables that represent different quantities cannot be named the same thing.
(3) We have to follow the graph -- each blue box creates its own table. In this case, the Verifier has its own table in the report, and the InlierSupportProcessor also has its own table in the report.

Screen Shot 2021-11-05 at 8 56 37 AM

travisdriver:
Ok, but now `two_view_report` and `two_view_report_pp` are now just exact copies of each other (with `two_view_report_pp` being `None` if there are not enough inliers). 

I'm not sure that just copying a variable and changing its name solves any of the action items you presented in an effective way.

akshay-krishnan:
I think we are lacking a good concept/meaning/purpose for the TwoViewEstimationReport. So far, it has included (almost?) everything from the two view estimator. 

IF the intended purpose of the report is to store metrics, then all data members that have a bigger use case than a metric can be moved to a different data structure. By that I mean the verified correspondences, inlier ratio, rotation, translation etc. 
This PR removes the rotation and translation, which is good, but can we clean up the rest too?

akshay-krishnan (commented):
Can we remove `v_corr_idxs: np.ndarray` and `v_corr_idxs_inlier_mask_gt: Optional[np.ndarray] = None` too? 
They are not metrics. They are used to compute the inlier ratio metric which is already another member.

johnwlambert:
Recap of decisions:
each blue box creates both metrics (pink box?) and a result (gray box). These are 2 separate data structures. At some point, each blue box could inherit from a class, and have a method def __produce_metrics(self):

travisdriver:
I agree that `v_corr_idxs` and `reproj_error_gt_model` should probably be removed. 

I'm ambivalent about removing `v_corr_idxs_inlier_mask_gt` because it's used to [visualize the metrics](https://github.com/borglab/gtsfm/blob/a7462f07a77250da0f0e913a16d23ea053ed800d/gtsfm/utils/viz.py#L264).

Issue #306, opened 2021-09-19
Investigate further inlier ratio w.r.t. GT model and w.r.t estimated model


johnwlambert:
Currently, the inlier ratio with respect to GT model is greater than that with respect to estimated model for USAC. This is perhaps expected, as the GT should explain the data most appropriately. But our inability to estimate a model with close to as many inliers is unfortunate, and perhaps could be explored further on some representative image pairs.

We see this with USAC on Skydio-8:
<img width="954" alt="Screen Shot 2021-09-18 at 10 56 36 PM" src="https://user-images.githubusercontent.com/16724970/133913943-a5f4d284-8c6e-477b-a5f0-4b47ed5a7430.png">

However, we do not see this with LORANSAC on Skydio-8
<img width="895" alt="Screen Shot 2021-09-18 at 11 01 24 PM" src="https://user-images.githubusercontent.com/16724970/133914006-b20a4aea-6801-436b-a30f-0a144054b9ab.png">

cc @akshay-krishnan @ayushbaid

akshay-krishnan:
It is expected to have a high inlier ratio for the GT model (if the matcher output is good), but I don't understand why the inlier ratio with respect to the estimated model is lesser than the GT model. 
The GT model was not estimated from our data, and if we had enough verifier iterations, we should have the model with the highest number of inliers. I'm not familiar with how USAC works though.

Another reason this could happen is if when we have degenerate cases (all inliers on a plane) and our estimated model is not good, but skydio-8 is not such a dataset right?

PR #374, opened 2021-11-05
Add report from matcher with correct inlier ratio computation

johnwlambert:
This actually computes inlier ratio w.r.t. gt model with # putatives as the denominator.

- Add MatcherReport class and .evaluate() method of base matcher class

Solves https://github.com/borglab/gtsfm/issues/306

### DIFF
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -122,6 +122,7 @@ class SceneOptimizer:
             descriptors_graph_list += [delayed_descs]
 
         # Estimate two-view geometry and get indices of verified correspondences.
+        matcher_reports_dict = {}
         i2Ri1_graph_dict = {}
         i2Ui1_graph_dict = {}
         v_corr_idxs_graph_dict: Dict[Tuple[int, int], np.ndarray] = {}
@@ -142,6 +143,7 @@ class SceneOptimizer:
                 v_corr_idxs,
                 two_view_report,
                 two_view_report_pp,
+                matcher_report,
             ) = self.two_view_estimator.create_computation_graph(
                 keypoints_graph_list[i1],
                 keypoints_graph_list[i2],
@@ -162,6 +164,7 @@ class SceneOptimizer:
             v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
             two_view_reports_dict[(i1, i2)] = two_view_report
             two_view_reports_pp_dict[(i1, i2)] = two_view_report_pp
+            matcher_reports_dict[(i1, i2)] = matcher_report
 
             # Visualize verified two-view correspondences.
             if self._save_two_view_correspondences_viz:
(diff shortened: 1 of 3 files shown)


travisdriver:
I'm not sure I really agree with any of these changes. I think spreading the evaluation tasks around to all the different modules just increases complexity. I think more responsibility should be given to the `TwoViewEstimationReport` (or other derived subclasses).

Moreover, we are now passing around three different reports (two of which are just copies of each other) that could easily be storing the same information.

gtsfm/scene_optimizer.py
```diff
@@ -122,6 +122,7 @@ def create_computation_graph(
             descriptors_graph_list += [delayed_descs]
 
         # Estimate two-view geometry and get indices of verified correspondences.
+        matcher_reports_dict = {}
```
travisdriver, line 125: Why are we creating and passing around an additional report that just holds the number of matches and inlier ratio, both of which are already stored in the `TwoViewEstimationReport`.

travisdriver:
How does this solve #306?

gtsfm/scene_optimizer.py
johnwlambert: The main idea behind this is the following direction from @dellaert :
(1) Each blue box in the graph should have its own table.
(2) We need to be totally functional in our programming. (We cannot modify objects)
(3) Each table can only be returned by one blue box.
(4) No two blue boxes can create a single gray box.
(5) The graph must determine the code/data types exactly.

Screen Shot 2021-11-05 at 8 56 37 AM

johnwlambert:
> How does this solve #306?

306 notes that the "inlier ratio w.r.t. GT model" was far higher than "inlier ratio w.r.t. est model". We would actually expect the following two fractions:
- Fraction 1: (# verified correspondences w.r.t. est model ) / (# putatives )
- Fraction 2: (# verified correspondences w.r.t. GT model ) / (# putatives )
to be very close to each other.

It turns out the "inlier ratio w.r.t. GT model" that we were reporting in the Verifier's output was actually
- Fraction 3: (# verified correspondences that are verified by GT model) / (# verified correspondences).

The denominator was different, explaining the issue we saw in 306. This PR provides the computation of Fraction 2 and adds it to the report.

gtsfm/scene_optimizer.py
travisdriver: As we talked about offline, I think all of this can be achieved without creating a redundant data type for each blue box. If we continue on this trajectory, we will be passing around 7-8 different reports, which will be extremely cumbersome. 

I think we should continue brainstorming.

akshay-krishnan (commented):
I like having a different section for each blue box and each box returning an instance of its metrics which don't have anything to do with other boxes. 
But I don't think we should have blue boxes defining custom data types for this. 
I think treating them all as metrics makes the system simpler and don't see any advantages to having additional types. Let me know if there are any.

akshay-krishnan:
Could you pull from master and update the PR? Or would you like one of us to "take it from here"? 

Everything in the PR looks good, except for the custom matcher report. This adds more boilerplate. We created the metrics struct to avoid such custom reports per module and to share operations across metrics. We could either return (num_inliers, inlier_ratio) as tuples (simpler), or return a GtsfmMetricsGroup and merge groups from different matchers later.

PR #376, opened 2021-11-05
[WIP] Expanding gtsfm data

travisdriver:
**Disclaimer: currently only working on AstroNet data**

The docstring for the `GtsfmData` class states that its purpose is "describing the complete 3D scene." Therefore, I thought it made sense to extend the class' functionality slightly to support describing the _ground truth_ scene as well. That way we can just track a single `GtsfmData` object as opposed to tracking the various components of the ground truth scene: `gt_cameras_graph`, `gt_poses_graph`, `gt_scene_mesh_graph`, etc.

This is definitely not in its final form, but I wanted to get some feedback before I sunk too much time into refactoring things.

### DIFF
--- a/gtsfm/common/gtsfm_data.py
+++ b/gtsfm/common/gtsfm_data.py
@@ -31,8 +31,8 @@ class GtsfmData:
     def __init__(
         self,
         number_images: int,
-        cameras: Dict[int, PinholeCameraCal3Bundler] = {},
-        tracks: List[SfmTrack] = [],
+        cameras: Optional[Dict[int, PinholeCameraCal3Bundler]] = None,
+        tracks: Optional[List[SfmTrack]] = None,
         scene_mesh: Optional[Trimesh] = None,
     ) -> None:
         """Initializes the class.
@@ -41,9 +41,17 @@ class GtsfmData:
             number_images: number of images/cameras in the scene.
         """
         self._number_images = number_images
-        self._cameras = cameras
-        self._tracks = tracks
-        self._scene_mesh = scene_mesh
+        self._cameras = cameras if cameras else {}
+        self._tracks = tracks if tracks else []
+        # self._cameras = cameras  # this caused an error for some reason??
+        # self._tracks = tracks
+        self.scene_mesh = scene_mesh
+
+    def get_two_view_data(self, i1: int, i2: int) -> "GtsfmData":
+        """Collects GtsfmData for a sinbgle image pair."""
+        # TODO (travisdriver): also collect tracks if available.
+        cameras_pair = {0: self.get_camera(i1), 1: self.get_camera(i2)}
+        return GtsfmData(2, cameras_pair, None, self.scene_mesh)
 
     def __eq__(self, other: object) -> bool:
         """Checks equality with the other object."""
--- a/gtsfm/common/gtsfm_data.py
+++ b/gtsfm/common/gtsfm_data.py
@@ -8,6 +8,7 @@ from typing import Any, Dict, List, Optional, Tuple
 
 import numpy as np
 from gtsam import PinholeCameraCal3Bundler, Pose3, SfmTrack, Similarity3
+from trimesh import Trimesh
 
 import gtsfm.utils.geometry_comparisons as geometry_comparisons
 import gtsfm.utils.graph as graph_utils
@@ -27,15 +28,22 @@ class GtsfmData:
     The situation of non-contiguous cameras can exists because of failures in front-end.
     """
 
-    def __init__(self, number_images: int) -> None:
+    def __init__(
+        self,
+        number_images: int,
+        cameras: Dict[int, PinholeCameraCal3Bundler] = {},
+        tracks: List[SfmTrack] = [],
+        scene_mesh: Optional[Trimesh] = None,
+    ) -> None:
         """Initializes the class.
 
         Args:
             number_images: number of images/cameras in the scene.
         """
-        self._cameras: Dict[int, PinholeCameraCal3Bundler] = {}
-        self._tracks: List[SfmTrack] = []
         self._number_images = number_images
+        self._cameras = cameras
+        self._tracks = tracks
+        self._scene_mesh = scene_mesh
 
     def __eq__(self, other: object) -> bool:
         """Checks equality with the other object."""
--- a/gtsfm/loader/astronet_loader.py
+++ b/gtsfm/loader/astronet_loader.py
@@ -8,12 +8,13 @@ from pathlib import Path
 from typing import Optional, Dict, Tuple, List
 
 import trimesh
-from gtsam import Cal3Bundler, Pose3, Rot3, Point3, SfmTrack
+from gtsam import PinholeCameraCal3Bundler, Cal3Bundler, Pose3, Rot3, Point3, SfmTrack
 
 import gtsfm.utils.io as io_utils
 import gtsfm.utils.logger as logger_utils
 from gtsfm.common.image import Image
 from gtsfm.loader.loader_base import LoaderBase
+from gtsfm.common.gtsfm_data import GtsfmData
 
 import thirdparty.colmap.scripts.python.read_write_model as colmap_io
 from thirdparty.colmap.scripts.python.read_write_model import Camera as ColmapCamera
@@ -78,7 +79,7 @@ class AstronetLoader(LoaderBase):
         if not Path(data_dir).exists():
             raise FileNotFoundError("No data found at %s." % data_dir)
         cameras, images, points3d = colmap_io.read_model(path=data_dir, ext=".bin")
-        self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
+        self._calibrations, self._wTi_list, img_fnames, self._sfmtracks, cameras_gtsfm = self.colmap2gtsfm(
             cameras, images, points3d, load_sfmtracks=use_gt_sfmtracks
         )
 
@@ -94,6 +95,9 @@ class AstronetLoader(LoaderBase):
             )
         else:
             self.gt_scene_trimesh = None
+
+        # Create GtsfmData object to hold GT data.
+        self.gt_gtsfm_data = GtsfmData(len(img_fnames), cameras_gtsfm, self._sfmtracks, self.gt_scene_trimesh)
 
         # Camera intrinsics are currently required due to absence of EXIF data and diffculty in approximating focal
         # length (usually 10000 to 100000 pixels).
@@ -124,7 +128,9 @@ class AstronetLoader(LoaderBase):
         images: Dict[int, ColmapImage],
         points3D: Dict[int, ColmapPoint3D],
         load_sfmtracks: bool = False,
-    ) -> Tuple[List[Cal3Bundler], List[Pose3], List[str], Optional[List[Point3]]]:
+    ) -> Tuple[
+        List[Cal3Bundler], List[Pose3], List[str], Optional[List[SfmTrack]], Dict[int, PinholeCameraCal3Bundler]
+    ]:
         """Converts COLMAP-formatted variables to GTSfM format.
 
         Args:
@@ -142,13 +148,15 @@ class AstronetLoader(LoaderBase):
         # Note: Assumes input cameras use `PINHOLE` model
         if len(images) == 0 and len(cameras) == 0:
             raise RuntimeError("No Image or Camera data provided to loader.")
-        cameras_gtsfm, images_gtsfm, img_fnames = [], [], []
+        calibrations_gtsfm, images_gtsfm, img_fnames = [], [], []
+        cameras_gtsfm = {}
         image_id_to_idx = {}  # keeps track of discrepencies between `image_id` and List index.
         for idx, img in enumerate(images.values()):
             images_gtsfm.append(Pose3(Rot3(img.qvec2rotmat()), img.tvec).inverse())
             img_fnames.append(img.name)
             fx, _, cx, cy = cameras[img.camera_id].params[:4]
-            cameras_gtsfm.append(Cal3Bundler(fx, 0.0, 0.0, cx, cy))
+            calibrations_gtsfm.append(Cal3Bundler(fx, 0.0, 0.0, cx, cy))
+            cameras_gtsfm[idx] = PinholeCameraCal3Bundler(images_gtsfm[-1], calibrations_gtsfm[-1])
             image_id_to_idx[img.id] = idx
 
         if len(points3D) == 0 and load_sfmtracks:
@@ -162,7 +170,7 @@ class AstronetLoader(LoaderBase):
                     sfmtrack.add_measurement(image_id_to_idx[image_id], images[image_id].xys[point2d_idx])
                 sfmtracks_gtsfm.append(sfmtrack)
 
-        return cameras_gtsfm, images_gtsfm, img_fnames, sfmtracks_gtsfm
+        return calibrations_gtsfm, images_gtsfm, img_fnames, sfmtracks_gtsfm, cameras_gtsfm
 
     def __len__(self) -> int:
         """The number of images in the dataset.
--- a/gtsfm/runner/run_scene_optimizer_astronet.py
+++ b/gtsfm/runner/run_scene_optimizer_astronet.py
@@ -58,7 +58,7 @@ class GtsfmRunnerAstronetLoader(GtsfmRunnerBase):
 
         with Client(cluster) as client, performance_report(filename="dask-report.html"):
             # Scatter surface mesh across all nodes to preserve computation time and memory.
-            gt_scene_trimesh_future = client.scatter(self.loader.gt_scene_trimesh, broadcast=True)
+            gt_gtsfm_data_future = client.scatter(self.loader.gt_gtsfm_data, broadcast=True)
 
             # Prepare computation graph.
             start_time = time.time()
@@ -68,8 +68,7 @@ class GtsfmRunnerAstronetLoader(GtsfmRunnerBase):
                 image_graph=self.loader.create_computation_graph_for_images(),
                 camera_intrinsics_graph=self.loader.create_computation_graph_for_intrinsics(),
                 image_shape_graph=self.loader.create_computation_graph_for_image_shapes(),
-                gt_cameras_graph=self.loader.create_computation_graph_for_cameras(),
-                gt_scene_mesh=gt_scene_trimesh_future,
+                gt_gtsfm_data=self.loader.gt_gtsfm_data,
             )
 
             # Run SfM pipeline.
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -103,10 +103,16 @@ class SceneOptimizer:
         image_graph: List[Delayed],
         camera_intrinsics_graph: List[Delayed],
         image_shape_graph: List[Delayed],
-        gt_cameras_graph: Optional[List[Delayed]] = None,
-        gt_scene_mesh: Optional[Trimesh] = None,
+        gt_gtsfm_data: Optional[GtsfmData] = None,
     ) -> Delayed:
         """The SceneOptimizer plate calls the FeatureExtractor and TwoViewEstimator plates several times."""
+        # Build cameras graph from GT GtsfmData.
+        # TODO (): remove later; this is just to conform to the required input of other functions.
+        gt_cameras_graph = (
+            [dask.delayed(gt_gtsfm_data.get_camera)(i) for i in gt_gtsfm_data._cameras.keys()]
+            if gt_gtsfm_data
+            else None
+        )
 
         # auxiliary graph elements for visualizations and saving intermediate
         # data for analysis, not returned to the user.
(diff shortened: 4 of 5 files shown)


akshay-krishnan:
Have not looked at code, but just from your description, the ground truth scene and the computed scene are actually not the same reconstruction, so would it be better if we used a different GtsfmData instance for ground truth? 

Gtsfm is currently in development and GT and metrics are everywhere. But once we get to a stable state, GtsfmData is our result which I think should be free from any GT members.

akshay-krishnan:
We could maybe put all GT members in a different GtsfmData or maybe have another container GTSfMGTData, curious what @ayushbaid  thinks

ayushbaid:
> We could maybe put all GT members in a different GtsfmData or maybe have another container GTSfMGtData, curious what @ayushbaid  thinks

Agree with Akshay. It doesn't make sense to have additional information on this datastructure. We can have two if needed. The API design and the dataclasses have to be clean, and just have the arguments/attributes which are obvious.

GtsfmData is our output, which describes a 3d scene completely. It should not have attributes which are not needed to visualize a 3d scene.

ayushbaid:
For any evaluation we need for the astronet dataset or any other dataset, we can load the gtsfmdata from disk, construct the loader, and compare them.

johnwlambert:
I like the idea of having 1 class that holds a complete result.
- Estimated result can be an instance of this.
- GT result can be a separate instance of this.

If MVS produces a mesh, that I think we could safely assume the mesh could also be an instance of a "complete result".

travisdriver:
Thank you guys for the feedback.

@akshay-krishnan Sorry, I think my description was a little misleading. The main change I made was having the loader create an instance of  `GtsfmData` that contained all of the ground truth data, and replaced the `gt_cameras_graph`, `gt_poses_graph`, and `gt_scene_mesh_graph` with this instance in the `SceneOptimizer` (which still generates a different `GtsfmData` instance with our estimated reconstruction). 

The only thing I added to the `GtsfmData` class was a `scene_mesh` attribute. While its only currently used in the GT for AstroNet, eventually GTSfM's MVS pipeline will populate it with our estimated mesh.

ayushbaid:
> Thank you guys for the feedback.
> 
> @akshay-krishnan Sorry, I think my description was a little misleading. The main change I made was having the loader create an instance of `GtsfmData` that contained all of the ground truth data, and replaced the `gt_cameras_graph`, `gt_poses_graph`, and `gt_scene_mesh_graph` with this instance in the `SceneOptimizer` (which still generates a different `GtsfmData` instance with our estimated reconstruction).
> 
> The only thing I added to the `GtsfmData` class was a `scene_mesh` attribute. While its only currently used in the GT for AstroNet, eventually GTSfM's MVS pipeline will populate it with our estimated mesh.

Sounds good. Does @codyly also plan to use the same trimesh format for dense reconstruction?

gtsfm/common/gtsfm_data.py
```diff
@@ -27,15 +28,30 @@ class GtsfmData:
...
         """Initializes the class.
 
         Args:
             number_images: number of images/cameras in the scene.
         """
-        self._cameras: Dict[int, PinholeCameraCal3Bundler] = {}
-        self._tracks: List[SfmTrack] = []
         self._number_images = number_images
+        self._cameras = cameras if cameras else {}
+        self._tracks = tracks if tracks else []
```
akshay-krishnan, line 45: why not just use those are defaults instead of None?

gtsfm/loader/astronet_loader.py
```diff
@@ -78,7 +79,7 @@ def __init__(
         if not Path(data_dir).exists():
             raise FileNotFoundError("No data found at %s." % data_dir)
         cameras, images, points3d = colmap_io.read_model(path=data_dir, ext=".bin")
-        self._calibrations, self._wTi_list, img_fnames, self._sfmtracks = self.colmap2gtsfm(
+        self._calibrations, self._wTi_list, img_fnames, self._sfmtracks, cameras_gtsfm = self.colmap2gtsfm(
```
akshay-krishnan, line 82: seeing wTi and cameras together is always a red flag ..

gtsfm/loader/astronet_loader.py
```diff
@@ -118,13 +122,19 @@ def __init__(
         self._num_imgs = len(self._image_paths)
         logger.info("AstroNet loader found and loaded %d images and %d tracks.", self._num_imgs, self.num_sfmtracks)
 
+    @property
+    def gt_gtsfm_data(self):
```
akshay-krishnan, line 126: nit on name: just gt_data or get_gt_data might suffice

gtsfm/loader/astronet_loader.py
```diff
@@ -118,13 +122,19 @@ def __init__(
...
+        return self._gt_gtsfm_data
+
     @staticmethod
     def colmap2gtsfm(
```
akshay-krishnan, line 130: why is this a static method in this class? like what does it have to do with astronet specifically?

gtsfm/loader/colmap_loader.py
```diff
@@ -70,10 +71,10 @@ def __init__(
         self._use_gt_extrinsics = use_gt_extrinsics
         self._max_frame_lookahead = max_frame_lookahead
 
+        # TODO in future PR: if img_fnames is None, default to using everything inside image directory
```
akshay-krishnan, line 74: Use # TODO(username) or # TODO(#issue number)

gtsfm/scene_optimizer.py
```diff
@@ -103,10 +102,16 @@ def create_computation_graph(
...
         image_shape_graph: List[Delayed],
-        gt_cameras_graph: Optional[List[Delayed]] = None,
-        gt_scene_mesh: Optional[Trimesh] = None,
+        gt_gtsfm_data: Optional[GtsfmData] = None,
     ) -> Delayed:
         """The SceneOptimizer plate calls the FeatureExtractor and TwoViewEstimator plates several times."""
+        # Build cameras graph from GT GtsfmData.
+        # TODO (): remove later; this is just to conform to the required input of other functions.
+        gt_cameras_graph = (
+            [dask.delayed(gt_gtsfm_data.get_camera)(i) for i in gt_gtsfm_data._cameras.keys()]
```
akshay-krishnan, line 111: Maybe I should ask @ayushbaid , but what is our "standard representation" for cameras? Is it dicts or lists? why are converting one to another here?

akshay-krishnan (changes requested):
Some comments, please let us know when the PR is ready to review/submit

gtsfm/common/gtsfm_data.py
johnwlambert: Default arguments that are mutable have led me to some very nasty bugs. https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/

gtsfm/loader/astronet_loader.py
travisdriver: Can you elaborate?

gtsfm/loader/astronet_loader.py
travisdriver: Usually with properties, you name it after the attribute you're returning.

gtsfm/loader/astronet_loader.py
travisdriver: AstroNet data is in COLMAP format. I could move this to the base class since it would also be useful for the Colmap data?

gtsfm/loader/colmap_loader.py
travisdriver: I didn't write this originally, I just moved it.

gtsfm/loader/astronet_loader.py
akshay-krishnan: cameras contain poses, camera.pose(), so wTi is not necessary

gtsfm/loader/astronet_loader.py
akshay-krishnan: I dont feel strongly, so you can keep it this way, though I tend to name functions based on what they are returning semantically, not the datatype.

gtsfm/loader/astronet_loader.py
akshay-krishnan: I dont feel like it is related to the loader, maybe it can be moved to a util file. Why would a user of a loader want to convert colmap data to gtsfm format? I would assume it just provides the gtsfm format, using whatever underneath.

gtsfm/loader/colmap_loader.py
akshay-krishnan: Oh thats worse, so we dont know who wrote this already

gtsfm/common/gtsfm_data.py
akshay-krishnan: Interesting, TIL. Thanks for sharing!

gtsfm/loader/astronet_loader.py
travisdriver: Because the AstroNet data is in Colmap format and it needs to be converted to GTSfM format before it can be used by the pipeline.

gtsfm/loader/colmap_loader.py
travisdriver: Yeah I think it's pretty old

gtsfm/scene_optimizer.py
travisdriver: GtsfmData uses dicts to account for noncontiguous cameras ids, as some frames may be rejected during the verification process.

I convert the to lists here because the rest of the code expects it as a list, but I do agree that it's dangerous to have the same data as different data types.

gtsfm/loader/astronet_loader.py
akshay-krishnan: I mean the loader is supposed to abstract that away. I think this PR https://github.com/borglab/gtsfm/pull/346 is moving it to a util? Which is better I think

akshay-krishnan: The loader can call a util, convert it to GTSfM format and provide it, the user of the loader does not have to know what format Astronet data is natively in.

gtsfm/scene_optimizer.py
akshay-krishnan: Yeah, so maybe we should update the rest of the code

gtsfm/loader/astronet_loader.py
travisdriver: Same could be said for the calibrations.

gtsfm/loader/astronet_loader.py
travisdriver: **The loader is responsible for reading in the specific data and converting it to a format usable by GTSfM.** 

I don't think it makes sense to have to add a util in some other file for every dataset used, as this will just congest the other modules. However, I think it's fine to move this specific function to a util, as we work with the Colmap format a lot.

When other users want to apply GTSfM to their specific data, all they should need to do is fill in a new loader (and maybe a runner) without touching any other code. It's infeasible to expect our main code base to handle converting _every single_ dataset to the expected GTSfM format.

gtsfm/scene_optimizer.py
travisdriver: I spent a while trying to update the code so that we do not rely on converting the dictionary of cameras in `GtsfmData` to a List. However, almost all of the metric computation functions rely on Lists of Pose derived from the camera dictionary of GtsfmData.

I'm assuming this is why the `_number_images` attribute of the `GtsfmData` class was added in the first place: so that the cameras data could be converted to a list such that its the same length as the original ground truth data. I think this is should be saved for another PR.

gtsfm/common/gtsfm_data.py
```diff
@@ -27,15 +28,30 @@ class GtsfmData:
...
+        # self._cameras = cameras  # this caused an error for some reason??
```
akshay-krishnan, line 46: remove these commented code?

gtsfm/common/gtsfm_data.py
```diff
@@ -27,15 +28,30 @@ class GtsfmData:
...
+        self,
+        number_images: int,
+        cameras: Optional[Dict[int, PinholeCameraCal3Bundler]] = None,
+        tracks: Optional[List[SfmTrack]] = None,
+        scene_mesh: Optional[Trimesh] = None,
+    ) -> None:
         """Initializes the class.
 
         Args:
             number_images: number of images/cameras in the scene.
```
akshay-krishnan, line 41: args section needs update

gtsfm/loader/astronet_loader.py
akshay-krishnan: true, so why are we returning everything here?

gtsfm/loader/astronet_loader.py
akshay-krishnan: > However, I think it's fine to move this specific function to a util, as we work with the Colmap format a lot.

 yes, that is my reason as well. 
I don't see why anyone would use the Astronet loader to convert colmap data to gtsfm. It already provides data in gtsfm format. And even if we did want to convert colmap to gtsfm format for some other reason, it does not need a AstroNet loader.

gtsfm/runner/run_scene_optimizer_astronet.py
```diff
@@ -49,33 +45,32 @@ def construct_loader(self) -> LoaderBase:
...
-                image_shape_graph=self.loader.create_computation_graph_for_image_shapes(),
-                gt_cameras_graph=self.loader.create_computation_graph_for_cameras(),
-                gt_scene_mesh=gt_scene_trimesh_future,
-            )
-
-            # Run SfM pipeline.
-            sfm_result = sfm_result_graph.compute()
-        assert isinstance(sfm_result, GtsfmData)
-        logger.info("GTSFM took %.2f minutes to compute sparse multi-view result.", (time.time() - start_time) / 60)
+    # def run(self) -> None:
```
akshay-krishnan, line 48: This needs to be uncommented or removed before the final review.

gtsfm/scene_optimizer.py
akshay-krishnan: @ayushbaid

gtsfm/loader/astronet_loader.py
akshay-krishnan: https://github.com/borglab/gtsfm/pull/346 is trying to move it to io_utils.py, which I think is better than a static method in AstroNet loader (my point that its being used in a non-loader specific context).

travisdriver: Maybe I should just leave it as is here to avoid conflicts with #346?

gtsfm/loader/astronet_loader.py
travisdriver: Sorry, I forgot to push a commit that removed `cameras_gtsfm`; will do that.

What I meant was that even in the SceneOptimizer we are passing in calibrations and cameras. However, this is hard to avoid since it is not possible to set either the Pose or Calibration to None in `PinholeCameraCal3Bundler`. 

So, the two options are (1) pass around Calibration(s) and Pose(s) separately and construct cameras as needed, or (2) pass around Calibration(s) and Camera(s) which have redundant information.

gtsfm/scene_optimizer.py
travisdriver: Similarly, I think `_tracks` should be a dictionary as well.

gtsfm/common/gtsfm_data.py
```diff
@@ -27,15 +28,30 @@ class GtsfmData:
...
+        self._tracks = tracks if tracks else []
+        # self._cameras = cameras  # this caused an error for some reason??
+        # self._tracks = tracks
+        self.scene_mesh = scene_mesh
+
+    def get_two_view_data(self, i1: int, i2: int) -> "GtsfmData":
+        """Collects GtsfmData for a sinbgle image pair."""
+        # TODO (travisdriver): also collect tracks if available.
+        cameras_pair = {0: self.get_camera(i1), 1: self.get_camera(i2)}
+        return GtsfmData(2, cameras_pair, None, self.scene_mesh)
```
ayushbaid, line 54: Whats the reason for introducing this API?

Also nit: typo in the docstring.

travisdriver: So that we don't need to pass the entire GtsfmData object to the TwoViewEstimator, as it could contain thousands of cameras. Also, it conforms with the previous API.

gtsfm/runner/run_scene_optimizer_astronet.py
travisdriver: removed

gtsfm/common/gtsfm_data.py
travisdriver: fixed typo

gtsfm/common/gtsfm_data.py
travisdriver: updated

gtsfm/common/gtsfm_data.py
travisdriver: removd

gtsfm/loader/astronet_loader.py
travisdriver: removed `cameras_gtsfm`

gtsfm/loader/astronet_loader.py
akshay-krishnan: SInce youre changing the API and Jon isn't I think we should merge this is first. 

@womackj1 Can #346 wait until this is merged? I don't see any updates on it recently.

akshay-krishnan (approved):

gtsfm/two_view_estimator.py
```diff
@@ -94,6 +92,13 @@ def create_computation_graph(
             Two view report w/ verifier metrics wrapped as Delayed.
             Two view report w/ post-processor metrics wrapped as Delayed.
         """
+        # Unpack GT data.
+        if gt_gtsfm_data is not None:
+            gt_wTi1_graph = gt_gtsfm_data.get_camera(0).pose()
```
johnwlambert, line 97: instead of hard-coding camera 0 and 1, we should pull out `i1` and `i2` here

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -86,27 +88,49 @@ def construct_scene_optimizer(self) -> SceneOptimizer:
 
         return scene_optimizer
 
+    def scatter_scene_mesh(self, client: Client) -> Optional[GtsfmData]:
+        """Scatters ground truth scene mesh into distributed memory to preserve computation time and memory.
+
+        Ref: http://distributed.dask.org/en/stable/api.html#distributed.Client.scatter
+        """
+        # TODO(travisdriver): find a way to scatter mesh that doesn't require copying the GtsfmData object.
```
johnwlambert, line 96: where do we do the copy?

gtsfm/common/gtsfm_data.py
```diff
@@ -27,15 +29,32 @@ class GtsfmData:
...
-        self._tracks: List[SfmTrack] = []
         self._number_images = number_images
+        self._cameras = cameras if cameras else {}
+        self._tracks = tracks if tracks else []
+        self.scene_mesh = scene_mesh
+
+    def get_two_view_data(self, i1: int, i2: int) -> "GtsfmData":
+        """Collects GtsfmData for a single image pair."""
+        # TODO (travisdriver): also collect tracks if available.
+        cameras_pair = {0: self.get_camera(i1), 1: self.get_camera(i2)}
```
johnwlambert, line 56: I think we should always use `i1` and `i2` directly, instead of remapping to 0 and 1, since this can create some confusion/bugs downstream for the user.

@ayushbaid @akshay-krishnan  what are your thoughts?

johnwlambert:
Thanks for the PR, Travis. Agreed that this needed some consolidation/ clean up.

johnwlambert:
I'm seeing a few flake8 failures:
```
Run flake8 --max-line-length 120 --ignore E201,E202,E203,E231,W291,W293,E303,W391,E402,W503,E731 gtsfm
gtsfm/runner/run_scene_optimizer_astronet.py:6:1: F401 'time' imported but unused
gtsfm/runner/run_scene_optimizer_astronet.py:8:1: F401 'dask.distributed.Client' imported but unused
gtsfm/runner/run_scene_optimizer_astronet.py:8:1: F401 'dask.distributed.LocalCluster' imported but unused
gtsfm/runner/run_scene_optimizer_astronet.py:8:1: F401 'dask.distributed.performance_report' imported but unused
gtsfm/runner/run_scene_optimizer_astronet.py:11:1: F401 'gtsfm.common.gtsfm_data.GtsfmData' imported but unused
gtsfm/common/gtsfm_data.py:12:1: F401 'dask.distributed.Client' imported but unused
```

gtsfm/two_view_estimator.py
travisdriver: If so, I would have to pass in `i1` and `i2`.

johnwlambert: I'm good with passing in i1 and i2. I think whenever we use a camera index in GTSFM, we need to be able to assume it really means the index we expect

travisdriver: I don't see the utility in passing in two indices for a single object as opposed to just using a known mapping (i.e., i1 -> 0 and i2 -> 1), especially since none of the other objects use the indices.

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Any reason we can’t use copy.deepcopy?

gtsfm/two_view_estimator.py
akshay-krishnan: I partially agree with John. I think it might be best to accept gt_i1Ti2 and the gt_mesh as two arguments. 
Passing `i1` and `i2` I think makes the API too messy. But yes, as you said (also in another comment), assuming the input has only 2 cameras at 0 and 1 could lead to bugs.

PR #380, opened 2021-11-16
Add Image masking

travisdriver:
Added an optional mask to the `Image` class, and also added functionality for filtering extracted keypoints with respect to this mask.

An example where this may be useful are when images are consistently obstructed (e.g., [a spacecraft rendezvous with the ISS](https://twitter.com/wonderofscience/status/1442123944236965891?s=12), [landing on the Moon](https://twitter.com/theplanetaryguy/status/1440772536438980612?s=12)).

Moreover, from a previous project, it has also been shown that masking out dark regions (pixel value of ~0) of images captured during small body relative navigation scenarios (e.g., the interface between the foreground body and the background deep space, self-shadowing) substantially increases the reliability of the visual measurements:

### No masking
Screenshot from 2021-11-16 00-25-16

### Masking
Screenshot from 2021-11-16 00-23-12


gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
```diff
@@ -43,9 +43,26 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
...
+        # Sort by responses.
+        sort_idxs = np.argsort(-keypoints.responses)
+        sort_idxs = sort_idxs[: self.max_keypoints]
+
+        return keypoints.extract_indices(sort_idxs), descriptors[sort_idxs]
+
+    @staticmethod
+    def filter_by_mask(mask: np.ndarray, keypoints: Keypoints, descriptors: np.ndarray) -> Tuple[Keypoints, np.ndarray]:
+        """Filter features with respect to a binary mask of the image."""
+        valid_idxs = np.flatnonzero([mask[round(v), round(u)] == 1 for (u, v) in keypoints.coordinates])
```
johnwlambert, line 60: thanks for the PR. Any reason not to just use `np.round(keypoints.coordinates)`?

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
```diff
@@ -43,9 +43,26 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
             Corr. descriptors, of shape (N, D) where D is the dimension of each descriptor.
         """
 
+    def filter_by_response(self, keypoints: Keypoints, descriptors: np.ndarray) -> Tuple[Keypoints, np.ndarray]:
```
johnwlambert, line 46: why not just use `Keypoints.get_top_k()`?

https://github.com/borglab/gtsfm/blob/master/gtsfm/common/keypoints.py#L89

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
johnwlambert: seems could be vectorized with no for loop in that case

gtsfm/frontend/detector_descriptor/sift.py
```diff
@@ -39,28 +39,21 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
...
-        # convert to GTSFM's keypoints
+        # Convert to GTSFM's keypoints.
         keypoints = feature_utils.cast_to_gtsfm_keypoints(cv_keypoints)
 
-        # sort the features and descriptors by the score
-        # (need to sort here as we need the sorting order for descriptors)
-        sort_idx = np.argsort(-keypoints.responses)[: self.max_keypoints]
-
-        keypoints = Keypoints(
-            coordinates=keypoints.coordinates[sort_idx],
```
johnwlambert, line 59: @ayushbaid do you remember why we weren't just using `Keypoints.get_top_k()` here?

https://github.com/borglab/gtsfm/blob/master/gtsfm/common/keypoints.py#L89

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -16,11 +16,13 @@
 import torch
 
 import gtsfm.utils.images as image_utils
+import gtsfm.utils.logger as logger_utils
 from gtsfm.common.image import Image
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.frontend.detector_descriptor.detector_descriptor_base import DetectorDescriptorBase
 from thirdparty.SuperGluePretrainedNetwork.models.superpoint import SuperPoint
 
+logger = logger_utils.get_logger()
```
johnwlambert, line 25: what's the logger for here?

gtsfm/loader/astronet_loader.py
```diff
@@ -8,10 +8,13 @@
 from typing import Optional, Dict, Tuple, List
 
 import trimesh
+import cv2 as cv
```
johnwlambert, line 11: I think the ordering here should be 
```python
import cv2 as cv
import numpy as np
import trimesh
```

gtsfm/loader/astronet_loader.py
```diff
@@ -187,8 +190,18 @@ def get_image_full_res(self, index: int) -> Image:
...
             raise IndexError(f"Image index {index} is invalid")
 
+        # Read in image.
         img = io_utils.load_image(self._image_paths[index])
-        return img
+
+        # Generate mask to separate background deep space from foreground target body
+        # based on image intensity values.
+        gray_image = image_utils.rgb_to_gray_cv(img)
+        _, binary_image = cv.threshold(gray_image.value_array, 5, 255, cv.THRESH_BINARY)
```
johnwlambert, line 199: let's make this a standalone function. Perhaps Image.py would be a good place for it. But here is ok too, if you think it's too domain specific to be useful elsewhere.

```python
def get_nonzero_intensity_mask(img: Image) -> np.ndarray:
      """Generate mask of where image intensity values are non-zero.
      
      Args:
            img: ...
      """
       gray_image = image_utils.rgb_to_gray_cv(img)
       _, binary_image = cv.threshold(gray_image.value_array, 5, 255, cv.THRESH_BINARY)
       mask = cv.erode(binary_image, np.ones((15, 15), np.uint8))
       return mask
```

johnwlambert: keyword args on the 15x15 kernel and for the params 5 and 255 would also help make it even more readable

gtsfm/utils/images.py
```diff
@@ -115,7 +115,17 @@ def resize_image(image: Image, new_height: int, new_width: int) -> Image:
         interpolation=cv.INTER_CUBIC,
     )
 
-    return Image(value_array=resized_value_array, file_name=image.file_name)
+    # Resize the mask using nearest-neighbor inerpolation.
```
johnwlambert, line 118: nit: typo on interpolation

tests/frontend/detector_descriptor/test_detector_descriptor_base.py
```diff
@@ -74,6 +74,32 @@ def test_computation_graph(self):
...
+        """Test the `filter_by_mask` method."""
+        # Create a (9, 9) mask with ones in a (5, 5) square in the center of the mask and zeros everywhere else.
+        mask = np.zeros((9, 9)).astype(np.uint8)
+        mask[2:7, 2:7] = 1
+
+        # Test coordinates near corners of square of ones and along the diagonal.
+        coordinates = np.array(
+            [
+                [1.4, 1.4],
+                [1.4, 6.4],
```
johnwlambert, line 87: any reason to have the subpixel coordinates? i think makes it a bit less readable. but if there's a specific use case you're targeting, maybe that's ok.

I don't think any of our current detectors provide subpixel resolution.

tests/frontend/detector_descriptor/test_detector_descriptor_base.py
```diff
@@ -74,6 +74,32 @@ def test_computation_graph(self):
...
+                [5.0, 5.0],
+                [0.0, 0.0],
+                [8.0, 8.0],
+            ]
+        )
+
+        # Create keypoints from coordinates and dummy descriptors.
+        keypoints = Keypoints(coordinates, scales=None, responses=None)
+        descriptors = np.ones((coordinates.shape[0], 10))  # dummy descriptors
+        filtered_keypoints, filtered_descriptors = self.detector_descriptor.filter_by_mask(mask, keypoints, descriptors)
```
johnwlambert, line 99: would be nice to check against the values that remain after filtering, in addition to checking the cardinality

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
```diff
@@ -43,9 +43,26 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
...
+        """Filter features according to their responses."""
```
akshay-krishnan, line 47: also both these methods are missing args and returns, if you still need them.

akshay-krishnan (commented):
I think John is addressing most concerns I had, I am also not very familiar with the frontend APIs. I can take a final look once all other comments are addressed, if necessary.

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: Because we need the indices as well for the descriptors.

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: Good idea.

gtsfm/frontend/detector_descriptor/sift.py
travisdriver: Because we need the indices for the descriptors.

gtsfm/frontend/detector_descriptor/superpoint.py
travisdriver: Was debuggging; will remove

gtsfm/loader/astronet_loader.py
travisdriver: Yeah I'm not sure sure what other instances it could be used for; think it just makes sense to keep it here. 

Yes I should make those keyword args.

tests/frontend/detector_descriptor/test_detector_descriptor_base.py
travisdriver: Most feature keypoints are subpixel (including SIFT), as detections are usually computed by aggregating information at multiple scales.

gtsfm/loader/astronet_loader.py
johnwlambert: Sure, we can keep it here. Let's make it a separate function though, to improve readability.

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
ayushbaid: This should technically be a part of `describe` operation? For example the opencv sift implementation takes in the mask too. It will try to find more points in the non-masked region. Post-detection filtering with a mask will be suboptimal in that case.

gtsfm/frontend/detector_descriptor/sift.py
```diff
@@ -39,28 +39,21 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
...
-            scales=keypoints.scales[sort_idx],
-            responses=keypoints.responses[sort_idx],
-        )
-
-        descriptors = descriptors[sort_idx]
+        # Filter features.
+        if image.mask is not None:
+            keypoints, descriptors = self.filter_by_mask(image.mask, keypoints, descriptors)
```
ayushbaid, line 56: you can use opencv itself for the mask.

gtsfm/frontend/detector_descriptor/superpoint.py
```diff
@@ -49,27 +51,23 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
         model.eval()
         # TODO: fix inference issue #110
 
+        # Compute features.
```
ayushbaid, line 54: I think the comments you added here are too verbose.

gtsfm/frontend/detector_descriptor/sift.py
travisdriver: I'm not aware of an OpenCV function that masks keypoints. Can you point me to the function you're referring to?

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: The variable names are self-explanatory IMHO.

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: Yes, I believe you're right about SIFT. I will use their built-in function.

However, data-driven feature extractors typically return all features output by the network that are above some confidence (set by the user), as the hidden layers cannot be manipulated at test time. The output features are then post-processed (usually just taking the top k features wrt their scores). I'm unaware of any data-driven pipelines that take in a mask as input and change their output with respect to this mask.

johnwlambert: @ayushbaid could you add a pointer/reference to where in the SIFT code they are looking for additional keypoint during `describe()`?

I think it would be cleaner to put this masking in just a single place, and not spread it out across detect/describe.

gtsfm/frontend/detector_descriptor/superpoint.py
travisdriver: I think it's useful for users that are unfamiliar with SuperPoint. It's not obvious what's being done in that code block to the untrained eye.

tests/frontend/detector_descriptor/test_detector_descriptor_base.py
travisdriver: added

gtsfm/utils/images.py
travisdriver: fixed

gtsfm/loader/astronet_loader.py
travisdriver: moved to separate function and added keyword args

gtsfm/loader/astronet_loader.py
travisdriver: fixed

gtsfm/frontend/detector_descriptor/superpoint.py
travisdriver: removed

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: changed to using `np.round(keypoints.coordinates)` instead of for-loop.

akshay-krishnan:
Would prefer Ayush and John's reviews for this one. No additional comments from me.

gtsfm/configs/deep_front_end.yaml
```diff
@@ -47,7 +47,7 @@ SceneOptimizer:
       min_track_len: 2
       mode:
         _target_: gtsfm.data_association.data_assoc.TriangulationParam
-        value: 0 # 0 corresponds to NO_RANSAC
+        value: 1 # 0 corresponds to NO_RANSAC
```
ayushbaid, line 50: why change in this PR?

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
ayushbaid: The inputs, yes. but the return types could do with some clarification.

You are right, it makes perfect sense to us who use GtSfM regularly, but having the details (like shapes, lengths) in docstring helps new users I guess

gtsfm/frontend/detector_descriptor/sift.py
ayushbaid: Snippet: "sift.detect() function finds the keypoint in the images. You can pass a mask if you want to search only a part of image" from https://docs.opencv.org/3.4/da/df5/tutorial_py_sift_intro.html.

Hence I prefer the detection algorithm to handle the masks rather than us doing it. The function you wrote to filter out detections in the masked region can go to utils, and we can call it in the `detect_and_describe` function for algorithms like `Superpoint` which do not provide a direct API to apply a mask. Let's not make it a post-processing step for every algorithm.

ayushbaid (changes requested):

gtsfm/frontend/detector_descriptor/sift.py
travisdriver: Yes, I'm aware of the one for features implemented by OpenCV (e.g., SIFT); I thought you were referring to a general keypoint masking function.

You're right about the OpenCV masking for the _handcrafted_ features (e.g., ORB, SIFT) being different than just masking the output, as they mask the image at each level in their scale pyramid. Thanks for bringing this up.

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: @johnwlambert It seems as though OpenCV masks the _input_ before computing features as opposed to masking the output. I'm not sure if this is any different than computing _all_ features and then masking them based on their keypoint, but it probably makes more sense to just use OpenCV's since it's built into the `detectAndCompute` function.

gtsfm/configs/deep_front_end.yaml
johnwlambert: same comment -- let's leave as 0 here

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
johnwlambert: I think the shapes of input and return args would be useful to new users

gtsfm/loader/astronet_loader.py
```diff
@@ -253,3 +262,21 @@ def is_valid_pair(self, idx1: int, idx2: int) -> bool:
             validation result.
         """
         return super().is_valid_pair(idx1, idx2) and abs(idx1 - idx2) <= self._max_frame_lookahead
+
+
+def get_nonzero_intensity_mask(img: Image, eps: int = 5, kernel: Tuple[int, int] = (15, 15)) -> np.ndarray:
```
johnwlambert, line 267: I think the argument name should be `kernel_size`, not `kernel`

gtsfm/loader/astronet_loader.py
```diff
@@ -253,3 +262,21 @@ def is_valid_pair(self, idx1: int, idx2: int) -> bool:
...
+    """Generate mask of where image intensity values are non-zero.
+
+    Args:
+        img: input Image to be masked (values in range [0, 255]).
+        eps: minimum allowable intesnity value, i.e., values below this value withh be masked out.
```
johnwlambert, line 272: nit: typo on "intensity" and "withh"

gtsfm/loader/astronet_loader.py
```diff
@@ -253,3 +262,21 @@ def is_valid_pair(self, idx1: int, idx2: int) -> bool:
             validation result.
         """
         return super().is_valid_pair(idx1, idx2) and abs(idx1 - idx2) <= self._max_frame_lookahead
+
+
+def get_nonzero_intensity_mask(img: Image, eps: int = 5, kernel: Tuple[int, int] = (15, 15)) -> np.ndarray:
+    """Generate mask of where image intensity values are non-zero.
```
johnwlambert, line 268: Let's mention that we are using a morphological operator on top of the thresholding in the docstring, e.g.

```python
"""
...
After thresholding, we use an erosion kernel to .... because ...
```

Without the erosion kernel, is the result worse? or maybe we can add a brief justification here in the docstring

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: I think the return statement is also very easy to read with respect to understanding the outputs: `return keypoints.extract_indices(valid_idxs), descriptors[valid_idxs]`

I would pretty much just be copying and pasting the explanation from the docstring of `detect_and_describe` which is two lines above this function. I can do that if you feel strongly about a novice user not being able to understand the two lines of code.

gtsfm/utils/images.py
```diff
@@ -88,7 +88,7 @@ def vstack_image_list(imgs: List[np.ndarray]) -> Image:
     vstack_img = np.zeros((sum(all_heights), img_w, 3), dtype=np.uint8)
 
     running_h = 0
-    for i, img in enumerate(imgs):
+    for img in imgs:
```
johnwlambert, line 91: why the change here to handle non-standard image height? if we're concerned about different image aspect ratios, e.g. Notre Dame, shouldn't we check both height and width? 

In MVS, we have this hard-coded to assume constant image size, so this is a bigger issue.

gtsfm/configs/deep_front_end.yaml
travisdriver: Yes, was testing something; will change back

gtsfm/utils/images.py
travisdriver: Hmm, the only thing I changed was removing the iterator `i` which wasn't being used. I also removed `img_h`, which also wasn't being used.

gtsfm/loader/astronet_loader.py
travisdriver: I erode the mask to add a small buffer, because theoretically there could still be border points if I didn't.

gtsfm/configs/deep_front_end.yaml
travisdriver: fixed

gtsfm/loader/astronet_loader.py
travisdriver: Added explanation

gtsfm/frontend/detector_descriptor/sift.py
travisdriver: Now passing mask to `detectAndCompute`

gtsfm/loader/astronet_loader.py
travisdriver: changed

gtsfm/loader/astronet_loader.py
travisdriver: fixed

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: Now passing mask to `detectAndCompute`

travisdriver: Adding link to OpenCV source where masking is executed for future reference: https://github.com/opencv/opencv/blob/4.x/modules/features2d/src/orb.cpp

johnwlambert:
@travisdriver thanks for the updates. Can you add the masked and unmasked jobs to the CI like we discussed in our last meeting? That way we can track improvements even without the mask, and we can ensure we retain good performance on the masked job without degradation

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
johnwlambert: Let's go ahead and add the full docstring. I think almost always a good idea.

gtsfm/utils/images.py
johnwlambert: Oh I see, nevermind. Thanks

travisdriver:
> @travisdriver thanks for the updates. Can you add the masked and unmasked jobs to the CI like we discussed in our last meeting? That way we can track improvements even without the mask, and we can ensure we retain good performance on the masked job without degradation

I'd like to save that for a different PR if possible, as I want to change a couple things about the AstroNet benchmarks.

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
```diff
@@ -43,9 +46,46 @@ def detect_and_describe(self, image: Image) -> Tuple[Keypoints, np.ndarray]:
...
+
+        Returns:
+            N <= M keypoints, and their corresponding desciptors as an (N, D) array, such that their (rounded)
+                coordinates corresponded to a 1 in the input mask array.
         """
-        Generates the computation graph for detections and their descriptors.
+        rounded_coordinates = np.round(keypoints.coordinates).astype(int)
+        valid_idxs = np.flatnonzero(mask[rounded_coordinates[:, 1], rounded_coordinates[:, 0]] == 1)
+
+        return keypoints.extract_indices(valid_idxs), descriptors[valid_idxs]
```
ayushbaid, line 85: Lets modify the get_top_k method to return the keypoints as well as indices and use it here

ayushbaid: Also, lets move the mask application fucntion to keypoints class

johnwlambert:
> > @travisdriver thanks for the updates. Can you add the masked and unmasked jobs to the CI like we discussed in our last meeting? That way we can track improvements even without the mask, and we can ensure we retain good performance on the masked job without degradation
> 
> I'd like to save that for a different PR if possible, as I want to change a couple things about the AstroNet benchmarks.

@travisdriver Sure thing, can you open an issue so we don't lose track of it later?

Could you also upload the visual dashboard file here in the comments, so we can see the diff?

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
travisdriver: Added

johnwlambert:
> > > @travisdriver thanks for the updates. Can you add the masked and unmasked jobs to the CI like we discussed in our last meeting? That way we can track improvements even without the mask, and we can ensure we retain good performance on the masked job without degradation
> > 
> > 
> > I'd like to save that for a different PR if possible, as I want to change a couple things about the AstroNet benchmarks.
> 
> @travisdriver Sure thing, can you open an issue so we don't lose track of it later?
> 
> Could you also upload the visual dashboard file here in the comments, so we can see the diff?

Hi @travisdriver, I think this PR is almost ready to merge : - )
Just these last two things here above

gtsfm/common/keypoints.py
```diff
@@ -99,18 +99,32 @@ def get_top_k(self, k: int) -> "Keypoints":
             subset of current keypoints.
         """
         if k >= len(self):
-            return copy.deepcopy(self)
+            return copy.deepcopy(self), np.arange(self.__len__())
```
ayushbaid, line 102: nit: `len(self)` looks better. I am fine with this too.

ayushbaid (approved):

johnwlambert:
dashboard:
[visual_comparison_dashboard.html 2.zip](https://github.com/borglab/gtsfm/files/7686129/visual_comparison_dashboard.html.2.zip)

johnwlambert (approved):

PR #377, opened 2021-11-06
Add MVS to scene optimizer

johnwlambert:
Run MVS in the SceneOptimizer.

## How big are the artifacts?
With voxel downsampling, file sizes now reduced to :

results-deep_front_end-2011205_rc3-20-png-wget-astronet-1024-true.zip | 66.6 MB |  
results-deep_front_end-door-12-12-JPG-test_data-olsson-loader-1296-true.zip | 22.5 MB |  
results-deep_front_end-skydio-32-32-jpg-gdrive-colmap-loader-760-true.zip | 32.5 MB |  
results-sift_front_end-door-12-12-JPG-test_data-olsson-loader-1296-true.zip | 26 MB |  
results-sift_front_end-skydio-32-32-jpg-gdrive-colmap-loader-760-true.zip | 8.73 MB |  
results-sift_front_end-skydio-8-8-jpg-gdrive-colmap-loader-760-true.zip | 4.34 MB

## What do the results look like?

Door-12, sift front end:
Screen Shot 2021-11-27 at 7 03 38 PM

Skydio-32, deep front end:
Screen Shot 2021-11-27 at 7 05 26 PM

Screen Shot 2021-11-27 at 7 05 04 PM

## How big were the artifacts before we added voxel-based downsampling of the point cloud?

PLY sizes:
- Door-8, deep-front end: 576 MB from 21 million vertices (tons of redundancy).
Screen Shot 2021-11-08 at 9 37 21 PM
Screen Shot 2021-11-08 at 9 37 59 PM

- Skydio-8, deep front end: 5 MB (@codyly this looks much worse than the ones you showed earlier?)
Screen Shot 2021-11-08 at 9 34 33 PM

- Skydio-32: 732 KB (27K vertices)
Screen Shot 2021-11-09 at 10 18 34 AM

-

PR #381, opened 2021-11-16
Colored MVS result

codyly:
this PR is based on #377. RGB colors are added to the result ply file

### DIFF
--- a/gtsfm/densify/mvs_patchmatchnet.py
+++ b/gtsfm/densify/mvs_patchmatchnet.py
@@ -64,7 +64,7 @@ class MVSPatchmatchNet(MVSBase):
         min_conf_thresh: float = MIN_CONFIDENCE_THRESH,
         min_num_consistent_views: float = MIN_NUM_CONSISTENT_VIEWS,
         num_workers: int = 0,
-    ) -> np.ndarray:
+    ) -> Dict[str, np.ndarray]:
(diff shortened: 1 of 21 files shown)


johnwlambert:
@codyly can you change the upstream branch of your PR to be [377](https://github.com/borglab/gtsfm/pull/377)? That way the diff is very small and it's easier to review.

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -302,4 +308,4 @@ def filter_depth(
                 joint_mask.mean(),
             )
 
-        return np.concatenate(vertices, axis=0)
+        return {"points": np.concatenate(vertices, axis=0), "rgb": np.concatenate(vertex_colors, axis=0)}
```
johnwlambert, line 311: @ayushbaid @akshay-krishnan could you all take a look at this line?

I think we might be "using a dictionary as a class", which can lead to a bunch of issues.

Would everyone be ok with defining a new class
```python
class MvsResult(NamedTuple)
       points: np.ndarray
       rgb: np.ndarray
```

That we return instead?

ayushbaid: Whats are the pros/cons of storing the color along with the coordinates vs storing storing them separate as you have done here?

PR #385, opened 2021-11-21
Add visual dashboard for benchmarks

johnwlambert:
Provides visual insights  into all metrics of all benchmarks.

Red=regression
Green=improvement
Screen Shot 2021-11-21 at 1 30 27 AM

### DIFF
--- a/gtsfm/evaluation/visualize_benchmark_comparison.py
+++ b/gtsfm/evaluation/visualize_benchmark_comparison.py
@@ -1,34 +1,36 @@
 """
-Use Plotly to create a heatmap representing a 2d table, and add annotations to it.
+Use Plotly to create a dashboard that compares the metrics across all the benchmarks from the CI.
+
+The dashboard is a heatmap representing a 2d table, with text annotations added to it.
 
 Authors: John Lambert
 """
 
 import argparse
 from collections import defaultdict
+from pathlib import Path
 from typing import List
 
 import numpy as np
-import pandas as pd
-import matplotlib.pyplot as plt
-import plotly
 import plotly.graph_objects as go
 from matplotlib import colors
 from matplotlib.colors import LinearSegmentedColormap
-from plotly.graph_objs.layout import Font, Margin, XAxis, YAxis
+from plotly.graph_objs.layout import Annotation, Font, Margin, XAxis, YAxis
 
 import gtsfm.evaluation.merge_reports as report_utils
 import gtsfm.evaluation.metrics_report as metrics_report
+import gtsfm.utils.metrics as metrics_utils
 
 
 HEATMAP_WIDTH = 1500
 HEATMAP_HEIGHT = 900
 NUM_COLORS_COLORMAP = 100
-EPSILON = 1e-12
 MAX_NUM_CHARS_ARTIFACT_FNAME = 35
 
 MIN_RENDERABLE_PERCENT_CHANGE = -20
 MAX_RENDERABLE_PERCENT_CHANGE = 20
+
+DASHBOARD_HTML_SAVE_FPATH = Path(__file__).parent.parent.parent / "visual_comparison_dashboard.html"
 
 ZIP_FNAMES = [
     "deep_front_end-2011205_rc3-20-png-wget-astronet-1024-true.zip",
@@ -58,43 +60,44 @@ PALE_YELLOW_HEX = "#f5f6ce"
 GREEN_HEX = "#31b404"
 
 
-def colormap_to_colorscale(cmap):
-    # function that transforms a matplotlib colormap to a Plotly colorscale
-    return [colors.rgb2hex(cmap(k * 1 / NUM_COLORS_COLORMAP)) for k in range(NUM_COLORS_COLORMAP + 1)]
+def colorscale_from_list(colorlist: List[str]) -> List[str]:
+    """Create hex colorscale to interpolate between requested colors.
+
+    Args:
+        colorlist: requested colors.
+
+    Returns:
+        colorscale: list of length (NUM_COLORS_COLORMAP+1) representing a list of colors.
+    """
+    cmap = LinearSegmentedColormap.from_list(name="dummy_name", colors=colorlist)
+    colorscale = [colors.rgb2hex(cmap(k * 1 / NUM_COLORS_COLORMAP)) for k in range(NUM_COLORS_COLORMAP + 1)]
+    return colorscale
 
 
-def colorscale_from_list(alist, name):
-    # Defines a colormap, and the corresponding Plotly colorscale from the list alist
-    # alist=the list of basic colors
-    # name is the name of the corresponding matplotlib colormap
-
-    cmap = LinearSegmentedColormap.from_list(name, alist)
-    colorscale = colormap_to_colorscale(cmap)
-    return cmap, colorscale
-
-
-def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray, table_name: str) -> None:
+def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray) -> str:
     """Create an annotated heatmap.
 
     Args:
         X: labels for each column (column names).
         Y: labels for each row (row names).
         Z: 2d matrix, representing percentage changes from value for a metric on the master branch.
+
+    Returns:
+        string representing html for table.
     """
 
-    # clip Z to -100% and +100%, and then scale result to a linear scale [0,100]
+    # Clip Z to -20% and +20%. The clipping is only for the color -- the text will still display the correct numbers.
     Z_clipped = np.clip(Z, a_min=MIN_RENDERABLE_PERCENT_CHANGE, a_max=MAX_RENDERABLE_PERCENT_CHANGE)
 
     redgreen = [RED_HEX, PALE_YELLOW_HEX, GREEN_HEX]
-    fin_cmap, fin_cs = colorscale_from_list(redgreen, "fin_cmap")
+    colorscale = colorscale_from_list(redgreen)
     trace = go.Heatmap(
         z=Z_clipped,
         x=X,
         y=Y,
-        colorscale=fin_cs,  # fin_asymm_cs,
+        colorscale=colorscale,
         zmin=-MIN_RENDERABLE_PERCENT_CHANGE,
-        zmax=MAX_RENDERABLE_PERCENT_CHANGE
-        # colorbar=dict(thickness=20, tickvals=[-20+k*5 for k in range(5)], ticklen=4)
+        zmax=MAX_RENDERABLE_PERCENT_CHANGE,
     )
 
     layout = go.Layout(
@@ -106,8 +109,6 @@ def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray, table_name: st
             title="",
             autorange="reversed",
             showgrid=True,
-            # autotick=False,
-            # dtick=1
         ),
         autosize=False,
         height=HEATMAP_HEIGHT,
@@ -124,7 +125,7 @@ def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray, table_name: st
     for i in range(num_rows):
         for j in range(num_cols):
             annotations.append(
-                go.Annotation(
+                Annotation(
                     text=str(np.round(Z[i, j], 1)) + "%",
                     x=X[j],
                     y=Y[i],
@@ -141,13 +142,15 @@ def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray, table_name: st
 def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> None:
     """Generate a dashboard showing a visual representation of the diff against master on all benchmarks.
 
+    This script expects to find the metrics in CI artifact files and saves to the main repo directory.
+    TODO(johnwlambert): read metrics from JSON, instead of from the HTML report.
+
     Args:
-        curr_master_dirpath
-        new_branch_dirpath
+        curr_master_dirpath: path to directory containing benchmark artifacts for the master branch.
+        new_branch_dirpath: path to directory containing benchmark artifacts for a new branch.
     """
 
-    html_path = "visual_comparison_dashboard.html"
-    f = open(html_path, mode="w")
+    f = open(DASHBOARD_HTML_SAVE_FPATH, mode="w")
 
     # Write HTML headers.
     f.write("<!DOCTYPE html>" "<html>")
@@ -179,7 +182,7 @@ def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> Non
                 if branch_val is None:
                     percentage_change = np.nan
                 else:
-                    percentage_change = compute_percentage_change(float(master_val), float(branch_val))
+                    percentage_change = metrics_utils.compute_percentage_change(float(master_val), float(branch_val))
 
                 if "error" in metric_name and "outlier" not in metric_name:
                     # smaller is better, so this will flip the color to green for reduced values, instead of red
@@ -196,56 +199,27 @@ def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> Non
             Y.append(metric_name)
 
         Z = np.array(Z_rows)
-        html = plot_colored_table(X, Y, Z, table_name=table_name)
+        table_html = plot_colored_table(X, Y, Z)
 
         # Write name of the metric group in human readable form.
         f.write(metrics_report.get_html_metric_heading(table_name))
-        f.write(html)
+        f.write(table_html)
 
         # Close HTML tags.
         f.write("</html>")
 
 
-def compute_percentage_change(x: float, y: float) -> float:
-    """Return percentage in representing the regression or improvement of a value x, for new value y."""
-    return (y - x) / (x + EPSILON) * 100
-
-
-def test_compute_percentage_change_improve() -> None:
-    """ """
-    x = 100
-    y = 150
-    change_percent = compute_percentage_change(x, y)
-    assert np.isclose(change_percent, 50)
-
-
-def test_compute_percentage_change_static() -> None:
-    """ """
-    x = 100
-    y = 100
-    change_percent = compute_percentage_change(x, y)
-    assert np.isclose(change_percent, 0)
-
-
-def test_compute_percentage_change_regression() -> None:
-    """ """
-    x = 100
-    y = 1
-    change_percent = compute_percentage_change(x, y)
-    assert np.isclose(change_percent, -99)
-
-
 if __name__ == "__main__":
     parser = argparse.ArgumentParser()
-    parser.add_argument(
-        "--new_branch_dirpath",
-        required=True,
-        help="Path to directory containing benchmark artifacts for a new branch.",
-    )
     parser.add_argument(
         "--curr_master_dirpath",
         required=True,
         help="Path to directory containing benchmark artifacts for the master branch.",
+    )
+    parser.add_argument(
+        "--new_branch_dirpath",
+        required=True,
+        help="Path to directory containing benchmark artifacts for a new branch.",
     )
     args = parser.parse_args()
     generate_dashboard(
--- a/gtsfm/evaluation/visualize_benchmark_comparison.py
+++ b/gtsfm/evaluation/visualize_benchmark_comparison.py
@@ -119,7 +119,6 @@ def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray) -> str:
     fig = go.Figure(data=go.Data([trace]), layout=layout)
 
     annotations = go.Annotations()
-    alist = list(Z)
 
     num_rows, num_cols = Z.shape
     for i in range(num_rows):
--- a/gtsfm/evaluation/visualize_benchmark_comparison.py
+++ b/gtsfm/evaluation/visualize_benchmark_comparison.py
@@ -74,27 +74,28 @@ def colorscale_from_list(colorlist: List[str]) -> List[str]:
     return colorscale
 
 
-def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray) -> str:
+def plot_colored_table(row_labels: List[str], col_labels: List[str], tab_data: np.ndarray) -> str:
     """Create an annotated heatmap.
 
     Args:
-        X: labels for each column (column names).
-        Y: labels for each row (row names).
-        Z: 2d matrix, representing percentage changes from value for a metric on the master branch.
+        row_labels: labels for each column (column names) in the "x" direction.
+        col_labels: labels for each row (row names) in the "y" direction.
+        tab_data: 2d matrix, representing table data. Entries of the table represent percentage changes
+            from a value for a metric on the master branch. Values can be considered in the "z" direction.
 
     Returns:
-        string representing html for table.
+        string representing HTML code for the generated Plotly table.
     """
 
-    # Clip Z to -20% and +20%. The clipping is only for the color -- the text will still display the correct numbers.
-    Z_clipped = np.clip(Z, a_min=MIN_RENDERABLE_PERCENT_CHANGE, a_max=MAX_RENDERABLE_PERCENT_CHANGE)
+    # Clip "Z" to -20% and +20%. The clipping is only for the color -- the text will still display the correct numbers.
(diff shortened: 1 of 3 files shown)


dellaert:
That’s awesome!

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+HEATMAP_HEIGHT = 900
+NUM_COLORS_COLORMAP = 100
+EPSILON = 1e-12
+MAX_NUM_CHARS_ARTIFACT_FNAME = 35
+
+MIN_RENDERABLE_PERCENT_CHANGE = -20
+MAX_RENDERABLE_PERCENT_CHANGE = 20
+
+ZIP_FNAMES = [
+    "deep_front_end-2011205_rc3-20-png-wget-astronet-1024-true.zip",
```
dellaert, line 34: should this list come from some configuration file associated with the CI? Or just a directory read?

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
+"""
+Use Plotly to create a heatmap representing a 2d table, and add annotations to it.
```
akshay-krishnan, line 2: Maybe add that it compares the metrics across all the benchmarks from the CI? Currently lacks any description of what the 2d table is.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+def colormap_to_colorscale(cmap):
+    # function that transforms a matplotlib colormap to a Plotly colorscale
+    return [colors.rgb2hex(cmap(k * 1 / NUM_COLORS_COLORMAP)) for k in range(NUM_COLORS_COLORMAP + 1)]
+
+
+def colorscale_from_list(alist, name):
+    # Defines a colormap, and the corresponding Plotly colorscale from the list alist
+    # alist=the list of basic colors
+    # name is the name of the corresponding matplotlib colormap
```
akshay-krishnan, line 69: Dont these need documentation in our standard format?

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray, table_name: str) -> None:
+    """Create an annotated heatmap.
+
+    Args:
+        X: labels for each column (column names).
+        Y: labels for each row (row names).
+        Z: 2d matrix, representing percentage changes from value for a metric on the master branch.
+    """
+
+    # clip Z to -100% and +100%, and then scale result to a linear scale [0,100]
```
akshay-krishnan, line 85: is this comment outdated? looks like the constants are set to -20 and +20.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+    Z_clipped = np.clip(Z, a_min=MIN_RENDERABLE_PERCENT_CHANGE, a_max=MAX_RENDERABLE_PERCENT_CHANGE)
```
akshay-krishnan, line 86: Just to clarify, the clipping is only for the color right? It will still display the correct numbers?

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+    # Defines a colormap, and the corresponding Plotly colorscale from the list alist
+    # alist=the list of basic colors
+    # name is the name of the corresponding matplotlib colormap
+
+    cmap = LinearSegmentedColormap.from_list(name, alist)
+    colorscale = colormap_to_colorscale(cmap)
+    return cmap, colorscale
+
+
+def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray, table_name: str) -> None:
```
akshay-krishnan, line 76: Looks like the returned type is not correct? Also please update documentation.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> None:
+    """Generate a dashboard showing a visual representation of the diff against master on all benchmarks.
+
+    Args:
+        curr_master_dirpath
+        new_branch_dirpath
+    """
+
+    html_path = "visual_comparison_dashboard.html"
```
akshay-krishnan, line 149: should this be made a constant or an argument?

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+        Y = []
+
+        benchmark_table_vals = defaultdict(dict)
+
+        # loop over each benchmark result (columns of table)
+        for zip_fname in ZIP_FNAMES:
+            # use just the first 35 chars
+            X.append(zip_fname[:MAX_NUM_CHARS_ARTIFACT_FNAME])
+
+            report1_fpath = f"{curr_master_dirpath}/results-{zip_fname}/result_metrics/gtsfm_metrics_report.html"
```
akshay-krishnan, line 169: /result_metrics/gtsfm_metrics_report.html should also probably be made a constant or an argument. 
Instead of using the "results-{zip_fname}" directory, can we use results/{zip_fname}? That way, we can bake the "results/" into curr_master_dirpath, and let zip_fname be equal to the directory name.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+    change_percent = compute_percentage_change(x, y)
+    assert np.isclose(change_percent, 0)
+
+
+def test_compute_percentage_change_regression() -> None:
+    """ """
+    x = 100
+    y = 1
+    change_percent = compute_percentage_change(x, y)
+    assert np.isclose(change_percent, -99)
```
akshay-krishnan, line 235: why are these tests here? I am surprised to see so many tests for a simple formula. If you really need them, can we move the function to a util and the tests to its tests? I think there is a util for evaluation or metrics

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+        # Write name of the metric group in human readable form.
+        f.write(metrics_report.get_html_metric_heading(table_name))
+        f.write(html)
+
+        # Close HTML tags.
+        f.write("</html>")
+
+
+def compute_percentage_change(x: float, y: float) -> float:
+    """Return percentage in representing the regression or improvement of a value x, for new value y."""
```
akshay-krishnan, line 210: Document args and returns?

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+                    font=dict(color="rgb(25,25,25)"),
+                    showarrow=False,
+                )
+            )
+    fig["layout"].update(annotations=annotations)
+    return fig.to_html(full_html=False, include_plotlyjs="cdn")
+
+
+def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> None:
+    """Generate a dashboard showing a visual representation of the diff against master on all benchmarks.
```
akshay-krishnan, line 142: Document where the function looks for these metrics and where it saves the dashboard.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+        # loop over each benchmark result (columns of table)
+        for zip_fname in ZIP_FNAMES:
+            # use just the first 35 chars
+            X.append(zip_fname[:MAX_NUM_CHARS_ARTIFACT_FNAME])
+
+            report1_fpath = f"{curr_master_dirpath}/results-{zip_fname}/result_metrics/gtsfm_metrics_report.html"
+            tables_dict1 = report_utils.extract_tables_from_report(report1_fpath)
+
+            report2_fpath = f"{new_branch_dirpath}/results-{zip_fname}/result_metrics/gtsfm_metrics_report.html"
+            tables_dict2 = report_utils.extract_tables_from_report(report2_fpath)
```
akshay-krishnan, line 173: this is okay if you dont have more time to work on this now, so maybe you could add a TODO? 
Ideally we should fetch information from the json files with the metrics. That is the source of truth. The report will probably change in future, might contain other errors etc.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+            # loop over each metric within this table (rows of table)
+            for i, (metric_name, master_val, branch_val) in enumerate(merged_tables_dict[table_name]):
+
+                if branch_val is None:
+                    percentage_change = np.nan
+                else:
+                    percentage_change = compute_percentage_change(float(master_val), float(branch_val))
+
+                if "error" in metric_name and "outlier" not in metric_name:
```
akshay-krishnan, line 184: Im not sure if this is consistent across all metrics, and will be enforced in future. But I agree there's no better way to get that information now. If this becomes important we can add some information to the GtsfmMetric.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--new_branch_dirpath",
+        required=True,
+        help="Path to directory containing benchmark artifacts for a new branch.",
+    )
+    parser.add_argument(
+        "--curr_master_dirpath",
```
akshay-krishnan, line 246: Does this have to be master? Or can it be any "base" branch?

akshay-krishnan (changes requested):
Nice dashboard, thanks!

ayushbaid (commented):
LGTM overall. Can we take care of `Nan%` in the output

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: thanks for the review Akshay -- yes you're right, I'll clarify that

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: good catch, i'll update

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: could be any base branch. I can rename it if you think that would be clearer.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+Authors: John Lambert
+"""
+
+import argparse
+from collections import defaultdict
+from typing import List
+
+import numpy as np
+import pandas as pd
```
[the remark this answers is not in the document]
johnwlambert: thanks for the review @ayushbaid . i'll update

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: sure, i moved them to utils/metrics.py and the corresponding unit test file now

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: thanks, updated now

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: I made it a global constant to make it clearer where the file would be saved

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: updated, thanks

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: updated -- thanks

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+
+def colorscale_from_list(alist, name):
+    # Defines a colormap, and the corresponding Plotly colorscale from the list alist
+    # alist=the list of basic colors
+    # name is the name of the corresponding matplotlib colormap
+
+    cmap = LinearSegmentedColormap.from_list(name, alist)
+    colorscale = colormap_to_colorscale(cmap)
+    return cmap, colorscale
```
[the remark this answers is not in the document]
johnwlambert: Not sure. They could be used instead of the `colour` library in the Open3d frustum color visualization, for example. We could consolidate later, also.

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: yes, thanks -- updated

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+    "Rotation Cycle Consistency Metrics",
+    "Cycle Consistent Frontend Summary",
+    "Averaging Metrics",
+    "Data Association Metrics",
+    "Bundle Adjustment Metrics",
+]
+
+RED_HEX = "#df0101"
+PALE_YELLOW_HEX = "#f5f6ce"
+GREEN_HEX = "#31b404"
```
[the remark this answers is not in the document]
johnwlambert: I'm a bit ambivalent about this. Not sure if there is a better file than here.

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: thanks for the review Frank. We have a few options we could go with here. The zip artifact names are auto-generated, so we don't have them stored somewhere else yet. But we could put their names in a separate file somewhere.

If we glob the file names from a directory read, the directory might be missing some artifacts (if some of the CI runs failed, for example), so we might not now the dashboard is incomplete.

@akshay-krishnan @ayushbaid any preferences on your end here? I think it's ok for now the way it is, but if you have a strong preference, I'm happy to update

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,254 @@
...
+
+TABLE_NAMES = [
+    "Verifier Summary",
+    "Inlier Support Processor Summary",
+    "Rotation Cycle Consistency Metrics",
+    "Cycle Consistent Frontend Summary",
+    "Averaging Metrics",
+    "Data Association Metrics",
+    "Bundle Adjustment Metrics",
+]
```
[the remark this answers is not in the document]
johnwlambert: We could. I don't think we look for or use these table titles anywhere else currently, though

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: (i put the TODO in the docstring)

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: Hmm the path prefix `results/*` wouldn't be valid though, right?

gtsfm/evaluation/visualize_benchmark_comparison.py
travisdriver: If it's not too difficult, generating the file names from the information in `config_dataset_info` of the `benchmark.yaml` config would make the most sense to me. That way, we don't have to edit anything in the main code base when we add an additional benchmark.

gtsfm/evaluation/visualize_benchmark_comparison.py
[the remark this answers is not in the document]
travisdriver: Global color definitions are fine IMO.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,227 @@
...
+    return colorscale
+
+
+def plot_colored_table(X: List[str], Y: List[str], Z: np.ndarray) -> str:
+    """Create an annotated heatmap.
+
+    Args:
+        X: labels for each column (column names).
+        Y: labels for each row (row names).
+        Z: 2d matrix, representing percentage changes from value for a metric on the master branch.
```
travisdriver, line 83: These variable names don't represent the information they hold very well. Maybe `row_labels`, `col_tables`, `data` or something would be clearer.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,227 @@
...
+        new_branch_dirpath: path to directory containing benchmark artifacts for a new branch.
+    """
+
+    f = open(DASHBOARD_HTML_SAVE_FPATH, mode="w")
+
+    # Write HTML headers.
+    f.write("<!DOCTYPE html>" "<html>")
+    f.write(metrics_report.get_html_header())
+
+    # loop over each table in the HTML report.
```
travisdriver, line 158: nit: capitalize first letter of sentence.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,227 @@
...
+    for table_name in TABLE_NAMES:
+
+        X = []
+        Y = []
+
+        benchmark_table_vals = defaultdict(dict)
+
+        # loop over each benchmark result (columns of table)
```
travisdriver, line 166: nit: capitalization and punctuation.

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -0,0 +1,227 @@
...
+            X.append(zip_fname[:MAX_NUM_CHARS_ARTIFACT_FNAME])
+
+            report1_fpath = f"{curr_master_dirpath}/results-{zip_fname}/result_metrics/gtsfm_metrics_report.html"
+            tables_dict1 = report_utils.extract_tables_from_report(report1_fpath)
+
+            report2_fpath = f"{new_branch_dirpath}/results-{zip_fname}/result_metrics/gtsfm_metrics_report.html"
+            tables_dict2 = report_utils.extract_tables_from_report(report2_fpath)
+            merged_tables_dict = report_utils.merge_tables(tables_dict1, tables_dict2)
+
+            # loop over each metric within this table (rows of table)
```
travisdriver, line 178: nit: capitalization and punctuation

gtsfm/utils/metrics.py
```diff
@@ -422,3 +423,16 @@ def save_metrics_as_json(metrics_groups: Delayed, output_dir: str) -> None:
...
+    """Return percentage in representing the regression or improvement of a value x, for new value y.
+
+    Args:
+        x: original value to compare against.
+        y: new value.
+
+    Returns:
+        percentage change.
+    """
+    return (y - x) / (x + EPSILON) * 100
```
travisdriver, line 438: `abs(y - x) / (abs(x) + EPSILON) * 100`?

johnwlambert: So I compute here a negative or positive percentage change (and use a red or green color based on the sign). I updated the docstring to clarify that.

gtsfm/utils/metrics.py
```diff
@@ -422,3 +423,16 @@ def save_metrics_as_json(metrics_groups: Delayed, output_dir: str) -> None:
     """
     for metrics_group in metrics_groups:
         metrics_group.save_to_json(os.path.join(output_dir, metrics_group.name + ".json"))
+
+
+def compute_percentage_change(x: float, y: float) -> float:
```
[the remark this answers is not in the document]
johnwlambert: It could be, but we couldn't unit test it in that case.

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: thanks, updated.

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: thanks, updated

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: thanks, updated

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: good point, updated now.

gtsfm/evaluation/visualize_benchmark_comparison.py
akshay-krishnan: Yeah we would have to change where these files are populated as well. What I’m trying to say is: this function assumes certain files to exist in certain locations which is not ideal. Ideally, we’d pass it arguments that provide the whole path and it does its job.

akshay-krishnan (approved):

gtsfm/evaluation/visualize_benchmark_comparison.py
akshay-krishnan: That would be better and more general, but I don’t feel so strongly about it.

gtsfm/evaluation/visualize_benchmark_comparison.py
travisdriver: @johnwlambert Would this be possible to do?

gtsfm/evaluation/visualize_benchmark_comparison.py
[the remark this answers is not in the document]
travisdriver: I agree with @ayushbaid. Maybe we should move these functions to `visualization/vis_utils.py` or something?

travisdriver (changes requested):
Would like [this comment](https://github.com/borglab/gtsfm/pull/385/files/d45d3ddd133026369f088f37604bccc4a82a2312#r754412618) addressed before merging.

gtsfm/evaluation/visualize_benchmark_comparison.py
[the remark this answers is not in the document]
travisdriver: Aren't these just the exact titles of the sections in `gtsfm_metrics_report.html`?

gtsfm/evaluation/visualize_benchmark_comparison.py
travisdriver: I'm still seeing three tests in this file.

gtsfm/evaluation/visualize_benchmark_comparison.py
johnwlambert: Sure, this is a good suggestion. I've updated the code now to parse the benchmark YAML and auto-generate the expected file names.

travisdriver (approved):
LGTM. Great work!

gtsfm/evaluation/visualize_benchmark_comparison.py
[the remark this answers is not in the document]
johnwlambert: I'll create an issue for it. Something we can consolidate in a future small PR, but I think out of scope here.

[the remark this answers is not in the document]
travisdriver: It's as easy as copying and pasting to another file. Think you should just resolve it here instead of opening an issue...

akshay-krishnan (approved):

gtsfm/evaluation/visualize_benchmark_comparison.py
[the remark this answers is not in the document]
johnwlambert: I prefer to have more small PRs that touch fewer files at once, to reduce the diff. The colormap won't be exactly the same, and I'd prefer to land this ASAP so we can start using this dashboard to help land other PRs more quickly

PR #391, opened 2021-11-27
add unit tests and test data for dashboard creation utilities

johnwlambert:
Add unit tests and test data for dashboard creation utilities

### DIFF
--- a/tests/data/sample_ci_benchmark/benchmark.yml
+++ b/tests/data/sample_ci_benchmark/benchmark.yml
@@ -1,0 +1,92 @@
+name: Benchmark GTSFM on select datasets using SIFT and Deep front-ends
+
+on: [pull_request, workflow_dispatch]
+
+jobs:
+  benchmark:
+    name: Benchmark
+    runs-on: ubuntu-latest
+    strategy:
+      # Prevents other configs from being canceled if one of them fails.
+      fail-fast: false
+      matrix:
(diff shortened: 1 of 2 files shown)


tests/data/sample_ci_benchmark/benchmark.yml
```diff
@@ -0,0 +1,92 @@
+name: Benchmark GTSFM on select datasets using SIFT and Deep front-ends
```
akshay-krishnan, line 1: is this file duplicated?

johnwlambert: it is. i'm including it in `tests/data/sample_ci_benchmark/benchmark.yml` so that we even if we change the benchmark datasets in the future, the unit test will still be valid

akshay-krishnan: is it not better to update the unit tests every time we change the benchmarks? It might not need update if we only removing or adding datasets, but would need update if we change the format or something.

johnwlambert: I think it will be too onerous to have to update the unit tests every time we update a benchmark.

In my ideal world, the unit test is written one time, and passes, we don't have to look at it again, we can focus on other things instead.

akshay-krishnan (approved):

PR #393, opened 2021-11-29
Image retriever

travisdriver:
Here is the image retrieval class I had put together a while back. At this point, I think it's futile to try to update this branch and merge with master. However, wanted to get some feedback on the general structure of the solution.


dellaert:
Why is it "futile"?

travisdriver:
> Why is it "futile"?

I meant that it would probably take longer to try to merge master into this branch as opposed to starting a _fresh_ branch off master (as the last commit was on August 31st). 

However, the general structure of everything is still the same, so I can incorporate any suggestions with respect to this draft into the fresh branch.

gtsfm/frontend/retriever/default_retriever.py
```diff
@@ -0,0 +1,50 @@
...
+
+from gtsfm.frontend.retriever.retriever_base import RetrieverBase
+import gtsfm.utils.logger as logger_utils
+from gtsfm.two_view_estimator import TwoViewEstimator
+
+
+logger = logger_utils.get_logger()
+
+
+class DefaultRetriever(RetrieverBase):
```
johnwlambert, line 18: thanks for the PR, @travisdriver. I think we should name this as `TemporalRetriever`, so that the name is a bit more descriptive

gtsfm/scene_optimizer.py
```diff
@@ -119,14 +126,27 @@ def create_computation_graph(
             keypoints_graph_list += [delayed_dets]
             descriptors_graph_list += [delayed_descs]
 
-        # estimate two-view geometry and get indices of verified correspondences.
+        # Add detections and descriptors to the Retriever and create local matching graph.
+        local_matching_graph_dict = self.image_retriever.create_computation_graph(
```
johnwlambert, line 130: It might make more sense to have the Retriever run after the feature extractor, and then have it spit out potential pairs, and then just keep the existing for loop over pairs (each one goes into TwoViewEstimator).

So I was thinking TwoViewEstimator would not be an attribute of the retriever (I think retriever is more minor and TwoViewEstimator is more major in class prominence)

gtsfm/frontend/retriever/default_retriever.py
travisdriver: It doesn't really use any temporal information. Maybe `SequentialRetriever` or `SlidingWindowRetriever`?

johnwlambert: True, we can't guarantee the sequence was captured in chronological order. I like the term "SequentialRetriever". If I remember correct, that's also the terminology COLMAP uses.

gtsfm/scene_optimizer.py
travisdriver: Is the method that you described parallelizable? Before we call the TwoViewEstimator, we have to know the image pair indices `i1` and `i2`. 

I believe the way I've coded it here allows for the pair retrieval to be delayed and parallelized. Although, I may be wrong.

travisdriver: Can retrieval be parallelized in the way that you describe? Before we call the TwoViewEstimator, the image indices `i1` and `i2` have to already have been computed.

I believe making delayed calls to the TwoViewEstimator from the Retriever allows for the pair proposal process to be done in parallel. Otherwise, we have to evaluate all of the features first and pass into the Retriever, creating a bottleneck. I'm not 100% sure on this though.

However, for the case when we're considering global descriptors, I'm not sure this bottleneck can be avoided (as we need to compute the global descriptors for all the images first before proposing image pairs for the TwoViewEstimator).

womackj1:
The Image Retriever is now [here](https://github.com/borglab/gtsfm/pull/431)

gtsfm/scene_optimizer.py
travisdriver: @johnwlambert Still wanted to get your opinion on this. The way I see it there are three options since the TwoViewEstimator will be called based on the output of the Retriever:

1. Make the TwoViewEstimator an attribute of the Retriever and only call it if the Retriever decides it's appropriate (this is what's implemented in this branch currently)
2. Pass the (Delayed) output of the Retriever to the TwoViewEstimator and let the estimator evaluate the output
3. Don't use any delayed operations in the Retriever and evaluate its output in the main loop of the the SceneOptimizer, which would explicitly create a bottleneck between feature extraction and the TwoViewEstimator

johnwlambert: Hi Travis, thanks for following up on this. I like option #2 most.

akshay-krishnan:
can we close this then?

PR #396, opened 2021-12-01
Use ransac-based Sim3 pose graph alignment

johnwlambert:
Account for occasional outliers, to still get correct pose errors w.r.t. reference/ground truth.


gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +43,122 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
...
+    valid_idxs = [i for i, bTi in enumerate(bTi_list_est) if bTi is not None]
+    num_to_delete = math.ceil(delete_frac * len(valid_idxs))
+    if len(valid_idxs) - num_to_delete < 2:
+        # run without RANSAC! want at least 2 frame pairs for good alignment.
+        aligned_bTi_list_est, aSb = align_poses_sim3_ignore_missing(
+            aTi_list_ref, bTi_list_est
+        )
+        return aligned_bTi_list_est, aSb
+
+    # randomly delete some elements
```
akshay-krishnan, line 81: Why is this implemented this way? Do we know how many minimum number of points are needed for a uniqueue Sim3 estimate? Then we should select those many points.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +43,122 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
...
+        aligned_bTi_list_est, aSb = align_poses_sim3_ignore_missing(
+            aTi_list_ref_subset, bTi_list_est_subset
+        )
+
+        # evaluate inliers.
+        rot_error, trans_error, _, _ = compute_pose_errors_3d(aTi_list_ref, aligned_bTi_list_est)
+        if verbose:
+            print("Deleted ", delete_idxs, f" -> trans_error {trans_error:.1f}, rot_error {rot_error:.1f}")
+
+        if trans_error <= best_trans_error and rot_error <= best_rot_error:
```
akshay-krishnan, line 101: im not sure that is the best option. Basically if we have a hypothesis with a slightly worse rotation error but _much better_ translation error, we don't use it?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +43,122 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
     return [aRb.compose(bRi) for bRi in bRi_list]
 
 
+def ransac_align_poses_sim3_ignore_missing(
+    aTi_list_ref: List[Optional[Pose3]],
+    bTi_list_est: List[Optional[Pose3]],
+    num_iters: int = 1000,
+    delete_frac: float = DEFAULT_RANSAC_ALIGNMENT_DELETE_FRAC,
+    verbose: bool = False,
```
akshay-krishnan, line 51: I think we should use the powerful verbosity settings in the logger instead of creating more arguments.

akshay-krishnan (changes requested):
Just some high level comments to start with, we can discuss lower-level details after these are resolved

gtsfm/utils/geometry_comparisons.py
johnwlambert: Thanks for the review Akshay. Indeed there are other ways to implement this. In my experience thus far, this option works reasonably well.

gtsfm/utils/geometry_comparisons.py
johnwlambert: Good point. I'll update

gtsfm/utils/geometry_comparisons.py
johnwlambert: So we need at least 2 poses to align the pose graphs. Since it's least squares, we would expect that with more than 2, we could get a less noisy estimate, though.

gtsfm/utils/geometry_comparisons.py
johnwlambert: Updated.

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: I think we need 3, there is a rotational ambiguity if we use two. I am not sure about 3 either, just a guess. Could you check that we have at least those many cameras before finding the Sim3?

gtsfm/utils/geometry_comparisons.py
akshay-krishnan: Are there any metrics or plots that might be more convincing? I am surprised, because it really depends on the order of hypotheses. And there is a clear reason why it would not work.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +43,115 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
...
+    num_to_delete = math.ceil(delete_frac * len(valid_idxs))
+    if len(valid_idxs) - num_to_delete < 2:
+        # run without RANSAC! want at least 2 frame pairs for good alignment.
+        aligned_bTi_list_est, aSb = align_poses_sim3_ignore_missing(
+            aTi_list_ref, bTi_list_est
+        )
+        return aligned_bTi_list_est, aSb
+
+    # randomly delete some elements
+    for _ in range(num_iters):
```
akshay-krishnan, line 80: We don’t need more than num_cameras choose num_cameras *delete_frac. So many use the min of that and num_iters?

gtsfm/utils/geometry_comparisons.py
```diff
@@ -40,6 +43,115 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
     return [aRb.compose(bRi) for bRi in bRi_list]
 
 
+def ransac_align_poses_sim3_ignore_missing(
```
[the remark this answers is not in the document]
johnwlambert: Thanks for the review Ayush. I think that making a RANSAC wrapper for this will make the code a bit less readable. Since we'll have to have a very generic interface. So I'd prefer we leave it as is, for now.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -38,6 +41,119 @@ def align_rotations(aRi_list: List[Optional[Rot3]], bRi_list: List[Optional[Rot3
...
+    best_trans_error = float("inf")
+    best_rot_error = float("inf")
+
+    valid_idxs = [i for i, bTi in enumerate(bTi_list_est) if bTi is not None]
+    num_to_delete = math.ceil(delete_frac * len(valid_idxs))
+    if len(valid_idxs) - num_to_delete < 2:
+        # Run without RANSAC! Want at least 2 frame pairs for good alignment.
+        aligned_bTi_list_est, aSb = align_poses_sim3_ignore_missing(
+            aTi_list_ref, bTi_list_est
+        )
```
ayushbaid, line 74: Why not run with just 2 pairs as input (the minimum required to align)?

johnwlambert:
Superseded by https://github.com/borglab/gtsfm/pull/778, so closing (which samples just 2 poses to fit Sim3)

PR #397, opened 2021-12-02
Use Huber loss for Shonan

travisdriver:


johnwlambert:
Hi Travis, thanks for the PR. Just wanted to make sure you've seen this other closed PR: https://github.com/borglab/gtsfm/pull/344

dellaert:
Don't understand actually why we closed #344 without merging. I think we could have just merged it with a config flag set to OFF? Shonan is *really* bad at handling outliers so in theory it should be ON by default, but maybe tuned such that it does not negatively affect datasets without outlier pairwise edges.

travisdriver:
@johnwlambert #344 only presented results with respect to skydio-32. The dashboard will help with compare the results for all

johnwlambert:
> Don't understand actually why we closed #344 without merging. I think we could have just merged it with a config flag set to OFF? Shonan is _really_ bad at handling outliers so in theory it should be ON by default, but maybe tuned such that it does not negatively affect datasets without outlier pairwise edges.

@dellaert The main reason I closed #344  was that robust Shonan didn't appear to give any improvement on any dataset over standard Shonan. In my experiences, the results were always inferior or identical

ayushbaid:
Do we still want to merge this?

travisdriver:
> Do we still want to merge this?

I think having the option of using the robust optimization is useful. Even though it does not currently show any improvements, it may just be that the hyperparameters are not tuned properly.

PR #409, opened 2021-12-13
remove cameras with no measurements

codyly:
remove cameras with no measurements in `GtsfmData.filter_landmarks`

### DIFF
--- a/gtsfm/common/gtsfm_data.py
+++ b/gtsfm/common/gtsfm_data.py
@@ -168,7 +168,10 @@ class GtsfmData:
         """
         if camera is None:
             raise ValueError("Camera cannot be None, should be a valid camera")
-        self._cameras[index] = camera
+
+        # if camera with the given index has not been added, add this new camera
+        if index not in self._cameras:
+            self._cameras[index] = camera
 
     def get_track_length_statistics(self) -> Tuple[float, float]:
         """Compute mean and median lengths of all the tracks.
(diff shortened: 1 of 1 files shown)


gtsfm/common/gtsfm_data.py
```diff
@@ -351,14 +354,14 @@ def filter_landmarks(self, reproj_err_thresh: float = 5) -> "GtsfmData":
...
-            filtered_data.add_camera(i, self.get_camera(i))
-
         for j in range(self.number_tracks()):
             track = self.get_track(j)
 
             if self.__validate_track(track, reproj_err_thresh):
+                # check if all cameras with measurement in this track have already been added
+                for k in range(track.number_measurements()):
+                    i, _ = track.measurement(k)
+                    filtered_data.add_camera(i, self.get_camera(i))
```
johnwlambert, line 364: thanks for the PR @codyly . Won't this add the same camera over and over again? shouldn't we add just once?

codyly: I also modified the `add_camera` function to check whether the same camera has been added

codyly: If we do not modify the `add_camera`, it can be
```py
    for k in range(track.number_measurements()):
        i, _ = track.measurement(k)
        if i not in self._cameras:
            filtered_data.add_camera(i, self.get_camera(i))
```

johnwlambert: Ah I see, ok what you have already works for me then.

johnwlambert (approved):

PR #412, opened 2021-12-20
fix broken MVS-related unit tests

johnwlambert:
Two tests are failing currently: `TestEllipsoidUtils.test_get_right_singular_vectors` and `TestPatchmatchNetData.test_select_src_views`.

The first is due to the fact that we aren't returning singular values, so we need to correct the terminology. The other test looks like it is flaky due to some random ordering of the output.
```
>       assert np.allclose(singular_values, S)
E       AssertionError: assert False
E        +  where False = <function allclose at 0x7fefe2b41f70>(array([7.73928148, 2.82711821, 1.20177289]), array([13.40482874,  4.89671238,  2.08153171]))
```


johnwlambert:
This is a hotfix, so I'm merging it now to make the CI pass for others, and opening an issue to resolve these asap

PR #414, opened 2021-12-29
Add front-end runner

johnwlambert:
This PR adds a runner just for the front-end. Previously, all of our runners have run both the front-end and back-end sequentially.

Having an isolated runner allows us to use less boilerplate for unit tests just on the front-end.


gtsfm/runner/frontend_runner.py
```diff
@@ -0,0 +1,70 @@
...
+
+    Note: Copied from SceneOptimizer class, without back-end code.
+
+    Args:
+        loader: image loader.
+        feature_extractor: feature extractor module to use.
+        two_view_estimator: two-view estimator module to use.
+
+    Returns:
+        keypoints_list_results: detected keypoints for each image.
```
akshay-krishnan, line 28: why the _results suffix ? 
Actually we don't even need variable names for returns.

gtsfm/runner/frontend_runner.py
```diff
@@ -0,0 +1,70 @@
...
+            camera_intrinsics_graph[i2],
+            image_shape_graph[i1],
+            image_shape_graph[i2],
+        )
+        i2Ri1_graph_dict[(i1, i2)] = i2Ri1
+        i2Ui1_graph_dict[(i1, i2)] = i2Ui1
+        v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
+
+    with dask.config.set(scheduler="single-threaded"):
+        keypoints_list_results, i2Ri1_results, i2Ui1_results, corr_idxs_dict = dask.compute(
```
akshay-krishnan, line 66: could you consider removing the _results suffix?

akshay-krishnan: maybe just remove the _graph from the input names to get the output names?

gtsfm/runner/frontend_runner.py
```diff
@@ -0,0 +1,70 @@
...
+            camera_intrinsics_graph[i1],
+            camera_intrinsics_graph[i2],
+            image_shape_graph[i1],
+            image_shape_graph[i2],
+        )
+        i2Ri1_graph_dict[(i1, i2)] = i2Ri1
+        i2Ui1_graph_dict[(i1, i2)] = i2Ui1
+        v_corr_idxs_graph_dict[(i1, i2)] = v_corr_idxs
+
+    with dask.config.set(scheduler="single-threaded"):
```
akshay-krishnan, line 65: should the number of threads be configurable?

gtsfm/runner/frontend_runner.py
```diff
@@ -0,0 +1,70 @@
...
+
+import dask
+from dask.delayed import Delayed
+
+from gtsfm.feature_extractor import FeatureExtractor
+from gtsfm.loader.loader_base import LoaderBase
+from gtsfm.two_view_estimator import TwoViewEstimator
+
+
+def run_frontend(
```
akshay-krishnan, line 15: My most important comment here is why is this a function? My understanding is that a `runner` refers to a class with a run() method. Can we stick to that design ?

gtsfm/runner/frontend_runner.py
johnwlambert: Thanks for the review @akshay-krishnan.

#threads could be configurable. But for the purposes of tests, I propose we make this as simple as possible for now. We could update later.

gtsfm/runner/frontend_runner.py
johnwlambert: Sure, I've removed the `_results` suffix now. Good suggestion.

gtsfm/runner/frontend_runner.py
johnwlambert: Sure, I've removed the `_results` suffix now. Good suggestion.

gtsfm/runner/frontend_runner.py
johnwlambert: I'm a bit ambivalent about this. We could do:
```python
class SceneFrontendOptimizer:
     def run(
        self, loader: LoaderBase, feature_extractor: FeatureExtractor, two_view_estimator: TwoViewEstimator
     ) -> Tuple[List[Keypoints], Dict[Tuple[int,int], Rot3], Dict[Tuple[int,int], Unit3], Dict[Tuple[int,int], np.ndarray]]:
          """ """"
```
or
```python
@dataclass
class SceneFrontendOptimizer:
     loader: LoaderBase
     feature_extractor: FeatureExtractor
     two_view_estimator: TwoViewEstimator

     def run(self) -> Tuple[List[Keypoints], Dict[Tuple[int,int], Rot3], Dict[Tuple[int,int], Unit3], Dict[Tuple[int,int], np.ndarray]]:
          """ """"
```
or we can keep it purely functional, as it is now (1 line of code instead of 2 lines, since we would have to instantiate the object).

akshay-krishnan: I think design consistency is more important. It helps to not have exceptions and let all runners be classes with some common structure. But it looks like this is already not the case (the argoverse runner is not a class). 
Other runners always construct a scene optimizer (there is a virtual method for it), but here we are not doing so. If we can change that about runners, we could have a base class design that could be used for this frontend runner as well. Otherwise it does not make much difference. We could make it a class so that users do the same thing for all runners, but since it is not derived from the runner base class anyway, I don't feel strongly about it.

gtsfm/runner/frontend_runner.py
akshay-krishnan: sounds good.

akshay-krishnan (approved):

gtsfm/runner/frontend_runner.py
johnwlambert: Thanks Akshay. Good points. I agree more consistency is better. I think we didn't make the Argoverse runner a child of the base runner since the arguments were so different. @ayushbaid is that right?

I'll merge for now, but let's keep in mind how we can improve this in the long run.

travisdriver: I think having this as a class would have been consistent with our design policies thus far.

PR #417, opened 2022-01-02
Add generic GRIC-based two-view geometry verifier, for unit tests

johnwlambert:
Implements pycolmap's GRIC verifier as an additional verifier option.


johnwlambert:
Dashboard:
[visual_comparison_dashboard 2.html.zip](https://github.com/borglab/gtsfm/files/7805061/visual_comparison_dashboard.2.html.zip)

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
...
+            uv_i1: array of shape (N3,2) representing coordinates of 2d points in image 1.
+            uv_i2: array of shape (N3,2) representing corresponding coordinates of 2d points in image 2.
+            camera_intrinsics_i1: intrinsics for image #i1.
+            camera_intrinsics_i2: intrinsics for image #i2.
+
+        Returns:
+            dictionary containing result status code, estimated relative pose (R,t), and inlier mask.
+        """
+
+        def get_pycolmap_camera_dict(camera_intrinsics: Cal3Bundler) -> pycolmap.Camera:
```
akshay-krishnan, line 105: if we are treating it as pycolmap.Camera, I dont know why there are so many _dict here. IIUC, its basically a dict, but that is an implementation detail that can be explain in comment without much change in variable names.

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
...
+            logger.info("[GRIC] Not enough correspondences for verification.")
+            return self._failure_result
+
+        uv_i1 = keypoints_i1.coordinates[match_indices[:, 0]]
+        uv_i2 = keypoints_i2.coordinates[match_indices[:, 1]]
+
+        result_dict = self.__estimate_two_view_geometry(uv_i1, uv_i2, camera_intrinsics_i1, camera_intrinsics_i2)
+
+        success = result_dict["success"]
+        if not success:
```
akshay-krishnan, line 177: I dont think writing `if not result_dict["success"]` is losing any readability.

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
...
+            logger.info(f"[GRIC] matrix estimation unsuccessful.")
+            return self._failure_result
+
+        logger.info("Two view configuration: {ConfigurationType(result_dict['configuration_type']}")
+
+        num_inliers = result_dict["num_inliers"]
+        inlier_ratio_est_model = num_inliers / match_indices.shape[0]
```
akshay-krishnan, line 184: same here, why not directly use result_dict["num_inliers"]

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
...
+            height = int(cy * 2)
+            logger.info("Estimated height and width: %d,%d", height, width)
+
+            camera_dict = pycolmap.Camera(
+                model="SIMPLE_PINHOLE",
+                width=width,
+                height=height,
+                params=[focal_length, cx, cy],
+            )
+            return camera_dict
```
akshay-krishnan, line 125: could we move this to a util instead of copying the function?

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
+"""
```
akshay-krishnan, line 1: not sure if this file is named appropriately, technically loransac is also being used from colmap. so why not call this gric_verifer?

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: thanks for the review Akshay. Didn't quite understand this comment -- could you explain a bit more?

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: Good point -- I'll rename to `get_pycolmap_camera()`?

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: You're right. I'll rename to `GricVerifier` and `gric_verifier.py`

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: Good point, I'll update

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
...
+        estimation_threshold_px: float,
+    ) -> None:
+        """Initializes the verifier.
+
+        (See https://github.com/mihaidusmanu/pycolmap/blob/master/essential_matrix.cc#L98)
+
+        Args:
+            use_intrinsics_in_verification: Flag to perform keypoint normalization and compute the essential matrix
+                instead of fundamental matrix. This should be preferred when the exact intrinsics are known as opposed
+                to approximating them from exif data.
```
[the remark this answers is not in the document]
johnwlambert: Sure, just updated.

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
```diff
@@ -0,0 +1,195 @@
...
+        camera_intrinsics_i1: Cal3Bundler,
+        camera_intrinsics_i2: Cal3Bundler,
+    ) -> Dict[str, Any]:
+        """Use the pycolmap Pybind wrapper to estimate an Essential matrix using LORANSAC.
+
+        Args:
+            uv_i1: array of shape (N3,2) representing coordinates of 2d points in image 1.
+            uv_i2: array of shape (N3,2) representing corresponding coordinates of 2d points in image 2.
+            camera_intrinsics_i1: intrinsics for image #i1.
+            camera_intrinsics_i2: intrinsics for image #i2.
```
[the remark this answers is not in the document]
johnwlambert: Thanks, updated

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: Good point, updated now

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
akshay-krishnan: The exact same function is also used in the loransac verifier (except for the log statement, which I think is unnecessary). Can we move it to a util file and use in both places?

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
akshay-krishnan: Yeah, please also refer to my other comment on moving it to a util file.

gtsfm/configs/deep_front_end.yaml
```diff
@@ -26,7 +26,7 @@ SceneOptimizer:
         use_outdoor_model: True
 
     verifier:
-      _target_: gtsfm.frontend.verifier.colmap_two_view_geometry_verifier.ColmapTwoViewGeometryVerifier
+      _target_: gtsfm.frontend.verifier.ransac.Ransac
```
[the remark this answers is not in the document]
johnwlambert: We should try it out and analyze it more after we fix some of the determinism issues.

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
+"""
+Wrapper about COLMAP's GRIC estimators, using pycolmap's pybind API.
+Uses LORANSAC for Essential, Fundamental, and Homography matrix estimation.
+
+LORANSAC paper:
+ftp://cmp.felk.cvut.cz/pub/cmp/articles/matas/chum-dagm03.pdf
+
+On Linux and Mac, a python wheel is available:
+https://pypi.org/project/pycolmap/#files
```
ayushbaid, line 9: Can you explicitly mention this verifier is not available on windows?

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
...
+            See https://colmap.github.io/cameras.html#camera-models for info about the COLMAP camera models.
+            Both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length.
+            """
+            focal_length = camera_intrinsics.fx()
+            cx, cy = camera_intrinsics.px(), camera_intrinsics.py()
+
+            # TODO (johnwlambert): use more accurate proxy?
+            width = int(cx * 2)
+            height = int(cy * 2)
+            logger.info("Estimated height and width: %d,%d", height, width)
```
ayushbaid, line 115: Do we need to log this at _info_ level?

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
...
+            dictionary containing result status code, estimated relative pose (R,t), and inlier mask.
+        """
+
+        def get_pycolmap_camera(camera_intrinsics: Cal3Bundler) -> pycolmap.Camera:
+            """Convert Cal3Bundler intrinsics to a pycolmap-compatible format (a dictionary).
+
+            See https://colmap.github.io/cameras.html#camera-models for info about the COLMAP camera models.
+            Both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length.
+            """
+            focal_length = camera_intrinsics.fx()
```
ayushbaid, line 109: Take mean of fx and fy?

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
...
+            uv_i1: array of shape (N3,2) representing coordinates of 2d points in image 1.
+            uv_i2: array of shape (N3,2) representing corresponding coordinates of 2d points in image 2.
+            camera_intrinsics_i1: intrinsics for image i1.
+            camera_intrinsics_i2: intrinsics for image i2.
+
+        Returns:
+            dictionary containing result status code, estimated relative pose (R,t), and inlier mask.
+        """
+
+        def get_pycolmap_camera(camera_intrinsics: Cal3Bundler) -> pycolmap.Camera:
```
ayushbaid, line 103: Private function?

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
...
+            """Convert Cal3Bundler intrinsics to a pycolmap-compatible format (a dictionary).
+
+            See https://colmap.github.io/cameras.html#camera-models for info about the COLMAP camera models.
+            Both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length.
```
ayushbaid, line 107: Why do we need this line in the docstring?

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
...
+            return camera_dict
+
+        camera_dict1 = get_pycolmap_camera(camera_intrinsics_i1)
+        camera_dict2 = get_pycolmap_camera(camera_intrinsics_i2)
+
+        result_dict = pycolmap.two_view_geometry_estimation(
+            points2D1=uv_i1,
+            points2D2=uv_i2,
+            camera1=camera_dict1,
+            camera2=camera_dict2,
```
ayushbaid, line 132: just call the `get_pycolmap_camera` here instead of assigning to local variables? I feel `camera_dict` is not very clear on what it is, and we can just avoid the confusion.

gtsfm/frontend/verifier/gric_verifier.py
```diff
@@ -0,0 +1,191 @@
...
+
+        (See https://github.com/mihaidusmanu/pycolmap/blob/master/essential_matrix.cc#L98)
+
+        Args:
+            use_intrinsics_in_verification: Flag to perform keypoint normalization and compute the essential matrix
+                instead of fundamental matrix. This should be preferred when the exact intrinsics are known.
+            estimation_threshold_px: maximum distance (in pixels) to consider a match an inlier, under squared
+                Sampson distance.
+        """
+        self._use_intrinsics_in_verification = use_intrinsics_in_verification
```
ayushbaid, line 73: Are we switching between E/F estimation?

ayushbaid (approved):
LGTM except minor comments. Can you add unit tests for GRIC too?

gtsfm/frontend/verifier/gric_verifier.py
johnwlambert: Changed to `debug` level now, thanks

gtsfm/frontend/verifier/gric_verifier.py
johnwlambert: Sure, just added an inline comment with that info.

gtsfm/frontend/verifier/gric_verifier.py
johnwlambert: Cal3Bundler only has a single common focal length (https://github.com/borglab/gtsam/blob/develop/gtsam/geometry/Cal3Bundler.h#L61), so we don't know fy

gtsfm/frontend/verifier/gric_verifier.py
johnwlambert: Good idea, thanks. Updated now.

gtsfm/frontend/verifier/gric_verifier.py
johnwlambert: GRIC automatically checks E. vs. F vs. H

johnwlambert: (I just added a comment to explain that better)

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: Thanks for the review @akshay-krishnan. I wasn't sure which util file would be appropriate for this. I think may be ok as is. Is there any file you think is a good place for it? Not sure if it warrants its own util file.

travisdriver:
> Dashboard: [visual_comparison_dashboard 2.html.zip](https://github.com/borglab/gtsfm/files/7805061/visual_comparison_dashboard.2.html.zip)

Great PR! Could you elaborate a bit more about what's being compared in this dashboard?

johnwlambert:
> > Dashboard: [visual_comparison_dashboard 2.html.zip](https://github.com/borglab/gtsfm/files/7805061/visual_comparison_dashboard.2.html.zip)
> 
> Great PR! Could you elaborate a bit more about what's being compared in this dashboard?

Hi @travisdriver, this is comparing the previous master w/ this branch

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
akshay-krishnan: maybe create pycolmap_utils if we are planning on using it longer term? I would not have a problem with it being here if it was not in another file already. 
It is detail that can be removed from the verifier + it is duplicated code, so I think it should be moved to a util. It would also be okay to put it in the verification utils file.

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: Should be fixed now.

gtsfm/frontend/verifier/colmap_two_view_geometry_verifier.py
johnwlambert: Updated now.

gtsfm/frontend/verifier/gric_verifier.py
johnwlambert: Just as an analogy between GTSAM camera type and corresponding COLMAP camera type.

gtsfm/utils/pycolmap_utils.py
```diff
@@ -0,0 +1,29 @@
...
+
+    See https://colmap.github.io/cameras.html#camera-models for info about the COLMAP camera models.
+    Both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length.
+    """
+    focal_length = camera_intrinsics.fx()
+    cx, cy = camera_intrinsics.px(), camera_intrinsics.py()
+
+    # TODO (johnwlambert): use more accurate proxy?
+    width = int(cx * 2)
+    height = int(cy * 2)
```
akshay-krishnan, line 21: I had not noticed this: will this not affect results? Maybe you need more than the Cal3Bundler here?

johnwlambert: Currently we always set cx,cy as half W,H so it’s identical. We don’t pass the image shape to the verifier, so we can’t read it off directly

akshay-krishnan: you mean the loader sets it as half the image size? do we not have the principal point in the intrinsics for some datasets?

akshay-krishnan (approved):

gtsfm/utils/pycolmap_utils.py
johnwlambert: @akshay-krishnan Good point. That's true that for some datasets we have intrinsics (including principal point) from another source, so image height, image width would not necessarily be 2 * cy, 2 * cx.

However, these values here are just dummy values (image width and image height aren't used anywhere in the F / E / H estimation; rather cx and cy are used in the essential matrix estimation here: https://github.com/colmap/colmap/blob/9f3a75ae9c72188244f2403eb085e51ecf4397a8/src/base/camera_models.h#L629)

So we could technically put any values there. I'll add a comment explaining that.

dellaert:
Dashboard impact?

PR #418, opened 2022-01-04
Increase 1DSfM projections

travisdriver:
Increases the number of random projections in 1DSfM (originally 200) in an attempt to decrease entropy in the final solution.


akshay-krishnan:
Do you find any decrease in randomness? I don't think it changes 1dsfm output. Maybe 1 or 2 edges may flip from outlier to inlier or vice versa in the best case. Ive experimented with this before. 200 is already higher than the number used by 1dsfm authors (50).

akshay-krishnan:
If we want to reduce entropy we must first find out which module has outputs with high variance, i.e run the pipeline several times on the same input and measure the variance of each module's output. I suspect this is the frontend, not the backend.

travisdriver:
@akshay-krishnan From what I've gathered, the entropy seems to come from the averaging module. Here is the comparison between two subsequent commits (https://github.com/borglab/gtsfm/actions/runs/1598155743 and https://github.com/borglab/gtsfm/actions/runs/1603167384) where the only change was fixing a broken unit test:

[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/7848858/visual_comparison_dashboard.html.zip)

There are no changes in any of the metrics until we get to averaging (and specifically 1DSfM), and then there are some pretty significant discrepancies between the two. In this case, the results for Skydio are much worse than the reference commit.

newplot

The random projections was the only place I saw that could result in this amount of entropy between runs, but let me know if you have any other ideas for what might be causing this.

dellaert:
I’m not following this closely, but I remember from the previous meeting, the last one in December, that indeed 1D SFM introduce variability. I’m glad it’s being investigated.

Not sure the question about whether this increase to 2000 decreased variability was answered. It would be good to provide evidence on that in the PR comment.

dellaert:
So, @travisdriver, does this increase to 2000 decrease variability? It would be good to provide evidence on that in the PR comment.

travisdriver:
> So, @travisdriver, does this increase to 2000 decrease variability? It would be good to provide evidence on that in the PR comment.

I will need to run the benchmarks workflow a couple of more times before I can conclude anything about variability between runs. I will update the PR comment when I gather more evidence and change this PR to a draft in the meantime.

akshay-krishnan:
@travisdriver I think you mentioned offline that this does not help. Can we close the PR? We can discussing switching to alternate sampling methods in #427

travisdriver:
> @travisdriver I think you mentioned offline that this does not help. Can we close the PR? We can discussing switching to alternate sampling methods in #427 

Yes, it does not help with performance, but does seem to significantly decrease entropy. Let me post of benchmarks that demonstrate the decreased entropy for documentation, and then I will close the PR.

johnwlambert:
Hi @travisdriver, can we go ahead and close this PR?

PR #419, opened 2022-01-05
moving open3d to pip in the Mac environment file

akshay-krishnan:
The conda forge version does not support ARM M1 Macs. https://github.com/isl-org/Open3D/issues/2675#issuecomment-996530610


johnwlambert (approved):
thanks!

PR #424, opened 2022-01-17
Add partial dataset support to COLMAP loader

johnwlambert:
Previously, if some images from a COLMAP dataset were missing, the image path would not be added to the list of image paths. This meant that image paths and calibration lists could become out of sync. This PR fixes this, allowing us to work with any subset of a COLMAP dataset, e.g. a triplet of images, and obtain the correct GT calibration info.

### DIFF
--- a/gtsfm/loader/colmap_loader.py
+++ b/gtsfm/loader/colmap_loader.py
@@ -70,7 +70,7 @@ class ColmapLoader(LoaderBase):
         self._use_gt_extrinsics = use_gt_extrinsics
         self._max_frame_lookahead = max_frame_lookahead
 
-        self._wTi_list, img_fnames = io_utils.read_images_txt(fpath=os.path.join(colmap_files_dirpath, "images.txt"))
+        wTi_list, img_fnames = io_utils.read_images_txt(fpath=os.path.join(colmap_files_dirpath, "images.txt"))
         self._calibrations = io_utils.read_cameras_txt(fpath=os.path.join(colmap_files_dirpath, "cameras.txt"))
 
         # TODO in future PR: if img_fnames is None, default to using everything inside image directory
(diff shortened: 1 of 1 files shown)


gtsfm/loader/colmap_loader.py
```diff
@@ -83,16 +83,32 @@ def __init__(
...
             if not Path(img_fpath).exists():
                 continue
+            self._img_fnames.append(img_fname)
             self._image_paths.append(img_fpath)
+            self._wTi_list.append(wTi)
 
         self._num_imgs = len(self._image_paths)
         logger.info("Colmap image loader found and loaded %d images", self._num_imgs)
 
+    def get_image_fname(self, idx: int) -> str:
```
akshay-krishnan, line 104: unused function?

gtsfm/loader/colmap_loader.py
```diff
@@ -83,16 +83,32 @@ def __init__(
...
+        """Given an image index, provide the corresponding image filename."""
+        return Path(self._image_paths[idx]).name
+
+    def get_image_index_from_filename(self, fname: str) -> int:
```
akshay-krishnan, line 108: unused function?

akshay-krishnan (commented):
do we need the two additional functions? Maybe we should add them when we need them.

gtsfm/loader/colmap_loader.py
johnwlambert: Thanks for taking a look @akshay-krishnan. I forgot to mention I'm using these for testing in another branch.

gtsfm/loader/colmap_loader.py
johnwlambert: makes it easy to get poses given an image filename (and vice versa)

akshay-krishnan (approved):

PR #425, opened 2022-01-18
Adding multiple sampling options in TA

akshay-krishnan:
This PR adds all the options to sample projection directions in translation averaging as an Enum. This allows us to easily experiment with different options.

### DIFF
--- a/gtsfm/averaging/translation/averaging_1dsfm.py
+++ b/gtsfm/averaging/translation/averaging_1dsfm.py
@@ -12,7 +12,7 @@ Authors: Jing Wu, Ayush Baid, Akshay Krishnan
 """
 import random
 from collections import defaultdict
-from enum import Enum
+from enum import auto, Enum
 from typing import Dict, List, Optional, Tuple
 
 import gtsam
@@ -46,9 +46,9 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
 
     class ProjectionSamplingMethod(Enum):
         """Used to select how the projection directions in 1DSfM are sampled."""
-        SAMPLE_INPUT_MEASUREMENTS = 1   # Randomly choose projection directions from input measurements.
-        SAMPLE_WITH_INPUT_DENSITY = 2   # Fit a Gaussian density to input measurements and sample from it.
-        SAMPLE_WITH_UNIFORM_DENSITY = 3 # Uniformly sample 3D directions at random.
+        SAMPLE_INPUT_MEASUREMENTS = auto()   # Randomly choose projection directions from input measurements.
+        SAMPLE_WITH_INPUT_DENSITY = auto()   # Fit a Gaussian density to input measurements and sample from it.
+        SAMPLE_WITH_UNIFORM_DENSITY = auto() # Uniformly sample 3D directions at random.
 
     def __init__(self, robust_measurement_noise: bool = True) -> None:
         """Initializes the 1DSFM averaging instance.
@@ -82,7 +82,7 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
             sampled_indices = np.random.choice(w_i2Ui1_measurements, num_samples, replace=False)
             return [w_i2Ui1_measurements[idx].measured() for idx in sampled_indices]
 
-        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_INPUT_DENSITY:\
+        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_INPUT_DENSITY:
             return _sample_kde_directions(w_i2Ui1_measurements, num_samples=self._max_1dsfm_projection_directions)
 
         elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_UNIFORM_DENSITY:
@@ -108,6 +108,7 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
             wRi_list: global rotations for each camera pose in the world coordinates.
             scale_factor: non-negative global scaling factor.
             gt_wTi_list: ground truth poses for computing metrics.
+            projection_sampling_method: ProjectionSamplingMethod to be used for directions to run 1DSfM.
 
         Returns:
             Global translation wti for each camera pose. The number of entries in the list is `num_images`. The list
@@ -115,7 +116,6 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
                 or ill-constrained system).
             A GtsfmMetricsGroup of 1DSfM metrics.
         """
-        logger.info("Running translation averaging")
         noise_model = gtsam.noiseModel.Isotropic.Sigma(NOISE_MODEL_DIMENSION, NOISE_MODEL_SIGMA)
         if self._robust_measurement_noise:
             huber_loss = gtsam.noiseModel.mEstimator.Huber.Create(HUBER_LOSS_K)
@@ -148,6 +148,7 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
         # compute average outlier weight
         avg_outlier_weights = defaultdict(float)
         for outlier_weight_dict in outlier_weights:
+            # TODO(akshay-krishnan): use keys from outlier weight dict once we can iterate over it (gtsam fix).
             for index_pair in valid_i2_i1:
                 avg_outlier_weights[index_pair] += outlier_weight_dict[index_pair]
 
@@ -200,7 +201,8 @@ def _sample_kde_directions(w_i2Ui1_measurements: BinaryMeasurementsUnit3, num_sa
      """
     w_i2Ui1_list = [w_i2Ui1.measured() for w_i2Ui1 in w_i2Ui1_measurements]
     if len(w_i2Ui1_list) > MAX_KDE_SAMPLES:
-        w_i2Ui1_list = [w_i2Ui1_list[i] for i in random.sample(range(len(w_i2Ui1_list)), MAX_KDE_SAMPLES)]
+        w_i2Ui1_subset_indices = np.random.choice(range(len(w_i2Ui1_list)), MAX_KDE_SAMPLES, replace=False).tolist()
+        w_i2Ui1_list = [w_i2Ui1_list[i] for i in w_i2Ui1_subset_indices]
 
     w_i2Ui1_spherical = conversion_utils.cartesian_to_spherical_directions(w_i2Ui1_list)
 
--- a/gtsfm/averaging/translation/averaging_1dsfm.py
+++ b/gtsfm/averaging/translation/averaging_1dsfm.py
@@ -49,46 +49,52 @@ class TranslationAveraging1DSFM(TranslationAveragingBase):
         SAMPLE_WITH_INPUT_DENSITY = auto()   # Fit a Gaussian density to input measurements and sample from it.
         SAMPLE_WITH_UNIFORM_DENSITY = auto() # Uniformly sample 3D directions at random.
 
-    def __init__(self, robust_measurement_noise: bool = True) -> None:
+    def __init__(
+        self, 
+        robust_measurement_noise: bool = True,
+        projection_sampling_method: ProjectionSamplingMethod = ProjectionSamplingMethod.SAMPLE_WITH_UNIFORM_DENSITY
+    ) -> None:
         """Initializes the 1DSFM averaging instance.
 
         Args:
             robust_measurement_noise: Whether to use a robust noise model for the measurements, defaults to true.
+            projection_sampling_method: ProjectionSamplingMethod to be used for directions to run 1DSfM.
         """
         super().__init__(robust_measurement_noise)
 
         self._max_1dsfm_projection_directions = MAX_PROJECTION_DIRECTIONS
         self._outlier_weight_threshold = OUTLIER_WEIGHT_THRESHOLD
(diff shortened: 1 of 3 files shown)


gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -10,33 +10,46 @@
...
 HUBER_LOSS_K = 1.345  # default value from GTSAM
 
 MAX_INLIER_MEASUREMENT_ERROR_DEG = 5.0
 
+logger = logger_utils.get_logger()
 
 class TranslationAveraging1DSFM(TranslationAveragingBase):
     """1D-SFM translation averaging with outlier rejection."""
 
+    class ProjectionSamplingMethod(Enum):
```
johnwlambert, line 47: very nice. I like it. would a string enum value be a little clearer? maybe not. or is `auto()` simpler?
```python
class ProjectionSamplingMethod(str, Enum):
     SAMPLE_INPUT_MEASUREMENTS: str = "SAMPLE_INPUT_MEASUREMENTS"
```

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -139,6 +188,28 @@ def run(
...
+    Args:
+        w_i2Ui1_measurements: List of BinaryMeasurementUnit3 direction measurements.
+        num_samples: Number of samples to be sampled from the kernel.
+
+    Returns:
+        List of sampled Unit3 directions.
+     """
+    w_i2Ui1_list = [w_i2Ui1.measured() for w_i2Ui1 in w_i2Ui1_measurements]
+    if len(w_i2Ui1_list) > MAX_KDE_SAMPLES:
+        w_i2Ui1_list = [w_i2Ui1_list[i] for i in random.sample(range(len(w_i2Ui1_list)), MAX_KDE_SAMPLES)]
```
johnwlambert, line 203: maybe cleaner to only use random number generator from numpy instead of `random` and `numpy`.

could we use
```python
random.choice(a=w_i2Ui1_list, size=MAX_KDE_SAMPLES, replace=False)
```

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -48,13 +61,44 @@ def __init__(self, robust_measurement_noise: bool = True) -> None:
...
+            List of sampled Unit3 projection directions.
+        """
+        num_measurements = len(w_i2Ui1_measurements)
+
+        if projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_INPUT_MEASUREMENTS:
+            num_samples = min(num_measurements, self._max_1dsfm_projection_directions)
+            sampled_indices = np.random.choice(w_i2Ui1_measurements, num_samples, replace=False)
+            return [w_i2Ui1_measurements[idx].measured() for idx in sampled_indices]
+
+        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_INPUT_DENSITY:\
```
johnwlambert, line 85: @akshay-krishnan is there a stray backslash here?

johnwlambert (approved):
LGTM, pending what the CI says for the dashboard, and some minor comments i left

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -48,13 +61,44 @@ def __init__(self, robust_measurement_noise: bool = True) -> None:
...
+            raise ValueError("Unsupported sampling method!")
+
     def run(
         self,
         num_images: int,
         i2Ui1_dict: Dict[Tuple[int, int], Optional[Unit3]],
         wRi_list: List[Optional[Rot3]],
         scale_factor: float = 1.0,
         gt_wTi_list: Optional[List[Optional[Pose3]]] = None,
+        projection_sampling_method: ProjectionSamplingMethod = ProjectionSamplingMethod.SAMPLE_WITH_UNIFORM_DENSITY,
```
ayushbaid, line 101: docstring change here?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -71,6 +115,7 @@ def run(
                 or ill-constrained system).
             A GtsfmMetricsGroup of 1DSfM metrics.
         """
+        logger.info("Running translation averaging")
```
ayushbaid, line 118: Is there a reason for adding this?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -81,14 +126,16 @@ def run(
 
         # convert translation direction in global frame using rotations.
         w_i2Ui1_measurements = BinaryMeasurementsUnit3()
+        valid_i2_i1 = []
```
ayushbaid, line 129: Can we rename to something more expressive? `nonnull_global_rotations_i2i1_indices`? Or is it too long?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -99,20 +146,22 @@ def run(
...
-            for index_pair, weight in outlier_weight_dict.items():
-                if index_pair in avg_outlier_weights:
-                    avg_outlier_weights[index_pair] += weight / len(outlier_weights)
-                else:
-                    avg_outlier_weights[index_pair] = weight / len(outlier_weights)
+            for index_pair in valid_i2_i1:
+                avg_outlier_weights[index_pair] += outlier_weight_dict[index_pair]
+
+        for index_pair in avg_outlier_weights:
+            avg_outlier_weights[index_pair] /= len(projection_directions)
```
ayushbaid, line 155: Can we move this logic to a new private function?

ayushbaid (approved):

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: I think auto is better. A string is probably inefficient, I dont see a use for it.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: Im keeping it as is for now, the API for that method will likely change, using valid_i1_i2 is a (hopefully temporary) workaround. Im not able to iterate on some gtsam structures like BinaryMeasurementsUnit3 in the M1 python build.

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: I think maybe a bit too long

akshay-krishnan: Its too long :O 
i1_i2 -> means indices, just like i2Ri1 means a rotation (also not very expressive IMO). I dont mind changing it, but I cannot think of a shorter and better name. Besides, it might be removed once we fix the gtsam issue I mentioned in another comment.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: was debugging locally, but I realized we need better logging statements. I think Ill create a PR where I just add and remove some logging statements across the pipeline.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: ah dang, why did i not see that

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -48,13 +62,42 @@ def __init__(self, robust_measurement_noise: bool = True) -> None:
...
+            return [w_i2Ui1_measurements[idx].measured() for idx in sampled_indices]
+
+        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_INPUT_DENSITY:
+            return _sample_kde_directions(w_i2Ui1_measurements, num_samples=self._max_1dsfm_projection_directions)
+
+        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_UNIFORM_DENSITY:
+            return _sample_random_directions(num_samples=self._max_1dsfm_projection_directions)
+
+        else:
+            raise ValueError("Unsupported sampling method!")
```
travisdriver, line 91: Is this if-else statement unnecessary since there are return statements?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -139,6 +188,29 @@ def run(
         return wti_list, ta_metrics
 
 
+def _sample_kde_directions(w_i2Ui1_measurements: BinaryMeasurementsUnit3, num_samples: int) -> List[Unit3]:
+    """Fits a Gaussian density kernel to the provided measurements, and then samples num_samples from this kernel.
```
travisdriver, line 192: Is there a reference for this?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -48,13 +62,42 @@ def __init__(self, robust_measurement_noise: bool = True) -> None:
...
+        if projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_INPUT_MEASUREMENTS:
+            num_samples = min(num_measurements, self._max_1dsfm_projection_directions)
+            sampled_indices = np.random.choice(w_i2Ui1_measurements, num_samples, replace=False)
+            return [w_i2Ui1_measurements[idx].measured() for idx in sampled_indices]
+
+        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_INPUT_DENSITY:
+            return _sample_kde_directions(w_i2Ui1_measurements, num_samples=self._max_1dsfm_projection_directions)
+
+        elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_UNIFORM_DENSITY:
+            return _sample_random_directions(num_samples=self._max_1dsfm_projection_directions)
```
travisdriver, line 88: Add these options to the config?

akshay-krishnan: I would not want to, because for that we will need the parameter in the constructor of the base class, and this is very implementation specific (should not go in the base class). 
Unless there is a way in the config to provide options for derived classes.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: not really, doesnt hurt either, I can change later. I thought it would be better to place the exception inside an `else` instead of just default.

gtsfm/averaging/translation/averaging_1dsfm.py
travisdriver: You could just make it an attribute of the derived class, similar to how you did it for the [`CycleConsistentRotationViewGraphEstimator`](https://github.com/borglab/gtsfm/blob/3eaa76e6439f9a4f6e8565bb0c42b8ce92fcdb13/gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py#L47).

gtsfm/averaging/translation/averaging_1dsfm.py
travisdriver: Maybe define a variable and set it in the if-else statement and return at the end?

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: yeah i think it might be nice to put that in the constructor as default

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: im ok with it as it is, kind of nice to catch the undefined types

akshay-krishnan: I think both are good ways. I am not too sure which one to choose. 
I prefer returning immediately since there is no more relevant code for that branch. The elif ... else.. makes it explicit and slightly easier to understand at first glance. So I'll keep it as is.

travisdriver: I think reformatting to the following is preferable to avoid pylint warnings:

```python
if projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_INPUT_MEASUREMENTS:
    num_samples = min(num_measurements, self._max_1dsfm_projection_directions)
    sampled_indices = np.random.choice(w_i2Ui1_measurements, num_samples, replace=False)
    projections = [w_i2Ui1_measurements[idx].measured() for idx in sampled_indices]
elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_INPUT_DENSITY:
    projections = _sample_kde_directions(w_i2Ui1_measurements, num_samples=self._max_1dsfm_projection_directions)
elif projection_sampling_method == self.ProjectionSamplingMethod.SAMPLE_WITH_UNIFORM_DENSITY:
    projections = _sample_random_directions(num_samples=self._max_1dsfm_projection_directions)
else:
    raise ValueError("Unsupported sampling method!")

return projections
```

akshay-krishnan: done.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: updated

PR #426, opened 2022-01-18
vectorized triangulation angle computation

johnwlambert:
Used in ViewGraphEstimators.

### DIFF
--- a/gtsfm/densify/mvs_utils.py
+++ b/gtsfm/densify/mvs_utils.py
@@ -26,6 +26,7 @@ def calculate_triangulation_angle_in_degrees(
     C1      C2
     References:
     - https://github.com/colmap/colmap/blob/dev/src/base/triangulation.cc#L122
+    
     Args:
         camera_1: the first camera.
         camera_2: the second camera.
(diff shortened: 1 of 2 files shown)


gtsfm/densify/mvs_utils.py
```diff
@@ -44,6 +45,49 @@ def calculate_triangulation_angle_in_degrees(
     return geometry_utils.compute_relative_unit_translation_angle(Unit3(ray_1), Unit3(ray_2))
 
 
+def calculate_triangulation_angles_in_degrees(
```
ayushbaid, line 48: Do we need the non-vectorized function anywhere?

gtsfm/densify/mvs_utils.py
```diff
@@ -44,6 +45,49 @@ def calculate_triangulation_angle_in_degrees(
...
+    N = points_3d.shape[0]
+    # ensure broadcasting is in the correct direction
+    rays1 = points_3d - camera_center_1.reshape(1, 3)
+    rays2 = points_3d - camera_center_2.reshape(1, 3)
+
+    # normalize rays to unit length
+    rays1 /= np.linalg.norm(rays1, axis=1).reshape(N, 1)
+    rays2 /= np.linalg.norm(rays2, axis=1).reshape(N, 1)
+
+    dot_products = np.multiply(rays1, rays2).sum(axis=1)
```
ayushbaid, line 84: Why not use `np.dot`?

ayushbaid (approved):
LGTM except minor comments

gtsfm/densify/mvs_utils.py
johnwlambert: Thanks for taking a look @ayushbaid. I'm using it in another branch, and wanted to keep the other branch lean. COLMAP relies pretty heavily on minimum triangulation angles -- we can discuss more tomorrow.

gtsfm/densify/mvs_utils.py
johnwlambert: rays1 and rays2 are (N,3) and (N,3), so if we do `rays1 @ rays2.T` we get a `N x N` matrix, and only the diagonal is useful. This way we keep things efficient.

PR #438, opened 2022-01-26
Update 1dsfm sampling type

akshay-krishnan:
This PR updates the sampling strategy in 1dsfm to reduce the randomness in the outlier rejection step.


akshay-krishnan:
Translation averaging metrics after this change: 
newplot

akshay-krishnan:
BA metrics:
newplot (1)

akshay-krishnan:
This change is required to ensure determinism of translation averaging but looks like it produces regressions in TA and BA metrics, particularly on the skydio32 dataset. Not sure if we should proceed with this.

travisdriver:
> This change is required to ensure determinism of translation averaging but looks like it produces regressions in TA and BA metrics, particularly on the skydio32 dataset. Not sure if we should proceed with this.


Have you tried tuning the outlier threshold? According to the dashboard, it seems that 1DSfM is accepting more (potentially outlier) measurements than before. Maybe we should make the threshold more restrictive?

johnwlambert:
Good points Travis. It looks like maybe the regression is significant enough that we should investigate a bit more.

Would the KDE be more deterministic?

akshay-krishnan:
> Have you tried tuning the outlier threshold? According to the dashboard, it seems that 1DSfM is accepting more (potentially outlier) measurements than before. Maybe we should make the threshold more restrictive?

That is only the case for SIFT-skydio32. For all the others it is rejecting more measurements (note the increase in outliers and the drop in recall).

akshay-krishnan:
> Good points Travis. It looks like maybe the regression is significant enough that we should investigate a bit more.
> 
> Would the KDE be more deterministic?

KDE also does not pass @ayushbaid 's reproducibility tests.

travisdriver:
> > Have you tried tuning the outlier threshold? According to the dashboard, it seems that 1DSfM is accepting more (potentially outlier) measurements than before. Maybe we should make the threshold more restrictive?
> 
> That is only the case for SIFT-skydio32. For all the others it is rejecting more measurements (note the increase in outliers and the drop in recall).

Yeah sorry I meant the other way around. You should try _relaxing_ the threshold.

akshay-krishnan:
I found slightly better results with a threshold of 0.125, but this is still a regression compared to master. 

translation averaging metrics: 
newplot (2)

akshay-krishnan:
bundle adjustment metrics
newplot (3)

akshay-krishnan:
I am also not too sure about these numbers, looking at the result poses for skydio-32, they look bad both on master as well as the new branch..

akshay-krishnan:
@johnwlambert  @travisdriver  please let me know if the PR looks good to be merged. was there anything else to be checked?

ayushbaid (approved):
Palace-of-fine-arts dataset has a huge improvement. Any speculation on why that might be the case?

akshay-krishnan:
I would not consider that one, the frontend outputs between baseline and test are not the same.

travisdriver (approved):
LGTM!

PR #441, opened 2022-01-29
Update docstring about image height,width approximation from principal point

johnwlambert:
Update docstring about image height,width approximation from principal point.

### DIFF
--- a/gtsfm/utils/pycolmap_utils.py
+++ b/gtsfm/utils/pycolmap_utils.py
@@ -15,8 +15,8 @@ def get_pycolmap_camera(camera_intrinsics: Cal3Bundler) -> pycolmap.Camera:
     
     Note: the image width and image height values approximated below are dummy placeholder values.
     For some datasets we have intrinsics (including principal point) where image height, image width
-    would not necessarily be 2 * cy, 2 * cx. However, image width and image height aren't used anywhere
-    in the F / E / H estimation; rather cx and cy are used in the essential matrix estimation:
+    would not necessarily be 2 * cy, 2 * cx. However, image dimensions aren't used anywhere
+    in the F / E / H estimation; rather cx and cy are used in the Essential matrix estimation:
     https://github.com/colmap/colmap/blob/9f3a75ae9c72188244f2403eb085e51ecf4397a8/src/base/camera_models.h#L629)
     
     Args:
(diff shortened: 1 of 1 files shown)


gtsfm/utils/pycolmap_utils.py
```diff
@@ -12,11 +12,19 @@ def get_pycolmap_camera(camera_intrinsics: Cal3Bundler) -> pycolmap.Camera:
 
     See https://colmap.github.io/cameras.html#camera-models for info about the COLMAP camera models.
     Both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length.
+    
+    Note: the image width and image height values approximated below are dummy placeholder values.
+    For some datasets we have intrinsics (including principal point) where image height, image width
+    would not necessarily be 2 * cy, 2 * cx. However, image width and image height aren't used anywhere
+    in the F / E / H estimation; rather cx and cy are used in the essential matrix estimation:
+    https://github.com/colmap/colmap/blob/9f3a75ae9c72188244f2403eb085e51ecf4397a8/src/base/camera_models.h#L629)
```
akshay-krishnan, line 20: so datasets without cx and cy dont support E matrix estimation? do we check for that?

gtsfm/utils/pycolmap_utils.py
```diff
@@ -12,11 +12,19 @@ def get_pycolmap_camera(camera_intrinsics: Cal3Bundler) -> pycolmap.Camera:
 
     See https://colmap.github.io/cameras.html#camera-models for info about the COLMAP camera models.
     Both SIMPLE_PINHOLE and SIMPLE_RADIAL use 1 focal length.
+    
+    Note: the image width and image height values approximated below are dummy placeholder values.
+    For some datasets we have intrinsics (including principal point) where image height, image width
+    would not necessarily be 2 * cy, 2 * cx. However, image width and image height aren't used anywhere
```
akshay-krishnan, line 18: nit: might be a bit too verbose. you may also consider using "image dimensions" instead of "image height and image width"

akshay-krishnan (approved):

gtsfm/utils/pycolmap_utils.py
johnwlambert: We have to approximate cx,cy in this case as w/2,h/2 in this case.

PR #442, opened 2022-01-29
Fix Olsson dataset loading bug for Palace of Fine Arts dataset.

johnwlambert:
Previously, we assumed the 0th pose was always identity, and this was not true, we derived incorrect intrinsic matrices.

The loader now uses a proper decomposition of the GT camera projection matrix.


gtsfm/configs/deep_front_end.yaml
```diff
@@ -26,7 +26,7 @@ SceneOptimizer:
         use_outdoor_model: True
 
     verifier:
-      _target_: gtsfm.frontend.verifier.ransac.Ransac
+      _target_: gtsfm.frontend.verifier.gric_verifier.GricVerifier
```
johnwlambert, line 29: this is another change I'm validating; I'll provide dashboard results when they finish.

gtsfm/loader/loader_base.py
```diff
@@ -183,6 +183,13 @@ def get_camera_intrinsics(self, index: int) -> Cal3Bundler:
             u0=intrinsics_full_res.px() * scale_u,
             v0=intrinsics_full_res.py() * scale_v,
         )
+
+        if rescaled_intrinsics.fx() <= 0:
+            raise RuntimeError("Focal length must be positive.")
+
+        if rescaled_intrinsics.px() <= 0 or rescaled_intrinsics.py() <= 0:
+            raise RuntimeError("Principal point must have positive coordinates.")
```
akshay-krishnan, line 191: I think these checks should happen above, for intrinsics_full_res - and `get_camera_intrinsics_full_res` would be the culprit.

gtsfm/loader/olsson_loader.py
```diff
@@ -76,20 +81,25 @@ def __init__(
 
         # stores camera poses (extrinsics) and intrinsics as 3x4 projection matrices
         # 'P' array will have shape (1,num_imgs), and each element will be a (3,4) matrix
-        data = loadmat(cam_matrices_fpath)
+        data = scipy.io.loadmat(cam_matrices_fpath)
+
+        if len(data["P"][0]) != self._num_imgs:
+            raise RuntimeError("Found more or less images than ground truth provides.")
```
akshay-krishnan, line 87: I think "Number of images not equal to ground truth (images?)" would be a better (less confusing) message.

gtsfm/loader/loader_base.py
johnwlambert: good catch, will update

gtsfm/visualization/view_scene_olsson.py
```diff
@@ -0,0 +1,81 @@
+"""
+Script to render a GTSFM scene using Open3d.
```
akshay-krishnan, line 2: sorry I did not see this file earlier. how is it different from existing visualizers?

gtsfm/visualization/view_scene_olsson.py
```diff
@@ -0,0 +1,81 @@
...
+def view_scene(args: argparse.Namespace) -> None:
+    """Read Olsson Dataset ground truth from a data.mat file and render the scene to the GUI.
+
+    Args:
+        args: rendering options.
+    """
+    loader = OlssonLoader(
+        args.dataset_root,
+        image_extension=args.image_extension,
+        max_frame_lookahead=args.max_frame_lookahead,
```
akshay-krishnan, line 30: why is max_frame_lookahead relevant here?

gtsfm/loader/olsson_loader.py
johnwlambert: Good point. Updated now.

gtsfm/visualization/view_scene_olsson.py
johnwlambert: You’re right, it’s not relevant, but it’s a required arg to the loader. Now I made it a global dummy constant value

gtsfm/loader/loader_base.py
johnwlambert: Updated now

gtsfm/loader/olsson_loader.py
```diff
@@ -76,20 +81,25 @@ def __init__(
...
         # 'P' array will have shape (1,num_imgs), and each element will be a (3,4) matrix
-        data = loadmat(cam_matrices_fpath)
+        data = scipy.io.loadmat(cam_matrices_fpath)
+
+        if len(data["P"][0]) != self._num_imgs:
+            raise RuntimeError("Number of images found on disk not equal to number of ground truth images.")
 
         # M = K [R | t]
         # in GTSAM notation, M = K @ cTw
         M_list = [data["P"][0][i] for i in range(self._num_imgs)]
```
ayushbaid, line 91: Can we rename this to `projection_matrices`? For readable imo

gtsfm/loader/olsson_loader.py
```diff
@@ -76,20 +81,25 @@ def __init__(
...
+        for M in M_list:
+            K, wTc = verification_utils.decompose_camera_projection_matrix(M)
+            self._wTi_list.append(wTc)
 
-        # decode camera poses as:
-        #    K^{-1} @ M = cTw
-        iTw_list = [Kinv @ M_list[i] for i in range(self._num_imgs)]
-        self._wTi_list = [Pose3(Rot3(iTw[:3, :3]), iTw[:, 3]).inverse() for iTw in iTw_list]
+        # GT 3d structure (point cloud)
+        self._point_cloud = data["U"].T[:, :3]
```
ayushbaid, line 102: Any API to use this in GTSFM right now?

ayushbaid (approved):

gtsfm/visualization/view_scene_olsson.py
akshay-krishnan: could you add more documentation on how this is different?

akshay-krishnan (approved):

gtsfm/loader/loader_base.py
```diff
@@ -171,6 +171,12 @@ def get_camera_intrinsics(self, index: int) -> Cal3Bundler:
         if intrinsics_full_res is None:
             raise ValueError(f"No intrinsics found for index {index}.")
 
+        if intrinsics_full_res.fx() <= 0:
+            raise RuntimeError("Focal length must be positive.")
+
+        if intrinsics_full_res.px() <= 0 or intrinsics_full_res.py() <= 0:
+            raise RuntimeError("Principal point must have positive coordinates.")
+
```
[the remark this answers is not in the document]
johnwlambert: Yeah it would have caught the bug

gtsfm/configs/deep_front_end.yaml
ayushbaid: @johnwlambert I think having this change in a separate PR will enable this to merge quickly. We can also analyze the performance of palace dataset with the correct ground truth before changing the verifier.

gtsfm/loader/olsson_loader.py
johnwlambert: Sure, renamed now.

gtsfm/loader/olsson_loader.py
johnwlambert: Yeah, I'm using in the `gtsfm/visualization/view_scene_olsson.py` file now to access the GT 3d structure.

gtsfm/visualization/view_scene_olsson.py
johnwlambert: Sure, I just re-wrote the docstring at the top of the file.

gtsfm/configs/deep_front_end.yaml
johnwlambert: Good idea. I removed that change from this PR.

gtsfm/visualization/view_scene_olsson.py
```diff
@@ -0,0 +1,78 @@
...
+from gtsfm.loader.olsson_loader import OlssonLoader
+
+
+TEST_DATA_ROOT = Path(__file__).resolve().parent.parent.parent / "tests" / "data"
+
+# The Olsson loader requires we set a value for the frame lookahead, but it will not affect the visualization.
+DUMMY_MAX_FRAME_LOOKAHEAD = 1
+
+
+def view_scene(args: argparse.Namespace) -> None:
```
travisdriver, line 24: Maybe we could make this into a more general module in a future PR? Since this just called the respective dataloader to read in the GT point cloud, maybe we don't need a separate file for each dataset?

PR #446, opened 2022-01-31
Use multi-model fitting check in front-end

johnwlambert:


ayushbaid:
@johnwlambert are we waiting on some code for this or do we need to analyze the metrics?

PR #449, opened 2022-02-02
Increase #Sinkhorn Iters

johnwlambert:


johnwlambert:
[visual_comparison_dashboard.html 2.zip](https://github.com/borglab/gtsfm/files/7988112/visual_comparison_dashboard.html.2.zip)

No clear benefit, for now.

PR #452, opened 2022-02-09
Adding reproducibility tests for point3d initializers

ayushbaid:
Adding the reproducibility tests, with the test being flaky.

The test for no-ransac configuration fails occasionally. Will require digging into the GTSAM code.

### DIFF
--- a/tests/repro_tests/data_association/test_point3dinitializer_reproducibility.py
+++ b/tests/repro_tests/data_association/test_point3dinitializer_reproducibility.py
@@ -27,6 +27,8 @@ GROUND_TRUTH_DATA: GtsfmData = io_utils.read_bal(gtsam.findExampleDataFile(GTSAM
 
 MEASUREMENTS_NOISE_STANDARD_DEVIATION = 1  # in pixels
 
+TRIANGULATION_REPROJ_ERROR_THRESHOLD = 10  # using a high value to account for random noise in input measurements
+
 TRIANGULATION_RESULT_TYPE = Tuple[Optional[SfmTrack], Optional[float], TriangulationExitCode]
 
 
@@ -36,7 +38,7 @@ class TestPoint3dInitializerNoRansac(ReproducibilityTestBase, unittest.TestCase)
     def setUp(
         self,
         triangulation_options: TriangulationOptions = TriangulationOptions(
-            reproj_error_threshold=3, mode=TriangulationSamplingMode.NO_RANSAC
+            reproj_error_threshold=TRIANGULATION_REPROJ_ERROR_THRESHOLD, mode=TriangulationSamplingMode.NO_RANSAC
         ),
     ) -> None:
         camera_dict: Dict[int, PinholeCameraCal3Bundler] = {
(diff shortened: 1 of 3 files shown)


ayushbaid:
> Thanks for working on this, Ayush. Was curious why the threshold affects the reproducibility.

The threshold matters because we were using some random noise in the input measurements. As a result, the threshold turned out to be too strict.

tests/repro_tests/data_association/test_point3dinitializer_reproducibility.py
```diff
@@ -0,0 +1,120 @@
...
+    TriangulationSamplingMode,
+    TriangulationExitCode,
+)
+from tests.repro_tests.test_repro_base import ReproducibilityTestBase
+
+# Load ground truth data with 3 images and 7 tracks.
+GTSAM_EXAMPLE_FILE: str = "dubrovnik-3-7-pre"
+GROUND_TRUTH_DATA: GtsfmData = io_utils.read_bal(gtsam.findExampleDataFile(GTSAM_EXAMPLE_FILE))
+
+MEASUREMENTS_NOISE_STANDARD_DEVIATION = 1  # in pixels
```
johnwlambert, line 28: looks good. `MEASUREMENTS_NOISE_STD_DEV` would be a little more readable imho

tests/repro_tests/data_association/test_point3dinitializer_reproducibility.py
```diff
@@ -0,0 +1,120 @@
...
+
+    def run_once(self) -> TRIANGULATION_RESULT_TYPE:
+        return self._point3d_initializer.triangulate(self._track_2d)
+
+    def assert_results(self, results_a: TRIANGULATION_RESULT_TYPE, results_b: TRIANGULATION_RESULT_TYPE) -> None:
+        self.__assert_track3d(results_a[0], results_b[0])
+        self.assertAlmostEqual(results_a[1], results_b[1])
+        self.assertEqual(results_a[2], results_b[2])
+
+    def __assert_track3d(self, track3d_a: Optional[SfmTrack], track3d_b: Optional[SfmTrack]) -> None:
```
johnwlambert, line 68: the fn name sounds a bit like we are asserting this is a track. i think we also want to assert they're equal tracks, right?
would `__assert_track3d_equal()` make sense here?

johnwlambert: or should we add this on the GTSAM side as a .equals() method?

tests/repro_tests/data_association/test_point3dinitializer_reproducibility.py
```diff
@@ -0,0 +1,120 @@
...
+        triangulation_options: TriangulationOptions = TriangulationOptions(
+            reproj_error_threshold=TRIANGULATION_REPROJ_ERROR_THRESHOLD,
+            mode=TriangulationSamplingMode.RANSAC_SAMPLE_UNIFORM,
+        ),
+    ) -> None:
+        super().setUp(triangulation_options)
+
+        # replace one measurement with an outlier
+        measurements_with_one_outlier = self._track_2d.measurements
+        measurements_with_one_outlier[2] = SfmMeasurement(measurements_with_one_outlier[2].i, np.array([50, 30]))
```
johnwlambert, line 95: maybe an inline comment stating what the correct value was before?

johnwlambert:
thanks for the PR, @ayushbaid.

johnwlambert:
> > Thanks for working on this, Ayush. Was curious why the threshold affects the reproducibility.
> 
> The threshold matters because we were using some random noise in the input measurements. As a result, the threshold turned out to be too strict.

Hmm not sure I follow still. Could you give a brief example here of what you mean?

PR #453, opened 2022-02-09
NetVLAD retriever & GlobalDescriptor classes

johnwlambert:
Skydio loop repeats at image index 35 and 76 (see repeated narrow diagonal bands):

Screen Shot 2022-02-09 at 1 41 31 PM

Screen Shot 2022-02-09 at 1 29 02 PM


gtsfm/retriever/retriever_base.py
```diff
@@ -0,0 +1,27 @@
...
+    def run(loader: LoaderBase, num_images: int, num_matched: int = 2) -> List[Tuple[int, int]]:
+        """
+        Args:
+            loader: image loader.
+            num_images: total number of images for exhaustive global descriptor matching.
+            num_matched: number of K potential matches to provide per query. These are the top "K" matches per query.
+
+        Return:
+            pair_indices: (i1,i2) image pairs.
+        """
```
travisdriver, line 27: Should there be a `create_computation_graph()` method to allow for delayed computation like the other frontend modules? Otherwise, I believe it will create an explicit bottleneck in the pipeline.

johnwlambert: Hi @travisdriver, thanks for taking a look. I believe we would only be able to return a single `Delayed` object, which would not be iterable. Since the retrieval list is the result of a single algorithm call, not separate calls.

So I don't think Dask will work here, at least in our current paradigm.

travisdriver: You can address this by returning a dictionary where the pair indices are the keys and the values are the delayed calls to the retriever. This is what's done in #393.

johnwlambert: Thanks for the reference Travis. In this scenario, if we aren't validating many potential pairs (each validation as a Delayed object), I think we will have an issue. Because we will be returning a single Delayed object (result of 1 algo), and we can't iterate over that. Let me know if I'm misunderstanding, though.

travisdriver: You would return `Dict[Tuple[int, int], Delayed]`, which you can iterate over. You could then pass this delayed object to the TwoViewEstimator to be computed when Dask seems fit.

johnwlambert: Correct me if I'm wrong, but if we generate a NxN similarity matrix, and then select the top entries in each row, we would not be able to wrap the selection for each (i1,i2) pair in Dask's Delayed. This is because generating the entire list is one big Delayed object, and we generate it in a single shot from a single matrix.

I played around with this a decent amount of the retrieval impl, and kept running into this limitation.

travisdriver: Yeah there may be situations where the bottleneck is unavoidable. I was trying to avoid explicitly enforcing the bottleneck for all retrievers, but since it's unavoidable for the NetVLAD retriever we can revisit it later if we have to.

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,167 @@
+"""Retriever implementation which provides a NetVLAD global image descriptor to suggest potential image pairs.
+
+Note: Similarity computation based off of Paul-Edouard Sarlin's HLOC:
+Reference: https://github.com/cvg/Hierarchical-Localization/blob/master/hloc/pairs_from_retrieval.py
```
travisdriver, line 4: Referencing the [NetVLAD paper](https://openaccess.thecvf.com/content_cvpr_2016/papers/Arandjelovic_NetVLAD_CNN_Architecture_CVPR_2016_paper.pdf) may be more appropriate here.

johnwlambert: Good point, I'll add it.

johnwlambert:
These show the overhead passes over the crane mast:
skydio2-200-netvlad_similarity_matrix

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+        invalid: (K1,K2) boolean array indicating invalid match pairs (e.g. self-matches).
+        num_select: number of matches to select, per query.
+        min_score: minimum allowed similarity score.
+
+    Returns:
+        pairs: tuples representing pairs (i1,i2) of images.
+    """
+    N, _ = scores.shape
+    # if there are only N images to choose from, selecting more than N is not allowed
+    num_select = min(num_select, N)
```
travisdriver, line 159: `N - 1` since we don't want to match image with itself.

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+                sim_block = torch.einsum("id,jd->ij", block_i_query_descs.to(device), block_j_query_descs.to(device))
+                sim[i_start:i_end, j_start:j_end] = sim_block
+
+        return sim
+
+
+def pairs_from_score_matrix(
+    scores: torch.Tensor, invalid: np.array, num_select: int, min_score: Optional[float] = None
+) -> List[Tuple[int, int]]:
+    """Identify image pairs from a score matrix.
```
travisdriver, line 146: It might be a good idea to reference the HLoc function this is based off here.

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+        device = "cuda" if torch.cuda.is_available() else "cpu"
+
+        if num_images > MAX_NUM_IMAGES:
+            raise RuntimeError("Cannot construct similarity matrix of this size.")
+
+        sim = torch.zeros((num_images, num_images))
+        num_blocks = math.ceil(num_images / self._blocksize)
+
+        for block_i in range(num_blocks):
+            for block_j in range(num_blocks):
```
travisdriver, line 105: Can this be parallelized?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+from gtsfm.retriever.retriever_base import RetrieverBase
+from gtsfm.frontend.cacher.global_descriptor_cacher import GlobalDescriptorCacher
+from gtsfm.frontend.global_descriptor.netvlad_global_descriptor import NetVLADGlobalDescriptor
+
+
+logger = logger_utils.get_logger()
+
+PLOT_SAVE_DIR = Path(__file__).parent.parent.parent / "plots"
+
+MAX_NUM_IMAGES = 10000
```
travisdriver, line 30: How was this number chosen?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+            plt.imshow(np.triu(sim.detach().cpu().numpy()))
+            os.makedirs(PLOT_SAVE_DIR, exist_ok=True)
+            plt.title("Image Similarity Matrix")
+            np.savetxt(
+                fname=os.path.join(PLOT_SAVE_DIR, "netvlad_similarity_matrix.txt"),
+                X=sim.detach().cpu().numpy(),
+                fmt="%.2f",
+                delimiter=",",
+            )
+            plt.savefig(os.path.join(PLOT_SAVE_DIR, "netvlad_similarity_matrix.jpg"), dpi=500)
```
travisdriver, line 77: Unrelated to this PR, but maybe we could generate a similar figure after the TwoViewEstimator where the z-axis is the pose error?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+    Returns:
+        pairs: tuples representing pairs (i1,i2) of images.
+    """
+    N, _ = scores.shape
+    # if there are only N images to choose from, selecting more than N is not allowed
+    num_select = min(num_select, N)
+
+    assert scores.shape == invalid.shape
+    invalid = torch.from_numpy(invalid).to(scores.device)
+    if min_score is not None:
```
travisdriver, line 163: Should a minimum score be required? If it's not provided, does it just result in exhaustive matching?

gtsfm/retriever/sequential_retriever.py
```diff
@@ -0,0 +1,46 @@
...
+        num_images = len(loader)
+
+        pairs = []
+        for i1 in range(num_images):
+            for i2 in range(num_images):
+                if i1 >= i2:
+                    continue
+                if (i2 - i1) > self._max_frame_lookahead:
+                    continue
+                pairs.append((i1, i2))
```
travisdriver, line 43: Maybe the following would be better:

```python
for i1 in range(num_images):
    for i2 in range(i1 + 1, num_images):
        ...
```

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -86,11 +106,30 @@ def construct_scene_optimizer(self) -> SceneOptimizer:
...
+            retriever = NetVLADRetriever(num_matched=self.parsed_args.num_matched)
+
+        elif self.parsed_args.matching_regime == "sequential":
+            retriever = SequentialRetriever(max_frame_lookahead=self.parsed_args.max_frame_lookahead)
+
+        elif self.parsed_args.matching_regime == "sequential_with_retrieval":
+            retriever = JointNetVLADSequentialRetriever(
+                num_matched=self.parsed_args.num_matched, max_frame_lookahead=self.parsed_args.max_frame_lookahead
+            )
+        return retriever
```
travisdriver, line 124: Maybe the config would be a better place for this?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+    Args:
+        scores: (K1,K2) for matching K1 images against K2 images.
+        invalid: (K1,K2) boolean array indicating invalid match pairs (e.g. self-matches).
+        num_select: number of matches to select, per query.
+        min_score: minimum allowed similarity score.
+
+    Returns:
+        pairs: tuples representing pairs (i1,i2) of images.
+    """
+    N, _ = scores.shape
```
travisdriver, line 157: nit: `scores.shape[0]`

gtsfm/retriever/netvlad_retriever.py
travisdriver, line 163: Should we at least require that the scores are greater than some epsilon? It might not be a good idea to return pairs whose descriptors are almost perpendicular.

gtsfm/retriever/netvlad_retriever.py
johnwlambert: that's a good idea, would be an interesting visualization

gtsfm/retriever/netvlad_retriever.py
johnwlambert: Thanks for the comments, @travisdriver. Good point, I'll add a reference here

gtsfm/retriever/netvlad_retriever.py
johnwlambert: We probably should. I plotted what theshold would be good for skydio-250, and saw 0.5 as a good threshold. That threshold doesn't work for Door-12, though. So a bit hard to tune.
Screen Shot 2022-02-10 at 10 49 30 AM
Screen Shot 2022-02-10 at 10 49 17 AM

gtsfm/retriever/sequential_retriever.py
johnwlambert: that's cleaner, thanks. I'll update.

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: This was something I went back and forth on. If we want it configured by the config, it would need to go into the SceneOptimizer class. But we run most of the SceneOptimizer with dask, whereas this would not be run with Dask.

The other thing was that I wanted this to be mainly configured by CLI for flexibility.

But I'm open to rethinking this.

gtsfm/retriever/netvlad_retriever.py
johnwlambert: It could be. One way would be to have a dask `.compute()` call right before we set up the SceneOptimizer's computational graph.

We could have the computation of each sub-block be a separate Delayed dask.

gtsfm/retriever/netvlad_retriever.py
travisdriver: These plots are great; if they're not already generated automatically in the codebase we should add them.

Is it Skydio-250 top and door-12 bottom?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -86,11 +106,30 @@ def construct_scene_optimizer(self) -> SceneOptimizer:
 
         return scene_optimizer
 
+    def construct_retriever(self):
+        """Set up retriever module."""
+        if self.parsed_args.matching_regime == "exhaustive":
+            retriever = SequentialRetriever(max_frame_lookahead=MAX_POSSIBLE_FRAME_LOOKAHEAD)
+
```
travisdriver, line 113: I think it would be cleaner if this was it's own class.

gtsfm/retriever/netvlad_retriever.py
johnwlambert: Thanks These are actually both from Skydio-250 (U error plot, and then R error plot).

Sure, I can add that -- it's only about 10 lines of code to call at the end.

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Gotcha -- could you share a bit more about the overall structure you're envisioning?

travisdriver: It could just be a child of the `SequentialRetriever` and it would set `max_frame_lookahead=MAX_POSSIBLE_FRAME_LOOKAHEAD` by default.

gtsfm/retriever/netvlad_retriever.py
johnwlambert: @travisdriver I went ahead and converted this part to use Dask -- I can get a 10x + speedup on this section now with parallelization. Nice suggestion.

gtsfm/retriever/netvlad_retriever.py
johnwlambert: I added this code now to `scene_optimizer.py` as a free function, called at the end of execution.

womackj1:
> These show the overhead passes over the crane mast: skydio2-200-netvlad_similarity_matrix

gtsfm/runner/gtsfm_runner_base.py
womackj1: Since exhaustive matching can occur outside of sequential image sets, I see it as separate from the SequentialRetriever (except that SequentialRetriever allows for reuse of code).

gtsfm/frontend/cacher/global_descriptor_cacher.py
```diff
@@ -0,0 +1,86 @@
...
+    def __get_cache_path(self, cache_key: str) -> Path:
+        """Gets the file path to the cache bz2 file from the cache key."""
+        return CACHE_ROOT_PATH / "global_descriptor" / "{}.pbz2".format(cache_key)
+
+    def __generate_cache_key(self, image: Image) -> str:
+        """Generates the cache key from the input image and underlying global descriptor."""
+        input_key = cache_utils.generate_hash_for_image(image)
+        return "{}_{}".format(self._global_descriptor_obj_cache_key, input_key)
+
+    def __load_result_from_cache(self, image: Image) -> Optional[np.ndarray]:
```
womackj1, line 49: Would __load_descriptor_from_cache be more descriptive? Or is there a case where the result would not be a global descriptor?

gtsfm/frontend/cacher/global_descriptor_cacher.py
```diff
@@ -0,0 +1,86 @@
...
+        """Load cached result, if they exist."""
+        cache_path = self.__get_cache_path(cache_key=self.__generate_cache_key(image=image))
+        cached_data = io_utils.read_from_bz2_file(cache_path)
+        if cached_data is None:
+            return None
+        return cached_data["global_descriptor"]
+
+    def __save_result_to_cache(self, image: Image, global_descriptor: np.ndarray) -> None:
```
womackj1, line 57: Same as above comment - __save_descriptor_to_cache() ?

gtsfm/retriever/joint_netvlad_sequential_retriever.py
```diff
@@ -0,0 +1,43 @@
...
+            loader: image loader. The length of this loader will provide the total number of images
+                for exhaustive global descriptor matching.
+
+        Return:
+            pair_indices: (i1,i2) image pairs.
+        """
+        sim_pairs = self._similarity_retriever.run(loader)
+        seq_pairs = self._seq_retriever.run(loader)
+
+        pairs = list(set(sim_pairs).union(seq_pairs))
```
womackj1: Should we add an argument for taking the intersection of `sim_pairs` and `seq_pairs`?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+def pairs_from_score_matrix(
+    scores: torch.Tensor, invalid: np.array, num_select: int, min_score: Optional[float] = None
+) -> List[Tuple[int, int]]:
+    """Identify image pairs from a score matrix.
+
+    Args:
+        scores: (K1,K2) for matching K1 images against K2 images.
+        invalid: (K1,K2) boolean array indicating invalid match pairs (e.g. self-matches).
```
womackj1, line 236: Where 1s are invalid and 0s are valid?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -12,9 +12,16 @@
...
 from gtsfm.scene_optimizer import SceneOptimizer
 
+from gtsfm.retriever.joint_netvlad_sequential_retriever import JointNetVLADSequentialRetriever
+from gtsfm.retriever.netvlad_retriever import NetVLADRetriever
+from gtsfm.retriever.sequential_retriever import SequentialRetriever
+
 
 logger = logger_utils.get_logger()
 
+# For exhaustive matching, we limit the lookahead to 10,000 images.
```
womackj1, line 23: Out of curiosity, how was this limit decided? I would think the effect of additional matches to a particular image would become marginal before 10,000 images.

gtsfm/frontend/global_descriptor/global_descriptor_base.py
```diff
@@ -0,0 +1,17 @@
...
+"""
+
+import abc
+
+import numpy as np
+
+from gtsfm.common.image import Image
+
+
+class GlobalDescriptorBase:
```
akshay-krishnan, line 13: Missing documentation here, for the class and for the args of describe()

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
```diff
@@ -0,0 +1,45 @@
...
+import numpy as np
+import torch
+from torch import nn
+
+from gtsfm.common.image import Image
+from gtsfm.frontend.global_descriptor.global_descriptor_base import GlobalDescriptorBase
+from thirdparty.hloc.netvlad import NetVLAD
+
+
+class NetVLADGlobalDescriptor(GlobalDescriptorBase):
```
akshay-krishnan, line 25: no documentation?

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
```diff
@@ -0,0 +1,45 @@
...
+    def __init__(self) -> None:
+        """ """
+        pass
+
+    def describe(self, image: Image) -> np.ndarray:
+        """
```
akshay-krishnan, line 31: small 1 line summary for method would be helpful I think

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Was just an arbitrary number that I didn't think we could reach / handle

gtsfm/retriever/netvlad_retriever.py
johnwlambert: it's actually a boolean matrix (I cast np.ones() to bool type before passing it in).

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
```diff
@@ -0,0 +1,45 @@
...
+        Args:
+            image: input image.
+
+        Returns:
+            img_desc: array of shape (D,) representing global image descriptor.
+        """
+        model: nn.Module = NetVLAD()
```
akshay-krishnan, line 38: why is this not done in `__init__()` ?

gtsfm/retriever/joint_netvlad_sequential_retriever.py
```diff
@@ -0,0 +1,60 @@
...
+        sim_pairs = self._similarity_retriever.create_computation_graph(loader)
+        seq_pairs = self._seq_retriever.create_computation_graph(loader)
+
+        pairs = dask.delayed(self.aggregate_pairs)(sim_pairs=sim_pairs, seq_pairs=seq_pairs)
+        return pairs
+
+    def aggregate_pairs(
+        self, sim_pairs: List[Tuple[int, int]], seq_pairs: List[Tuple[int, int]]
+    ) -> List[Tuple[int, int]]:
+        """
```
akshay-krishnan, line 50: missing method summary

gtsfm/retriever/joint_netvlad_sequential_retriever.py
```diff
@@ -0,0 +1,60 @@
...
+from gtsfm.retriever.retriever_base import RetrieverBase
+from gtsfm.retriever.netvlad_retriever import NetVLADRetriever
+from gtsfm.retriever.sequential_retriever import SequentialRetriever
+
+
+logger = logger_utils.get_logger()
+
+
+class JointNetVLADSequentialRetriever(RetrieverBase):
+    """Note: this class contains no .run() method."""
```
akshay-krishnan, line 21: missing class documentation

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,261 @@
...
+            loader: image loader. The length of this loader will provide the total number of images
+                for exhaustive global descriptor matching.
+            visualize:
+
+        Return:
+            Delayed task which evaluates to a list of (i1,i2) image pairs.
+        """
+        num_images = len(loader)
+        sim = self.compute_similarity_matrix(loader, num_images)
+        return dask.delayed(self.compute_pairs_from_similarity_matrix)(sim=sim, loader=loader, visualize=visualize)
```
akshay-krishnan, line 82: not sure how this is done in other places (@ayushbaid )
isnt the Dask.delayed supposed to be created in create_computation_graph? so that run() returns the actual type.

akshay-krishnan:
I got halfway through. Ideally this should be multiple PRs, there seem to be multiple concepts being introduced - Sequential retriever, global descriptor, NetVLAD, joint sequential + netvlad. 
Its difficult for a reviewer to understand the big picture + the details for the entire PR. And the quality of the review depends on how well the PR was understood.

gtsfm/frontend/global_descriptor/global_descriptor_base.py
johnwlambert: thanks, updated

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
johnwlambert: initializing in the constructor leads to OOM w/ our current Dask paradigm.

gtsfm/retriever/joint_netvlad_sequential_retriever.py
johnwlambert: Updated now, thanks

thirdparty/netvlad/netvlad.py
```diff
@@ -0,0 +1,186 @@
+"""Source: Adapted from "Hierarchical-Localization" toolbox.
```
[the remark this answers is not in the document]
johnwlambert: I worry that if we use all submodules, we'll start to have a pretty heavyweight set of dependencies, since some of these libraries are big (for example, hloc has many submodules itself). But I'm open to the idea in the future, if we feel like the current solution isn't working.

gtsfm/retriever/joint_netvlad_sequential_retriever.py
akshay-krishnan: Do we have a standard for this? I prefer the class documentation to be more verbose, because it will show up in IDEs (if you hover over an instance of the class) and other doc tools we may want to use in the future. If not more verbose at least somewhat descriptive.

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
akshay-krishnan: how/why? 🤔

johnwlambert: I haven't dug into it too much to be honest. But I tested it both ways, and using it in the constructor was essentially unrunnable. I believe it has to do with the amount of memory being allocated and kept around per worker.

gtsfm/retriever/joint_netvlad_sequential_retriever.py
johnwlambert: Hmm I hadn't considered that option. Is there a scenario where you think would be especially useful?

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
akshay-krishnan: unrunnable on the CI or locally as well? 
I find issues when running with a high max lookahead from distributed utils complaining that 20-50% of CPU is being used for garbage collection. This could be another instance where that might happen. We should not have inefficient code just to fit a 500 image dataset in 16GB RAM, which is kind of a long shot.

johnwlambert: unrunnable locally for me

gtsfm/retriever/joint_netvlad_sequential_retriever.py
johnwlambert: Hmm didn't realize that, I use Sublime Text : - )

Happy to put more info in the class docstring here, though.

akshay-krishnan:
In the CL description could you provide a list of modules this CL adds and a brief one-liner on what they do (this can be skipped if its totally obvious). 

As a reviewer without much insight into retrievers, I still lack a high-level understanding of what is changing.

gtsfm/scene_optimizer.py
```diff
@@ -432,11 +443,62 @@ def save_full_frontend_metrics(
...
+
+        sim_scores.append(sim_score)
+        R_errors.append(R_error)
+        U_errors.append(U_error)
+
+    plt.scatter(sim_scores, R_errors, 10, color="r", marker=".")
+    plt.xlabel("Similarity score")
+    plt.ylabel("Rotation error w.r.t. GT (deg.)")
+    plt.savefig(os.path.join(PLOT_BASE_PATH, "gt_rot_error_vs_similarity_score.jpg"), dpi=500)
+    plt.close("all")
```
akshay-krishnan, line 491: isnt this to be called only once?

gtsfm/scene_optimizer.py
```diff
@@ -432,11 +443,62 @@ def save_full_frontend_metrics(
...
+        # must come after two-view report file is written to disk in the Dask dependency graph.
+        _save_retrieval_two_view_metrics()
+
+
+def _save_retrieval_two_view_metrics() -> None:
+    """Compare 2-view similarity scores with their 2-view pose errors after viewgraph estimation."""
+    sim_fpath = os.path.join(PLOT_BASE_PATH, "netvlad_similarity_matrix.txt")
+    sim = np.loadtxt(sim_fpath, delimiter=",")
+
+    json_fpath = os.path.join(METRICS_PATH, "two_view_report_VIEWGRAPH_2VIEW_REPORT.json")
```
akshay-krishnan, line 469: why not pass the two view report here?

gtsfm/scene_optimizer.py
```diff
@@ -432,11 +443,62 @@ def save_full_frontend_metrics(
...
+
+    if matching_regime not in [ImageMatchingRegime.RETRIEVAL, ImageMatchingRegime.SEQUENTIAL_WITH_RETRIEVAL]:
+        return
+    if "VIEWGRAPH_2VIEW_REPORT" in filename:
+        # must come after two-view report file is written to disk in the Dask dependency graph.
+        _save_retrieval_two_view_metrics()
+
+
+def _save_retrieval_two_view_metrics() -> None:
+    """Compare 2-view similarity scores with their 2-view pose errors after viewgraph estimation."""
```
akshay-krishnan, line 465: I am not sure how the pose errors are expected to capture visual similarity in image, especially the translation error. 
- cameras may be far away looking at the same object. 
- repetitive structure can have good visual similarity but high pose error.

akshay-krishnan: also why are we using the pose error from view graph estimation instead of ground truth? Im not sure if we want to evaluate a module using products downstream.

gtsfm/retriever/retriever_base.py
```diff
@@ -0,0 +1,53 @@
...
+from dask.delayed import Delayed
+
+from gtsfm.loader.loader_base import LoaderBase
+
+
+class ImageMatchingRegime(str, Enum):
+    SEQUENTIAL: str = "sequential"
+    RETRIEVAL: str = "retrieval"
+    EXHAUSTIVE: str = "exhaustive"
+    SEQUENTIAL_WITH_RETRIEVAL: str = "sequential_with_retrieval"
```
akshay-krishnan, line 20: could you use all caps, matching with the enum? that is how it has been done in a few other places, just for uniformity.

gtsfm/retriever/retriever_base.py
```diff
@@ -0,0 +1,53 @@
...
+        Args:
+            loader: image loader. The length of this loader will provide the total number of images
+                for exhaustive global descriptor matching.
+
+        Return:
+            pair_indices: (i1,i2) image pairs.
+        """
+
+    def create_computation_graph(self, loader: LoaderBase) -> Delayed:
+        """Compute potential image pairs.
```
akshay-krishnan, line 44: the doc for this method generally hints at creating a dask graph, and is not the same as the previous method.

gtsfm/retriever/retriever_base.py
```diff
@@ -0,0 +1,53 @@
...
+
+
+class ImageMatchingRegime(str, Enum):
+    SEQUENTIAL: str = "sequential"
+    RETRIEVAL: str = "retrieval"
+    EXHAUSTIVE: str = "exhaustive"
+    SEQUENTIAL_WITH_RETRIEVAL: str = "sequential_with_retrieval"
+
+
+class RetrieverBase:
```
akshay-krishnan, line 23: brief class docstring please! :)

akshay-krishnan (commented):
comments set 2/N :( 
Sorry for the fragmented reviews!

gtsfm/scene_optimizer.py
```diff
@@ -413,13 +422,15 @@ def save_full_frontend_metrics(
         # Note: if GT is unknown, then R_error_deg, U_error_deg, and inlier_ratio_gt_model will be None
         metrics_list.append(
             {
-                "i1": i1,
-                "i2": i2,
+                "i1": int(i1),
```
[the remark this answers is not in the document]
johnwlambert: turns out they are int64, not int. So json can't serialize it otherwise. It's strange, and I'm not 100% sure why. But this fixes the serialization error.

gtsfm/scene_optimizer.py
johnwlambert: so we are plotting error w.r.t. GT, for only the pairs that remain after viewgraph estimation, to get a better sense of how our 2-view errors are linked to the temporal distance in the sequence. doesn't make sense for all datasets, but for temporally ordered ones, i think it's still useful.

thirdparty/hloc/netvlad.py
```diff
@@ -0,0 +1,186 @@
...
+    # https://github.com/uzh-rpg/netvlad_tf_open/blob/master/matlab/net_class2struct.m.
+    dir_models = {
+        "VGG16-NetVLAD-Pitts30K": "https://cvg-data.inf.ethz.ch/hloc/netvlad/Pitts30K_struct.mat",
+        "VGG16-NetVLAD-TokyoTM": "https://cvg-data.inf.ethz.ch/hloc/netvlad/TokyoTM_struct.mat",
+    }
+
+    def __init__(self, conf: Dict[str, Any] = default_conf):
+        """
+        Args:
+            config: model config w/ inference settings.
```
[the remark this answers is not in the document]
johnwlambert: so this file is actually taken verbatim from another repo: https://github.com/cvg/Hierarchical-Localization/blob/master/hloc/extractors/netvlad.py

I did add a few type hints and docstrings, though.

thirdparty/hloc/netvlad.py
```diff
@@ -0,0 +1,186 @@
...
+
+logger = logging.getLogger(__name__)
+
+# path to /thirdparty/netvlad/weights/{CHECKPOINT}.mat
+netvlad_path = Path(__file__).resolve().parent / "weights"
+
+EPS = 1e-6
+
+
+class NetVLADLayer(nn.Module):
```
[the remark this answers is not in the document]
johnwlambert: sure, just added it

gtsfm/scene_optimizer.py
travisdriver: I think it would be more informative to plot the results of all the pairs provided by the Retriever.

gtsfm/scene_optimizer.py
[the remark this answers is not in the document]
akshay-krishnan: wasnt this working before this change?

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
johnwlambert: Updated, so marking as resolved.

gtsfm/retriever/joint_netvlad_sequential_retriever.py
johnwlambert: Marking as resolved.

thirdparty/hloc/netvlad.py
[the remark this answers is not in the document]
johnwlambert: Updated, so marking as resolved.

tests/retriever/test_sequential_retriever.py
```diff
@@ -0,0 +1,60 @@
...
+
+DATA_ROOT_PATH = Path(__file__).resolve().parent.parent / "data"
+DEFAULT_FOLDER = DATA_ROOT_PATH / "set1_lund_door"
+
+
+def test_sequential_retriever() -> None:
+    """Assert that we get 30 total matches with a lookahead of 3 frames on the Door Dataset."""
+    loader = OlssonLoader(str(DEFAULT_FOLDER), image_extension="JPG", max_frame_lookahead=12)
+    
+    retriever = SequentialRetriever(max_frame_lookahead=3)
```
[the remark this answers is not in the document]
johnwlambert: Updated now, thanks.

gtsfm/retriever/retriever_base.py
johnwlambert: yeah, the argparse argument is of type string here: https://github.com/borglab/gtsfm/pull/453/files#diff-dabbe43e473aeebd439bd77b3d2cafbd366a4da78dbc171661d01e9e9a2f2142R70

gtsfm/retriever/retriever_base.py
johnwlambert: Updated now, thanks

gtsfm/retriever/retriever_base.py
johnwlambert: Added one now, thanks.

gtsfm/scene_optimizer.py
johnwlambert: Hmm would need to be cleared after every plot call, to prevent aggregation on the same background plot.

gtsfm/retriever/joint_netvlad_sequential_retriever.py
johnwlambert: @womackj1 can I mark this comment as resolved?

gtsfm/scene_optimizer.py
johnwlambert: Hi guys, thanks for the feedback here.

So I was interested in seeing how retrieval affects the final performance / back-end. If pairs dont have enough support, they never come out of the front-end, so they would just be noise in our system analysis. So I think the other plot could potentially lead to some incorrect conclusions.

gtsfm/scene_optimizer.py
johnwlambert: seemed roughly equivalent to me, happy to change it though if you prefer

gtsfm/scene_optimizer.py
travisdriver: I still think visualizing the raw Retriever output would be the most useful. Having both may be even better to see how the solution evolves following the viewgraph?

gtsfm/retriever/retriever_base.py
akshay-krishnan: You could have the flag as lower case and convert it to upper before calling the enum (that way the user can use whatever case), and the enum is still consistent.

gtsfm/scene_optimizer.py
akshay-krishnan: > So I was interested in seeing how retrieval affects the final performance / back-end. 
 
we can compare the final metrics with and without the retriever to evaluate that. 

this plot is a kind of validation that we are doing the right thing.. retrieved edges -> edges with low error in the view graph. but I don't think it is a metric that we want to monitor long-term. As Travis mentioned, it ignores the edges that were retrieved but were rejected during view graph estimation. 
If this is useful for debugging, we can have a separate viz script that reads these jsons and creates these plots, but doing that in the pipeline seems a little hacky to me

gtsfm/scene_optimizer.py
akshay-krishnan: for both files, could you handle the case that file was not found? 
its safer to just pass in the reports instead of reading them from memory.

gtsfm/scene_optimizer.py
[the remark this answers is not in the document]
akshay-krishnan: they were working before, right? so what changed?

thirdparty/hloc/netvlad.py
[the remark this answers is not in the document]
akshay-krishnan: okay, never mind then. :)

gtsfm/retriever/joint_netvlad_sequential_retriever.py
akshay-krishnan: Yeah, it would be nice if you could add a short summary now, but for future reference I think class and method documentation should be preferred over file documentation. In GTSAM for example, we just have the copyright and a brief one or two liner for the file.

travisdriver:
@johnwlambert There's a small Flake8 error thats preventing the unit tests from running: `gtsfm/retriever/exhaustive_retriever.py:18:65: W292 no newline at end of file`

.github/workflows/benchmark.yml
```diff
@@ -19,7 +19,7 @@ jobs:
             [deep_front_end,           skydio-8,              8,   jpg,  gdrive,     colmap-loader,  760, true],
             [sift_front_end,           skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true],
             [deep_front_end,           skydio-32,             32,  jpg,  gdrive,     colmap-loader,  760, true],
-            [deep_front_end,           skydio-501,            15,  jpg,  wget,       colmap-loader,  760, true],
+            [deep_front_end,           skydio-501,             5,  jpg,  wget,       colmap-loader,  760, true],
```
travisdriver, line 22: Why this change?

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
travisdriver: Just a short description is fine to avoid mypy warnings.

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,262 @@
...
+                X=sim.detach().cpu().numpy(),
+                fmt="%.2f",
+                delimiter=",",
+            )
+            plt.savefig(os.path.join(PLOT_SAVE_DIR, "netvlad_similarity_matrix.jpg"), dpi=500)
+            plt.close("all")
+
+        named_pairs = [(query_names[i], query_names[j]) for i, j in pairs]
+        logger.info("Found %d pairs from the NetVLAD Retriever.", len(pairs))
+        logger.info("Image Name Pairs:" + str(named_pairs))
```
travisdriver, line 223: Should probably remove this before merging

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -72,6 +90,7 @@ def construct_loader(self) -> LoaderBase:
         pass
 
     def construct_scene_optimizer(self) -> SceneOptimizer:
+        """ """
```
travisdriver, line 93: Short description

dellaert:
All, this PR seems close to done. I added a commit to use the unittest framework as in all other tests.

dellaert:
@akshay-krishnan , @johnwlambert , we split this PR at great expense of our time :-)

gtsfm/retriever/joint_netvlad_sequential_retriever.py
akshay-krishnan: Unresolved :)

gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py
akshay-krishnan: unresolved

gtsfm/retriever/joint_netvlad_sequential_retriever.py
```diff
@@ -0,0 +1,62 @@
...
+from gtsfm.loader.loader_base import LoaderBase
+from gtsfm.retriever.retriever_base import RetrieverBase
+from gtsfm.retriever.netvlad_retriever import NetVLADRetriever
+from gtsfm.retriever.sequential_retriever import SequentialRetriever
+
+
+logger = logger_utils.get_logger()
+
+
+class JointNetVLADSequentialRetriever(RetrieverBase):
```
akshay-krishnan, line 20: High-level design question:

Can we bake in this retrieving logic into whatever calls this class? Ideally it would be nice to specify a list of retrievers in the config and an aggregation method - union, intersection etc. and get them aggregated. Or we could create an aggregator that accepts a list of retrievers. 

Can also be done in another PR. Please create an issue maybe?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,262 @@
...
+
+logger = logger_utils.get_logger()
+
+PLOT_SAVE_DIR = Path(__file__).parent.parent.parent / "plots"
+
+MAX_NUM_IMAGES = 10000
+
+
+@dataclass
+class SubBlockSimilarityResult:
```
akshay-krishnan, line 37: what do the fields mean? what is this dataclass for?

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,262 @@
...
+            plt.savefig(os.path.join(PLOT_SAVE_DIR, "netvlad_similarity_matrix.jpg"), dpi=500)
+            plt.close("all")
+
+        named_pairs = [(query_names[i], query_names[j]) for i, j in pairs]
+        logger.info("Found %d pairs from the NetVLAD Retriever.", len(pairs))
+        logger.info("Image Name Pairs:" + str(named_pairs))
+        return pairs
+
+
+def pairs_from_score_matrix(
```
akshay-krishnan, line 227: is this unit tested? looks like a util function to me..

gtsfm/retriever/netvlad_retriever.py
akshay-krishnan: unresolved

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,262 @@
...
+class SubBlockSimilarityResult:
+    i_start: int
+    i_end: int
+    j_start: int
+    j_end: int
+    subblock: torch.Tensor
+
+
+class NetVLADRetriever(RetrieverBase):
+    def __init__(self, num_matched: int, blocksize: int = 50) -> None:
```
akshay-krishnan, line 46: is this class specific to netvlad? 
Can this be done with other descriptors too? maybe ones @travisdriver was looking into?

akshay-krishnan: if you think it can be made generic, please create another issue before merging.

gtsfm/scene_optimizer.py
akshay-krishnan: please move this to a util file? feel free to create one if it doesn't exist already

gtsfm/scene_optimizer.py
akshay-krishnan: unresolved, although I dont remember what my earlier comment was about.

akshay-krishnan (approved):

gtsfm/retriever/netvlad_retriever.py
```diff
@@ -0,0 +1,262 @@
...
+        """Compute potential image pairs.
+
+        Args:
+            loader: image loader. The length of this loader will provide the total number of images
+                for exhaustive global descriptor matching.
+
+        Return:
+            Delayed task that evaluates to a list of (i1,i2) image pairs.
+        """
+        return self.run(loader=loader)
```
ayushbaid, line 67: This shoud be delayed? And this should just be in the base class, right? Maybe needs a rebase.

gtsfm/retriever/netvlad_retriever.py
ayushbaid: Lets have the run method not use any dask delayed. We can duplicate the logic in create_computation_graph if we like. The reason is that non-dask pathways can be very useful for large datasets (e.g. Hilti).

gtsfm/retriever/netvlad_retriever.py
travisdriver: This class is specific to NetVLAD

gtsfm/retriever/netvlad_retriever.py
travisdriver: This is only logged once so I'm fine with leaving it

travisdriver:
Here's the final dashboard with NetVLAD turned off by default, where all the changes seem to be due to noise (on the order of 0.1 or 0.01):

newplot (15)

[visual_comparison_dashboard.zip](https://github.com/borglab/gtsfm/files/9199101/visual_comparison_dashboard.zip)

PR #454, opened 2022-02-15
Adding ratio test for 2-way matcher

ayushbaid:
Adding ratio test to match colmap's default setting


johnwlambert:
(the deep front end results must just be noise, since they are not affected here)

travisdriver:
> (the deep front end results must just be noise, since they are not affected here)

Must have been a lot of noise. Was the front-end summary removed from the metrics report?

ayushbaid:
> > (the deep front end results must just be noise, since they are not affected here)
> 
> Must have been a lot of noise. Was the front-end summary removed from the metrics report?

This means our reproducibility tests doesn't capture this case?

ayushbaid:
Major improvements observed:
frontend_post_ba_2view:
frontend_post_ba_2view

final:
final

travisdriver:
> > thanks Ayush. I think the code might be simpler if we made just one TwoWayMatcher, with ratio test as an optional argument. I think most of the times we would want the TwoWayMatcher w/ the ratio test turned on. Have you checked results w/ and w/o it? Could you share here on our benchmarks?
> 
> The logic for the two-way with ratio-test is a bit complicated, and I having having two seperate classes instead of an if-else is better. Also we will need two different unit-test classes so I think having two different implementation classes works fine.

I agree with @johnwlambert, I think just having one optional parameter would greatly simplify things. 

Also, would you mind posting the actual HTML dashboard along with the screenshots?

ayushbaid:
[visual_comparison_dashboard.pdf](https://github.com/borglab/gtsfm/files/8101383/visual_comparison_dashboard.pdf)

ayushbaid:
Regarding two classes vs one class I think the pros and cons are:

Pros:
1. Easy to switch between using and not using the ratio test.

Cons:
1. If else statements in the code.
2. Having None in the config file to turn off ratio test if we turn on the ratio test by default (which we should)

I think having two separate classes is better, but I am not completely against having an optional argument too. Let me know if I am missing something in pros or cons.

Looking at the results, I think we will keep the ratio test turned on by default and I see no reason to switch back and forth.

travisdriver:
@ayushbaid Would it be possible to turn it off for AstroNet for now? It seems to lead to some pretty significant regression.

Have you tried increasing the number of computed features? It might be worth trying again with the ratio test.

gtsfm/frontend/matcher/twoway_matcher.py
```diff
@@ -86,3 +85,55 @@ def match(
...
+    def __init_opencv_matcher(self) -> cv.DescriptorMatcher:
+        """Initialize the OpenCV matcher.
+
+        Returns:
+            Matcher object.
+        """
+        if self._distance_type is MatchingDistanceType.EUCLIDEAN:
+            distance_metric = cv.NORM_L2
+        elif self._distance_type is MatchingDistanceType.HAMMING:
+            distance_metric = cv.NORM_HAMMING
```
travisdriver, line 98: Do we have any binary descriptors in GTSFM?

gtsfm/frontend/matcher/twoway_matcher.py
```diff
@@ -86,3 +85,55 @@ def match(
...
+        match_indices_2to1: Dict[int, int] = self.__perform_oneway_matching(descriptors_2, descriptors_1)
+
+        match_indices_1to2to1 = {
+            idx1: match_indices_2to1[idx2] for idx1, idx2 in match_indices_1to2.items() if idx2 in match_indices_2to1
+        }
+
+        twoway_match_indices = np.array(
+            [[idx1, idx2] for idx1, idx2 in match_indices_1to2.items() if match_indices_1to2to1.get(idx1) == idx1],
+            dtype=np.uint32,
+        )
```
travisdriver, line 124: I think this can be accomplished without looping through all the matches twice.

```python
matches = [(idx1, idx2) for idx1, idx2 in match_ind_1to2.items() if match_ind_2to1[idx2] == idx1]
```

gtsfm/frontend/matcher/twoway_matcher.py
```diff
@@ -86,3 +85,55 @@ def match(
...
+        return twoway_match_indices
+
+    def __perform_oneway_matching(self, descriptors_1: np.ndarray, descriptors_2: np.ndarray) -> Dict[int, int]:
```
travisdriver, line 127: docstring

gtsfm/frontend/matcher/twoway_matcher.py
```diff
@@ -86,3 +85,55 @@ def match(
...
+        opencv_matcher = self.__init_opencv_matcher()
+
+        if self._ratio_test_threshold is not None:
+            all_matches = opencv_matcher.knnMatch(descriptors_1, descriptors_2, k=2)
+            matches = [m1 for m1, m2 in all_matches if m1.distance <= self._ratio_test_threshold * m2.distance]
+        else:
+            matches = opencv_matcher.match(descriptors_1, descriptors_2)
```
travisdriver, line 134: You should perform cross check first, and then you only need to do the ratio test once.

travisdriver: Here's what I did in the previous branch for reference: https://github.com/borglab/gtsfm/blob/12c8869b5ef155875585e2f6b65d1742c2a3ec9d/gtsfm/frontend/matcher/twoway_matcher.py#L92

ayushbaid: Doing ratio test twice is not the same as doing it once after crosscheck. It's a stricter test.

travisdriver: You're right nevermind

gtsfm/frontend/matcher/twoway_matcher.py
ayushbaid: We don't right now. But I prefer to just keep it and add an issue to track removal if we want it removed.

gtsfm/frontend/matcher/twoway_matcher.py
```diff
@@ -86,3 +85,60 @@ def match(
...
+        """
+        opencv_matcher = self.__init_opencv_matcher()
+
+        if self._ratio_test_threshold is not None:
+            all_matches = opencv_matcher.knnMatch(descriptors_1, descriptors_2, k=2)
+            matches = [m1 for m1, m2 in all_matches if m1.distance <= self._ratio_test_threshold * m2.distance]
+        else:
+            matches = opencv_matcher.match(descriptors_1, descriptors_2)
+
+        matches = sorted(matches, key=lambda r: r.distance)
```
akshay-krishnan, line 141: nit: use the same `m` inside the lambda? what does r stand for?

tests/frontend/matcher/test_twoway_withoutratiotest_matcher.py
```diff
@@ -10,31 +10,58 @@
...
         super().setUp()
 
-        self.matcher = TwoWayMatcher()
+        self.matcher = TwoWayMatcher(ratio_test_threshold=None)
 
     def test_on_dummy_data(self):
         """Test using dummy 1D descriptors to verify correctness."""
 
-        image_shape_i1 = (300, 100) # as (H,W)
+        image_shape_i1 = (300, 100)  # as (H,W)
```
akshay-krishnan, line 27: why these formatting changes in this file? are your black settings consistent?

tests/frontend/matcher/test_twoway_withratiotest_matcher.py
```diff
@@ -0,0 +1,79 @@
...
+                ]
+            )
+            .reshape(-1, 1)
+            .astype(np.float32)
+        )
+        keypoints_i2 = feature_utils.generate_random_keypoints(descriptors_i2.shape[0], image_shape_i2)
+        expected_matches = np.array([[9, 5], [2, 4], [3, 2], [0, 3]])
+
+        result = self.matcher.match(
+            keypoints_i1, keypoints_i2, descriptors_i1, descriptors_i2, image_shape_i1, image_shape_i2
```
akshay-krishnan, line 72: just curious, are the keypoints and image shapes used by any matcher? I guess we would need geometry to use the key points, but we dont have any geometry before matching happens?

tests/frontend/matcher/test_twoway_withratiotest_matcher.py
```diff
@@ -0,0 +1,79 @@
+"""Tests for the two-way-with-ratio-test matcher.
+
+Authors: Ayush Baid
+"""
+import unittest
+import numpy as np
+
+import tests.frontend.matcher.test_matcher_base as test_matcher_base
+import gtsfm.utils.features as feature_utils
```
akshay-krishnan, line 9: is the order here correct?

akshay-krishnan (approved):
looks neat and readable, Im not too familiar with matching-specific logic.

tests/frontend/matcher/test_twoway_withoutratiotest_matcher.py
ayushbaid: I am not using any specific black formatter setting except the line length to 120.

From the PEP 8 guidelines, it seems like 2 spaces before the inline comment is the recommendation: https://www.python.org/dev/peps/pep-0008/#inline-comments

tests/frontend/matcher/test_twoway_withratiotest_matcher.py
ayushbaid: correcting it now.

tests/frontend/matcher/test_twoway_withratiotest_matcher.py
ayushbaid: Yes they are used by superglue.

travisdriver (approved):
LGTM! Please post a final dashboard before merging.

ayushbaid:
Sparse reconstruction metrics:

final

PR #456, opened 2022-02-16
Add graph topology visualization

johnwlambert:
## Deep Front End AstroNet
Input:
view_graph_estimator_input_topology_gt_locations

Output:
view_graph_estimator_output_topology_gt_locations

## Sift Front End AstroNet
Input:
view_graph_estimator_input_topology_gt_locations

Output:
view_graph_estimator_output_topology_gt_locations

## Skydio-32 Sift Front End
Input:
view_graph_estimator_input_topology_gt_locations

Output:
view_graph_estimator_output_topology_gt_locations

## Skydio-32 Deep Front End 
Input:
view_graph_estimator_input_topology_gt_locations

Output:
view_graph_estimator_output_topology_gt_locations

### DIFF
--- a/gtsfm/utils/graph.py
+++ b/gtsfm/utils/graph.py
@@ -5,8 +5,15 @@ Authors: Ayush Baid, John Lambert, Akshay Krishnan
 from collections import defaultdict
 from typing import DefaultDict, Dict, List, Optional, Set, Tuple
 
+import matplotlib.pyplot as plt
 import networkx as nx
-from gtsam import Rot3, Unit3
+import numpy as np
+from gtsam import PinholeCameraCal3Bundler, Rot3, Unit3
+
+from gtsfm.common.two_view_estimation_report import TwoViewEstimationReport
+
+GREEN = [0, 1, 0]
+RED = [1, 0, 0]
 
 
 def get_nodes_in_largest_connected_component(edges: List[Tuple[int, int]]) -> List[int]:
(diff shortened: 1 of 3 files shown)


tests/utils/test_graph_utils.py
```diff
@@ -229,6 +230,35 @@ def test_create_adjacency_list_empty(self) -> None:
         assert len(adj_list.keys()) == 0
         assert isinstance(adj_list, defaultdict)
 
+    def test_draw_graph_topology(self) -> None:
```
ayushbaid, line 233: This won't automatically test anything, right?

I suggest we mock the call to networkx, and verify the args passed to the mocked method..

ayushbaid (commented):
Does this PR also plans to integrate this topology plots into the pipeline?

johnwlambert:
> Does this PR also plans to integrate this topology plots into the pipeline?

I was thinking to add that in a separate PR, but could add it here too if you'd like.

gtsfm/utils/graph.py
```diff
@@ -119,3 +126,60 @@ def extract_cyclic_triplets_from_edges(edges: List[Tuple[int, int]]) -> List[Tup
                 triplets.add(cycle_nodes)
 
     return list(triplets)
+
+
+def draw_graph_topology(
```
travisdriver, line 131: `draw_graph_topology` seems a little ambiguous; maybe `draw_view_graph` would be more descriptive?

ayushbaid:
Also, having the nodes at their ground truth position might lead to weird visualizations depending on the actual post location? Like if some of them are too close to each other or stacked on top of each other?

I am fine with the current logic, just want to discuss briefly if there are other approaches.

johnwlambert:
> Also, having the nodes at their ground truth position might lead to weird visualizations depending on the actual post location? Like if some of them are too close to each other or stacked on top of each other?
> 
> I am fine with the current logic, just want to discuss briefly if there are other approaches.

This is true. I've provided an option for Networkx to choose the node locations automatically, though (by passing in `cameras_gt=None`.

tests/utils/test_graph_utils.py
johnwlambert: So this unit test is a "smokescreen" test IMHO. It makes sure that the code can run without crashing. It is a weak test, but I've inspected the plots and they look good.

gtsfm/utils/graph.py
johnwlambert: Good point -- I renamed to `draw_view_graph_topology()`, let me know if that works.

johnwlambert:
> This is really cool :) I also think it'd probably be cleaner to just integrate the plots into the pipeline in this PR as @ayushbaid mentioned.
> 
> It might also be more informative to color the edges depending on the magnitude of the pose error as opposed to a binary labeling with respect to a threshold.

I like the idea about coloring the edges based on the magnitude of the pose error. If we use a linear scale from [0,180], and we had 98 errors near 0 degree error, and 1 error near 20 degree error, and 1 error near 170 degrees error ,then the 20 degree error would be green in any colormap. So it would make sense maybe to do a uniform scale from [0,10], and then clamp. But unfortunately I don't have more time currently to play with that. I think we could add in a future PR.

travisdriver:
> > This is really cool :) I also think it'd probably be cleaner to just integrate the plots into the pipeline in this PR as @ayushbaid mentioned.
> 
> > 
> 
> > It might also be more informative to color the edges depending on the magnitude of the pose error as opposed to a binary labeling with respect to a threshold.
> 
> 
> 
> I like the idea about coloring the edges based on the magnitude of the pose error. If we use a linear scale from [0,180], and we had 98 errors near 0 degree error, and 1 error near 20 degree error, and 1 error near 170 degrees error ,then the 20 degree error would be green in any colormap. So it would make sense maybe to do a uniform scale from [0,10], and then clamp. But unfortunately I don't have more time currently to play with that. I think we could add in a future PR.

I like the clamping idea, but yeah we can save it for a future PR. 

Everything else looks great to me.

travisdriver (approved):

johnwlambert:
thanks for the review @travisdriver. it looks like the deep front end would do better on AstroNet if we can complete the loop (exhaustive matching), so maybe we should upload the cache of precomputed superpoint/superglue for RC3 to the CI? Have we done that yet?

PR #457, opened 2022-02-16
Add interface from OpenMVG output to GTSFM for benchmarking/comparison

johnwlambert:


ayushbaid:
Should be keep files like these in the GTSFM repo, or have a different repo?

dellaert:
Let’s keep here for now

PR #461, opened 2022-02-23
Try min track length 3

johnwlambert:


johnwlambert:
[visual_comparison_dashboard_3_length.html.zip](https://github.com/borglab/gtsfm/files/8127749/visual_comparison_dashboard_3_length.html.zip)
Screen Shot 2022-02-23 at 3 56 03 PM

johnwlambert:
This change doesn't look so promising right now @dellaert @ayushbaid @travisdriver

travisdriver:
@johnwlambert Yeah it seems a majority of our tracks are of length 2; we'll need to figure out a way to increase recall. Maybe through enforcing transitivity between correspondences? The [CLEAR algorithm](https://github.com/mit-acl/clear) seems like a good candidate for this.

PR #463, opened 2022-03-01
Add assert_all_results to repro test base

akshay-krishnan:
Adding an optional method to the repro test base class to check statistics on all results. This is useful for translation averaging.

### DIFF
--- a/tests/repro_tests/test_repro_base.py
+++ b/tests/repro_tests/test_repro_base.py
@@ -34,7 +34,7 @@ class ReproducibilityTestBase(metaclass=abc.ABCMeta):
             all_results.append(current_result)
             self.assert_results(current_result, reference_result)
         
-        self.assert_all_results(all_results)
+        self.assert_results_statistics(all_results)
 
     def assert_results_statistics(self, all_results: List[Any]) -> None:
         """Check for any statistics across all_results.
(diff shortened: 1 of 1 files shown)


ayushbaid (commented):
The idea looks good.

However, I do not agree with the name `assert_all_results`. What kind of comparisons do we think are possible. I think percentile based checks are what we need right now, `__assert_percentile_of_results` or something of similar fashion is better?

akshay-krishnan:
It does not have to be percentile, cant we leave that bit to the user? Maybe they use mean, std? For example, in translation averaging I am checking standard deviation.

akshay-krishnan:
I am open to other names though.

akshay-krishnan:
`assert_results_statistics` ? I dont know ..

ayushbaid:
> 

`assert_results_statistics` looks good. The other alternative is to let the one function remain, and the individual tests can override as needed. For example, the T.A. test can override the current method and check for std deviations instead of equality for each result.

akshay-krishnan:
Yeah, but then no logic from the base class would be used. Thought it would be nice to have some common structure. And lesser code in derived class. But Im fine with either option.

johnwlambert (approved):
LGTM, thanks Akshay

PR #466, opened 2022-03-13
Adding rootsift Detector-Descriptor

ayushbaid:


ayushbaid:
We see moderate improvements after the verification stage, but the final result is worse.


Verifier summary pre-2view-ba
rootsift_pre2viewba

Final bundle adjustment
rootsift_final

PR #473, opened 2022-05-05
Retriever classes

dellaert:
@womackj1 and I split the big PR, so NetVlad is easier to review @akshay-krishnan  :-)

The base retriever class and two instantiations, with tests.

(Copied from netvlad branch, who will now target this branch)

### DIFF
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -12,6 +12,10 @@ from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.loader.loader_base import LoaderBase
 from gtsfm.scene_optimizer import SceneOptimizer
 
+from gtsfm.retriever.exhaustive_retriever import ExhaustiveRetriever
+from gtsfm.retriever.retriever_base import ImageMatchingRegime
+from gtsfm.retriever.sequential_retriever import SequentialRetriever
+
 
 logger = logger_utils.get_logger()
 
@@ -23,6 +27,7 @@ class GtsfmRunnerBase:
         self.parsed_args: argparse.Namespace = argparser.parse_args()
 
         self.loader: LoaderBase = self.construct_loader()
+        self.retriever = self.construct_retriever()
         self.scene_optimizer: SceneOptimizer = self.construct_scene_optimizer()
 
     def construct_argparser(self) -> argparse.ArgumentParser:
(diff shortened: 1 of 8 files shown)


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -86,20 +96,39 @@ def construct_scene_optimizer(self) -> SceneOptimizer:
...
+        """Run the SceneOptimizer."""
         start_time = time.time()
+
+        # create dask client
+        cluster = LocalCluster(
+            n_workers=self.parsed_args.num_workers, threads_per_worker=self.parsed_args.threads_per_worker
+        )
+
+        pairs_graph = self.retriever.create_computation_graph(self.loader)
+        with Client(cluster), performance_report(filename="dask-report.html"):
```
dellaert, line 121: @johnwlambert why do we now have a second `with Client` clause here? The compute only happens in line 135, right?

tests/retriever/test_sequential_retriever.py
```diff
@@ -0,0 +1,63 @@
...
+DATA_ROOT_PATH = Path(__file__).resolve().parent.parent / "data"
+DEFAULT_FOLDER = DATA_ROOT_PATH / "set1_lund_door"
+
+
+class TestSequentialRetriever(unittest.TestCase):
+    def test_sequential_retriever(self) -> None:
+        """Assert that we get 30 total matches with a lookahead of 3 frames on the Door Dataset."""
+
+        max_frame_lookahead = 3
+        loader = OlssonLoader(str(DEFAULT_FOLDER), image_extension="JPG", max_frame_lookahead=max_frame_lookahead)
```
dellaert, line 21: @johnwlambert why do we still need this lookahead parameter now?

dellaert:
Just changed (sole) reviewer of this to @travisdriver , so @akshay-krishnan can concentrate on the Netvlad one (which is less of a priority, BTW).

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Hi @dellaert, the graph is now partitioned into 2 separate computation graphs.

There is a `.compute()` call on line 122, and also on 135. Dask cannot iterate over a sequence of unknown length, which the image pair indices end up being after the Retriever.

I don't think there is any way around this, unfortunately, in our Dask paradigm.

tests/retriever/test_sequential_retriever.py
johnwlambert: We need it because sequential matching needs a specification of the sliding window size.

johnwlambert (approved):
LGTM

tests/retriever/test_sequential_retriever.py
dellaert: Yes, but why is that not just a retriever parameter?

dellaert:
All, where can I find that "overview report" in the CI, showing green/red grid?

tests/retriever/test_sequential_retriever.py
johnwlambert: You're right, good catch. Can be removed. I recommend to do it in a small follow-up PR.

dellaert:
I ran comparison. Weird that this change, which is supposed to be neutral, has changes.

image

Full results here: [visual_comparison_dashboard.html.pdf](https://github.com/borglab/gtsfm/files/8645104/visual_comparison_dashboard.html.pdf)

dellaert:
New metrics no longer include skipped artifacts (see #475) and look better (but still not neutral):
image

Still, after email thread I will merge.

PR #474, opened 2022-05-06
RigRetriever class

dellaert:
A retriever class that reads a file of constraints, including a set of 5 x 5 counts, indicating which cameras on the rig in two successive poses are likely to have visual matches. These counts are matching features from the lidar, in the case of the Hilti challenge, and serve us a proxy to see how good we expect visual matching to work between two cameras.

Currently hardcoded to 5 cameras. There is no need to generalize yet.


dellaert:
@womackj1 FYI
@jerredchen I updated your code and unit tests. I suggest you copy back to the SLAM repo.

dellaert:
PS @ayushbaid how do I check in a new data directory? Tee tests will fail as somehow the files I added in `tests/data/hilti` are not seen as added files by git.

ayushbaid:
> PS @ayushbaid how do I check in a new data directory? Tee tests will fail as somehow the files I added in `tests/data/hilti` are not seen as added files by git.

I use the `git add -f` to add a small subset of the test directory.

ayushbaid:
@dellaert I assume the final version of this PR will be on the Hilti loader (which I will push once I we fix the time sync issue)?

gtsfm/retriever/rig_retriever.py
```diff
@@ -0,0 +1,98 @@
...
+from gtsfm.retriever.retriever_base import RetrieverBase
+
+
+@dataclass
+class Constraint:
+    a: int = 0
+    b: int = 1
+    aTb: gtsam.Pose3 = gtsam.Pose3()
+    cov: np.ndarray = np.eye(6)
+    counts: np.ndarray = np.zeros((5, 5))
```
ayushbaid, line 22: Can you add a docstring for the args. `counts` for example is non-trivial to understand.

gtsfm/retriever/rig_retriever.py
```diff
@@ -0,0 +1,98 @@
...
+import gtsam
+import numpy as np
+from gtsfm.loader.loader_base import LoaderBase
+from gtsfm.retriever.retriever_base import RetrieverBase
+
+
+@dataclass
+class Constraint:
+    a: int = 0
+    b: int = 1
```
ayushbaid, line 19: Should we skip default values for a and b?

gtsfm/retriever/rig_retriever.py
```diff
@@ -0,0 +1,98 @@
...
+    @staticmethod
+    def read(fname: str) -> List["Constraint"]:
+        """Read constraints from text file."""
+        constraint_matrix = np.loadtxt(fname)
+        return [Constraint.from_row(row) for row in constraint_matrix]
+
+    def edges(self, threshold=30):
+        """Return pairs of cameras with over `threshold` count."""
+        return [
+            (self.a * 5 + camera_index_in_a, self.b * 5 + camera_index_in_b)
```
ayushbaid, line 72: Nit: moving 5 to a global variable.

gtsfm/retriever/rig_retriever.py
```diff
@@ -0,0 +1,98 @@
...
+    def __init__(self, constraints_path: Path, threshold: int = 100):
+        """Initialize with path to a constraints file."""
+        self._constraints_path = constraints_path
+        self._threshold = threshold
+
+    def run(self, loader: LoaderBase) -> List[Tuple[int, int]]:
+        """Compute potential image pairs.
+
+        Args:
+            loader: image loader.
```
ayushbaid, line 91: Loader is not being used right now?

gtsfm/retriever/rig_retriever.py
dellaert: done

gtsfm/retriever/rig_retriever.py
dellaert: Done

gtsfm/retriever/rig_retriever.py
dellaert: Hardcoding 5 until we do another rig. Building for the present.

gtsfm/retriever/rig_retriever.py
dellaert: Indeed. But needed by base class

dellaert:
@ayushbaid I'd like to integrate the real HiltiLoader in a separate PR. I think we need to make changes to it to read the constraints in this PR, as well.

ayushbaid:
> @ayushbaid I'd like to integrate the real HiltiLoader in a separate PR. I think we need to make changes to it to read the constraints in that PR, as well.

Constraints will be read in the retriever class, right?

ayushbaid (approved):
LGTM

PR #476, opened 2022-05-07
Adding loader for the hilti dataset

ayushbaid:


dellaert (approved):
Awesome !
I think some of this code might not be used in all our runs given our discussion yesterday, and odometry reading will likely change, but it's good to have it around.

PR #477, opened 2022-05-07
Remove bad tracks from 2-view BA

dellaert:
I enabled filtering after the bundle adjustment step, that filters the verified correspondences. 
I will do a comparison report after the CI finishes.

PR also includes 
- a test on triangulate_two_view_correspondences
- some re-formatting in `tests/common/test_gtsfm_data.py`, sorry about the noise.


gtsfm/common/gtsfm_data.py
```diff
@@ -375,26 +375,29 @@ def __validate_track(self, track: SfmTrack, reproj_err_thresh: float) -> bool:
...
         # TODO: move this function to utils or GTSAM
         filtered_data = GtsfmData(self.number_images())
 
-        for j in range(self.number_tracks()):
-            track = self.get_track(j)
+        valid_mask = [self.__validate_track(track, reproj_err_thresh) for track in self._tracks]
 
-            if self.__validate_track(track, reproj_err_thresh):
+        for track, valid in zip(self._tracks, valid_mask):
+            if valid:
```
johnwlambert, line 393: prefer to `continue` early rather than nest logic here

gtsfm/two_view_estimator.py
```diff
@@ -60,6 +60,7 @@ def __init__(
         bundle_adjust_2view: bool,
         eval_threshold_px: float,
         bundle_adjust_2view_maxiters: int = 100,
+        ba_reproj_error_thresh: float = 0.5,
```
ayushbaid, line 63: Isn't this too strict?

ayushbaid: For reference, we use 3px/4px threshold in verification and the final BA.

ayushbaid (approved):
LGTM. Please check the effect on performance as we are using a very strict threshold of 0.5px by default.

gtsfm/two_view_estimator.py
dellaert: I'll let @johnwlambert comment, as he told me he fine-tuned these for his thesis. But this is 2-view bundle adjustment, which also optimizes for calibration, I think, so it should be able to overfit very closely. In reichstag it does not filter a lot of correspondences away.

dellaert:
I ran a comparison and eager to hear your comments. Seems better overall, although several datasets did not run (time outs?). Some colors should maybe be reversed?
The entire file is [visual_comparison_dashboard.html.pdf](https://github.com/borglab/gtsfm/files/8647137/visual_comparison_dashboard.html.pdf) and the BA result:
image

dellaert:
Update: failing CI was not a time out but a bug in a rare BA case, fixed, re-running.

dellaert:
I now ran comparisons on 8 datasets, and also flipped some colors to better reflect what is "good" and what is "bad". Those edits are in another PR. I think everything is more positive, so I will merge after resolving conflicts, ok?

Full report: [visual_comparison_dashboard.html.pdf](https://github.com/borglab/gtsfm/files/8647867/visual_comparison_dashboard.html.pdf)

ayushbaid (approved):

dellaert:
For posterity:
image

PR #478, opened 2022-05-07
Plumbing pose priors and supporting fisheye calibration

ayushbaid:
Main changes:
1. Plumbing absolute and relative pose priors to all the plates. Note that we decided on not using absolute priors for now, but I am plumbing them just in case we need to use them
2. Supporting Cal3Fisheye calibration in the pipeline
3. Making view graph estimator optional
4. Making dense reconstruction optional
5. Other miscellaneous changes

### DIFF
--- a/gtsfm/bundle/bundle_adjustment.py
+++ b/gtsfm/bundle/bundle_adjustment.py
@@ -4,7 +4,7 @@ Authors: Xiaolong Wu, John Lambert, Ayush Baid
 """
 from collections import Counter
 from pathlib import Path
-from typing import List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple
 
 import dask
 import gtsam
@@ -12,19 +12,24 @@ import numpy as np
 from dask.delayed import Delayed
 from gtsam import (
     GeneralSFMFactor2Cal3Bundler,
+    GeneralSFMFactor2Cal3Fisheye,
     NonlinearFactorGraph,
     PinholeCameraCal3Bundler,
+    PinholeCameraCal3Fisheye,
     PriorFactorCal3Bundler,
+    PriorFactorCal3Fisheye,
     PriorFactorPose3,
     SfmTrack,
     Values,
     symbol_shorthand,
 )
 
+import gtsfm.common.types as gtsfm_types
 import gtsfm.utils.logger as logger_utils
 import gtsfm.utils.metrics as metrics_utils
 import gtsfm.utils.tracks as track_utils
 from gtsfm.common.gtsfm_data import GtsfmData
+from gtsfm.common.pose_prior import PosePrior
 from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
 
 METRICS_GROUP = "bundle_adjustment_metrics"
@@ -41,6 +46,7 @@ K = symbol_shorthand.K  # calibration
 
 CAM_POSE3_DOF = 6  # 6 dof for pose of camera
 CAM_CAL3BUNDLER_DOF = 3  # 3 dof for f, k1, k2 for intrinsics of camera
+CAM_CAL3FISHEYE_DOF = 9
 IMG_MEASUREMENT_DIM = 2  # 2d measurements (u,v) have 2 dof
 POINT3_DOF = 3  # 3d points have 3 dof
 
@@ -48,6 +54,7 @@ POINT3_DOF = 3  # 3d points have 3 dof
 # noise model params
 CAM_POSE3_PRIOR_NOISE_SIGMA = 0.1
 CAM_CAL3BUNDLER_PRIOR_NOISE_SIGMA = 1e-5  # essentially fixed
+CAM_CAL3FISHEYE_PRIOR_NOISE_SIGMA = 1e-5  # essentially fixed
 MEASUREMENT_NOISE_SIGMA = 1.0  # in pixels
 
 logger = logger_utils.get_logger()
@@ -85,7 +92,12 @@ class BundleAdjustmentOptimizer:
     def __map_to_calibration_variable(self, camera_idx: int) -> int:
         return 0 if self._shared_calib else camera_idx
 
-    def __construct_factor_graph(self, initial_data: GtsfmData) -> NonlinearFactorGraph:
+    def __construct_factor_graph(
+        self,
+        initial_data: GtsfmData,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
+    ) -> NonlinearFactorGraph:
         graph = NonlinearFactorGraph()
 
         # noise model for measurements -- one pixel in u and v
@@ -93,9 +105,12 @@ class BundleAdjustmentOptimizer:
         if self._robust_measurement_noise:
             measurement_noise = gtsam.noiseModel.Robust(gtsam.noiseModel.mEstimator.Huber(1.345), measurement_noise)
 
+        is_fisheye_calibration = isinstance(initial_data.get_camera(0), PinholeCameraCal3Fisheye)
+
         # Create a factor graph
 
         # Add measurements to the factor graph
+        sfm_factor_class = GeneralSFMFactor2Cal3Fisheye if is_fisheye_calibration else GeneralSFMFactor2Cal3Bundler
         for j in range(initial_data.number_tracks()):
             track = initial_data.get_track(j)  # SfmTrack
             # retrieve the SfmMeasurement objects
@@ -104,7 +119,7 @@ class BundleAdjustmentOptimizer:
                 i, uv = track.measurement(m_idx)
                 # note use of shorthand symbols C and P
                 graph.add(
-                    GeneralSFMFactor2Cal3Bundler(
+                    sfm_factor_class(
                         uv,
                         measurement_noise,
                         X(i),
@@ -114,8 +129,9 @@ class BundleAdjustmentOptimizer:
                 )
 
         # get all the valid camera indices, which need to be added to the graph.
-        valid_camera_indices = initial_data.get_valid_camera_indices()
+        valid_camera_indices: List[int] = initial_data.get_valid_camera_indices()
 
+        # TODO: add pose priors to the graph
         # Add a prior on first pose. This indirectly specifies where the origin is.
         graph.push_back(
             PriorFactorPose3(
@@ -126,12 +142,17 @@ class BundleAdjustmentOptimizer:
         )
 
         # add prior on all calibrations
+        calibration_prior_factor_class = PriorFactorCal3Fisheye if is_fisheye_calibration else PriorFactorCal3Bundler
+        calibration_prior_factor_dof = CAM_CAL3FISHEYE_DOF if is_fisheye_calibration else CAM_CAL3BUNDLER_DOF
+        calibration_prior_noise_sigma = (
+            CAM_CAL3FISHEYE_PRIOR_NOISE_SIGMA if is_fisheye_calibration else CAM_CAL3BUNDLER_PRIOR_NOISE_SIGMA
+        )
         for i in valid_camera_indices[: 1 if self._shared_calib else len(valid_camera_indices)]:
             graph.push_back(
-                PriorFactorCal3Bundler(
+                calibration_prior_factor_class(
                     K(self.__map_to_calibration_variable(i)),
                     initial_data.get_camera(i).calibration(),
-                    gtsam.noiseModel.Isotropic.Sigma(CAM_CAL3BUNDLER_DOF, CAM_CAL3BUNDLER_PRIOR_NOISE_SIGMA),
+                    gtsam.noiseModel.Isotropic.Sigma(calibration_prior_factor_dof, calibration_prior_noise_sigma),
                 )
             )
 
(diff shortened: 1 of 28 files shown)


gtsfm/bundle/bundle_adjustment.py
```diff
@@ -177,13 +198,16 @@ def __optimize_factor_graph(self, graph: NonlinearFactorGraph, initial_values: V
     def run(
         self,
         initial_data: GtsfmData,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
akshay-krishnan, line 202: I think since we decided on adding the IMU pose variables in bundle adjustment, we will need another input for the hard constraint between imu and camera. I think we can call it fixed_body_T_camera? 
Can we add those in the same PR? Might have to modify loader as well.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -243,50 +271,59 @@ def evaluate(
...
-            poses_gt = [cam.pose() for cam in cameras_gt]
+        poses_gt = [cam.pose() if cam is not None else None for cam in cameras_gt]
 
-            # align the sparse multi-view estimate after BA to the ground truth pose graph.
-            aligned_filtered_data = filtered_data.align_via_Sim3_to_poses(wTi_list_ref=poses_gt)
-            ba_pose_error_metrics = metrics_utils.compute_ba_pose_metrics(
-                gt_wTi_list=poses_gt, ba_output=aligned_filtered_data
-            )
-            ba_metrics.extend(metrics_group=ba_pose_error_metrics)
+        non_none_poses_count = len(poses_gt) - poses_gt.count(None)
```
akshay-krishnan, line 276: nit : valid_poses_count?

gtsfm/data_association/data_assoc.py
```diff
@@ -133,6 +128,7 @@ def run(
         ] / len(tracks_2d)
 
         # pick only the largest connected component
+        # TODO: remove this for hilti as disconnected components not an issue?
```
akshay-krishnan, line 131: not sure what you mean, disconnected components are an issue right?

gtsfm/scene_optimizer.py
```diff
@@ -112,7 +114,10 @@ def create_computation_graph(
         image_graph: List[Delayed],
         camera_intrinsics_graph: List[Delayed],
         image_shape_graph: List[Delayed],
-        gt_cameras_graph: Optional[List[Delayed]] = None,
+        relative_pose_priors: Dict[Tuple[int, int], Delayed],
+        absolute_pose_priors: List[Delayed],
+        gt_cameras_graph: List[Delayed],
+        gt_poses_graph: List[Delayed],
```
akshay-krishnan, line 120: the pose is already in the camera, so you would not need an additional argument for it I think.

gtsfm/scene_optimizer.py
```diff
@@ -225,24 +227,19 @@ def create_computation_graph(
...
+                    metric_group_name="verifier_summary_{}".format(tag),
                 )
+            )
 
         # aggregate metrics for multiview optimizer
         if optimizer_metrics_graph is not None:
             metrics_graph_list.extend(optimizer_metrics_graph)
 
         # Modify BA input, BA output, and GT poses to have point clouds and frustums aligned with x,y,z axes.
-        gt_poses_graph = (
```
akshay-krishnan, line 242: remove the argument and keep this code?

gtsfm/scene_optimizer.py
```diff
@@ -225,24 +227,19 @@ def create_computation_graph(
...
                 )
             )
-            if gt_cameras_graph is not None:
-                metrics_graph_list.append(
-                    dask.delayed(two_view_estimator.aggregate_frontend_metrics)(
-                        report_dict,
-                        self._pose_angular_error_thresh,
-                        metric_group_name="verifier_summary_{}".format(tag),
-                    )
+            metrics_graph_list.append(
```
akshay-krishnan, line 230: does this work well if gt_cameras == None?

gtsfm/two_view_estimator.py
```diff
@@ -257,56 +296,52 @@ def create_computation_graph(
...
-                gt_wTi2_graph,
-                gt_scene_mesh_graph,
-            )
-        else:
-            pre_ba_R_error_deg, pre_ba_U_error_deg = None, None
-            post_ba_R_error_deg, post_ba_U_error_deg = None, None
-            pre_ba_inlier_mask_wrt_gt, pre_ba_reproj_error_wrt_gt = None, None
-            post_ba_inlier_mask_wrt_gt, post_ba_reproj_error_wrt_gt = None, None
+
+        pre_ba_R_error_deg, pre_ba_U_error_deg = dask.delayed(compute_relative_pose_metrics, nout=2)(
```
akshay-krishnan, line 313: do these metrics provide the expected output when ground truth is None?

gtsfm/utils/graph.py
```diff
@@ -53,6 +55,8 @@ def prune_to_largest_connected_component(
         Subset of unit_translations which are in the largest connected components.
     """
     input_edges = [k for (k, v) in rotations.items() if v is not None]
+    # TODO: start using pose priors by uncommenting the following line
+    # input_edges += [k for (k, v) in pose_priors.items() if v is not None]
```
akshay-krishnan, line 59: are you merging before this TODO is done?

gtsfm/two_view_estimator.py
```diff
@@ -398,10 +436,14 @@ def compute_relative_pose_metrics(
         Rotation error, in degrees
         Unit translation error, in degrees
     """
-    R_error_deg = comp_utils.compute_relative_rotation_angle(i2Ri1_computed, i2Ti1_expected.rotation())
-    U_error_deg = comp_utils.compute_relative_unit_translation_angle(
-        i2Ui1_computed, Unit3(i2Ti1_expected.translation())
-    )
+    if wTi1_expected is not None and wTi2_expected is not None:
+        i2Ti1_expected = wTi2_expected.between(wTi1_expected)
```
akshay-krishnan, line 440: cant the relative pose be passed in here, instead of computing it here? this function only needs relative pose, why ask for more?

gtsfm/scene_optimizer.py
ayushbaid: Yep. Individual functions are supposed to handle missing GT cameras now.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -85,17 +92,25 @@ def __init__(
     def __map_to_calibration_variable(self, camera_idx: int) -> int:
         return 0 if self._shared_calib else camera_idx
 
-    def __construct_factor_graph(self, initial_data: GtsfmData) -> NonlinearFactorGraph:
+    def __construct_factor_graph(
+        self,
+        initial_data: GtsfmData,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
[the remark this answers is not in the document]
ayushbaid: I'd like it to be just a dict without the None entries, but I cannot do it in the current dask node construction.

```
return {(i1, i2): dask.delayed(self.get_relative_pose_prior)(i1, i2) for i1, i2 in pairs}
```

To not have None entries, I have to filter before making them delayed. What do you think?

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Do you propose adding it as a third argument?

gtsfm/data_association/data_assoc.py
ayushbaid: If we have priors between all IMU poses, even if there is are disconnected components at the data association step, a single BA can still work.

gtsfm/scene_optimizer.py
ayushbaid: Its tricky to handle that if some cameras are None. I had to handle helper functions to extract poses. For now, having a dedicated pose graph is a stopgap.

gtsfm/two_view_estimator.py
ayushbaid: They give Nones I think.

gtsfm/utils/graph.py
ayushbaid: Yes. Any logic related change is out-of-scope here. I just want to get all the required inputs to the classes for now.

gtsfm/two_view_estimator.py
ayushbaid: I have to compute it somewhere. I do not want to have this computation in the scene_optimizer class like we were doing before.

dellaert (commented):
Cool, I’ll trust Akshay for the detailed review. 
How do we know it works and how to use it? Can you paste a command line and maybe some output? If you have a notebook feel free to check it in as well, otherwise I’ll probably create it.

dellaert:
Seems CI is failing, BTW

gtsfm/data_association/data_assoc.py
akshay-krishnan: Yeah, so we'd have to make that check configurable.

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: Yes

akshay-krishnan: it would be nice if we could indicate that these priors are in the body frame, not the camera frame. Or maybe leave this change to the bundle adjustment PR and only plumb the prior until MVO? 

I was not thinking of the two-view BA use case, only the final BA use. So it would be tricky. In two-view BA, the relative pose prior will be between cameras and in final BA between IMU's? Or will we change two-view BA to include IMU poses as well?

gtsfm/bundle/bundle_adjustment.py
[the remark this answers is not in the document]
akshay-krishnan: `return {(i1, i2): dask.delayed(self.get_relative_pose_prior)(i1, i2) for i1, i2 in pairs}` 
Is that from the loader? Yes, I think the iteration over pairs can be done inside the delayed function. 
But I dont feel too strongly about it, upto you.

gtsfm/scene_optimizer.py
akshay-krishnan: we have to revisit this, we dont need gt_cameras then, and only gt_calibration would suffice? I think @travisdriver unified them in some PR?

gtsfm/two_view_estimator.py
akshay-krishnan: Yeah just wanted to be sure. Those if statements might have been there for a reason.

gtsfm/two_view_estimator.py
akshay-krishnan: okay

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Yeah its a bit tricky. Let's list down what priors are to be used:

## 2-view BA:
- prior on i2Ti1: available as relative_pose_priors. Hence, relative pose priors are always between cameras.


## BA:
- prior/hard-constraint between camera and IMU
- relative constraints between IMU.

I think we need a way of representing IMU in the BA graph, maybe write something specific to hilti.

gtsfm/scene_optimizer.py
ayushbaid: It doesnt work when the camera is None.

gtsfm/data_association/data_assoc.py
ayushbaid: Sounds good. Let me do that in a separate PR though.

ayushbaid:
Command used: `python gtsfm/runner/run_scene_optimizer_hilti.py --dataset_dirpath /media/ayush/cross_os1/dataset/hilti  --max_length 100 --max_frame_lookahead 1 --config deep_front_end --matching_regime sequential_hilti --threads_per_worker 2 --num_workers 1`

Report:
[hilti_plumbingPr.pdf](https://github.com/borglab/gtsfm/files/8646355/hilti_plumbingPr.pdf)

dellaert:
> Report: [hilti_plumbingPr.pdf](https://github.com/borglab/gtsfm/files/8646355/hilti_plumbingPr.pdf)

Yikes. We should probably meet to look at image pairs, inliers, etc. Also concerning that with a supposedly strong backbone from SLAM we don't get all rotations.

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: Yeah, I think have another class for rig BA might be better, let us discuss that in the BA PR. Thanks for making the note here!

akshay-krishnan (approved):
Looks good, feel free to merge if you have verified that it runs without any GT and the CI is producing some valid metrics (just to be sure nothing is broken).

ayushbaid:
> 

Two points about this:
1. We are not using the prior information anywhere in the PR.
2. We are using the FAST-LIO output as ground truth, which may not be as accurate as the real ground truth.

ayushbaid:
> Looks good, feel free to merge if you have verified that it runs without any GT and the CI is producing some valid metrics (just to be sure nothing is broken).

Yep its working fine. Let me merge it and add followup PRs.

akshay-krishnan:
Actually Ayush, looks like some datasets are failing due to the API change: https://github.com/borglab/gtsfm/runs/6339199083?check_suite_focus=true

ayushbaid:
I ran all the runners locally (olsson, hilti, astronet, colmap) and checked the artifacts. Things look good. Hence I am merging.

PR #479, opened 2022-05-08
Disabling skydio-501

ayushbaid:
Disabling skydio 501 for quicker turnaround on the CI


dellaert:
If we can make good decisions on merging PRs without this dataset, maybe it *is* a good idea to turn off until after Hilti...

dellaert (approved):
I will stop CI on this an merge immediately.

ayushbaid:
Thanks Frank!

dellaert:
> Thanks Frank!

Still interested in answer to question: why are we *saving* a lot of images? And can we turn that off, for *all* benchmark runs?

ayushbaid:
We are not sharing images of 2-view correspondences. There should be just
plots of poses and landmarks. <10 I think. Let me check again.

dellaert:
Maybe it *was* downloading the dataset to the compute node.

PR #480, opened 2022-05-08
Using relative pose priors in bundle adjustment

ayushbaid:
In this PR, we start using relative pose priors in BA by adding `BetweenFactorsPose3` in the factor graph.

Some details:
1. Right now, we distinguish priors as hard (coming from rig geometry) and soft (cameras at different timestamp). As we do not have the covariance information from FASTLIO output, we are using a covariance of `1e-3` for hard priors and `3e-2` for soft priors.
  1.1 Should we use robust noise model for soft priors? 
2. There were two places in the multi-view optimizer where we were reducing the pose graph to a single largest connected component. I have added the use of edges from relative priors at one place, and adding them in the data association module is WIP.

Some notes about performance:
1. Two view results seem to 'improve'. However, improve here means have lower error compared to FASTLIO output (which we use as ground truth) as we are using the between factor in the 2-view BA factor graph.
2.Averaging is performing really poorly. Maybe we should try an experiment where we bypass the averaging and just use FASTLIO output as global pose initializations?

Report:
[hilti_relativePriorsInBA.pdf](https://github.com/borglab/gtsfm/files/8647629/hilti_relativePriorsInBA.pdf)


dellaert:
Question: is this with the new "star graph" ?

image

akshay-krishnan:
> Question: is this with the new "star graph" ?
> 
> image

I think Ayush's changes here are with 2-view bundle adjustment in mind. The pose priors are still not passed to the final bundle adjustment. 
These two have quite a few differences for Hilti and it may not be feasible to use the same class for both. Ayush's current suggestion is to use a new class RigBundleAdjustment for the final bundle adjustment using the "star graph".

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -56,6 +57,8 @@
 CAM_CAL3BUNDLER_PRIOR_NOISE_SIGMA = 1e-5  # essentially fixed
 CAM_CAL3FISHEYE_PRIOR_NOISE_SIGMA = 1e-5  # essentially fixed
 MEASUREMENT_NOISE_SIGMA = 1.0  # in pixels
+HARD_POSE_PRIOR_SIGMA = 1e-3  # 1e-5 did not work as well
+SOFT_POSE_PRIOR_SIGMA = 3e-2  # 1e-5 did not work as well
```
akshay-krishnan, line 61: can these sigmas instead be defined wherever we create the pose prior? 
We could create an artificial covariance matrix using these instead of hardcoding here.

dellaert: Is this not almost automatic with Hydra? It would be nice to have an explanation of how Hydra works in GTSFM for collaborators.

ayushbaid: Yes I can define them where I am creating the `PosePrior` object. I thought to just hardcode it as its a temporary solution.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -130,6 +133,25 @@ def __construct_factor_graph(
 
         # get all the valid camera indices, which need to be added to the graph.
         valid_camera_indices: List[int] = initial_data.get_valid_camera_indices()
+        for (i1, i2), i2Ti1_prior in relative_pose_priors.items():
```
dellaert, line 136: This is becoming a very long function. Factor graphs contain shared pointers and hence it is easy to say:
```
graph.push_back(self._reprojection_factors(...))
graph.push_back(self._pose_priors(...))
graph.push_back(self._between_factors(...))
```

ayushbaid:
@dellaert can you please comment on things we can do to improve Shonan?

Input to shonan:
image


Output of shonan:
image


2-view BA with priors from FASTLIO gives very low errors in relative rotations, but global rotations estimated by Shonan are far off from the rotations computed by FASTLIO.

gtsfm/bundle/bundle_adjustment.py
dellaert: PS, saw your commit. Stylewise, maybe avoid `get`, `construct`, `construction` (?).

`graph.push_back(self._pose_priors(...))`

is more readable than

`graph.push_back(self._construct_pose_priors(...))`

PS, why do y'all have *double* `_` in front of private methods? I thought it was just one `_`

dellaert:
> 2-view BA with priors from FASTLIO gives very low errors in relative rotations, but global rotations estimated by Shonan are far off from the rotations computed by FASTLIO.

Hmmm. Gut feeling, you're giving mixed messages to Shonan :-) I.e., a bug (transpose somewhere?) in input. A lot changed so that would be my first guess.

dellaert:
> > 2-view BA with priors from FASTLIO gives very low errors in relative rotations, but global rotations estimated by Shonan are far off from the rotations computed by FASTLIO.
> 
> Hmmm. Gut feeling, you're giving mixed messages to Shonan :-) I.e., a bug (transpose somewhere?) in input. A lot changed so that would be my first guess.

But then again, I am assuming you are evaluating the input error wrpt to your input. So, hmmmmm.

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Thanks. Will do.

`__` is used to model private functions, which should not be overriden. `_` is meant for functions which are designed to be overriden in the children classes.

ayushbaid:
> Question: is this with the new "star graph" ?
> 
> image

This star graph is not modeled in class `BundleAdjustmentHiltiOptimizer`. Please check.

@dellaert @akshay-krishnan

dellaert:
> > > 2-view BA with priors from FASTLIO gives very low errors in relative rotations, but global rotations estimated by Shonan are far off from the rotations computed by FASTLIO.
> > 
> > 
> > Hmmm. Gut feeling, you're giving mixed messages to Shonan :-) I.e., a bug (transpose somewhere?) in input. A lot changed so that would be my first guess.
> 
> But then again, I am assuming you are evaluating the input error wrpt to your input. So, hmmmmm.

To debug, I'd maybe run with a limited number of poses and print out
- the shonan graph and errors before and after Shonan the last time it succeeds
- the shonan graph and errors before and after Shonan the first time it fails
You can create the shonan graph (if not already created) using shonan.CreateGraphAt(3) and then graph.printErrors()

I can help debug if you have time to meet at 6pm your time, but if you'd rather wait until tomorrow that's fine, too.

dellaert:
> > Question: is this with the new "star graph" ?
> > image
> 
> This star graph is now modeled in class `BundleAdjustmentHiltiOptimizer`. Please check.
> 
> @dellaert @akshay-krishnan

@akshay-krishnan has the latest version from our meeting on Friday. Akshay send a link?

ayushbaid:
> > > > 2-view BA with priors from FASTLIO gives very low errors in relative rotations, but global rotations estimated by Shonan are far off from the rotations computed by FASTLIO.
> > > 
> > > 
> > > Hmmm. Gut feeling, you're giving mixed messages to Shonan :-) I.e., a bug (transpose somewhere?) in input. A lot changed so that would be my first guess.
> > 
> > 
> > But then again, I am assuming you are evaluating the input error wrpt to your input. So, hmmmmm.
> 
> To debug, I'd maybe run with a limited number of poses and print out
> 
>     * the shonan graph and errors before and after Shonan the last time it succeeds
> 
>     * the shonan graph and errors before and after Shonan the first time it fails
>       You can create the shonan graph (if not already created) using shonan.CreateGraphAt(3) and then graph.printErrors()
> 
> 
> I can help debug if you have time to meet at 6pm your time, but if you'd rather wait until tomorrow that's fine, too.

Thanks Frank. 6pm works for me.

akshay-krishnan:
The updated factor graph from our meeting last week is [here](https://drive.google.com/file/d/1i1lPWu-T2tZ1UwQWQEkKKrZFcCUQAN4N/view?ts=62729d6b). 

Sorry for the delay in my response!

ayushbaid:
@dellaert @akshay-krishnan should be ready to review. Please vet the choice of noise models I made in the `BundleAdjusmentHiltiOptimizer`.

dellaert:
@ayushbaid the CI failed on  test_shonan.py. I merged in master and fixed what I could but still fails:

```bash
(gtsfm-v1)$ python tests/averaging/rotation/test_shonan.py
[True, True, True, True]
.[True, True, True, True]
.F[True, True, True]
.[True, True, True]
.[True, True, True]
..[True, True, True]
.
======================================================================
FAIL: test_computation_graph (__main__.TestShonanRotationAveraging)
Test the dask computation graph execution using a valid collection of relative poses.
----------------------------------------------------------------------
  File "tests/averaging/rotation/test_shonan.py", line 113, in test_computation_graph
    self.assertTrue(
AssertionError: False is not true

----------------------------------------------------------------------
Ran 8 tests in 0.011s

FAILED (failures=1)
```

I also could not figure out where the printing of `[True, True, True, True]` etc came from above.

dellaert:
Canceled all bechmark CI as already passed.

gtsfm/utils/geometry_comparisons.py
```diff
@@ -28,16 +27,16 @@ def align_rotations(aRi_list: List[Rot3], bRi_list: List[Rot3]) -> List[Rot3]:
...
-    bRi0 = bRi_list[0]
-    i0Rb = bRi0.inverse()
-
-    # origin_transform -- map the origin of the input list to the reference list
-    aRb = aRi0.compose(i0Rb)
+    aRb_list = [
+        aRi.compose(bRi.inverse()) for aRi, bRi in zip(aRi_list, bRi_list) if aRi is not None and bRi is not None
+    ]
+    if len(aRb_list) > 0:
+        aRb = gtsam.FindKarcherMean(Rot3Vector(aRb_list))
```
dellaert, line 34: awesomesauce!

gtsfm/utils/geometry_comparisons.py
```diff
@@ -189,13 +188,12 @@ def compare_rotations(
...
-        [
-            compute_relative_rotation_angle(aRi, aRi_) < angular_error_threshold_degrees
-            for (aRi, aRi_) in zip(aRi_list, aRi_list_)
-        ]
-    )
+    error = [
+        compute_relative_rotation_angle(aRi, aRi_) < angular_error_threshold_degrees
+        for (aRi, aRi_) in zip(aRi_list, aRi_list_)
+    ]
+    print(error)
```
dellaert, line 195: remove !

tests/data_association/test_data_association.py
```diff
@@ -138,7 +138,12 @@ def verify_triangulation_sharedCal_2poses(self, triangulation_mode: Triangulatio
         )
         da = DataAssociation(min_track_len=3, triangulation_options=triangulation_options)
         triangulated_landmark_map, _ = da.run(
-            len(cameras), cameras, matches_dict, keypoints_list, cameras_gt=[None] * len(cameras)
+            len(cameras),
+            cameras,
+            matches_dict,
+            keypoints_list,
+            cameras_gt=[None] * len(cameras),
+            relative_pose_priors={},
```
dellaert, line 146: Oh, we can just do that? I thought dask needed dict of same size. In that case, please feel free to simplify the `test_shonan` code

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -19,28 +24,58 @@ class RotationAveragingBase(metaclass=abc.ABCMeta):
...
             Global rotations for each camera pose, i.e. wRi, as a list. The number of entries in the list is
                 `num_images`. The list may contain `None` where the global rotation could not be computed (either
                 underconstrained system or ill-constrained system).
         """
 
-    def create_computation_graph(self, num_images: int, i2Ri1_graph: Delayed) -> Delayed:
+    def create_computation_graph(
+        self,
+        num_images: int,
+        i2Ri1_graph: Delayed,
```
dellaert, line 49: BIG question: what is the difference between the graph and the priors? Why don't we just add the priors to the graph?

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -19,28 +24,58 @@ class RotationAveragingBase(metaclass=abc.ABCMeta):
...
         """
 
-        return dask.delayed(self.run)(num_images, i2Ri1_graph)
+        wRis = dask.delayed(self.run)(num_images, i2Ri1_graph, i2Ri1_priors)
+        metrics = dask.delayed(self.evaluate)(wRis, gt_wTi_list)
+
+        return wRis, metrics
+
+    def evaluate(self, wRi_computed: List[Optional[Rot3]], wTi_gt: List[Optional[Pose3]]) -> GtsfmMetricsGroup:
+        wRi_gt = [wTi.rotation() if wTi is not None else None for wTi in wTi_gt]
```
dellaert, line 71: docstring

gtsfm/averaging/rotation/shonan.py
```diff
@@ -87,7 +88,12 @@ def __run_with_consecutive_ordering(
 
         return wRi_list_consecutive
 
-    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
+    def run(
+        self,
+        num_images: int,
+        i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]],
+        i2Ri1_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
dellaert, line 95: adapt docstring below (if we keep priors)

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -128,33 +119,107 @@ def __construct_factor_graph(
...
+        return graph
+
+    def __construct_factor_graph(
+        self,
+        cameras_to_model: List[int],
+        initial_data: GtsfmData,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
+    ) -> NonlinearFactorGraph:
+        is_fisheye_calibration = isinstance(initial_data.get_camera(cameras_to_model[0]), PinholeCameraCal3Fisheye)
```
dellaert, line 204: docstring

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -128,33 +119,107 @@ def __construct_factor_graph(
...
+                    gtsam.noiseModel.Isotropic.Sigma(CAM_POSE3_DOF, CAM_POSE3_PRIOR_NOISE_SIGMA),
+                )
+            )
+
+        return graph
+
+    def __calibration_priors(
+        self, initial_data: GtsfmData, cameras_to_model: List[int], is_fisheye_calibration: bool
+    ) -> NonlinearFactorGraph:
+        graph = NonlinearFactorGraph()
```
dellaert, line 170: docstring

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -128,33 +119,107 @@ def __construct_factor_graph(
...
-        # add prior on all calibrations
+        return graph
+
+    def _pose_priors(
+        self,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        initial_data: GtsfmData,
+        camera_for_origin: gtsfm_types.CAMERA_TYPE,
+    ) -> NonlinearFactorGraph:
+        graph = NonlinearFactorGraph()
```
dellaert, line 150: docstring

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -195,6 +260,18 @@ def __optimize_factor_graph(self, graph: NonlinearFactorGraph, initial_values: V
         result_values = lm.optimize()
         return result_values
 
+    def __cameras_to_model(
+        self,
+        initial_data: GtsfmData,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
+    ) -> List[int]:
+        """Get the cameras which are to be modelled in the factor graph. We are using ability to add initial values as
```
dellaert, line 269: "modeled". But I don;t understand this function

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def __get_camera_type(self, camera_idx: int) -> int:
+        return camera_idx % 5
+
+    def _between_factors(
+        self, relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]], cameras_to_model: List[int]
+    ) -> NonlinearFactorGraph:
+        graph = NonlinearFactorGraph()
+
+        # add a between factor for each camera with its IMU
```
dellaert, line 96: this should be the docstring :-)

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+
+    def _pose_priors(
+        self,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        initial_data: GtsfmData,
+        camera_for_origin: gtsfm_types.CAMERA_TYPE,
+    ) -> NonlinearFactorGraph:
+        graph = NonlinearFactorGraph()
+
+        # TODO(Ayush): start using absolute prior factors.
```
dellaert, line 155: wait, so what *is* this function doing? docstring missing

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+                    gtsam.noiseModel.Isotropic.Sigma(CAM_POSE3_DOF, CAM_POSE3_PRIOR_NOISE_SIGMA),
+                )
+            )
+
+        return graph
+
+    def _initial_values(self, initial_data: GtsfmData) -> Values:
+        initial_values: Values = super()._initial_values(initial_data)
+
+        # add initial values for IMU poses
```
dellaert, line 176: = docstring, say how you compute this

dellaert (commented):
Biggest issue: do we have a need to have two identical inout arguments for shonan? Why not just add the "priors" to the input dict?

About doctrsing: just add one liners at least. No need to expound on all arguments

ayushbaid:
> 

If this is something which needs more discussion, I can just remove it for now. Anyway we are not using priors from input args.

ayushbaid:
> @ayushbaid the CI failed on test_shonan.py. I merged in master and fixed what I could but still fails:
> 
> ```shell
> (gtsfm-v1)$ python tests/averaging/rotation/test_shonan.py
> [True, True, True, True]
> .[True, True, True, True]
> .F[True, True, True]
> .[True, True, True]
> .[True, True, True]
> ..[True, True, True]
> .
> ======================================================================
> FAIL: test_computation_graph (__main__.TestShonanRotationAveraging)
> Test the dask computation graph execution using a valid collection of relative poses.
> ----------------------------------------------------------------------
> Traceback (most recent call last):
>   File "tests/averaging/rotation/test_shonan.py", line 113, in test_computation_graph
>     self.assertTrue(
> AssertionError: False is not true
> 
> ----------------------------------------------------------------------
> Ran 8 tests in 0.011s
> 
> FAILED (failures=1)
> ```
> 
> I also could not figure out where the printing of `[True, True, True, True]` etc came from above.

Both issues should be fixed now.

gtsfm/averaging/rotation/rotation_averaging_base.py
ayushbaid: i2Ri1_graph is coming from the 2-view plate. priors are coming from the loader data. We might have logic in the future which uses both these sources of information?

gtsfm/bundle/bundle_adjustment.py
ayushbaid: This is to figure out what cameras we will model in the factor graph.

For example: lets say we only receive 90 of the 100 input cameras after averaging and data association. If we do have prior information available on the remaining 10 cameras (lets assume the priors connect them to the clique of 90 cameras). Should we add them to the factor graph? 

Right now I am keeping the logic the same, i.e. add only those cameras which are part of a track. But created a function out of it for experimentation.

gtsfm/averaging/rotation/rotation_averaging_base.py
akshay-krishnan: I think graph here refers to the fact that it is a delayed object (and not the pose graph)? Its a terminology we had when we started, I wasnt a big fan.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -195,6 +260,18 @@ def __optimize_factor_graph(self, graph: NonlinearFactorGraph, initial_values: V
         result_values = lm.optimize()
         return result_values
 
+    def __cameras_to_model(
+        self,
+        initial_data: GtsfmData,
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
akshay-krishnan, line 267: why do we need the priors here?

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+from typing import Dict, List, Optional, Tuple
+
+import gtsam
+from gtsam import (
+    BetweenFactorPose3,
+    NonlinearFactorGraph,
+    Pose3,
+    PriorFactorPose3,
+    Values,
+    symbol_shorthand,
```
akshay-krishnan, line 15: nit: move symbol_shorthand above?

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+    def __init__(
+        self,
+        output_reproj_error_thresh: Optional[float] = None,
+        robust_measurement_noise: bool = False,
+        shared_calib: bool = False,
+        max_iterations: Optional[int] = None,
+    ) -> None:
+        super().__init__(output_reproj_error_thresh, robust_measurement_noise, shared_calib, max_iterations)
+
+        # Temporary hack to get relative pose between cameras and IMUs
```
akshay-krishnan, line 76: Do you plan on passing these as inputs?

akshay-krishnan: I think we need a body_T_cam: List[PosePrior] ? The length of the list would provide the number of cameras in the rig. So we would not have to hard code it.

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+
+            b2Tb1 = i2Tb2.inverse() * i2Ti1 * i1Tb1
+            imu_relative_pose_priors[(b1, b2)] = BetweenFactorPose3(
+                B(b2),
+                B(b1),
+                b2Tb1,
+                gtsam.noiseModel.Diagonal.Sigmas(i2Ti1_prior.covariance),
+            )
+
+        logger.info("Adding %d between factors for IMUs", len(imu_relative_pose_priors))
```
akshay-krishnan, line 140: nit: Added?

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+    ) -> NonlinearFactorGraph:
+        graph = NonlinearFactorGraph()
+
+        # TODO(Ayush): start using absolute prior factors.
+        num_priors_added = 0
+
+        if num_priors_added == 0:
+            # Adding a prior to fix origin as no absolute prior exists.
+            rig_for_origin = self.__get_rig_idx(camera_for_origin)
+            wTi = initial_data.get_camera(camera_for_origin).pose()
```
akshay-krishnan, line 161: you could pass this in as the absolute pose prior, so that you dont have to change the API later.

gtsfm/bundle/bundle_adjustment_hilti.py
```diff
@@ -0,0 +1,189 @@
...
+                PriorFactorPose3(
+                    B(rig_for_origin),
+                    wTb,
+                    gtsam.noiseModel.Isotropic.Sigma(CAM_POSE3_DOF, CAM_POSE3_PRIOR_NOISE_SIGMA),
+                )
+            )
+
+        return graph
+
+    def _initial_values(self, initial_data: GtsfmData) -> Values:
```
akshay-krishnan, line 173: rename to _imu_initial_values?

gtsfm/data_association/point3d_initializer.py
```diff
@@ -193,9 +193,9 @@ def execute_ransac_variant(self, track_2d: SfmTrack2d) -> np.ndarray:
                 )
             except RuntimeError:
                 # TODO: handle cheirality exception properly?
-                logger.debug(
```
akshay-krishnan, line 196: I think we may never return the TODO above if we comment this out :)

gtsfm/utils/graph.py
```diff
@@ -55,8 +55,7 @@ def prune_to_largest_connected_component(
         Subset of unit_translations which are in the largest connected components.
     """
     input_edges = [k for (k, v) in rotations.items() if v is not None]
-    # TODO(Ayush): start using pose priors by uncommenting the following line
-    # input_edges += [k for (k, v) in pose_priors.items() if v is not None]
+    input_edges += [k for (k, v) in pose_priors.items() if v is not None]
```
akshay-krishnan, line 58: and also remove the line above?

akshay-krishnan: what above further below, do we use rotations or pose_priors?

gtsfm/bundle/bundle_adjustment_hilti.py
ayushbaid: Yeah that would be optimal, but I guess it requires a little more work.

ayushbaid: And some unknowns too. For example, we will have prior on relative poses of cameras (which we use in 2-view BA) and on relative poses of imu (which we use here)

gtsfm/bundle/bundle_adjustment_hilti.py
ayushbaid: Just adding the prior on 1 IMU to define the origin of the world frame.

gtsfm/bundle/bundle_adjustment_hilti.py
ayushbaid: We could do this 'if' the absolute priors start having priors on IMUs too.

gtsfm/bundle/bundle_adjustment_hilti.py
ayushbaid: Can't. This is an overriden method.

gtsfm/averaging/rotation/rotation_averaging_base.py
ayushbaid: Yes. I think we should not use `_graph` prefix anymore, and hence I have dropped it since a long time. However, we have not done the migration of existing code.

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Ideally I'd like some sophisticated logic here, which decides if the prior information is good enough to optimize for a camera, even if its missing in the initial data (due to upstream failures for that camera)

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -19,28 +24,58 @@ class RotationAveragingBase(metaclass=abc.ABCMeta):
 
     # ignored-abstractmethod
     @abc.abstractmethod
-    def run(self, num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]]) -> List[Optional[Rot3]]:
+    def run(
+        self,
+        num_images: int,
+        i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]],
+        i2Ri1_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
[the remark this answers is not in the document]
ayushbaid: Changed it to i2Ti1. Thanks!

gtsfm/data_association/point3d_initializer.py
ayushbaid: Yes true. But what is the appropriate way to handle it. We are running into so many cheirality exception and the terminal is full of this logging.

gtsfm/utils/graph.py
ayushbaid: This filters the rotations and unit translations down to the largest connected component. The priors are used to add the extra edges when figuring out the connected components. 

For example, the upward facing camera is not connected to any other camera in the pairs given to the two view estimator. But we have strong priors on cam2 to other cams in the same timestamp and hence we should not throw away the connected component of cam2s.

gtsfm/data_association/point3d_initializer.py
dellaert: Let's keep for now, I'm definitely motivated by the logging to try and fix it.

gtsfm/bundle/rig_bundle_adjustment.py
```diff
@@ -0,0 +1,186 @@
...
+            graph.push_back(
+                BetweenFactorPose3(
+                    X(i),
+                    B(rig_idx),
+                    self._cam_T_imu[camera_type],
+                    gtsam.noiseModel.Isotropic.Sigma(CAM_POSE3_DOF, CAM_IMU_POSE_PRIOR_SIGMA),
+                )
+            )
+
+        # translate the relative pose priors between cams to IMUs, and add if not already present
```
dellaert, line 96: Why do we translate them? Camera poses are also variables, right?

gtsfm/bundle/rig_bundle_adjustment.py
```diff
@@ -0,0 +1,186 @@
...
+                PriorFactorPose3(
+                    B(rig_for_origin),
+                    wTb,
+                    gtsam.noiseModel.Isotropic.Sigma(CAM_POSE3_DOF, CAM_POSE3_PRIOR_NOISE_SIGMA),
+                )
+            )
+
+        return graph
+
+    def _initial_values(self, initial_data: GtsfmData) -> Values:
```
dellaert, line 164: Actually, I think we should get the initial values from the SLAM team and initialize the cameras from the IMU pose, right?

dellaert (approved):
I have some questions re. the initialization but I'm OK leaving this to the next PR. Perhaps we could meet to do that next PR together, as we now have the constraints file from Jerred.

gtsfm/loader/hilti_loader.py
```diff
@@ -224,13 +230,13 @@ def get_relative_pose_prior(self, i1: int, i2: int) -> Optional[PosePrior]:
             i2_T_imu: Pose3 = self._cam_T_imu_poses[cam_idx_for_i2]
             i2Ti1 = i2_T_imu.inverse().between(i1_T_imu.inverse())
             # TODO: add covariance
-            return PosePrior(value=i2Ti1, covariance=None, type=PosePriorType.HARD_CONSTRAINT)
+            return PosePrior(value=i2Ti1, covariance=HARD_RELATIVE_POSE_PRIOR_SIGMA, type=PosePriorType.HARD_CONSTRAINT)
```
dellaert, line 233: docstring:
```
If the images are on the same rig, then creates a hard pose prior between the two images, to be used by the two-view estimator. If they are not, we create a soft pose prior derived from the absolute poses (w_T_imu) passed in the constructor.
```

gtsfm/bundle/rig_bundle_adjustment.py
dellaert: I get it now. But we should get them from the constraints file.

gtsfm/bundle/rig_bundle_adjustment.py
ayushbaid: This is resolved now as we don't have IMU pose anywhere. All camera poses will be available from averaging and we will use them.

gtsfm/bundle/rig_bundle_adjustment.py
ayushbaid: Resolved as we do not use IMU anymore.

PR #481, opened 2022-05-08
Split loading and execution

dellaert:
Benchmark workflow is really cool, but I thought it would be good to isolate downloading the dataset and then running GTSFM on it. Several datasets are downloaded twice, but they're the smaller once so might not be worth trying to cache?

### DIFF
--- a/gtsfm/evaluation/visualize_benchmark_comparison.py
+++ b/gtsfm/evaluation/visualize_benchmark_comparison.py
@@ -228,7 +228,8 @@ def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> Non
                 else:
                     percentage_change = metrics_utils.compute_percentage_change(float(master_val), float(branch_val))
 
-                # For some metrics, smaller is better, so this will flip the color to green for reduced values, instead of red:
+                # For some metrics, smaller is better.
+                # Hence, below we flip the color to green for reduced values, instead of red:
                 # exception are outlier errors, which we want to get larger.
                 if "error" in metric_name and "outlier" not in metric_name:
                     percentage_change *= -1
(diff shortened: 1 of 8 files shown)


dellaert:
PS did you all notice Open3D warning, saying should be installed via pip instead?

ayushbaid:
> PS did you all notice Open3D warning, saying should be installed via pip instead?

Yes it has been there for atleast a month. I did not fix it yet though.

ayushbaid (approved):
LGTM

dellaert:
> > PS did you all notice Open3D warning, saying should be installed via pip instead?
> 
> Yes it has been there for atleast a month. I did not fix it yet though.

I could easily add to this PR. Just adding `open3d` to pip list in the environment files, right?

ayushbaid:
Colmap loaders are failing because of `run_scene_optimizer_colmaploader.py: error: argument --colmap_files_dirpath: expected one argument`

dellaert:
Yeah, I think I fixed it.

dellaert:
@ayushbaid should be ready now

gtsfm/evaluation/visualize_benchmark_comparison.py
```diff
@@ -226,9 +228,17 @@ def generate_dashboard(curr_master_dirpath: str, new_branch_dirpath: str) -> Non
...
-                    # smaller is better, so this will flip the color to green for reduced values, instead of red
-                    # exception are outlier errors, which we want to get larger.
+                    percentage_change *= -1
+                elif "outlier" in metric_name and "error" not in metric_name:
+                    percentage_change *= -1
+                elif "EXCEEDS" in metric_name:
+                    percentage_change *= -1
+                elif "failure_ratio" in metric_name:
+                    percentage_change *= -1
+                elif "CHEIRALITY_FAILURE" in metric_name:
```
ayushbaid, line 241: Might not be straightforward as we have reprojection result w.r.t GT cams and w.r.t computed CAMS. Anyway, I am fine with it.

ayushbaid (approved):

dellaert:
Uhg, another flake issue. Will cancel workflows, and also disable the reproducibility workflow until it is fixed. It seems a mostly an module import issue, but some other problems as well.

PR #482, opened 2022-05-10
 Adding special rig-shonan class for the hilti dataset 

ayushbaid:
Using the normal shonan class, we get a disconnected component composed of cam2s in the factor graph, leading to poor rotation averaging results.

After this class, the results are much better. Final BA result on 100 timestamps from Hilti's exp4:
rig_shonan


dellaert (approved):

gtsfm/averaging/rotation/rig_shonan.py
```diff
@@ -0,0 +1,93 @@
...
+from gtsfm.common.pose_prior import PosePrior
+
+
+logger = logger_utils.get_logger()
+
+TEST_DATA_ROOT: Path = Path(__file__).resolve().parent.parent.parent.parent / "tests" / "data"
+HILTI_TEST_DATA_PATH: Path = TEST_DATA_ROOT / "hilti_exp4_small"
+
+
+class RigShonanRotationAveraging(ShonanRotationAveraging):
```
ayushbaid, line 28: TODO: add between factor between cam2s across timestamp.

dellaert (approved):
Cool!

gtsfm/averaging/rotation/rig_shonan.py
ayushbaid: Done

PR #484, opened 2022-05-10
Don't delay priors

dellaert:
I might be mis-understanding dask, but the priors are relatively small objects that we just get from the loader, so it does not make sense to make them delayed. 

- for the relative priors, they are *just* used to instantiate the plate, so we might as well pass in their value when we create those plates.
- the absolute priors: they are all passed as a single block to the bottleneck BA step, so there also no benefit.

I think get_camera_pose, get_camera_shape, get_camera_intrinsics should possibly all be handled the same way, but I wanted to start with this one as relative pose priors were *just* introduced and they introduce $O(N^2)$ "tasks" in Dask.


gtsfm/loader/loader_base.py
```diff
@@ -204,14 +204,26 @@ def get_relative_pose_prior(self, i1: int, i2: int) -> Optional[PosePrior]:
...
+        """Get *all* relative pose priors for i2Ti1
+
+        Args:
+            pairs: all (i1,i2) pairs of image pairs
+
+        Returns:
+            A dictionary of PosePriors (or None) for all pairs.
+        """
+
+        return {pair: self.get_relative_pose_prior(*pair) for pair in pairs}
```
dellaert, line 225: Maybe we need to filter out None just as in Hilti loader

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -306,16 +306,16 @@ def evaluate(
     def create_computation_graph(
         self,
         sfm_data_graph: Delayed,
-        absolute_pose_priors: List[Delayed],
-        relative_pose_priors: Dict[Tuple[int, int], Delayed],
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
dellaert, line 310: Maybe we should insist the Dict has non None values. We now (since this PR) check for the key when creating the plate anyway.

gtsfm/scene_optimizer.py
```diff
@@ -158,7 +159,7 @@ def create_computation_graph(
                 camera_intrinsics_graph[i2],
                 image_shape_graph[i1],
                 image_shape_graph[i2],
-                relative_pose_priors[(i1, i2)],
+                relative_pose_priors[(i1, i2)] if (i1, i2) in relative_pose_priors else None,
```
dellaert, line 162: I have no idea how this worked *before* this PR

gtsfm/bundle/bundle_adjustment.py
ayushbaid: Yes I agree. We can kill the Optional in relative priors.

gtsfm/loader/hilti_loader.py
```diff
@@ -273,22 +271,8 @@ def map_index_to_camera(self, index: int) -> int:
...
-                    pairs.add((i, j))
-                else:
-                    break
-
-        return {(i1, i2): dask.delayed(self.get_relative_pose_prior)(i1, i2) for i1, i2 in pairs}
-
-    def get_all_relative_pose_priors(self) -> Dict[Tuple[int, int], Optional[PosePrior]]:
-        pairs = set(self.get_valid_pairs())
+    def get_relative_pose_priors(self, pairs: List[Tuple[int, int]]) -> Dict[Tuple[int, int], Optional[PosePrior]]:
+        pairs = set(self.get_valid_pairs())  # overwites argument !
```
ayushbaid, line 275: We should not do this. Instead have `pairs = set(pairs)`. In the future, the `is_valid_pair` and `get_valid_pairs` will be killed from the loader and we will have retriever taking care of providing pairs.

gtsfm/loader/loader_base.py
ayushbaid: Yes we do.

gtsfm/multi_view_optimizer.py
```diff
@@ -45,8 +46,8 @@ def create_computation_graph(
         i2Ui1_graph: Dict[Tuple[int, int], Delayed],
         v_corr_idxs_graph: Dict[Tuple[int, int], Delayed],
         intrinsics_graph: List[Delayed],
-        absolute_pose_priors: List[Delayed],
-        relative_pose_priors: Dict[Tuple[int, int], Delayed],
+        absolute_pose_priors: List[Optional[PosePrior]],
+        relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
```
ayushbaid, line 50: We have killed optional here.

gtsfm/loader/loader_base.py
dellaert: Done

gtsfm/scene_optimizer.py
dellaert: Done

gtsfm/loader/hilti_loader.py
dellaert: Done

gtsfm/multi_view_optimizer.py
dellaert: Done

gtsfm/bundle/bundle_adjustment.py
dellaert: Done

dellaert:
Review and approve? (if you still have comments I'll implement them first).

gtsfm/scene_optimizer.py
```diff
@@ -158,7 +159,7 @@ def create_computation_graph(
                 camera_intrinsics_graph[i2],
                 image_shape_graph[i1],
                 image_shape_graph[i2],
-                relative_pose_priors[(i1, i2)],
+                relative_pose_priors.get((i1, i2), default=None),
```
ayushbaid, line 162: I think you can't use the named argument. The CI is failing because of this.

ayushbaid (approved):

PR #487, opened 2022-05-12
Support for priors and initial values in translation averaging

akshay-krishnan:
The priors are currently added by a method in the translation averaging module that assumes they are from the specific rig. 

A better design would be to have the correct priors input to the module after pre-processing them separately for the two-view estimator and all other MVO modules.


gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -184,7 +275,14 @@ def run(
                 w_i2Ui1_inlier_measurements.append(w_i2Ui1)
 
         # Run the optimizer
-        wti_values = TranslationRecovery(w_i2Ui1_inlier_measurements).run(scale_factor)
+        algorithm = TranslationRecovery()
+        w_i2ti1_priors = self.__get_priors_hilti_rig(i2Ti1_priors, wRi_list)
```
dellaert, line 279: We should ensure this is only called for Hilti datasets.

dellaert (approved):
Cool! Make sure you don't break other datasets.

gtsfm/averaging/translation/translation_averaging_base.py
```diff
@@ -33,6 +34,8 @@ def run(
         num_images: int,
         i2Ui1_dict: Dict[Tuple[int, int], Optional[Unit3]],
         wRi_list: List[Optional[Rot3]],
+        i2Ti1_priors: Optional[Dict[Tuple[int, int], PosePrior]] = None,
+        wTi_initial: Optional[Dict[int, PosePrior]] = None,
```
ayushbaid, line 38: You would not get a dictionary here. Atleast not how its currently present in the multiview optimizer. Can we use `wTi_priors=List[Optional[PosePrior]]` instead?

gtsfm/averaging/translation/translation_averaging_base.py
```diff
@@ -56,6 +59,8 @@ def create_computation_graph(
         num_images: int,
         i2Ui1_graph: Delayed,
         wRi_graph: Delayed,
+        i2Ti1_priors: Dict[Tuple[int, int], Delayed] = None,
+        wTi_initial: Dict[Tuple[int], Delayed] = None,
```
ayushbaid, line 63: some comment for here.

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -140,11 +150,91 @@ def compute_inlier_mask(
...
+        
+        If the priors are hard constraints (in the same rig), a hard-coded noise model is used. 
+        If the priors are soft constraints, the covariance from the PosePrior is used.
+
+        Soft constraints are only added between the 3rd rig cameras. 
+        Hard constraints are only added between the 3rd camera and other cameras in same rig.
+
+        Args: 
+            i2Ti1_priors: Relative pose priors between cameras, could be a hard or soft prior.
+            wRi_list: Absolute rotation estimates from Shonan averaging.
```
ayushbaid, line 168: We should drop Shonan averaging and instead say rotation averaging.

gtsfm/averaging/translation/averaging_1dsfm.py
ayushbaid: Maybe create a new class like how we did for Hilti?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -140,11 +150,91 @@ def compute_inlier_mask(
...
+            gtsam.BinaryMeasurementsPoint3 containing Point3 priors in world frame.
+        """
+        if i2Ti1_priors is None:
+            return gtsam.BinaryMeasurementsPoint3()
+
+        NUM_CAMERAS_IN_RIG = 5
+        BODY_FRAME_CAMERA = 2
+
+        def get_prior_in_world_frame(i2, i2Ti1_prior):
+            return wRi_list[i2].rotate(i2Ti1_prior.value).translation()
```
ayushbaid, line 180: I am fine with this for now, but ideally we should have a function which gets the translation in world frame from relative pose. We are doing this operation twice now: once for measurements and once for priors.

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -140,11 +150,91 @@ def compute_inlier_mask(
...
+
+        HARD_CONSTRAINT_NOISE_MODEL = gtsam.noiseModel.Constrained.All(1e4)
+        VALID_HARD_CONSTRAINT_EDGES = [(0, 2), (1, 2), (2, 3), (2, 4)]
+
+        w_i2ti1_priors = gtsam.BinaryMeasurementsPoint3()
+        priors_added = set()
+        for (i1, i2), i2Ti1_prior in i2Ti1_priors.items():
+            if i2Ti1_prior.type == PosePriorType.HARD_CONSTRAINT:
+                if (i1, i2) in VALID_HARD_CONSTRAINT_EDGES:
```
ayushbaid, line 189: You have to get the camera types from i1 and i2, right?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -140,11 +150,91 @@ def compute_inlier_mask(
...
+                            get_prior_in_world_frame(i2, i2Ti1_prior),
+                            HARD_CONSTRAINT_NOISE_MODEL,
+                        )
+                    )
+                    priors_added.add((i2, i1))
+            else:
+                r1 = i1 // NUM_CAMERAS_IN_RIG
+                r2 = i2 // NUM_CAMERAS_IN_RIG
+                c1 = r1 * NUM_CAMERAS_IN_RIG + BODY_FRAME_CAMERA
+                c2 = r2 * NUM_CAMERAS_IN_RIG + BODY_FRAME_CAMERA
```
ayushbaid, line 203: Nit: I think you can do a combined check like how I do in rig shonan. But its down to personal preference I guess.

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -140,11 +150,91 @@ def compute_inlier_mask(
...
+        return initial
+
+    # TODO(ayushbaid): Change wTi_initial to Pose3.
     def run(
         self,
         num_images: int,
         i2Ui1_dict: Dict[Tuple[int, int], Optional[Unit3]],
         wRi_list: List[Optional[Rot3]],
+        i2Ti1_priors: Optional[Dict[Tuple[int, int], PosePrior]] = None,
+        wTi_initial: Optional[Dict[Tuple[int], PosePrior]] = None,
```
ayushbaid, line 237: I think we should not make them optional.

akshay-krishnan:
@ayushbaid I am not sure that creating a Hilti-specific translation averaging class is the best solution. 
I think an algorithm implementation (or node in the gtsfm graph) should be dataset agnostic. This PR a temporary solution, it hardcodes the Hilti-rig parameters in this module and that is not desirable (another class would do the same). 

Instead I think we should have some logic to select the priors for the averaging modules for any dataset (both rotation and translation), along with logic to select priors for the two-view BA. These need not be same. 
The logic can perhaps live in the loader? And we can have an implementation of that logic for Hilti. 
WDYT?

dellaert:
> @ayushbaid I am not sure that creating a Hilti-specific translation averaging class is the best solution. I think an algorithm implementation (or node in the gtsfm graph) should be dataset agnostic. This PR a temporary solution, it hardcodes the Hilti-rig parameters in this module and that is not desirable (another class would do the same).
> 
> Instead I think we should have some logic to select the priors for the averaging modules for any dataset (both rotation and translation), along with logic to select priors for the two-view BA. These need not be same. The logic can perhaps live in the loader? And we can have an implementation of that logic for Hilti. WDYT?

Akshay, we are running out of time. I think the cleanest and fastest way *now* is to create a Hilti-specific class, which hydra automatically supports, right? Then we can think about integrating any logic into the pipeline later, or at least make it work for *all* rigs.

gtsfm/averaging/translation/averaging_1dsfm.py
dellaert: Done

gtsfm/averaging/translation/translation_averaging_base.py
dellaert: now `Optional[List[Optional[Pose3]]]`

gtsfm/averaging/translation/translation_averaging_base.py
dellaert: now `Optional[List[Optional[Pose3]]]`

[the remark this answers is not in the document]
dellaert: Done

dellaert:
OK, @akshay-krishnan , I made everything consistent and uploaded the wheels. We'll see what CI does...

dellaert:
Benchmarks seem fine but there is a unit test failure, will push fix

dellaert:
Canceling benchmarks, will merge after unit tests pass @akshay-krishnan

dellaert:
Works !!! Merging!

akshay-krishnan:
I just saw that I had a local change to multi_view_optimizer that calls translation averaging with the priors that I had not pushed in the hurry. Thank you for the fixes!

PR #488, opened 2022-05-12
Fix Vulnerabilities

dellaert:
I upgraded most packages, though this not work on my M1 mac because of numjs.
@adisingh50 Do we really need *all* of numjs? Or is there a smaller matrix package we could use?


adisingh50:
Hello, from what I remember numjs was the best, credible package I could find to support matrix operations in JS (there weren't many since JS is typically not used for that type of computation). So from my knowledge, I don't believe there is a lighter weight substitute for numjs. However, since I have not worked on the React visualization tool in a while, you can feel free to remove it if you don't plan on contuining to develop the React visualization tool in GTSFM.

PR #492, opened 2022-05-13
Fixing stack of delayed calls inside two-view-estimator

ayushbaid:
Replacing 10 `dask.delayed` task creations (and many more `getitem` tasks due to `nout` > 1 ) with a single `dask.delayed` use and 4 `getitem` calls. 

Anecdotally, my waiting time in building the dask graph for 100 rig indices (500 images) has dropped from ~10mins to 2.5 mins.

### DIFF
--- a/gtsfm/runner/frontend_runner.py
+++ b/gtsfm/runner/frontend_runner.py
@@ -62,9 +62,6 @@ def run_frontend(
             camera_intrinsics_i2=camera_intrinsics[i2],
             im_shape_i1=image_shapes[i1],
             im_shape_i2=image_shapes[i2],
-            i2Ti1_prior=dask.delayed(None),
-            gt_wTi1_graph=dask.delayed(None),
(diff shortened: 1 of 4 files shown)


dellaert (approved):
This is *amazing*. My compute time went from 1hr to 14 minutes. and *most* of it is transfer, I suspect of the two-view reports. But I have an idea about that as well.

ayushbaid:
> 

Kill the two-view reports for the hilti challenge?

dellaert:
> Kill the two-view reports for the hilti challenge?

No, i had another plan. But that might work as well. Is that not just a flag?

dellaert:
Note, fixing a small issue and then I will merge, ok?

ayushbaid:
> Note, fixing a small issue and then I will merge, ok?

Sounds good. The unit tests are failing because I did not fix them in the night. Sorry about that.

PR #493, opened 2022-05-14
Adding subsampling in hilti loader

ayushbaid:
Adding an optional subsampling flag to control subsampling by 3. This is to reduce number of pairs.

### DIFF
--- a/gtsfm/averaging/rotation/shonan.py
+++ b/gtsfm/averaging/rotation/shonan.py
@@ -10,7 +10,7 @@ References:
 
 Authors: Jing Wu, Ayush Baid, John Lambert
 """
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Set, Tuple
 
 import gtsam
 import numpy as np
@@ -112,6 +112,18 @@ class ShonanRotationAveraging(RotationAveragingBase):
 
         return wRi_list_consecutive
 
+    def _valid_nodes(
+        self, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]], i2Ti1_priors: Dict[Tuple[int, int], PosePrior]
+    ) -> Set[int]:
+        """Gets the nodes with edges which are to be modelled as between factors."""
+
+        unique_nodes_with_edges = set()
+        for (i1, i2) in i2Ri1_dict.keys():
+            unique_nodes_with_edges.add(i1)
+            unique_nodes_with_edges.add(i2)
+
+        return unique_nodes_with_edges
+
     def run(
         self,
         num_images: int,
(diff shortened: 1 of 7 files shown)


gtsfm/averaging/rotation/shonan.py
```diff
@@ -112,6 +112,18 @@ def _run_with_consecutive_ordering(
 
         return wRi_list_consecutive
 
+    def _valid_nodes(
```
akshay-krishnan, line 115: nit: maybe call it _unique_nodes?

gtsfm/loader/hilti_loader.py
```diff
@@ -49,13 +49,12 @@
 SOFT_RELATIVE_POSE_PRIOR_SIGMA = np.ones((6,)) * 3e-2
 SOFT_ABSOLUTE_POSE_PRIOR_SIGMA = np.ones((6,)) * 3e-2
 
+SUBSAMPLE_FACTOR: int = 3
```
akshay-krishnan, line 52: should we make this configurable?

gtsfm/loader/hilti_loader.py
```diff
@@ -301,7 +308,7 @@ def image_from_rig_and_camera(self, rig_index: int, camera_idx: int) -> int:
     def get_relative_pose_priors(self, pairs: List[Tuple[int, int]]) -> Dict[Tuple[int, int], PosePrior]:
         pairs = set(pairs)
         # For every rig index, add a "star" from camera 2 to 0,1,3,4:
-        for rig_index in range(self.num_rig_poses):
+        for rig_index in range(0, self.num_rig_poses, SUBSAMPLE_FACTOR if self._subsample else 1):
```
akshay-krishnan, line 311: why are we subsampling the relative between factors?

akshay-krishnan: AFAIK we only subsample the indices at which we match, but keep these relative poses? So that final result is at 10Hz?

ayushbaid: No these are the start factors. If I add them, the cameras would just be dangling cameras without any contribution to the factor graph.

All Cam2-cam2 edges are coming from retriever and are at 10Hz.

gtsfm/runner/run_scene_optimizer_hilti.py
```diff
@@ -34,13 +33,17 @@ def construct_argparser(self) -> argparse.ArgumentParser:
         )
 
         parser.add_argument("--max_length", type=int, default=None, help="Max number of timestamps to process")
+        parser.add_argument(
+            "--subsample", action="store_true", help="Subsample the timestamps by 5 (pick every 5th rig for visual SfM"
```
dellaert, line 37: Make into an int, default 1 which is no subsampling

dellaert (commented):
Not sure how it will work. What happens to all images on skipped rigs?
Very confused.

dellaert:
I think this is obsolete?

PR #496, opened 2022-05-15
Transform covariance from IMU to camera frame

akshay-krishnan:
Transforms the covariance from IMU to camera frame as derived here: https://colab.research.google.com/drive/1l3-dLcPVu3Qk9vmK67t078JmE7jaEVIk#scrollTo=pux69HEkWx2I

We store the Jacobian once in the constructor, and re-use it for all priors.


dellaert (approved):
Awesome!!!!

PR #498, opened 2022-05-15
Add method in loader to filter 1-step relative poses

akshay-krishnan:
Also added unit test, seems to work.


dellaert (approved):
Did not look too closely...

dellaert:
@akshay-krishnan merge?

PR #500, opened 2022-05-15
Feature/2 view

dellaert:
- Streamlined reprojection errors
- calculate mean outside

### DIFF
--- a/gtsfm/utils/io.py
+++ b/gtsfm/utils/io.py
@@ -476,7 +476,8 @@ def write_points(gtsfm_data: GtsfmData, save_dir: str, images: Optional[List[Ima
             track = gtsfm_data.get_track(j)
 
             r, g, b = image_utils.get_average_point_color(track, images)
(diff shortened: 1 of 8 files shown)


gtsfm/utils/io.py
```diff
@@ -438,7 +439,7 @@ def read_points_txt(fpath: str) -> Tuple[Optional[np.ndarray], Optional[np.ndarr
 
 def read_scene(
     images_fpath: str, cameras_fpath: str, points_fpath: str
-) -> Tuple[List[Pose3], List[str], List[Cal3Bundler], np.ndarray, np.ndarray]:
+) -> Tuple[Optional[List[Pose3]], Optional[List[str]], List[Cal3Bundler], Optional[np.ndarray], Optional[np.ndarray]]:
```
ayushbaid, line 442: Why is Cal3Bundler the only thing here which is not optional?

ayushbaid (approved):

gtsfm/utils/io.py
dellaert: No idea

PR #504, opened 2022-05-16
Loose covariance for filtered priors

akshay-krishnan:
Instead of removing outlier priors, and a high covariance edge, so that the graph does not break.

### DIFF
--- a/gtsfm/loader/hilti_loader.py
+++ b/gtsfm/loader/hilti_loader.py
@@ -129,7 +129,7 @@ class HiltiLoader(LoaderBase):
 
         filtered_constraints = {}
         rot_sigma = np.deg2rad(60)
-        trans_sigma = 50 # meters
+        trans_sigma = 10 # meters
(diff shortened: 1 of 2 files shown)


ayushbaid (approved):

gtsfm/loader/hilti_loader.py
```diff
@@ -121,20 +121,24 @@ def _filter_outlier_constraints(self, constraints: Dict[Tuple[int, int], Constra
...
+            ERROR_MARGIN = 0.1  # meters
             c, d = min(a, b), max(a, b)
             this_magnitude = all_magnitudes[(c, d)]
-            step_magnitude = all_magnitudes[(c, d+step_size)] if (c, d+step_size) in all_magnitudes else None
+            step_magnitude = all_magnitudes[(c, d + step_size)] if (c, d + step_size) in all_magnitudes else None
             return step_magnitude is not None and this_magnitude - step_magnitude > ERROR_MARGIN
 
         filtered_constraints = {}
+        rot_sigma = np.deg2rad(60)
+        trans_sigma = 50 # meters
```
dellaert, line 132: use 10?

dellaert (approved):

PR #507, opened 2022-05-16
Adding option to use GT poses as initial values in pose slam

ayushbaid:
The boolean is available in the config files as `use_gt_for_initialization`

### DIFF
--- a/gtsfm/pose_slam/pose_slam.py
+++ b/gtsfm/pose_slam/pose_slam.py
@@ -34,7 +34,7 @@ class PoseSlam:
             for i, wTi in enumerate(gt_wTi_list):
                 if wTi is None:
                     logger.error("None GT camera encountered at idx %d", i)
-                    continue
+                    raise ValueError("Need all GTs to work")
 
                 initial_values.insert(i, wTi)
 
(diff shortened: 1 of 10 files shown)


gtsfm/pose_slam/pose_slam.py
```diff
@@ -20,6 +20,27 @@
...
+        initial_values = gtsam.Values()
+        if self._use_gt_for_initialization and gt_wTi_list is not None:
+            logger.info("Using GT for initialization")
+            for i, wTi in enumerate(gt_wTi_list):
+                if wTi is None:
+                    continue
+
+                initial_values.insert(i, wTi)
+
+        initial_values = gtsam.InitializePose3.initialize(pose_init_graph, initial_values, useGradient=False)
```
dellaert, line 40: No, since this can fail, I think it has to be either or. So this should be in an else block

dellaert (approved):
Approve but crucial you fix my comment I think

ayushbaid:
Running into some issue s  where the pose init graph has some keys which are not present in values. Debugging

ayushbaid:
The PR *runs* now, but it has infinity errors:

> Pose SLAM initialization complete with error: inf`
> Pose SLAM optimization complete with error: inf`

ayushbaid:
The final bundle adjustment also has error infinity. Hence this code should *not* be merged.

PR #524, opened 2022-07-12
Restore working rtf viz tool

kfu02:
The packages and result_metrics/ dir had changed since the last time the frontend React JS website compiled. This PR gets the frontend to a compiling state. (Note: package-lock.json is auto-generated and does not need to be reviewed.)

## Testing
0. Install simplejson with `pip install simplejson`.
1. Run the gtsfm pipeline per the README: `python gtsfm/runner/run_scene_optimizer_olssonloader.py --dataset_root tests/data/set1_lund_door --image_extension JPG --config_name deep_front_end.yaml --num_workers 4`
2. Install and run the rtf tool per [rtf_vis_tool/README.md](https://github.com/borglab/gtsfm/pull/524/files#diff-29321d04f70019d4c6f7ca8ef25583de9ff1ff080c557c623600b0967530df1eR5). Simplified instructions are:
```
cd rtf_vis_tool/
npm install --legacy-peer-deps
npm start
(follow the link it generates)
```
3. Click around the UI that appears.

Installation note: if installing node for the first time, make sure to use the LTS version. If a different version of node is installed, you can use [nvm](https://github.com/nvm-sh/nvm) to switch to LTS:
```
nvm install --lts
nvm use --lts
```


gtsfm/utils/io.py
```diff
@@ -2,7 +2,7 @@
 
 Authors: Ayush Baid, John Lambert
 """
-import json
+import simplejson as json
```
akshay-krishnan, line 5: what happened to native json?

akshay-krishnan (commented):
Could you do a favor and document what needs to be changed in rtf_vis_tool when a metric file/name is changed/added? 
This could be part of the RTF vis tool readme.

gtsfm/utils/io.py
kfu02: Native JSON in Python outputs np.NaN as "NaN" to JSON files. This is not actually part of the JSON spec, so ReactJS throws an error when reading NaNs. The simplejson library has an ignore_nan flag that outputs "null" for all NaNs (and otherwise behaves nearly identically). 

Source: https://stackoverflow.com/questions/6601812/sending-nan-in-json

akshay-krishnan: How does numpy deal with nulls? Can we parse these JSONs? There are some tools that parse them and create plots, and we want to be sure they dont break with nulls.

akshay-krishnan: @travisdriver @johnwlambert I remember you running into some issues with NaNs in these metrics.

kfu02: The JSONs can be parsed, but any "null" is turned into a None in Python (see [here](https://stackoverflow.com/questions/49829286/convert-json-null-to-nan-in-simplejson-load)). This unit test I just wrote encapsulates the behavior in simplejson:
Screenshot from 2022-07-22 21-03-01

How are NaNs displayed in the existing tools? I assume they are not displayed; if so, I can fix the tools to check that the data they are attempting to display is not None. I looked very deeply into changing the way React parses JSON files and was unable to figure out how to force different parsing behavior, which is why I chose this method. (Also, I can check in that unit test once we resolve this to reduce future confusion.)

kfu02:
@akshay-krishnan [Here](https://github.com/borglab/gtsfm/pull/524/commits/51755da8d295e3832775a7f28cc51d36e57f7479) are the instructions you requested. I was unsure of how much detail to add, but I feel that this is adequate since the components I reference are simply hardcoded to translate the JSON into text labels and fields. It would be nicer if they were not coupled like this (which I can look into if that's of importance).

akshay-krishnan (approved):
thanks for adding the IO tests!

kfu02:
I messed up one of my unit tests...fixed now!

PR #526, opened 2022-07-13
Fix Colmap io utils

### LINKED ISSUE #527: Radial distortion ignored when loading colmap data
The implementation of `read_cameras_txt` in `io.py` ignores the radial distortion parameters k1 and k2 when parsing the lines of the `cameras.txt` file ([link](https://github.com/borglab/gtsfm/blob/f96bc07e99ad8b36e6f53c20be9221da97156eec/gtsfm/utils/io.py#L243))
Also, k1 and k2 have their values hardcoded to zero ([link](https://github.com/borglab/gtsfm/blob/f96bc07e99ad8b36e6f53c20be9221da97156eec/gtsfm/utils/io.py#L247))

### LINKED ISSUE #528: Sparse output in COLMAP format can't be used with COLMAP's GUI
The output generated by the `export_model_as_colmap_text` cannot be loaded in COLMAP's GUI. It crashes the application and generates a core dump.

It was identified that the `cameras.txt` file was not complying with the [colmap format](https://colmap.github.io/format.html) because the incompatibility between the number of parameters and camera model. 

According to the [colmap implementation](https://github.com/colmap/colmap/blob/d6f528ab59fd653966e857f8d0c2203212563631/src/base/camera_models.h) the following camera models are possible:
* RADIAL with 5 intrinsic parameters: f, cx, cy, k1, k2 
* SIMPLE_RADIAL with 4 intrinsic parameters:  f, cx, cy, k

[Current gtsfm implementation](https://github.com/borglab/gtsfm/blob/81f81af56cd38104de56a299357c3639a11b3c43/gtsfm/utils/io.py#L291) uses SIMPLE_RADIAL model with 5 parameters, which is incorrect.

luisfelipewb:
Some problems where observed with COLMAP io functions: 
1. Loss of radial distortion parameters.
2. Incompatible camera model tag on `cameras.txt` file.

This pull request includes:
* New unit tests for storing and loading camera data.
* Fix #527 
* Fix #528 
* Minor spelling and code style adjustments

IMPORTANT: only two camera models are supported after this change.

### DIFF
--- a/gtsfm/utils/io.py
+++ b/gtsfm/utils/io.py
@@ -244,7 +244,8 @@ def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
             _, _, img_w, img_h, fx, u0, v0, k1 = cam_params[:8]
             img_w, img_h, fx, u0, v0, k1 = int(img_w), int(img_h), float(fx), float(u0), float(v0), float(k1)
             # Convert COLMAP's SIMPLE_RADIAL to GTSAM's Cal3Bundler:
-            k2 = 0  # add sencond radial distorion coefficient of value zero
+            # Add second radial distortion coefficient of value zero.
+            k2 = 0
             calibrations.append(Cal3Bundler(fx, k1, k2, u0, v0))
         elif model == "RADIAL":
             _, _, img_w, img_h, fx, u0, v0, k1, k2 = cam_params[:9]
--- a/gtsfm/utils/io.py
+++ b/gtsfm/utils/io.py
@@ -17,12 +17,9 @@ import open3d
 from gtsam import Cal3Bundler, Point3, Pose3, Rot3, SfmTrack
 from PIL import Image as PILImage
 from PIL.ExifTags import GPSTAGS, TAGS
-from thirdparty.colmap.scripts.python.read_write_model import \
-    Camera as ColmapCamera
-from thirdparty.colmap.scripts.python.read_write_model import \
-    Image as ColmapImage
-from thirdparty.colmap.scripts.python.read_write_model import \
-    Point3D as ColmapPoint3D
+from thirdparty.colmap.scripts.python.read_write_model import Camera as ColmapCamera
+from thirdparty.colmap.scripts.python.read_write_model import Image as ColmapImage
+from thirdparty.colmap.scripts.python.read_write_model import Point3D as ColmapPoint3D
 
 import gtsfm.utils.images as image_utils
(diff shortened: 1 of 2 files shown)


luisfelipewb:
Adding link to issues:
fix #527
fix #528

dellaert:
@gchenfc take a look at this PR for an example of a GTDFMData roundtrip, and possibly speed up this code in C++ by 100 :-)

tests/utils/test_io_utils.py
```diff
@@ -84,6 +84,8 @@ def test_read_cameras_txt(self) -> None:
         self.assertEqual(K.fx(), 2332.47)
         self.assertEqual(K.px(), 2028)
         self.assertEqual(K.py(), 1520)
+        self.assertEqual(K.k1(), 0.00400066)
```
johnwlambert, line 87: Could you add a small inline that we read in a COLMAP `SIMPLE_RADIAL` model (with 1 radial distortion coefficient), and implicitly converted it to GTSAM's "Cal3Bundler" (with 2 radial distortion coefficients). (The file doesn't contain a k2).

gtsfm/utils/io.py
```diff
@@ -240,13 +237,26 @@ def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
...
-        k1 = 0
-        k2 = 0
-        calibrations.append(Cal3Bundler(fx, k1, k2, u0, v0))
+        model = cam_params[1]
+        # Currently only handles SIMPLE RADIAL and RADIAL camera models
+        assert model in ["SIMPLE_RADIAL", "RADIAL"]
+        if model == "SIMPLE_RADIAL":
+            _, _, img_w, img_h, fx, u0, v0, k1 = cam_params[:8]
+            img_w, img_h, fx, u0, v0, k1 = int(img_w), int(img_h), float(fx), float(u0), float(v0), float(k1)
+            k2 = 0
```
johnwlambert, line 246: Hi Luis, could you also add an inline comment here, saying that if the file contains a COLMAP `SIMPLE_RADIAL` model, we explicitly convert it to GTSAM's "Cal3Bundler" (adding a second radial distortion coefficient of value zero).

johnwlambert (approved):
Looks great, thanks for adding this. This situation occurred because we only ever imported COLMAP output, and only exported GTSFM output.

luisfelipewb:
Thank you @johnwlambert for the review. I added the two comments as suggested in a new commit. It seems that the changes triggered the CI workflow that are awaiting approval.

PR #533, opened 2022-07-16
Add ORB detector-descriptor for front-end.

johnwlambert:


gtsfm/frontend/detector_descriptor/orb.py
```diff
@@ -0,0 +1,57 @@
...
+        Returns:
+            Detected keypoints, with length N <= max_keypoints.
+            Corr. descriptors, of shape (N, D) where D is the dimension of each descriptor.
+        """
+
+        # Convert to grayscale.
+        gray_image = image_utils.rgb_to_gray_cv(image)
+
+        # Create OpenCV object.
+        opencv_obj = cv.ORB_create()
```
akshay-krishnan, line 45: how many detector-descriptors share this pattern? seems to the be same as KAZE, except for this line. Should we unify them with an enum for the method?

akshay-krishnan (approved):

gtsfm/frontend/detector_descriptor/orb.py
johnwlambert: I'm open to that, maybe in a follow-up PR?

PR #536, opened 2022-07-18
Add Least Median of Squares (LMEDS) verifier to front-end.

johnwlambert:
See https://web.ipac.caltech.edu/staff/fmasci/home/astro_refs/LeastMedianOfSquares.pdf

### DIFF
--- a/gtsfm/frontend/verifier/opencv_verifier_base.py
+++ b/gtsfm/frontend/verifier/opencv_verifier_base.py
@@ -1,0 +1,145 @@
+"""
+Base-class for OpenCV verifier implementations.
+
+Authors: John Lambert
+"""
+
+import abc
+from typing import Optional, Tuple
+
+import numpy as np
+from gtsam import Cal3Bundler, Rot3, Unit3
+
+import gtsfm.utils.features as feature_utils
+import gtsfm.utils.logger as logger_utils
+import gtsfm.utils.verification as verification_utils
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.frontend.verifier.verifier_base import NUM_MATCHES_REQ_E_MATRIX, NUM_MATCHES_REQ_F_MATRIX, VerifierBase
+
+logger = logger_utils.get_logger()
+
+
+class OpencvVerifierBase(VerifierBase):
+    def __init__(
+        self,
+        use_intrinsics_in_verification: bool,
+        estimation_threshold_px: float,
+    ) -> None:
+        """Initializes the verifier.
+
+        Args:
+            use_intrinsics_in_verification: Flag to perform keypoint normalization and compute the essential matrix
+                instead of fundamental matrix. This should be preferred when the exact intrinsics are known as opposed
+                to approximating them from exif data.
+            estimation_threshold_px: maximum distance (in pixels) to consider a match an inlier, under squared
+                Sampson distance.
+        """
+        self._use_intrinsics_in_verification = use_intrinsics_in_verification
+        self._estimation_threshold_px = estimation_threshold_px
+        self._min_matches = (
+            NUM_MATCHES_REQ_E_MATRIX if self._use_intrinsics_in_verification else NUM_MATCHES_REQ_F_MATRIX
+        )
+
+        # for failure, i2Ri1 = None, and i2Ui1 = None, and no verified correspondences, and inlier_ratio_est_model = 0
+        self._failure_result = (None, None, np.array([], dtype=np.uint64), 0.0)
+
+    def verify(
+        self,
+        keypoints_i1: Keypoints,
+        keypoints_i2: Keypoints,
+        match_indices: np.ndarray,
+        camera_intrinsics_i1: Cal3Bundler,
+        camera_intrinsics_i2: Cal3Bundler,
+    ) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray, float]:
+        """Performs verification of correspondences between two images to recover the relative pose and indices of
+        verified correspondences.
+
+        Args:
+            keypoints_i1: detected features in image #i1.
+            keypoints_i2: detected features in image #i2.
+            match_indices: matches as indices of features from both images, of shape (N3, 2), where N3 <= min(N1, N2).
+            camera_intrinsics_i1: intrinsics for image #i1.
+            camera_intrinsics_i2: intrinsics for image #i2.
+        
+        Returns:
+            Estimated rotation i2Ri1, or None if it cannot be estimated.
+            Estimated unit translation i2Ui1, or None if it cannot be estimated.
+            Indices of verified correspondences, of shape (N, 2) with N <= N3. These are subset of match_indices.
+            Inlier ratio w.r.t. the estimated model, i.e. #ransac inliers / # putative matches.
+        """
+        if match_indices.shape[0] < self._min_matches:
+            return self._failure_result
+
+        if self._use_intrinsics_in_verification:
+            uv_norm_i1 = feature_utils.normalize_coordinates(keypoints_i1.coordinates, camera_intrinsics_i1)
+            uv_norm_i2 = feature_utils.normalize_coordinates(keypoints_i2.coordinates, camera_intrinsics_i2)
+
+            # OpenCV can fail here, for some reason
+            if match_indices.shape[0] < 6:
+                return self._failure_result
+
+            # use stricter threshold, among the two choices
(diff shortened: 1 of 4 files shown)


gtsfm/frontend/verifier/opencv_verifier_base.py
```diff
@@ -0,0 +1,145 @@
...
+import gtsfm.utils.features as feature_utils
+import gtsfm.utils.logger as logger_utils
+import gtsfm.utils.verification as verification_utils
+from gtsfm.common.keypoints import Keypoints
+from gtsfm.frontend.verifier.verifier_base import NUM_MATCHES_REQ_E_MATRIX, NUM_MATCHES_REQ_F_MATRIX, VerifierBase
+
+logger = logger_utils.get_logger()
+
+
+class OpencvVerifierBase(VerifierBase):
```
ayushbaid, line 22: I didn't get the point of this class to be just OpenCV specific? Do we want these 2 functions for every verifier? If yes, then lets make the APIs part of the base verifier. Otherwise lets use the if-else in the verify function itself?

gtsfm/frontend/verifier/opencv_verifier_base.py
```diff
@@ -0,0 +1,145 @@
...
+            )
+
+        inlier_idxs = np.where(inlier_mask.ravel() == 1)[0]
+
+        v_corr_idxs = match_indices[inlier_idxs]
+        inlier_ratio_est_model = np.mean(inlier_mask)
+        (i2Ri1, i2Ui1) = verification_utils.recover_relative_pose_from_essential_matrix(
+            i2Ei1,
+            keypoints_i1.coordinates[match_indices[inlier_idxs, 0]],
+            keypoints_i2.coordinates[match_indices[inlier_idxs, 1]],
```
ayushbaid, line 106: nit: we can use v_corr_idxs directly here

gtsfm/frontend/verifier/ransac.py
```diff
@@ -11,122 +11,84 @@
...
-from gtsfm.frontend.verifier.verifier_base import VerifierBase, NUM_MATCHES_REQ_E_MATRIX, NUM_MATCHES_REQ_F_MATRIX
+from gtsfm.frontend.verifier.opencv_verifier_base import OpencvVerifierBase
 
 
 RANSAC_SUCCESS_PROB = 0.999999
 RANSAC_MAX_ITERS = 1000000
 
 logger = logger_utils.get_logger()
 
+"""Alternative options:
```
ayushbaid, line 29: These are options just for E matrix, right? Please add options for the F matrix too

ayushbaid (approved):

gtsfm/frontend/verifier/opencv_verifier_base.py
johnwlambert: Thanks, good catch. Updated.

gtsfm/frontend/verifier/opencv_verifier_base.py
```diff
@@ -0,0 +1,145 @@
...
+
+        if self._use_intrinsics_in_verification:
+            uv_norm_i1 = feature_utils.normalize_coordinates(keypoints_i1.coordinates, camera_intrinsics_i1)
+            uv_norm_i2 = feature_utils.normalize_coordinates(keypoints_i2.coordinates, camera_intrinsics_i2)
+
+            # OpenCV can fail here, for some reason
+            if match_indices.shape[0] < 6:
+                return self._failure_result
+
+            # use stricter threshold, among the two choices
```
akshay-krishnan, line 81: what threshold?

tests/frontend/verifier/test_lmeds.py
```diff
@@ -0,0 +1,23 @@
...
+
+class TestLmedsForFundamentalMatrix(test_verifier_base.TestVerifierBase):
+    """Unit tests for the LMEDS verifier w/o intrinsics in verification.
+
+    All unit test functions defined in TestVerifierBase are run automatically.
+    """
+
+    def setUp(self):
+        super().setUp()
+        self.verifier = LMEDS(use_intrinsics_in_verification=False, estimation_threshold_px=0.5)
```
akshay-krishnan, line 19: could you also test the case where use_intrinsics_in_verification = True?

gtsfm/frontend/verifier/opencv_verifier_base.py
```diff
@@ -0,0 +1,145 @@
...
+                print("Out of bounds access w/ keypoints", keypoints_i1.coordinates[:10])
+
+            i2Ei1, inlier_mask = self.estimate_E(
+                uv_norm_i1=uv_norm_i1, uv_norm_i2=uv_norm_i2, match_indices=match_indices, fx=fx
+            )
+        else:
+            i2Fi1, inlier_mask = self.estimate_F(
+                keypoints_i1=keypoints_i1, keypoints_i2=keypoints_i2, match_indices=match_indices
+            )
+            i2Ei1 = verification_utils.fundamental_to_essential_matrix(
```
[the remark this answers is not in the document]
johnwlambert: We actually do assume we know the intrinsics here to perform the conversion -- we use `camera_intrinsics_i1` and `camera_intrinsics_i2`.

gtsfm/frontend/verifier/opencv_verifier_base.py
johnwlambert: Thanks for catching this, was copied from before and not sure why we had that. Rephrased to `# Use larger focal length, among the two choices.`

gtsfm/frontend/verifier/opencv_verifier_base.py
```diff
@@ -0,0 +1,145 @@
...
+        if self._use_intrinsics_in_verification:
+            uv_norm_i1 = feature_utils.normalize_coordinates(keypoints_i1.coordinates, camera_intrinsics_i1)
+            uv_norm_i2 = feature_utils.normalize_coordinates(keypoints_i2.coordinates, camera_intrinsics_i2)
+
+            # OpenCV can fail here, for some reason
+            if match_indices.shape[0] < 6:
+                return self._failure_result
+
+            # use stricter threshold, among the two choices
+            fx = max(camera_intrinsics_i1.K()[0, 0], camera_intrinsics_i2.K()[0, 0])
```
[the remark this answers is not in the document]
johnwlambert: Updated by moving it below, thanks. That's cleaner.

tests/frontend/verifier/test_lmeds.py
johnwlambert: I actually found that setting (LMEDS w/ E-matrix estimation) yielded a result that couldnt pass the test error threshold, so I excluded it (`Angular error 9.3 vs. tol. 2.0`).

gtsfm/frontend/verifier/opencv_verifier_base.py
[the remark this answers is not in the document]
akshay-krishnan: but doesn't this branch run when use_intrinsics_in_verification is false?

tests/frontend/verifier/test_lmeds.py
akshay-krishnan: even with data that was generated using perfect intrinsics? 
If that is the case, it sounds like it doesn't work. Maybe we should not submit that part if it doesn't work yet.

gtsfm/frontend/verifier/opencv_verifier_base.py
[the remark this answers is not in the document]
johnwlambert: Indeed. However, we don't use the intrinsics during the verification, but we do afterwards.

One interpretation is that the intrinsics aren't accurate enough to compute errors with, but they are loosely accurate enough to lift from F->E.

Another interpretation is that we just want to see how well we can do with F estimation, and not E estimation.

tests/frontend/verifier/test_lmeds.py
johnwlambert: That BaseVerifier unit test is checking a planar case (2 planes), which LMEDS - EMat is no good at.

But it works fine for Door : - ). I'll add a test to show that.
Screen Shot 2022-07-23 at 1 16 15 AM

johnwlambert: Actually, I'll add this extra test in a follow-up PR. (I've tested locally, but have a few other changes to APIs I want to push first).

gtsfm/frontend/verifier/ransac.py
johnwlambert: Added more options as an Enum.

PR #540, opened 2022-07-28
Generate process graph

kfu02:
Generating a process graph like those found in [GTSFM Figures](https://docs.google.com/presentation/d/1P-U91z21yq_GOG9OQ3IedIYCBJIHA4avsNlu8bf59kM/edit?usp=sharing), but directly from the code. This PR outputs the graph to a .svg file which is then loaded in as a static image to the web interface. A future PR will make the SVG more interactive, potentially with d3.js.

process_graph_output

## Design

 * I used the [registry design pattern](https://charlesreid1.github.io/python-patterns-the-registry.html) to define an abstract GTSFMProcess class. 
 * Any class that serves as a base abstract class should subclass GTSFMProcess. This forces them to implement ``get_ui_metadata()``, which defines the display name, parent plate, and GTSFMProduct names that connect (all as strings) in a dataclass named UiMetadata. 
  * All children of GTSFMProcess are saved to a central registry when they are **defined** thanks to Python metaclass magic. 
  * ProcessGraphGenerator then iterates over this central registry to create a directional graph using [pydot](https://github.com/pydot/pydot), and saves the graph as an SVG file inside ``rtf_vis_tool/``. 
  * This SVG file is then read and displayed in JS.

I chose to define the graph via the Processes because those are the graph components that are guaranteed to have a uniquely associated file in the codebase. The Products are hardcoded strings because [it would be difficult](https://github.com/borglab/gtsfm/pull/540#discussion_r947897913) to have the types of Products automatically inferred from code. A future PR could perhaps make the graph instead be parent-plate driven, where each file that defines a parent plate has a list of the Processes it holds.

The current usage is explained at the top of gtsfm_process.py:

```
# When creating a new class, follow this template:

# 1: Import GTSFMProcess, UiMetadata
# ----------------------------------
# from gtsfm.ui.gtsfm_process import GTSFMProcess, UiMetadata

# 2: Subclass GTSFMProcess (this can replace ABCMeta, since it inherits from ABCMeta)
# ------------------------
# class ClassName(GTSFMProcess):

# 3: Implement get_ui_metadata()
# ------------------------------
#    def get_ui_metadata() -> UiMetadata:
#        """Returns data needed to display node and edge info for this process in the process graph."""
#
#        return UiMetadata(
#                   display_name="Display Name"
#                   parent_plate="Parent Plate Name"
#                   input_products=("Input Product 1", "Input Product 2")
#                   output_products=("Output Product 1", "Output Product 2")
#               )
```


akshay-krishnan:
Its a great start, but I have some high-level questions/comments first: 

- Do you plan to create another GreyNode class in another PR? I think this will be necessary. The blue nodes should not register string names but actual GreyNode objects. This will avoid the bi-directional edges you currently have. For example the "relative Ts" input to View Graph estimator are different from the relative Ts output by it. 

- I think we should call BlueNode and GreyNode something else. Maybe GtsfmProcessor, GtsfmProduct? I also remembered what the pink nodes we discussed were - they were metrics. 

- could you explain the solution at a high level? I thought the base classes (for example VerifierBase) would themselves inherit from BlueNode, but that doesn't seem to be the case.

kfu02:
@akshay-krishnan I edited the PR description to include a high-level description. Sorry, I intended to before requesting reviews but it must've slipped my mind. 

To respond to your questions:

1) The issue is that pydot, the graph generator tool I use, only cares about what string literals you give it to display. So a GrayNode class would not solve the underlying issue that the "relative Ts" that go in/out of View Graph Estimator have the same displayed name. A more descriptive name would fix that.
 
Also, only defining the names of GrayNodes as strings inside BlueNodes reduces the overhead for developers--rather than defining N GrayNode classes and referencing them inside the BlueNode class, one only has to define 1 BlueNode with a list of strings (implied GrayNodes) that connect to it. Since these GrayNodes would just be holding a string `display_name` I don't see the point in introducing the extra overhead.

2) I agree, but I wasn't sure what to change the name to. I was planning to ask in the meeting on Weds. GTSFMProcessor/Product/Metric makes sense to me. The ROS counterparts would be BlueNode = node, GrayNode = topic, as another option.
 
3) I originally tried to make the base cases inherit the BlueNode class. The issue is that when iterating over the central registry, I have to call `init()` on each class to get the useful UI metadata out of each, and often it makes the most sense to define the UI metadata in the abstract superclass (since all concrete implementations should have the same data in and out). So instead I chose the slightly weird option of creating a tiny UI-only class at the tops of files. 
 
I considered making the UI metadata class vars, but then there is no way to enforce that they are set via abstractmethod decorator. But that's another option. I couldn't think of an option that combines the best of both. What do you think?

gtsfm/ui/dot_graph_generator.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def _build_graph(self):
+        """Build graph based on RegistryHolder's REGISTRY."""
+
+        # TODO: remove this
+        print("!\n" * 100)
+        print(RegistryHolder.get_registry())
+
+        for cls_name, cls_type in RegistryHolder.get_registry().items():
+            # don't add the base class to the graph
```
dellaert, line 56: consider making a function. You should aim (yet not achieve) for functions/methods <= 4 lines long

dellaert (commented):
Nice ! what's left, or is this ready to merge?

kfu02:
@dellaert not ready yet, but GH doesn't allow me to un-request reviews. I'm in the process of adding more to the graph

tests/utils/test_io_utils.py
```diff
@@ -96,8 +96,11 @@ def test_read_cameras_txt_nonexistent_file(self) -> None:
         self.assertIsNone(calibrations)
 
     def test_round_trip_images_txt(self) -> None:
-        """Starts with a pose. Writes the pose to images.txt (in a temporary directory). Then reads images.txt to recover
-        that same pose. Checks if the original wTc and recovered wTc match up."""
+        """
+        Starts with a pose. Writes the pose to images.txt (in a temporary directory).
```
johnwlambert, line 100: Do you mind pulling in the changes from the master branch, Kevin? I fixed the formatting for these docstrings there.

gtsfm/densify/mvs_base.py
```diff
@@ -14,10 +14,15 @@
...
 
+from gtsfm.ui.registry import GTSFMProcess, UiMetadata
 
-class MVSBase(metaclass=abc.ABCMeta):
+
+class MVSBase(GTSFMProcess):
     """Base class for all multi-view stereo implementations."""
 
+    def get_ui_metadata() -> UiMetadata:
+        return UiMetadata("Multi-view Stereo", ("Images", "SFM Result"), ("Dense Points"), "")
```
johnwlambert, line 24: Do you mind adding a 1-line docstring here explaining what is being returned here?

gtsfm/ui/dot_graph_generator.py
```diff
@@ -0,0 +1,137 @@
...
+            output_products = (output_products,)
+
+        parent_plate = metadata.parent_plate
+
+        # add cleaned-up metadata to graph
+        self._add_nodes_and_edges(display_name, input_products, output_products, parent_plate)
+
+    def _add_nodes_and_edges(
+        self, display_name: str, input_products: Tuple[str], output_products: Tuple[str], parent_plate: str
+    ):
```
johnwlambert, line 104: do you mind adding the return type annotation here? `-> None`

gtsfm/ui/dot_graph_generator.py
```diff
@@ -0,0 +1,137 @@
...
+                pydot.Node(product_name, shape=style.node_shape, style=style.node_style, fillcolor=style.gray_fillcolor)
+            )
+
+        # add Edges for all input_products -> blue node -> all output_products
+        for input_product_name in input_products:
+            self._graph.add_edge(pydot.Edge(input_product_name, display_name, color=style.arrow_color))
+        for output_product_name in output_products:
+            self._graph.add_edge(pydot.Edge(display_name, output_product_name, color=style.arrow_color))
+
+    def save_graph(self, filepath=os.path.join(JS_ROOT, "src", "ui", "dot_graph_output.svg")):
```
johnwlambert, line 124: let's make the default arg a global var, like
```python
DEFAULT_GRAPH_VIZ_OUTPUT_PATH = JS_ROOT / "src" / "ui" / "dot_graph_output.svg"
```
and then using it here as def save_graph(self, filepath=DEFAULT_GRAPH_VIZ_OUTPUT_PATH) -> None`?

gtsfm/ui/dot_graph_generator.py
```diff
@@ -0,0 +1,137 @@
...
+from typing import Tuple
+from gtsfm.ui.registry import UiMetadata
+
+JS_ROOT = os.path.join(Path(__file__).resolve().parent.parent.parent, "rtf_vis_tool")
+
+
+class DotGraphGenerator:
+    """Generates and saves a graph of all the components in REGISTRY."""
+
+    def __init__(self, test_mode=False):
```
johnwlambert, line 24: let's add type annotations here, `test_mode: bool = False) -> None`

gtsfm/ui/dot_graph_generator.py
```diff
@@ -0,0 +1,137 @@
...
+        """Create DotGraphGenerator.
+
+        Args:
+            test_mode - boolean flag, only set True when unit testing.
```
johnwlambert, line 28: Per the [style guide](https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md), let's update this to colon separated. Thanks!

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,86 @@
...
+
+class RegistryHolder(type):
+    """
+    Class that defines central registry and automatically registers classes
+    that extend GTSFMProcess.
+    """
+
+    REGISTRY = {}
+
+    def __new__(cls, name, bases, attrs):
```
johnwlambert, line 36: Could you add type annotations here when you get a chance?

tests/ui/test_registry.py
```diff
@@ -0,0 +1,102 @@
...
+import unittest
+
+from gtsfm.ui.registry import RegistryHolder, GTSFMProcess, UiMetadata
+import abc
+
+
+class FakeImageLoader(GTSFMProcess):
+    """Test class."""
+
+    def __init__(self, fake_image_dir):
```
johnwlambert, line 16: nit: type annotation missing. Thanks!

gtsfm/ui/dot_graph_generator.py
```diff
@@ -0,0 +1,137 @@
...
+            # happens when concrete classes implement an abstract class without overwriting get_ui_metadata()
+            if metadata in seen_metadata:
+                continue
+
+            seen_metadata.add(metadata)
+
+            self._add_metadata_to_graph(metadata)
+
+    def _add_metadata_to_graph(self, metadata: UiMetadata):
+        """Given the UiMetadata from a GTSFMProcess, clean it up, then add to graph."""
```
johnwlambert, line 80: Could we rephrase the language about "clean it up" to be a bit more specific? Thanks!

gtsfm/frontend/inlier_support_processor.py
```diff
@@ -44,7 +44,8 @@ def run(
         v_corr_idxs: np.ndarray,
         two_view_report: TwoViewEstimationReport,
     ) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray, Optional[TwoViewEstimationReport]]:
-        """Check for sufficient support among correspondences to estimate a trustworthy relative pose for this image pair.
+        """
```
johnwlambert, line 47: Fixed in master -- please pull in when you get a chance. Thanks!

gtsfm/frontend/detector_descriptor/detector_descriptor_base.py
```diff
@@ -13,16 +13,20 @@
...
 
 
-class DetectorDescriptorBase(metaclass=abc.ABCMeta):
+class DetectorDescriptorBase(GTSFMProcess):
     """Base class for all methods which provide a joint detector-descriptor to work on an image.
 
     This class serves as a combination of individual detector and descriptor.
     """
 
+    def get_ui_metadata() -> UiMetadata:
```
johnwlambert, line 27: Let's add a 1-line docstring to all the `get_ui_metadata()` function defintions.

gtsfm/averaging/translation/translation_averaging_base.py
```diff
@@ -12,13 +12,29 @@
...
 
-class TranslationAveragingBase(metaclass=abc.ABCMeta):
+
+class TranslationAveragingBase(GTSFMProcess):
     """Base class for translation averaging.
 
     This class generates global unit translation estimates from pairwise relative unit translation and global rotations.
     """
 
+    def get_ui_metadata() -> UiMetadata:
```
johnwlambert, line 24: nit: docstring. Thanks!

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,99 @@
...
+class AbstractableRegistryHolder(abc.ABCMeta, RegistryHolder):
+    """Extra class to ensure GTSFMProcess can use both ABCMeta and RegistryHolder metaclasses."""
+
+    pass
+
+
+@dataclass(frozen=True)
+class UiMetadata:
+    """
+    Holds all info needed to display a GTSFMProcess in the process graph (see DotGraphGenerator).
```
kfu02, line 77:
```suggestion
    Holds all info needed to display a GTSFMProcess in the process graph (see ProcessGraphGenerator).
```

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,160 @@
+"""
+Creates a DOT graph based on the classes that subclass GTSFMProcess.
```
kfu02, line 2:
```suggestion
Creates a process graph based on the classes that subclass GTSFMProcess.
```

kfu02 (commented):
changing some references to DOT graph that I missed

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Holds all info needed to display a GTSFMProcess in the process graph (see ProcessGraphGenerator).
+
+    frozen=True makes this dataclass immutable and hashable.
+    """
+
+    display_name: str
+    input_products: Tuple[str]
+    output_products: Tuple[str]
+    parent_plate: str
```
akshay-krishnan, line 85: provide explanations to these in documentation? parent_plate was not clear to me.

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+class GTSFMProcess(metaclass=AbstractableRegistryHolder):
```
akshay-krishnan, line 88: should this class be in a different file?

akshay-krishnan: maybe along with UiMetadata

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,99 @@
...
+    """
+    Base type that all classes the REGISTRY can see must inherit from.
+
+    Built as a Mixin. For example usage see `tests/ui/test_registry.py`.
+    """
+
+    @staticmethod
+    @abc.abstractmethod
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display this process in the process graph. See gtsfm/ui/registry.py for more info."""
```
akshay-krishnan, line 98: weird that this pointer is in the same file.

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,160 @@
...
+    def __init__(self, test_mode: bool = False) -> None:
+        """Create ProcessGraphGenerator.
+
+        Args:
+            test_mode: boolean flag, only set True when unit testing.
+        """
+
+        # create empty directed graph
+        self._graph = pydot.Dot("my_graph", graph_type="digraph", bgcolor="white")
+        # TODO: fontname seems broken?
```
akshay-krishnan, line 41: add github usernames, ex: TODO(akshay-krishnan).

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        StyleTuple = namedtuple("StyleTuple", style_consts)
+        self._style = StyleTuple(**style_consts)
+
+        self._test_mode = test_mode
+
+    def _build_graph(self) -> None:
+        """Build graph based on RegistryHolder's REGISTRY."""
+
+        # TODO: remove this
+        print("!\n" * 100)
```
akshay-krishnan, line 62: remove this before merging?

gtsfm/ui/registry.py
akshay-krishnan, line 85: parent plate seems usually empty, can you make it optional or add a default value?

akshay-krishnan (commented):
Looking great, much better than the first version!

akshay-krishnan:
My biggest design comment at this point: 

The grey nodes are not just "UI metadata". This PR does a good job at recovering the blue nodes from code, but grey nodes are in a way hard-coded. All the Gtsfm classes you modified have a create_computation_graph() or run() method, that provide the outputs and their types. Should they all inherit from a base class, like GtsfmMetric? 

This is definitely for a future PR, I want to check if you've thought about this, and whether this design allows for it.

akshay-krishnan:
There seem to be many grey nodes without sources/users in the image. For example, no sources for matches, SfmResult, camera intrinsics, etc., and no users for global translations, view graph edges, etc.

gtsfm/ui/registry.py
kfu02: I just removed it everywhere, it was in the original GTSFM Figures, but I can't think of a good method of displaying it with GraphViz (and as you said it was mostly empty)

kfu02:
@akshay-krishnan for your last two comments, I believe this registry design allows for non-hardcoded GTSFMMetrics, which should also fix the gray nodes that are missing sources/users. However, this would require a few large changes:
 * All GTSFMProcess subclasses would need a common `run()` method that defines exactly what input/output products they need.
 * All products would need to subclass a new GTSFMProduct that has a static method `get_display_name()`. So even simple dicts of ints would need to be wrapped in a new class.
 * On launch, each GTSFMProcess would iterate over `run.__annotations__` which would give all the types that go in/out of a process. Then it would be able to save these types' display names and return them in `get_ui_metadata()`.

(Annotations is a Dict of input/output types of a function or class. See these [docs](https://docs.python.org/3/library/inspect.html#inspect.get_annotations) or this [answer](https://stackoverflow.com/questions/15200048/how-to-get-the-parameters-type-and-return-type-of-a-function).)
```
>>> def func(a: str) -> int:
       # code
>>> func.__annotations__["return"]
<class 'int'>
```

So it would require large structural changes, but it's possible.

johnwlambert:
Hi Kevin, do you mind sharing the latest version of what the generated graph visualizations look like?

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -14,14 +14,26 @@
...
 
     This class generates global rotation estimates from the pairwise relative
     rotations.
     """
 
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display this process in the process graph."""
+
+        return UiMetadata(
+            "Rotation Averaging",
```
johnwlambert, line 31: nit: could we use keyword args here for clarity?

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -14,14 +14,26 @@
...
+
+class RotationAveragingBase(GTSFMProcess):
     """Base class for rotation averaging.
 
     This class generates global rotation estimates from the pairwise relative
     rotations.
     """
 
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display this process in the process graph."""
```
johnwlambert, line 28: nit: I would prefer we say something a bit more informative like
"""Returns data necessary to display node and edge information for the current process in the process graph."""

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,56 @@
 logger = logger_utils.get_logger()
 
 
-class BundleAdjustmentOptimizer:
+class TwoViewBundleAdjustmentProcess(GTSFMProcess):
+    """Dummy class to get the UI graph correct.
```
johnwlambert, line 67: Let's refine this language a bit -- maybe something like
"""Dummy class that exists in order to visualize the 2-view BA process in a process graph visualization."""

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,56 @@
...
+
+    BundleAdjustmentOptimizer is used in two separate places (optimizing
```
johnwlambert, line 69: nit: let's break the lines at 120 chars (looks a bit off rn)

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,56 @@
...
     """Bundle adjustment using factor-graphs in GTSAM.
 
     This class refines global pose estimates and intrinsics of cameras, and also refines 3D point cloud structure given
     tracks from triangulation."""
 
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display this process in the process graph."""
+
+        return UiMetadata(
+            "Global Bundle Adjustment",
```
johnwlambert, line 103: same note here and throughout about keyword args, to explain what these args mean

gtsfm/densify/mvs_base.py
```diff
@@ -14,10 +14,22 @@
...
 
+from gtsfm.ui.gtsfm_process import GTSFMProcess, UiMetadata
 
-class MVSBase(metaclass=abc.ABCMeta):
+
+class MVSBase(GTSFMProcess):
     """Base class for all multi-view stereo implementations."""
 
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display this process in the process graph."""
```
johnwlambert, line 24: Same note here about updating docstring

gtsfm/densify/mvs_base.py
```diff
@@ -14,10 +14,22 @@
...
+
+        return UiMetadata(
```
johnwlambert, line 26: `UiMetadata` seems a little vague, but maybe ok. ProcessGraphMetadata? @akshay-krishnan thoughts?

gtsfm/ui/gtsfm_process.py
```diff
@@ -0,0 +1,68 @@
+"""
+Defines GTSFMProcess, which holds a UiMetadata object for displaying this process in the process graph.
+
+Author: Kevin Fu
+"""
+
+# When creating a new class, follow this template:
```
johnwlambert, line 7: multi-line comment should go into triple-quote separated lines

gtsfm/ui/gtsfm_process.py
```diff
@@ -0,0 +1,68 @@
...
+#        """Returns data needed to display this process in the process graph."""
+#
+#        return UiMetadata(
+#                   "Display Name"
+#                   "Parent Plate Name"
+#                   ("Input Product 1", "Input Product 2")
+#                   ("Output Product 1", "Output Product 2")
+#               )
+
+from gtsfm.ui.registry import AbstractableRegistryHolder
```
johnwlambert, line 29: nit: check the style guide for import order

gtsfm/ui/gtsfm_process.py
```diff
@@ -0,0 +1,68 @@
...
+
+from typing import Tuple
+
+
+@dataclass(frozen=True, order=True)
+class UiMetadata:
+    """
+    Holds all info needed to display a GTSFMProcess in the process graph (see ProcessGraphGenerator).
+
+    frozen=True makes this dataclass immutable and hashable.
```
johnwlambert, line 42: nit: i think no need to mention the immutable part here. if the hashable part is important, do you mind explaining a bit more about why, here in the comment?

gtsfm/ui/gtsfm_process.py
```diff
@@ -0,0 +1,68 @@
...
+import abc
+from dataclasses import dataclass
+
+from typing import Tuple
+
+
+@dataclass(frozen=True, order=True)
+class UiMetadata:
+    """
+    Holds all info needed to display a GTSFMProcess in the process graph (see ProcessGraphGenerator).
```
johnwlambert, line 40: nit: 1-sentence line should go in same line as the triple quote (see style guide)

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,167 @@
+"""
+Creates a process graph based on the classes that subclass GTSFMProcess.
+
+Specifically, all classes that subclass GTSFMProcess are added to a central
```
johnwlambert, line 4: nit: same note about breaking comment lines at 120 chars

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,167 @@
...
+        style_consts = {
+            "blue_fillcolor": "lightskyblue",
+            "gray_fillcolor": "gainsboro",
+            "node_style": '"rounded, filled, solid"',
+            "node_shape": "box",
+            "arrow_color": "gray75",
+        }
+        # namedtuple here is faster than default Dict and allows use of dot
+        # operator
+        StyleTuple = namedtuple("StyleTuple", style_consts)
```
johnwlambert, line 52: If we only use this NamedTuple definition once in this file, it could make sense to just make these global constants we use in this file

```python
BLUE_FILLCOLOR = "lightskyblue"
GRAY_FILLCOLOR = "gainsboro"
NODE_STYLE = '"rounded, filled, solid"'
NODE_SHAPE = "BOX"
ARROW_COLOR = "gray75"
```

Are we still using the "gray" and "blue" terminology though, or "PRODUCT" vs "PROCESS"?

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,167 @@
...
+        Args:
+            metadata: UiMetadata object to add nodes/edges for
+        """
+
+        display_name = metadata.display_name
+        style = self._style
+
+        # autocast strings to one-element tuples
+        input_products = metadata.input_products
+        if type(input_products) == str:
```
johnwlambert, line 130: Would it be cleaner to use a fixed type for `input_products`, e.g. always just return a Tuple?

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,46 @@
...
+Author: Kevin Fu
+"""
+
+import abc
+from typing import Tuple, Dict
+
+
+class RegistryHolder(type):
+    """
+    Class that defines central registry and automatically registers classes
```
johnwlambert, line 17: nit: let's put this on one line, on the same line as the triple quote in line 16

gtsfm/ui/registry.py
```diff
@@ -0,0 +1,46 @@
...
+    that extend GTSFMProcess.
+    """
+
+    REGISTRY = {}
+
+    def __new__(cls: type, name: str, bases: Tuple, attrs: Dict) -> None:
```
johnwlambert, line 23: nit: type hints should be `Tuple[something,something]` etc. and `Dict[something, something]`

tests/ui/test_process_graph_generator.py
```diff
@@ -0,0 +1,78 @@
...
+            process_graph_generator.save_graph(extra_dir_path)
+
+            # check that files were saved, not their correctness
+            self.assertTrue(os.path.exists(png_path))
+            self.assertTrue(os.path.exists(svg_path))
+            self.assertTrue(os.path.exists(extra_dir_path))
+
+    def test_no_fake_nodes(self):
+        """
+        Ensure that no test nodes from test_registry.py are in the final
```
johnwlambert, line 39: nit: check [style guide](https://github.com/borglab/gtsfm/blob/master/CONTRIBUTING.md). docstrings should have a 1-line description on the same line as the triple quote.

Can add additional info below that, separated by a line.

tests/ui/test_process_graph_generator.py
```diff
@@ -0,0 +1,78 @@
...
+        process_graph_generator._build_graph()
+
+        output_raw_dot = process_graph_generator._main_graph.to_string()
+
+        self.assertTrue("GTSFMProcess" not in output_raw_dot)
+        self.assertTrue("Fake" not in output_raw_dot)
+
+    def test_build_graph(self):
+        """
+        Check that the test nodes are in the graph and connected correctly.
```
johnwlambert, line 53: same note about the docstring here.

gtsfm/data_association/data_assoc.py
```diff
@@ -48,6 +51,16 @@ class DataAssociation(NamedTuple):
     triangulation_options: TriangulationOptions
     save_track_patches_viz: Optional[bool] = False
 
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display this process in the process graph."""
+
+        return UiMetadata(
+            "Data Association",
+            "Multi-View Optimizer",
+            ("View-Graph Correspondences", "Global Rotations", "Global Translations", "Camera Intrinsics"),
```
johnwlambert, line 60: optional: would it be possible to discover these automatically from the function definition for `create_computaton_graph()`? that way they wouldn't go stale if we change that function def?

gtsfm/densify/mvs_base.py
akshay-krishnan: Yeah, ProcessUiMetadata or ProcessGraphMetadata. but optional, im okay with UiMetadata too.

gtsfm/data_association/data_assoc.py
akshay-krishnan: This was my suggestion earlier too. The problem is that create_computation_graph does not have the type/name of returned arguments. run() would have types, but still not names. Without names, we won't know which output argument is fed as input to another node in the graph. Even IF we had argument names in these methods, we would have to keep them consistent across classes to infer the graph from them. 

Keeping this design, the only way would be make arguments and results derive a base class (for a grey node) which has some name metadata. I don't know if this is overkill.. depends on how we end up using this graph.

kfu02:
> Hi Kevin, do you mind sharing the latest version of what the generated graph visualizations look like?

Hey John, the latest graph visualization is in the PR description and has been checked in [here](https://github.com/borglab/gtsfm/pull/540/files#diff-dcdc3f8191a0020f9ec9c8c45491f630a308f1914c192ccc6427a9b12f871ffd).

gtsfm/ui/gtsfm_process.py
kfu02: I wanted to do that at first, but to make the method as easy to copy-paste as possible I included the docstring of `get_ui_metadata()` in the comment block. (See line 19-20 [below](https://github.com/borglab/gtsfm/pull/540/files/c7ae439db14f58a05e7754a8707bac0ee9dd136d#diff-2100374802706ad206882380c8fed0951f0edd7a4ca5c88eb890105f20b53c7dR19-R20).) 

Do you have any workarounds for that?

gtsfm/ui/process_graph_generator.py
kfu02: Yes. But I ran into this issue with one-element tuples (described in lines 115-119 [above](https://github.com/borglab/gtsfm/pull/540/files/c7ae439db14f58a05e7754a8707bac0ee9dd136d#diff-ce2386876f8582f4228575c840a96fc7350133a79df6a912b9b2bd2017020fdbR115-R119)). 

Since Python is not statically typed it can be easy to accidentally return a string instead of a one element tuple, which is why I added this autocast here. I can remove it but I feel that would be sacrificing a useful guardrail in pursuit of clean code.

gtsfm/ui/gtsfm_process.py
kfu02: >    frozen=True makes this dataclass hashable, used when adding it to a set to avoid duplicate nodes in the graph.

Added this^

gtsfm/ui/gtsfm_process.py
kfu02: Fixed. Might I suggest including [isort](https://pypi.org/project/isort/) as a part of GTSFM's linter tools (alongside black) to avoid this issue in the future?

gtsfm/ui/process_graph_generator.py
kfu02: Good point. Updated to be globals and use "PRODUCT_FILLCOLOR"/"PROCESS_FILLCOLOR"

gtsfm/data_association/data_assoc.py
kfu02: @dellaert mentioned last Weds that he thought this was overkill, given the large infrastructure changes necessary that @akshay-krishnan has described.

gtsfm/averaging/rotation/rotation_averaging_base.py
kfu02: To clarify, you want something like this?
```
        return UiMetadata(
            display_name="Translation Averaging",
            parent_plate="Multi-View Optimizer",
            input_products=(
                "View-Graph Relative Translations",
                "Global Rotations",
                "Absolute Pose Priors",
                "Relative Pose Priors",
            ),
            output_products=("Global Translations"),
        )
```

gtsfm/averaging/rotation/rotation_averaging_base.py
kfu02: changed to: 
> """Returns data needed to display node and edge info for this process in the process graph."""

gtsfm/densify/mvs_base.py
kfu02: I will keep UiMetadata then. I believe the improved docstring @johnwlambert [requested](https://github.com/borglab/gtsfm/pull/540#discussion_r947427117) will clarify things as well.

dellaert:
The new graph is pretty amazing. Let me know by email when it is ready for review. I have some time this weekend and might co-review it with Hayk.

gtsfm/ui/gtsfm_process.py
johnwlambert: Thanks! That would be great, I would welcome a PR for that!

gtsfm/averaging/rotation/rotation_averaging_base.py
johnwlambert: Yes -- thanks!

gtsfm/ui/gtsfm_process.py
kfu02: Done: https://github.com/borglab/gtsfm/pull/547

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,55 @@
...
+    """
+
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display node and edge info for this process in the process graph."""
+
+        return UiMetadata(
+            "Two-View Bundle Adjustment",
+            "Two-View Estimator",
+            (
+                "Relative Rotations",
```
dellaert, line 80: There is only one relative rotation and translation input. remove `s`

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,55 @@
...
+                "Relative Translations",
+                "Triangulated Points",
+                "Relative Pose Priors",
+                "Camera Intrinsics",
+                "Keypoints",
+                "Verified Correspondences",
+            ),
+            ("Optimized Relative Rotations", "Optimized Relative Translations", "Inlier Correspondences"),
```
dellaert, line 88: There is only one relative rotation and translation output. remove `s`

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,55 @@
 logger = logger_utils.get_logger()
 
 
-class BundleAdjustmentOptimizer:
+class TwoViewBundleAdjustmentProcess(GTSFMProcess):
```
dellaert, line 66: Should be called TwoViewBundleAdjustment and be in its own file I think

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -61,12 +63,55 @@
...
+            parent_plate="Multi-View Optimizer",
+            input_products=(
+                "Absolute Pose Priors",
+                "Relative Pose Priors",
+                "Camera Intrinsics",
+                "Global Translations",
+                "Global Rotations",
+                "3D Tracks",
+            ),
+            output_products=("Filtered Optimized Camera Poses", "Filtered Optimized 3D Tracks"),
```
dellaert, line 112: "Filtered Optimized" -> "Optimized" ?

dellaert (changes requested):
Initial review. very cool !
BUT: there is a big problem with the graph: the rule is **"products should be in the same plate as the process that _outputs_ them".** Currently, this is inconsistent. It's correct for 2 View bundle adjustment, but incorrect for matcher (which takes input products into the parent plate - bad :-))

tests/ui/test_gtsfm_process.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class FakeImageLoader(GTSFMProcess):
+    """Test class."""
+
+    def __init__(self, fake_image_dir: str) -> None:
+        self._fake_image_dir = fake_image_dir
+
+    def get_ui_metadata() -> UiMetadata:
+        """Returns data needed to display node and edge info for this process in the process graph."""
```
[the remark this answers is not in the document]
kfu02: Yes, it is under 120 chars long. Confirmed with `black -l 120 tests/ui/`.

tests/ui/test_gtsfm_process.py
```diff
@@ -0,0 +1,112 @@
...
+Unit tests for RegistryHolder and GTSFMProcess.
+
+Author: Kevin Fu
+"""
+
+import unittest
+
+from gtsfm.ui.registry import RegistryHolder
+from gtsfm.ui.gtsfm_process import GTSFMProcess, UiMetadata
+import abc
```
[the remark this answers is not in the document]
kfu02: Still needed, FakeOutputBase defines a dummy abstract method to show that subclassing GTSFMProcess gives access to abc.ABCMeta.

gtsfm/bundle/bundle_adjustment.py
kfu02: The issue with putting it in a separate file is that since that file is not imported anywhere, the registry never sees the dummy class. If I import it somewhere but don't use the dummy class, the linter will complain of an unused import.

dellaert: I’m confused, if it’s never used, why is it in our graph? Isn’t the fact that it is in our graph evidence for the fact that it is used?

kfu02: Both two-view and global bundle adjustment are done via the BundleAdjustmentOptimizer class, but there's no way to show that in the graph as currently constructed. Every process shown as a blue node must be a unique subclass of GTSFMProcess. So Akshay and I added the dummy class TwoViewBundleAdjustment to get a separate UiMetadata object. 

The reason this works is because the `bundle_adjustment.py` file is imported and used when running GTSFM, meaning both TwoViewBundleAdjustment and BundleAdjustmentOptimizer are visible to the central registry. Earlier I tried creating a separate file for TwoViewBundleAdjustment, but that renders it invisible to the registry unless the file is imported somewhere. However, importing the file somewhere will trigger a flake8 error (`F401 'gtsfm.bundle.two_view_ba' imported but unused`) because the dummy class in the file doesn't have any use besides exposing TwoViewBA-related nodes to the registry.

johnwlambert:
Looks quite nice. Can you merge in master? Was also wondering if we can prevent ‘image_pair_indices’ from getting orphaned in the visualization?

 By the way, we now support methods that generate matches directly from images (without descriptors). So technically there are two possible graphs now, or two separate pathways, because of the CorrespondenceGenerator. @ayushbaid @akshaykrishnan @dellaert any thoughts about how to visualize  that?

dellaert:
>  Was also wondering if we can prevent ‘image_pair_indices’ from getting orphaned in the visualization?

image_pair_indices is "orphaned" in some way, as its only use is creating the plate instances. I don't think it is consumed by any blue box.

akshay-krishnan:
> Looks quite nice. Can you merge in master? Was also wondering if we can prevent ‘image_pair_indices’ from getting orphaned in the visualization?
> 
> By the way, we now support methods that generate matches directly from images (without descriptors). So technically there are two possible graphs now, or two separate pathways, because of the CorrespondenceGenerator. @ayushbaid @akshaykrishnan @dellaert any thoughts about how to visualize that?

It will need more work on Kevin's end to update the graph, once he merges from master :/

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: TLDR: a class with the same name cannot be used in 2 places in the graph (due to the registration pattern), so we had to create a dummy one for the 2-view BA.

akshay-krishnan: ^Its hacky and Im not super happy with it, open to suggestions :)

dellaert: It seems to me it does not have to be dummy: you can have 2 derived classes from one base class that does not have the meta-data: 
BundleAdjustment (everything except metadata)
- TwoViewBundleAdjustment (just 2 view metadata)
- GlobalBundleAdjustment (just global metadata)
As per style guide, every class should be in its own file.

dellaert:
@kfu02 need to merge in master and resolve conflicts as well.

gtsfm/bundle/bundle_adjustment.py
kfu02: Done here: https://github.com/borglab/gtsfm/pull/540/commits/15dfd7849433ca0018e97340afe5e1386ac9804c

kfu02:
> > Looks quite nice. Can you merge in master? Was also wondering if we can prevent ‘image_pair_indices’ from getting orphaned in the visualization?
> > By the way, we now support methods that generate matches directly from images (without descriptors). So technically there are two possible graphs now, or two separate pathways, because of the CorrespondenceGenerator. @ayushbaid @akshaykrishnan @dellaert any thoughts about how to visualize that?
> 
> It will need more work on Kevin's end to update the graph, once he merges from master :/

I'm not sure how to visualize the two separate potential pathways, but I've created an issue for it (#550 ).

kfu02:
@dellaert I fixed the issue where output products are not in the same plate as their parent process (see PR desc above), it's ready for re-review now

dellaert:
> @dellaert I fixed the issue where output products are not in the same plate as their parent process (see PR desc above), it's ready for re-review now

Super-cool. Quick review based on SVG for now:
- Could we put loader and retriever in a plat as well, see how that looks? It's like a grouping plate, as with bundle adjustment, to get rid of messiness.
- where do "Triangulated Points" come from? Some blue box I presume?
- maybe rename "Multi-View Optimizer" to "Sparse Reconstruction" and add a "Dense Reconstruction" plate for the final step.

gtsfm/ui/process_graph_generator.py
```diff
@@ -0,0 +1,160 @@
...
+        for product_name in output_products:
+            cluster.add_node(pydot.Node(product_name, shape=NODE_SHAPE, style=NODE_STYLE, fillcolor=PRODUCT_FILLCOLOR))
+
+        # add process -> output_product edges
+        for output_product_name in output_products:
+            self._main_graph.add_edge(pydot.Edge(display_name, output_product_name, color=ARROW_COLOR))
+
+        # add input_product -> process edges
+        for input_product_name in input_products:
+            self._main_graph.add_edge(pydot.Edge(input_product_name, display_name, color=ARROW_COLOR))
```
ayushbaid, line 145: Do we detect cases where we have a typo and the input products of one Process does not match the output product of other process?

kfu02: No, though that will be obvious in the graph. In the future, people will likely only be adding a few nodes at a time, so I don't think typos will be a large issue.

kfu02:
> * Could we put loader and retriever in a plat as well, see how that looks? It's like a grouping plate, as with bundle adjustment, to get rid of messiness.
> * where do "Triangulated Points" come from? Some blue box I presume?
> * maybe rename "Multi-View Optimizer" to "Sparse Reconstruction" and add a "Dense Reconstruction" plate for the final step.

Triangulated Points are from a [class method](https://github.com/borglab/gtsfm/blob/15dfd7849433ca0018e97340afe5e1386ac9804c/gtsfm/two_view_estimator.py#L87) in the TwoViewEstimator plate, which is why the node doesn't have a parent.

I will pair program with Hayk to address the first and third comments so he can get a feel for the code as well.

dellaert (commented):
Looks great, except for CI failure. After you fix that, let me know and we can merge. Might need to enlist help from someone (not me :-))

akshay-krishnan (approved):

akshay-krishnan:
CI error is on the lettuce dataset which seems to be failing on master too. A problem reading those colmap ground truth files since we added the distortion check.  
https://github.com/borglab/gtsfm/actions/runs/2920158751

stepanyanhayk:
> CI error is on the lettuce dataset which seems to be failing on master too. A problem reading those colmap ground truth files since we added the distortion check. https://github.com/borglab/gtsfm/actions/runs/2920158751

does this mean I can merge to master now since the failure is not because of my changes?

akshay-krishnan:
Yes, it looks good to me. Not sure about other reviewers.

dellaert:
I will merge it :-)
But, @gchenfc could you possibly PR a change to take lettuce out of CI until this is fixed?

dellaert:
Thanks @kfu02 and @stepanyanhayk !!!

johnwlambert:
Looks good, thanks all.

@kfu02 looks like thirdparty/d2net/weights/d2_tf.pth got checked in by accident (30 MB). That file should be downloaded by the weight download script. Could you remove it in a follow up PR, and purge it from the git history? Thanks!

akshay-krishnan:
@johnwlambert I created a PR for that: https://github.com/borglab/gtsfm/pull/557

kfu02:
@akshay-krishnan it might be good to also add to the .gitignore file in your PR if this file is never supposed to be checked in

akshay-krishnan:
@kfu02 thats true, but I am not sure if we want a very specific third-party path in our gitignore file - lets discuss on that PR. Im happy to add it if John wants it.

PR #542, opened 2022-08-03
Allow passing custom output directory for results

johnwlambert:
Previously `results` always appeared in the top-level repo directory. This doesn't make sense for large-scale experiments, however, so this PR adds CLI to specify a custom output directory for results.

To get CI to pass, I also had to fix line-length overages in a few other files, which flake8 was complaining about.

### DIFF
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -297,7 +297,15 @@ class SceneOptimizer:
         )
 
         if self._save_3d_viz:
-            delayed_results.extend(save_visualizations(ba_input_graph, ba_output_graph, gt_wTi_list))
+            delayed_results.extend(
+                save_visualizations(
+                    ba_input_graph,
+                    ba_output_graph,
+                    gt_wTi_list,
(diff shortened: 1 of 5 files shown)


gtsfm/scene_optimizer.py
```diff
@@ -95,15 +87,30 @@ def __init__(
 
         self._save_gtsfm_data = save_gtsfm_data
         self._pose_angular_error_thresh = pose_angular_error_thresh
+        self.output_root = Path(output_root)
+        self._create_output_directories()
+
+    def _create_output_directories(self) -> None:
+        """Create various output directories for GTSFM results, metrics, and plots."""
+        # base paths for storage
+        self.plot_base_path = self.output_root / "plots"
```
ayushbaid, line 96: Nit: make them private members?

ayushbaid (approved):
LGTM except one suggestion

gtsfm/scene_optimizer.py
johnwlambert: Thanks, updated.

PR #557, opened 2022-08-28
Delete d2net weights file

akshay-krishnan:
The weights file was supposed to be downloaded from a script, and got accidentally checked in.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -143,6 +143,7 @@ thirdparty/models/
 thirdparty/patchmatchnet/checkpoints/
 thirdparty/SuperGluePretrainedNetwork/models/weights/
 thirdparty/hloc/weights/VGG16-NetVLAD-Pitts30K.mat
+thirdparty/**/*.pth
 
 # Data folder
 data/


johnwlambert (approved):
Thanks! @akshay-krishnan did you purge this file from the .git history also, to compress it by 30 MB?

johnwlambert:
(Let's add this path to the .gitignore, I think it will be a good fail-safe).

Or at least `thirdparty/**/*.pth`

akshay-krishnan:
Added to gitignore. 
For the purge from history, I did `git rm`. Are there better ways?

johnwlambert:
Thanks -- I believe something like the following would be necessary:

```bash
git filter-branch --index-filter \
    'git rm -rf --cached --ignore-unmatch thirdparty/d2net/weights/d2_tf.pth' HEAD

(See [here](https://stackoverflow.com/questions/43762338/how-to-remove-file-from-git-history).)

akshay-krishnan:
whoa, okay. running it now 🤞

akshay-krishnan:
I think this will need running on a fresh branch, unable to push to this branch, or pull from it anymore after I run that command.

PR #561, opened 2022-08-31
Generate Process Graph for ImageCorrespondenceGenerator

stepanyanhayk:
Since we now support methods that generate matches directly from images/without descriptors, there are two possible pathways for the process graph. This PR addresses this issue and makes the graph change its look based on the input config.

**DetDescCorrespondenceGenerator Config Run Example**
`python gtsfm/runner/run_scene_optimizer_olssonloader.py --dataset_root tests/data/set1_lund_door --image_extension JPG --config_name deep_front_end.yaml --num_workers 4`

process_graph_output

**ImageCorrespondenceGenerator Config Run Example**
`python gtsfm/runner/run_scene_optimizer_olssonloader.py --correspondence_generator_config_name loftr.yaml --max_resolution 200`

process_graph_output


gtsfm/ui/process_graph_generator.py
```diff
@@ -57,6 +58,11 @@ def _build_graph(self) -> None:
         sorted_metadata = sorted(list(unique_metadata))
 
         for metadata in sorted_metadata:
+            if self.is_image_correspondence and metadata.display_name in ["Matcher", "DetectorDescriptor"]:
```
akshay-krishnan, line 61: ideally I dont think we want to hard-code these in the graph generation code. 
These will break if we change the code.

akshay-krishnan: @dellaert WDYT? 
Everything else LGTM in the PR.

stepanyanhayk: @akshay-krishnan I agree that this might cause difficulties if we change the code. Do you think it would be better if I change the second part of the condition with `metadata.parent_plate == "DetDescCorrespondenceGenerator"`?

This way if we change/add new nodes as long as the parent_plate is the same the code will choose the right pathway for the graph.

akshay-krishnan: Yeah, we can expect the parent plate to change less often than the blue nodes.

akshay-krishnan: it may not change at all, but also could at some point..

akshay-krishnan (approved):

dellaert (approved):
I'm good with solution for now :-) Let's fix hardcoding when we need it.

PR #515, opened 2022-06-07
Merging changes from the hilti challenge

ayushbaid:

PR #566, opened 2022-09-10
Move init_cameras into DataAssociation

stepanyanhayk:


johnwlambert:
Thanks for the PR, Hayk. Could you describe a bit more about how this PR cleans up the Process Graphs we are generating?

akshay-krishnan:
@johnwlambert our process graph currently does not correctly reflect the code in some areas, this being one of them. We should move any computation happening in a plate into a blue node of its own or into one of the existing blue nodes.

tests/data_association/test_data_association.py
```diff
@@ -80,6 +82,19 @@ def generate_noisy_2d_measurements(
     return keypoints_list, img_idxs, cameras
 
 
+def get_camera_details(
```
akshay-krishnan, line 85: is this needed? generate_noisy_2d_measurements above can be modified to output the wRi, wti and intrinsics. 

There are one or two places where cameras are created without calling the above method. In those cases we directly create wRi, wti and intrinsics instead.

akshay-krishnan (approved):

stepanyanhayk:
> Thanks for the PR, Hayk. Could you describe a bit more about how this PR cleans up the Process Graphs we are generating?

Hi @johnwlambert, I guess my wording was a bit confusing. Instead of cleaning up the graph, this PR refactors the code so we do not have to complicate the graph. In this case, `init_cameras` function was a separate process in the `MultiViewOptimizer` class that was not visualized in the graph as a separate blue node. So, now I moved this small function under the DataAssociation class.

gtsfm/data_association/data_assoc.py
```diff
@@ -72,7 +72,9 @@ def __validate_track(self, sfm_track: Optional[SfmTrack]) -> bool:
     def run(
         self,
         num_images: int,
-        cameras: Dict[int, gtsfm_types.CAMERA_TYPE],
+        wRi_list: List[Optional[Rot3]],
+        wti_list: List[Optional[Point3]],
```
johnwlambert, line 76: @akshay-krishnan @dellaert @ayushbaid I think this will conflict with the changes I've been working to bring in from Ayush's hilti branch, where TA now returns a Pose3, instead of a Point3:

https://github.com/borglab/gtsfm/pull/515/files#diff-57f230607899ca8b4f6072976b6f231b322a3c7e6107287f2eb5aa9798e9d52aR266

Can we defer this change until I can get those other changes merged? Currently gtsfm is not functional with gtsam latest because the TA interface is now out of date.

akshay-krishnan: @johnwlambert sorry about that, wasnt aware. It looks like the PR you point to is pretty big. I don't think its a good idea to block this small PR until a large PR is reviewed. 

If you'd like to submit the Pose3 change in a quick small PR, we can wait for it.

johnwlambert: No problem -- sure, let me put those proposed changes into a quick PR in the next day or two. Might simplify this PR as well. Thanks!

PR #567, opened 2022-09-19
Adding Shonan averaging related changes from the hilti branch

ayushbaid:
Isolating the changes for rotation averaging (base class and shonan) which we used for the Hilti challenge.

Hilti super PR: #515


gtsfm/averaging/rotation/shonan.py
```diff
@@ -100,58 +112,75 @@ def _run_with_consecutive_ordering(
...
+            if result.exists(i):
+                wRi_list_consecutive[i] = result.atRot3(i)
 
         return wRi_list_consecutive
 
-    def run(
+    def _nodes_with_edges(
+        self, i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]], relative_pose_priors: Dict[Tuple[int, int], PosePrior]
+    ) -> Set[int]:
+        """Gets the nodes with edges which are to be modelled as between factors."""
```
akshay-krishnan, line 137: is this accurate? it looks like the function is finding a union over the edges in the two dicts..

akshay-krishnan (approved):
Usual reminder to check benchmarks before merging :)

gtsfm/averaging/rotation/shonan.py
```diff
@@ -71,12 +71,24 @@ def __between_factors_from_2view_relative_rotations(
...
     ) -> BetweenFactorPose3s:
         """Create between factors from the priors on relative poses."""
         between_factors = BetweenFactorPose3s()
 
-        # TODO(Ayush): use the priors, atleast between disconnected components.
+        def get_isotropic_noise_model_sigma(covariance: np.ndarray) -> float:
+            """Get the sigma to be used for the isotropic noise model.
+            We compute the average of the diagonal entries of the covariance matrix.
+            """
+            avg_cov = np.average(np.diag(covariance), axis=None)
```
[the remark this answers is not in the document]
ayushbaid: Good catch. The default for `np.average` is axis=None, but I dont think its the same for all APIs. I prefer it to be explicit that we are returning a scalar value. What do you think?

gtsfm/averaging/rotation/shonan.py
ayushbaid: Yes. We cannot have cameras as inputs which do not have any edge. And the input to rotation averaging has just 1 connected components. So the measurement edges and the prior edges gives us all the nodes which can be estimated.

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -35,18 +35,18 @@ def get_ui_metadata() -> UiMetadata:
 
     # ignored-abstractmethod
     @abc.abstractmethod
-    def run(
+    def run_rotation_averaging(
         self,
         num_images: int,
         i2Ri1_dict: Dict[Tuple[int, int], Optional[Rot3]],
-        i2Ti1_priors: Dict[Tuple[int, int], PosePrior],
+        relative_pose_priors: Dict[Tuple[int, int], PosePrior],
```
johnwlambert, line 42: Thanks for the PR, Ayush. Can we switch `relative_pose_priors` back to `i2Ti1_priors`? I think otherwise the ordering gets too confusing and it can lead to bugs.

akshay-krishnan: Dont we change it to i1Ti2 in Hilti?

gtsfm/averaging/rotation/shonan.py
akshay-krishnan: okay, I got confused because of the between factor in the documentation. this function has nothing do with between factors. You could also pass in .keys() of two dicts, and add them to a set here, call it _union_of_edges or something. But I don't have a preference, the current implementation is fine.

gtsfm/averaging/rotation/rotation_averaging_base.py
johnwlambert: Not sure. My point as to why we need to be explicit : - )

In addition to using explicit naming, can we add a unit test for this, so we know whether it should be i2Ti1 or i1Ti2?

ayushbaid: Yes we switched to i1Ti2 too. I agree that having the explicit naming convention helps.

gtsfm/averaging/rotation/shonan.py
ayushbaid: I call it _nodes_with_edges because we return nodes and not edges.

gtsfm/averaging/rotation/rotation_averaging_base.py
ayushbaid: Added one unit test too.

gtsfm/averaging/rotation/rotation_averaging_base.py
```diff
@@ -82,22 +82,22 @@ def create_computation_graph(
         self,
         num_images: int,
         i2Ri1_graph: Delayed,
-        i2Ti1_priors: Dict[Tuple[int, int], PosePrior],
+        i1Ti2_priors: Dict[Tuple[int, int], PosePrior],
```
akshay-krishnan, line 85: was it just for the priors that we used i1Ti2 or for the rotations too? 
I wasn't sure if it was already i1Ti2 on master, or was that change part of the Hilti PR.

akshay-krishnan (approved):

gtsfm/averaging/rotation/rotation_averaging_base.py
ayushbaid: We have to change it as part of Hilti

PR #575, opened 2022-09-25
Revive Frontend

stepanyanhayk:
This PR revives the GTSFM JS frontend, most of the files were last time edited about 2 years ago and were incompatible with new React versions. In particular, solved [Webpack 5 issues](https://web3auth.io/docs/troubleshooting/webpack-issues) that did not allow the frontend to run in the first place and fixed the SE3.test.js

Current look of frontend after the fixes:

Screenshot from 2022-09-24 22-02-29

### DIFF
--- a/rtf_vis_tool/config-overrides.js
+++ b/rtf_vis_tool/config-overrides.js
@@ -1,3 +1,11 @@
+/* While setting up a web3 project, we face multiple issues with webpack 5. The issue is caused due to
+the fact that the web3.js and ethers.js packages have certain dependencies, which are not present within
+the browser environment by webpack 5. This script defines the node polyfills required to override 
+the configurations and enable the usage of web3.js and ethers.js packages.
+
+Author: Hayk Stepanyan
+*/
+
 const webpack = require('webpack');
 module.exports = function override(config) {
     const fallback = config.resolve.fallback || {};
@@ -25,4 +33,3 @@ module.exports = function override(config) {
     })
     return config;
 }
-
--- a/rtf_vis_tool/config-overrides.js
+++ b/rtf_vis_tool/config-overrides.js
@@ -1,0 +1,28 @@
+const webpack = require('webpack');
+module.exports = function override(config) {
+    const fallback = config.resolve.fallback || {};
+    Object.assign(fallback, {
+        "crypto": require.resolve("crypto-browserify"),
+        "stream": require.resolve("stream-browserify"),
+        "assert": require.resolve("assert"),
+        "http": require.resolve("stream-http"),
+        "https": require.resolve("https-browserify"),
+        "os": require.resolve("os-browserify"),
+        "url": require.resolve("url")
+    })
+    config.resolve.fallback = fallback;
+    config.plugins = (config.plugins || []).concat([
+        new webpack.ProvidePlugin({
+            process: 'process/browser',
+            Buffer: ['buffer', 'Buffer']
+        })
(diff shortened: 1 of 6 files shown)


rtf_vis_tool/config-overrides.js
```diff
@@ -0,0 +1,28 @@
+const webpack = require('webpack');
```
johnwlambert, line 1: @stepanyanhayk thanks for the PR. Could you add a file global comment describing the purpose and contents of this file, and add inline comments about why the code is needed?

rtf_vis_tool/src/Components/frustum_classes/se3.js
```diff
@@ -41,7 +41,7 @@ class SE3 {
         Returns:
             Array, shape (1,3), representing the point in the w frame.
         */
-        var ones = nj.ones([1, 1]);
+        var ones = nj.ones([1]);
```
johnwlambert, line 44: curious what the shape was before and after, and why it worked before?

rtf_vis_tool/src/tests/SE3.test.js
```diff
@@ -21,13 +21,13 @@ describe('SE3 Class Unit Tests', () => {
                                    [1, 0, 0]]);
 
         const translation = nj.array([2,3,4]);
-        const testSE3 = SE3(rotation, translation);
+        const testSE3 = new SE3(rotation, translation);
```
johnwlambert, line 24: is `new` required here? heap allocation?

akshay-krishnan (approved):
Not a JS expert, Im happy CI is passing. You can merge if John approves.

rtf_vis_tool/src/tests/SE3.test.js
stepanyanhayk: yes, it was throwing `TypeError: Class constructor SE3 cannot be invoked without 'new'` error otherwise

stepanyanhayk: yes, it was throwing `TypeError: Class constructor SE3 cannot be invoked without 'new'` error otherwise

stepanyanhayk: > is `new` required here? heap allocation?

yes, it was throwing `TypeError: Class constructor SE3 cannot be invoked without 'new'` error otherwise

rtf_vis_tool/src/Components/frustum_classes/se3.js
stepanyanhayk: The variable ones actually continues to represent the same data ([1]) as before. However, this implementation changes the shape from [1, 1] into [1]. The previous implementation was throwing an error on the next line when calling `nj.concatenate` (`ValueError: all the input arrays must have same number of dimensions`).

To address your second question about why this was working before - when we run the JS application, these errors are ignored (i.e. the app might not show what we want but it will not break). The only way to check if this works is with the `se3.test.js` unittest which never passed (link to original [PR](https://github.com/borglab/gtsfm/actions/runs/1271046518/jobs/1552762259) failing the unittest). Also for some reason not all JS tests are run by github (see the screenshot attached it runs only 3 out of 8 tests available, not running the se3.test.js)

Screenshot from 2022-09-27 20-15-26

johnwlambert:
Changes look reasonable to me, feel free to merge. Thanks!

johnwlambert (approved):

PR #577, opened 2022-10-04
JS Frontend Runner

stepanyanhayk:
This PR creates a script that allows to run the JS Frontend from gtsfm directory either by running the script with
`python gtsfm/runner/run_vis_tool.py` or adding `--run_visualizer` argument in `gtsfm_runner_base`


gtsfm/runner/run_vis_tool.py
```diff
@@ -0,0 +1,22 @@
...
+
+import os
+
+from pynpm import NPMPackage
+from pathlib import Path
+
+VIS_TOOL_PATH = Path(__file__).resolve().parent.parent.parent / "rtf_vis_tool"
+
+
+def run_vis_tool():
```
akshay-krishnan, line 15: could you move this to a util? we don't want the runner base class depending on another runner.

gtsfm/runner/run_vis_tool.py
```diff
@@ -0,0 +1,22 @@
+"""Runner for the visualization tool only (back-end results are not desired or required).
```
akshay-krishnan, line 1: this is technically not a runner since it does not inherit from the runner base class. It might be good to  to reserve "runner" to anything that runs SfM. Could we call this file something else? And maybe move it to the viz folder?

akshay-krishnan (commented):
looks quite simple, neat! does it work already?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -118,6 +119,12 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             help="Root directory. Results, plots and metrics will be stored in subdirectories,"
             " e.g. {output_root}/results",
         )
+        parser.add_argument(
+            "--run_visualizer",
+            action="store_true",
+            default=False,
```
johnwlambert, line 125: Hi Hayk, did you have a chance to test out this CLI with both true and false args to make sure it works as expected? I've found argparse boolean defaults often don't work as expected, so want to double check this part

PR #578, opened 2022-10-05
Option to turn off MVS

travisdriver:
Running into OOM on Colab for larger datasets, so I thought it might be useful to give the user the option to turn off MVS without editing the config.

### DIFF
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -111,6 +111,7 @@ class GtsfmRunnerBase:
         parser.add_argument(
             "--share_intrinsics", action="store_true", help="Shares the intrinsics between all the cameras"
         )
+        parser.add_argument("--mvs_off", action="store_true", help="Turn off dense MVS reconstruction")
         parser.add_argument(
             "--output_root",
             type=str,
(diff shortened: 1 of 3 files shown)


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -111,6 +111,7 @@ def construct_argparser(self) -> argparse.ArgumentParser:
         parser.add_argument(
             "--share_intrinsics", action="store_true", help="Shares the intrinsics between all the cameras"
         )
+        parser.add_argument("--mvs_off", action="store_true", help="Turn off dense MVS reconstruction")
```
akshay-krishnan, line 114: nit: For flags it is more intuitive to have default false with the flag option to turn on. 
so having a flag like "--run_mvs, default = False" would let us do --run_mvs when we want to run it.  now we have to do --mvs_off=0 for running mvs.

akshay-krishnan (approved):
coincidentally I was doing this too locally today to avoid running MVS. I think its one of the changes in the large Hilti PR.

akshay-krishnan:
do we want to run MVS on the CI?

gtsfm/runner/gtsfm_runner_base.py
travisdriver: I believe having it this way will run it by default, and it can be turned off by calling `--mvs_off`, correct? I can change it to no run by default if you'd like.

akshay-krishnan: if the default is true, yeah.

akshay-krishnan:
shall we merge this?

akshay-krishnan:
@travisdriver Im merging it!

PR #583, opened 2022-10-25
Image loader visualization

stepanyanhayk:
This PR creates image viewer functionality in the JS visualization tool. Along with showing the images, we also represent the focal length and the image shapes. Screenshot showing the final result attached:

image


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -2,6 +2,7 @@
 
 import argparse
 import time
+import os
```
akshay-krishnan, line 5: should be in alphabetical order `os < time`

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -202,6 +205,23 @@ def run(self) -> None:
         process_graph_generator.save_graph()
 
         pairs_graph = self.retriever.create_computation_graph(self.loader)
+
+        image_vis_data = {}
+        for i in range(len(self.loader)):
```
akshay-krishnan, line 210: This should be optional, controlled by a flag. I think we also duplicate metrics and outputs in `REACT_METRICS_PATH`. Maybe we can use the same flag to decide whether we want to do this. It is non-trivial time for large datasets.

rtf_vis_tool/src/Components/ImageViewer.js
```diff
@@ -0,0 +1,55 @@
...
+      }
+    const images = importAll(require.context('../images', false, /\.(png|jpe?g|svg)$/));
+    const imageFileNames = Object.keys(images);
+
+    // sort file names
+    imageFileNames.sort((a, b) => {
+        a = parseInt(a.split(".")[0])
+        b = parseInt(b.split(".")[0])
+        console.log(a)
+        console.log(b)
```
akshay-krishnan, line 26: do we need to log here?

akshay-krishnan (commented):
did you actually change the svg or is it just regenerated ? 
We also need a TODO/issue to visualize large datasets. Please test all visualization code on larger datasets :)

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: addressed in a new commit

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: addressed in a new commit

rtf_vis_tool/src/Components/ImageViewer.js
stepanyanhayk: thanks for the catch, it was for debugging purposes, removed in the new commit

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -30,6 +32,7 @@
 logger = logger_utils.get_logger()
 
 DEFAULT_OUTPUT_ROOT = Path(__file__).resolve().parent.parent.parent
+REACT_METRICS_PATH = Path(__file__).resolve().parent.parent.parent / "rtf_vis_tool" / "src" / "images"
```
akshay-krishnan, line 35: that doesnt look right. why is react metrics pointing to a different directory now?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -118,6 +121,12 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             help="Root directory. Results, plots and metrics will be stored in subdirectories,"
             " e.g. {output_root}/results",
         )
+        parser.add_argument(
+            "--run_visualizer",
```
akshay-krishnan, line 125: nit: a shorter --run_viz might be better.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -202,6 +211,24 @@ def run(self) -> None:
...
+                )
+                image_vis_data[i] = {
+                    "shape": self.loader.get_image_shape(i),
+                    "focal_length": self.loader.get_camera_intrinsics(i).fx()
+                }
+
+            io_utils.save_json_file(
+                os.path.join(REACT_METRICS_PATH, "image_shapes.json"),
+                image_vis_data
+            )
```
akshay-krishnan, line 230: why do we need to save them again?

stepanyanhayk: so we can retrieve the images and shapes easily in the JS visualization

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: fixed

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: changes the name to REACT_IMAGES_PATH in a new commit. The path is different but it should not be in the same folder as the metrics in my opinion

gtsfm/runner/gtsfm_runner_base.py
akshay-krishnan: I get why you are creating the json file, but why duplicate the images? you could add the original image filename in the json too, and just open that image? am I missing something about reading images from the dataset?

PR #585, opened 2022-10-29
Fixing dependency version conflicts and moving to mamba

ayushbaid:
As argoverse pins the numpy version to 1.19, it is creating conflicts while running GTSFM. This PR also moves to mamba which is faster for conda env installs (and hence can do without caching packages)

### DIFF
--- a/.github/workflows/test-python.yml
+++ b/.github/workflows/test-python.yml
@@ -15,17 +15,9 @@ jobs:
 
     steps:
       - uses: actions/checkout@v3.1.0
-      - name: Cache conda env
-        uses: actions/cache@v3
-        env:
-          # Increase this value to reset cache if environment_linux_cpuonly.yml has not changed
-          CACHE_NUMBER: 0
-        with:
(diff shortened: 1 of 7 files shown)


.github/workflows/test-python.yml
```diff
@@ -41,7 +33,9 @@ jobs:
       - name: Unit tests
         run: |
           pytest tests --cov gtsfm \
-            --ignore tests/repro_tests
+            --ignore tests/repro_tests \
+            --ignore tests/loader/test_argoverse_dataset_loader.py \
+            --ignore tests/runner/test_frontend_runner.py
```
akshay-krishnan, line 38: does frontend runner also use argoverse?

akshay-krishnan (approved):

akshay-krishnan:
Lets also create an issue to add argoverse back once it is fixed.

akshay-krishnan:
I am also in favor of maintaining runners that need their own libraries in a separate repo, but we probably need further discussion for this.

PR #587, opened 2022-10-31
Using LOFTR in benchmark CI

ayushbaid:
In this PR, we introduce a new GTSFM config which uses LOFTR, and use it for few datasets in the CI)


ayushbaid:
Skydio-32 reconstruction: 3d

ayushbaid:
We see improvement in the number of tracks, but other metrics become worse. As a result, discarding this PR for now. 


## After verifier, but before 2-view BA
pre_2view_ba

## Sparse reconstruction
sparse_reconstruction

PR #589, opened 2022-11-02
Add a flag to change Dask's temp directory

akshay-krishnan:
The default `/tmp` may not be accessible to users without sudo access.


johnwlambert (approved):
LGTM, trusting you to test it out manually and confirm the fix : - )

akshay-krishnan:
Thanks! Yes, it worked for me. It was a local change I did not want to discard.

PR #594, opened 2022-11-21
Using tracks in translation averaging

akshay-krishnan:
We have the 2D tracks before translation averaging. These essentially give us 3D direction measurements from cameras to 3D points. Following 1DSfM, we add these to the translation averaging problem to provide more constraints. 
We only add a few tracks (this number needs to be tuned). 

This PR also refactors translation averaging, moves some functions to utils, and adds unit tests.


tests/utils/test_sampling_utils.py
```diff
@@ -28,3 +30,40 @@ def test_sample_points_on_plane() -> None:
...
+    mean_direction = np.mean(directions, axis=0)
+    std_direction = np.std(directions, axis=0)
+    std_direction_relative = std_direction / std_direction[0]
+
+    # Increase number of samples for a smaller threshold
+    assert np.allclose(mean_direction, np.zeros(3), atol=1e-1)
+    assert np.allclose(std_direction_relative, np.ones(3), atol=1e-1)
+
+
+def test_sample_kde_directions() -> None:
```
johnwlambert, line 51: nit: could we add a brief 1-sentence docstring to document in prose what we are testing here?

tests/utils/test_sampling_utils.py
```diff
@@ -28,3 +30,40 @@ def test_sample_points_on_plane() -> None:
     np.testing.assert_almost_equal(pts_residuals, np.zeros((num_points, 1)))
 
     assert pts.shape == (10, 3)
+
+
+def test_sample_random_directions() -> None:
```
johnwlambert, line 35: nit: could we add a brief 1-sentence docstring to document in prose what we are testing here?

gtsfm/common/sfm_track.py
```diff
@@ -69,6 +69,18 @@ def select_subset(self, idxs: List[int]) -> "SfmTrack2d":
...
+    def select_for_cameras(self, camera_idxs: Set[int]) -> "SfmTrack2d":
+        """Generates a new track with only those measurements which are in camera_idxs.
+
+        Unlike `select_subset`, this method does not require all camera_idxs to have a measurement in this track.
+
+        Returns:
+            Track with the subset of measurements.
+        """
+        measurements = [m for m in self.measurements if m.i in camera_idxs]
+
```
johnwlambert, line 81: nit: i propose we get rid of the newline between these two lines, to keep it a bit more compact

gtsfm/averaging/translation/translation_averaging_base.py
```diff
@@ -77,6 +81,8 @@ def create_computation_graph(
         num_images: int,
         i2Ui1_graph: Delayed,
         wRi_graph: Delayed,
+        tracks_2d: Delayed,
```
johnwlambert, line 84: nit: let's update the `Args` section below to document these two new arguments

gtsfm/averaging/translation/translation_averaging_base.py
```diff
@@ -50,6 +52,8 @@ def run_translation_averaging(
         num_images: int,
         i2Ui1_dict: Dict[Tuple[int, int], Optional[Unit3]],
         wRi_list: List[Optional[Rot3]],
+        tracks_2d: List[SfmTrack2d] = [],
```
johnwlambert, line 55: nit: let's update the `Args` section below to document these two new arguments

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -352,18 +469,19 @@ def _get_measurement_angle_errors(
...
         if (i1, i2) in i2Ui1_measurements and (i1, i2) in gt_i2Ui1_measurements:
-            errors.append(
-                comp_utils.compute_relative_unit_translation_angle(
-                    i2Ui1_measurements[(i1, i2)], gt_i2Ui1_measurements[(i1, i2)]
-                )
+            error = comp_utils.compute_relative_unit_translation_angle(
+                i2Ui1_measurements[(i1, i2)], gt_i2Ui1_measurements[(i1, i2)]
             )
+            if error is None:
+                raise ValueError("Unexpected None when computnig relative translation angle metric.")
```
johnwlambert, line 479: nit: typo on `computing`

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,227 @@ def _get_prior_measurements_in_world_frame(
...
+        w_i2Ui1_dict_tracks: RelativeDirectionsDict,
+        wRi_list,
+        i2Ti1_priors,
+        absolute_pose_priors,
+        scale_factor,
+    ):
+        """Runs the averaging algorithm.
+
+        Args:
+            w_i2Ui1_dict: Relative translations between cameras in world frame.
```
johnwlambert, line 340: nit: let's update this Args section : - )

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,227 @@ def _get_prior_measurements_in_world_frame(
...
+                cam_idx = measurement.i
+
+                if intrinsics[cam_idx] is None or wRi_list[cam_idx] is None:
+                    raise ValueError("Camera intrinsics or rotation cannot be None for input track measurements")
+
+                measurement_xy = intrinsics[cam_idx].calibrate(measurement.uv)
+                measurement_homog = Point3(measurement_xy[0], measurement_xy[1], 1.0)
+                w_cam_U_track = Unit3(wRi_list[cam_idx].rotate(Unit3(measurement_homog).point3()))
+
+                landmark_directions[(track_id, cam_idx)] = w_cam_U_track  # w_i2Ui1 here
```
johnwlambert, line 324: is the commented out line intentional? Can we add a few more words if so to make it clearer?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -17,41 +17,49 @@
...
 NOISE_MODEL_SIGMA = 0.01
 HUBER_LOSS_K = 1.345  # default value from GTSAM
 
 MAX_INLIER_MEASUREMENT_ERROR_DEG = 5.0
 
+TRACKS_TO_CAMERAS_RATIO = 20
+
 logger = logger_utils.get_logger()
 
+C = symbol_shorthand.A
```
johnwlambert, line 58: nit: can we add an inline coment saying `C represents camera variables, and L represents landmark variables`

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -17,41 +17,49 @@
...
 OUTLIER_WEIGHT_THRESHOLD = 0.125
-MAX_KDE_SAMPLES = 2000
 
 NOISE_MODEL_DIMENSION = 3  # chordal distances on Unit3
 NOISE_MODEL_SIGMA = 0.01
 HUBER_LOSS_K = 1.345  # default value from GTSAM
 
 MAX_INLIER_MEASUREMENT_ERROR_DEG = 5.0
 
+TRACKS_TO_CAMERAS_RATIO = 20
```
johnwlambert, line 54: nit: can we add a comment about how the number 20 was chosen

.github/workflows/benchmark.yml
```diff
@@ -42,7 +42,7 @@ jobs:
         with:
           path: cache
           key: ${{ matrix.config_dataset_info[0] }}-${{ matrix.config_dataset_info[1] }}-${{ matrix.config_dataset_info[2] }}-${{ matrix.config_dataset_info[6] }}-${{ env.CACHE_NUMBER_FRONTEND }}
-      - uses: conda-incubator/setup-miniconda@v2
+      - uses: mamba-org/provision-with-micromamba@main
```
johnwlambert, line 45: Have we profiled this change to see what influence it has?

gtsfm/averaging/translation/translation_averaging_base.py
```diff
@@ -98,5 +104,13 @@ def create_computation_graph(
             A GtsfmMetricsGroup with translation averaging metrics wrapped as Delayed.
         """
         return dask.delayed(self.run_translation_averaging, nout=2)(
-            num_images, i2Ui1_graph, wRi_graph, absolute_pose_priors, i2Ti1_priors, scale_factor, gt_wTi_list
+            num_images,
```
johnwlambert, line 107: nit: i think keyword args are allowed if I recall. could help here since we've got so many args these days

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Maybe "Unexpected `None` when computing ... " (single quote helps on var name)

.github/workflows/benchmark.yml
akshay-krishnan: I borrowed this from @ayushbaid

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: yet to do this process of picking the right number, I'll update this.

gtsfm/utils/sampling.py
```diff
@@ -39,3 +45,43 @@ def sample_points_on_plane(
 
     pts = np.hstack([x, y, z])
     return pts
+
+
+def sample_random_directions(num_samples: int) -> List[Unit3]:
+    """Samples num_samples Unit3 3D directions.
+    The sampling is done in 2D spherical coordinates (azimuth, elevation), and then converted to Cartesian coordinates.
```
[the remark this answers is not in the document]
akshay-krishnan: nope, good catch.

gtsfm/multi_view_optimizer.py
```diff
@@ -98,11 +99,14 @@ def create_computation_graph(
         delayed_wRi, rot_avg_metrics = self.rot_avg_module.create_computation_graph(
             num_images, pruned_i2Ri1_graph, i1Ti2_priors=relative_pose_priors, gt_wTi_list=gt_wTi_list
         )
+        tracks2d_graph = dask.delayed(get_2d_tracks)(viewgraph_v_corr_idxs_graph, keypoints_graph)
```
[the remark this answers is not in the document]
akshay-krishnan: Ive done the easy thing, and passed it to data association. I don't know if "data association" is the right name for that module anymore. And we probably need another blue node in the graph to denote 2D tracking.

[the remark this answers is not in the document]
akshay-krishnan: cc @johnwlambert

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+    def __run_averaging(
+        self,
+        num_images: int,
+        w_i2Ui1_dict: RelativeDirectionsDict,
+        w_i2Ui1_dict_tracks: RelativeDirectionsDict,
+        wRi_list: List[Optional[Rot3]],
+        i2Ti1_priors: Dict[Tuple[int, int], PosePrior],
+        absolute_pose_priors: List[Optional[PosePrior]],
+        scale_factor: float,
+    ):
```
johnwlambert, line 337: nit: missing return type `List[Optional[Point3]]`

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+                wti_list[i] = wti_values.atPoint3(C(i))
+        return wti_list
+
     # TODO(ayushbaid): Change wTi_initial to Pose3.
     def run_translation_averaging(
         self,
         num_images: int,
         i2Ui1_dict: Dict[Tuple[int, int], Optional[Unit3]],
         wRi_list: List[Optional[Rot3]],
+        tracks_2d: List[SfmTrack2d] = None,
```
johnwlambert, line 392: should this type be `Optional[List[SfmTrack2d]]`? Can we run mypy to see if we missed any?

gtsfm/averaging/translation/translation_averaging_base.py
johnwlambert: Noting as resolved, thanks.

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Noting as resolved, thanks.

gtsfm/common/sfm_track.py
johnwlambert: Noting as resolved, thanks.

tests/utils/test_sampling_utils.py
```diff
@@ -28,3 +30,42 @@ def test_sample_points_on_plane() -> None:
...
+
+    actual_samples = sampling_utils.sample_kde_directions(samples_normd, num_samples)
+    assert len(actual_samples) == num_samples
+
+    actual_samples_array = np.array([sample.point3() for sample in actual_samples])
+    samples_mean = np.mean(actual_samples_array, axis=0)
+    samples_mean_relative = samples_mean / samples_mean[0]
+    samples_cov = np.cov(actual_samples_array.T)
+
+    assert np.allclose(samples_mean_relative, mean, atol=1e-1)
```
johnwlambert, line 70: Should we set a random seed? For few samples, this assert may fail

tests/utils/test_sampling_utils.py
```diff
@@ -28,3 +30,42 @@ def test_sample_points_on_plane() -> None:
...
+    # centered at [1, 1, 1] with variance 0.1
+    mean = np.array([1, 1, 1])
+    cov = np.array([[0.1, 0, 0], [0, 0.1, 0], [0, 0, 0.1]])
+    samples = np.random.multivariate_normal(mean, cov, num_samples)
+    samples_normd = [Unit3(sample) for sample in samples]
+
+    actual_samples = sampling_utils.sample_kde_directions(samples_normd, num_samples)
+    assert len(actual_samples) == num_samples
+
+    actual_samples_array = np.array([sample.point3() for sample in actual_samples])
```
johnwlambert, line 65: Maybe would be helpful to add an inline comment like
`# Compute empirical sample mean and covariance and verify matches specified distribution.`

tests/utils/test_sampling_utils.py
```diff
@@ -28,3 +30,42 @@ def test_sample_points_on_plane() -> None:
...
+    samples_mean = np.mean(actual_samples_array, axis=0)
+    samples_mean_relative = samples_mean / samples_mean[0]
```
johnwlambert, line 67: Could e add a comment about why we need to divide by something? I didn't follow

tests/utils/test_sampling_utils.py
```diff
@@ -28,3 +30,42 @@ def test_sample_points_on_plane() -> None:
...
+    """Check that the sampled directions are uniformly distributed on the sphere, using their mean and std dev."""
+    num_samples = 1000
+
+    direction_list = sampling_utils.sample_random_directions(num_samples)
+    assert len(direction_list) == num_samples
+
+    directions = np.array([direction.point3() for direction in direction_list])
+    mean_direction = np.mean(directions, axis=0)
+    std_direction = np.std(directions, axis=0)
+    std_direction_relative = std_direction / std_direction[0]
```
johnwlambert, line 45: would it be helpful to add an inline comment explaining why we need to divide by standard dev here?

tests/data_association/test_data_association.py
```diff
@@ -299,7 +306,11 @@ def test_create_computation_graph(self):
         )
         da = DataAssociation(min_track_len=3, triangulation_options=triangulation_options)
         expected_sfm_data, expected_metrics = da.run(
-            len(cameras), cameras, corr_idxs_graph, keypoints_list, cameras_gt=cameras_gt, relative_pose_priors={}
+            len(cameras),
```
johnwlambert, line 309: nit: can we make these all keyword args, since I have a hard time remembering what represents what?

tests/data_association/test_data_association.py
```diff
@@ -177,7 +182,11 @@ def test_triangulation_individualCal_without_ransac(self):
         da = DataAssociation(min_track_len=2, triangulation_options=triangulation_options)
 
         sfm_data, _ = da.run(
-            len(cameras), cameras, matches_dict, keypoints_list, [None] * len(cameras), relative_pose_priors={}
+            len(cameras),
```
johnwlambert, line 185: nit: could we make these all keyword args? helps me follow

tests/data_association/test_data_association.py
```diff
@@ -80,6 +81,11 @@ def generate_noisy_2d_measurements(
     return keypoints_list, img_idxs, cameras
 
 
+def get_2d_tracks(corr_idxs_dict, keypoints_list):
```
johnwlambert, line 84: nit: could we add in the type hints here, presumably something like
```python
def get_2d_tracks(
    corr_idxs_dict: Dict[Tuple[int,int], np.ndarray], keypoints_list: List[Keypoints]
) -> List[SfmTrack2d]
```

tests/data_association/test_data_association.py
```diff
@@ -17,6 +17,7 @@
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.data_association.data_assoc import DataAssociation
 from gtsfm.data_association.point3d_initializer import TriangulationOptions, TriangulationSamplingMode
+from gtsfm.data_association.dsf_tracks_estimator import DsfTracksEstimator
```
johnwlambert, line 20: nit: by isort this line should go above by one line i believe

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+        intrinsics = [Cal3_S2(20, 20, 0.0, 0, 0)]
+        test_wRi_mat = np.eye(3)
+        test_wRi_mat[1, 1] = -1
+        wRi_list = [Rot3(test_wRi_mat)]
+
+        expected_w_i2Ui1_dict = {(0, 0): Unit3(Point3(1, -1, 1))}  # indices here are (track_idx, camera_idx)
+        actual_w_i2Ui1_dict = self.obj._get_landmark_directions(tracks, intrinsics, wRi_list)
+
+        self.assertEqual(expected_w_i2Ui1_dict.keys(), actual_w_i2Ui1_dict.keys())
+        self.assertTrue(np.array_equal(expected_w_i2Ui1_dict[(0, 0)].point3(), actual_w_i2Ui1_dict[(0, 0)].point3()))
```
johnwlambert, line 112: Maybe would be helpful to add a comment about why this is the expected result?

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+    def test_get_landmark_directions(self) -> None:
+        test_measurement = np.array([20, 20])
+        tracks = [SfmTrack2d(measurements=[SfmMeasurement(0, test_measurement)])]
+
+        intrinsics = [Cal3_S2(20, 20, 0.0, 0, 0)]
+        test_wRi_mat = np.eye(3)
+        test_wRi_mat[1, 1] = -1
+        wRi_list = [Rot3(test_wRi_mat)]
+
+        expected_w_i2Ui1_dict = {(0, 0): Unit3(Point3(1, -1, 1))}  # indices here are (track_idx, camera_idx)
```
johnwlambert, line 108: Maybe would be helpful to add a comment about why this is the expected result?

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+        }
+        self.assertEqual(expected_selected_tracks, actual_selected_track_cameras)
+
+    def test_get_landmark_directions(self) -> None:
+        test_measurement = np.array([20, 20])
+        tracks = [SfmTrack2d(measurements=[SfmMeasurement(0, test_measurement)])]
+
+        intrinsics = [Cal3_S2(20, 20, 0.0, 0, 0)]
+        test_wRi_mat = np.eye(3)
+        test_wRi_mat[1, 1] = -1
```
johnwlambert, line 105: Could we add an inline comment explaining what's going on with the euler angles here?

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+        actual_selected_track_cameras = {
+            track_idx: [m.i for m in track.measurements] for track_idx, track in enumerate(actual_selected_tracks)
+        }
+        self.assertEqual(expected_selected_tracks, actual_selected_track_cameras)
+
+    def test_get_landmark_directions(self) -> None:
+        test_measurement = np.array([20, 20])
+        tracks = [SfmTrack2d(measurements=[SfmMeasurement(0, test_measurement)])]
+
+        intrinsics = [Cal3_S2(20, 20, 0.0, 0, 0)]
```
johnwlambert, line 103: nit: prefer keyword args if possible, since I forget what the order of args is for Cal3_S2

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+        actual_w_i1ti2 = np.array([i1ti2.measured() for i1ti2 in w_i1ti2_measurements])
+        expected_w_i1ti2 = np.array([Point3(1, -1, 0) for (i1, i2) in i2Ti1_dict.keys()])
+
+        self.assertTrue(np.array_equal(actual_w_i1ti2, expected_w_i1ti2))
+
+    def test_select_tracks_for_averaging(self) -> None:
+        """Tests that the tracks are correctly selected to include in averaging."""
+        valid_cameras = set([0, 1, 2, 3])
+        test_measurement = np.array([20, 20])
+        test_tracks_cameras = {0: [0, 1, 2, 3, 5, 6], 1: [1, 2, 3], 2: [0, 1, 3, 4], 3: [0, 2]}
```
johnwlambert, line 85: nit: inline comment would help to explain what the values in the dictionary represent

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+        test_tracks = []
+        for track_idx, cameras in test_tracks_cameras.items():
+            measurements = [SfmMeasurement(i, test_measurement) for i in cameras]
+            test_tracks.append(SfmTrack2d(measurements=measurements))
+
+        intrinsics = [Cal3_S2(20, 20, 0.0, 0, 0)] * 7
+        expected_selected_tracks = {0: [0, 1, 2, 3], 1: [1, 2, 3], 2: [0, 1, 3]}
```
johnwlambert, line 92: Could we add an inline comment about why this is the expected result?

tests/averaging/translation/test_averaging_1dsfm.py
```diff
@@ -16,21 +17,99 @@
...
+from gtsfm.common.pose_prior import PosePrior, PosePriorType
+from gtsfm.common.sfm_track import SfmMeasurement, SfmTrack2d
 from gtsfm.loader.olsson_loader import OlssonLoader
 
 DATA_ROOT_PATH = Path(__file__).resolve().parent.parent.parent / "data"
 
 RELATIVE_ERROR_THRESHOLD = 3e-2
 ABSOLUTE_ERROR_THRESHOLD = 3e-1
 
+C = gtsam.symbol_shorthand.A
```
johnwlambert, line 29: inline comment here would help too
```python
C = symbol_shorthand.A  # for camera translation variables
L = symbol_shorthand.B  # for track (landmark) translation variables
```

gtsfm/utils/sampling.py
```diff
@@ -39,3 +43,42 @@ def sample_points_on_plane(
...
+    """Samples `num_samples` Unit3 3D directions, uniformly in Cartesian coordinates.
+
+    Args:
+        num_samples: Number of samples required.
+
+    Returns:
+        List of sampled Unit3 directions.
+    """
+    samples = np.random.normal(size=(num_samples, 3))
+    return [Unit3(sample / np.linalg.norm(sample)) for sample in samples]
```
johnwlambert, line 58: will Unit3 normalize these automatically? or is there a Unit3.normalize() we can call?

gtsfm/multi_view_optimizer.py
```diff
@@ -158,3 +161,8 @@ def init_cameras(
             cameras[idx] = camera_class(wTi, intrinsics_list[idx])
 
     return cameras
+
+
+def get_2d_tracks(corr_idxs_dict, keypoints_list):
```
johnwlambert, line 166: nit could we add type hints here:
```python
def get_2d_tracks(
    corr_idxs_dict: Dict[Tuple[int,int], np.ndarray], keypoints_list: List[Keypoints]
) -> List[SfmTrack2d]:
```

gtsfm/multi_view_optimizer.py
```diff
@@ -112,8 +116,7 @@ def create_computation_graph(
         ba_input_graph, data_assoc_metrics_graph = self.data_association_module.create_computation_graph(
             num_images,
             init_cameras_graph,
-            viewgraph_v_corr_idxs_graph,
-            keypoints_graph,
+            tracks2d_graph,
```
johnwlambert, line 119: nit: keyword args preferable since i get mixed up from all the args

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -429,15 +553,28 @@ def _compute_metrics(
...
         wRi2 = wRi_list[i2]
         if i2Ui1 is not None and wRi2 is not None:
-            # TODO: what if wRi2 is None, but wRi1 is not? Can we still transform.
-            w_i2Ui1_measurements.append(BinaryMeasurementUnit3(i2, i1, Unit3(wRi2.rotate(i2Ui1.point3())), noise_model))
-
-    return w_i2Ui1_measurements
+            w_i2Ui1_dict[(i1, i2)] = Unit3(wRi2.rotate(i2Ui1.point3()))
+            valid_cameras.add(i1)
+            valid_cameras.add(i2)
+        # should we retain a None in the dict?
```
johnwlambert, line 579: Didn't quite follow this question @akshay-krishnan

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -429,15 +553,28 @@ def _compute_metrics(
...
+    """Returns measurements for which both cameras have valid rotations, transformed to the world frame.
+
+    Args:
+        i2Ui1_dict: Relative translation directions between camera pairs.
+        wRi_list: List of estimated camera rotations.
 
-    w_i2Ui1_measurements = BinaryMeasurementsUnit3()
+    Returns:
+        Tuple of:
+            Relative translation directions between camera pairs, in world frame.
```
johnwlambert, line 567: could we note here what is pointing to what?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -233,114 +413,57 @@ def run_translation_averaging(
...
-        ]
-        # Combine estimated global rotations and global translations to Pose(3) objects.
-        wTi_list = [
-            Pose3(wRi, wti) if wRi is not None and wti is not None else None for wRi, wti in zip(wRi_list, wti_list)
-        ]
+        w_i2Ui1_dict_inliers, w_i2Ui1_dict_tracks_inliers, inlier_cameras = self.compute_inliers(
+            w_i2Ui1_dict, w_i2Ui1_dict_tracks
+        )
+
+        wti_list = self.__run_averaging(
```
johnwlambert, line 436: nit: keyword args would be nice here if possible because we've got a lot of args

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+            algorithm = TranslationRecovery()
+            if len(i2Ti1_priors) > 0:
+                # scale is ignored here.
+                w_i1ti2_priors = self._binary_measurements_from_priors(i2Ti1_priors, wRi_list)
+                wti_initial = self.__get_initial_values(absolute_pose_priors)
+                wti_values = algorithm.run(w_i1Ui2_measurements, 0.0, w_i1ti2_priors, wti_initial)
+            else:
+                wti_values = algorithm.run(w_i1Ui2_measurements, scale_factor)
+        except TypeError as e:
+            # TODO(akshay-krishnan): remove once latest gtsam pip wheels updated.
```
johnwlambert, line 374: can we modify this to explain which pip wheel does not support? which does 4.2a7 support? https://pypi.org/project/gtsam/#history

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+
+        Bad tracks are those that have fewer than 3 measurements from valid_cameras.
+        Sorts the remaining tracks in descending order by number of measurements, and returns as many tracks
+        as tracks_to_cameras_ratio * number of valid cameras.
+
+        Args:
+            tracks: List of all input tracks.
+            valid_cameras: Set of valid camera indices (these have direction measurements and valid rotations).
+            intrinsics: List of camera intrinsics.
+            tracks_to_cameras_ratio: Ratio of tracks to cameras to use for averaging.
```
johnwlambert, line 280: What happens if a camera does have X number of tracks? does it mean every camera needs to have N tracks?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+
+        Returns:
+            List of tracks to use for averaging.
+        """
+        max_tracks = int(len(valid_cameras) * tracks_to_cameras_ratio)
+        filtered_tracks = []
+        valid_cameras_with_intrinsics = set([c for c in valid_cameras if intrinsics[c] is not None])
+        for track in tracks:
+            valid_cameras_track = track.select_for_cameras(camera_idxs=valid_cameras_with_intrinsics)
+            if valid_cameras_track.number_measurements() < 3:
```
johnwlambert, line 290: nit: can we make 3 some sort of global var, like `MIN_NUM_TRACKS_FOR_BLAH`?

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: I think i'm misunderstanding the variable name since we use it below as (# cameras) * ratio. Maybe we could clarify that this var just means that we will use X times as many tracks as valid cameras exist.

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+        Returns:
+            Set of indices (i1, i2) which are inliers.
+        """
+
+        # Sample directions for projection
+        combined_measurements = list(w_i2Ui1_dict.values()) + list(w_i2Ui1_dict_tracks.values())
+        projection_directions = self.__sample_projection_directions(combined_measurements)
 
-    def __get_initial_values(self, wTi_initial: List[Optional[PosePrior]]):
+        # Convert to measurements - indexes to symbols.
```
johnwlambert, line 209: nit: what does the hyphen mean here? prefer something else since can be confused with minus

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+        inlier_w_i2Ui1_dict = {}
+        inlier_w_i2Ui1_dict_tracks = {}
+        inlier_cameras: Set[int] = set()
+        for (i1, i2) in w_i2Ui1_dict:
+            if (C(i2), C(i1)) in inliers:  # there is a flip in indices from w_i2Ui1_dict to inliers.
+                inlier_w_i2Ui1_dict[(i1, i2)] = w_i2Ui1_dict[(i1, i2)]
+                inlier_cameras.add(i1)
+                inlier_cameras.add(i2)
+
+        for (track_id, cam_id) in w_i2Ui1_dict_tracks:
```
johnwlambert, line 242: inline comment for lines 242-244 would be nice since lots going on here

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+        w_i2Ui1_dict_tracks: RelativeDirectionsDict,
+    ) -> Tuple[RelativeDirectionsDict, RelativeDirectionsDict, Set[int]]:
+        """Perform inlier detection for the relative direction measurements.
+
+        Args:
+            w_i2Ui1_dict: Dictionary of Unit3 relative translations between cameras.
+            w_i2Ui1_dict_tracks: Dictionary of Unit3 relative translations between cameras and landmarks.
+
+        Returns:
+            Set of indices (i1, i2) which are inliers.
```
johnwlambert, line 202: nit: i think there should be 3 return args here for `inlier_w_i2Ui1_dict, inlier_w_i2Ui1_dict_tracks, inlier_cameras`

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -88,81 +96,66 @@ def __init__(
         self._outlier_weight_threshold = OUTLIER_WEIGHT_THRESHOLD
         self._reject_outliers = reject_outliers
         self._projection_sampling_method = projection_sampling_method
+        self._use_tracks_for_averaging = use_tracks_for_averaging
 
     def __sample_projection_directions(
         self,
-        w_i2Ui1_measurements: BinaryMeasurementsUnit3,
+        w_i2Ui1_list: List[Unit3],
```
johnwlambert, line 103: just curious, why did we use `BinaryMeasurementsUnit3` before, but now we use `List[Unit3]`? is it because the BinaryMeasurements assumes we have (i1,i2) poses and now we have directions from points also, which have no camera index?

johnwlambert (approved):
LGTM, dashboard from CI after making last changes would be nice

johnwlambert:
btw CI seems to be failing

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -233,114 +413,57 @@ def run_translation_averaging(
...
+        ]
+        return wTi_list, ta_metrics
 
 
-def _get_measurement_angle_errors(
+def get_measurement_angle_errors(
     i1_i2_pairs: Set[Tuple[int, int]],
-    i2Ui1_measurements: Dict[Tuple[int, int], Unit3],
-    gt_i2Ui1_measurements: Dict[Tuple[int, int], Unit3],
-) -> List[Optional[float]]:
```
johnwlambert, line 343: why were there None's in here before, but not anymore?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +164,233 @@ def _get_prior_measurements_in_world_frame(
...
+            mfas_instance = MFAS(w_i1Ui2_measurements, direction)
+            outlier_weights.append(mfas_instance.computeOutlierWeights())
+        logger.debug("Computed outlier weights using MFAS.")
+
+        # Compute average outlier weight.
+        outlier_weights_sum: DefaultDict[Tuple[int, int], float] = defaultdict(float)
+        inliers = set()
+        for outlier_weight_dict in outlier_weights:
+            for w_i1Ui2 in w_i1Ui2_measurements:
+                i1, i2 = w_i1Ui2.key1(), w_i1Ui2.key2()
```
johnwlambert, line 226: Do `i1` and `i2` still mean camera indices, or can one be a landmark index? If not, can we use different variable names to disambiguate, or can we add an inline comment noting this?

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Can we also clarify what `i1` and `i2` represent in this context? can be a landmark index?

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: maybe `MIN_NUM_MEAUSUREMENTS_HIGH_QUALITY_TRACK`?

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: I am adding a comment that says: `'inliers' contains both camera-camera and camera-landmark inliers. We separate them here.`
Does it matter what i1 and i2 mean? within this outlier computation they are just points - could be either cameras or landmarks.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: thanks, yeah, I can rename inlier_w_i2Ui1_dict_tracks to `inlier_w_cUt_dict` , but then folks have to remember that the idxs are (t, c) not (c, t). Using i2 and i1 just lets them know what the keys are. 
Also there are places (very close to the C++ interface) where these two are in the same data structure, they are combined before calling C++ code.

gtsfm/utils/sampling.py
akshay-krishnan: yeah it normalizes internally.

akshay-krishnan: this was written by copilot and I felt it was okay :)

akshay-krishnan: changed it now anyway

tests/data_association/test_data_association.py
akshay-krishnan: its unrelated to this PR, but added them

tests/utils/test_sampling_utils.py
akshay-krishnan: magnitude is irrelevant here, added a comment.

tests/utils/test_sampling_utils.py
akshay-krishnan: that is what the test description means, but I can choose this exact wording if you prefer that :)

tests/utils/test_sampling_utils.py
akshay-krishnan: only magnitudes are relevant here. I updated the test.

tests/utils/test_sampling_utils.py
akshay-krishnan: doesnt np set a default set when imported? so the test would always run on the same seed?

johnwlambert:
Thanks for all the updates @akshay-krishnan. I think there are still two small flake errors in the CI:

```
Run flake8 --max-line-length 120 --ignore E201,E202,E203,E231,W291,W293,E303,W391,E402,W503,E731 gtsfm tests
gtsfm/data_association/data_assoc.py:2[8](https://github.com/borglab/gtsfm/actions/runs/3542445705/jobs/5947945179#step:5:9):1: F401 'gtsfm.common.keypoints.Keypoints' imported but unused
gtsfm/data_association/data_assoc.py:31:1: F401 'gtsfm.data_association.dsf_tracks_estimator.DsfTracksEstimator' imported but unused
```

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Noting as resolved, thanks

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Noting as resolved, thanks!

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Got it, maybe if we amend this slightly to 

```python
tracks_to_cameras_ratio: Ratio of total number of tracks to use to total number of cameras.
```

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Got it, thanks for the clarification!

gtsfm/multi_view_optimizer.py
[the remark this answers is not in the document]
johnwlambert: Looks good!

Could we open an issue to rename the `DataAssociation` class to `PointTriangulatorWrapperBlah`? I would only consider data association to be the 2d match graph connected component finding (which we've moved out of it now).

johnwlambert (approved):

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: Noting as resolved, thanks

gtsfm/multi_view_optimizer.py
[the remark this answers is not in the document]
akshay-krishnan: done

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: I think `TRACK_MIN_MEASUREMENTS_FOR_AVERAGING` sounds better.

akshay-krishnan:
Translation averaging metrics: 
image

 
Data association metrics: 
image

Bundle adjustment metrics:
image

akshay-krishnan:
All places where we see a negative impact are already small numbers in terms of absolute value. 
Also the color scale does not fix yellow at zero (so sometimes improvements can have a light red shade).

akshay-krishnan:
Latency for translation averaging is expected to increase by 2-3x. This is dominated by an increase in the outlier rejection time as we have more measurements. 
Palace of fine arts (281 images) originally took ~3min for translation averaging on a desktop, but now takes ~7-8min.

gtsfm/averaging/translation/averaging_1dsfm.py
ayushbaid: I agree with John that i1, i2 can be a bit confusing. But I think we should be fine here.

ayushbaid: Maybe idx1, idx2?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+        projection_directions = self.__sample_projection_directions(combined_measurements)
+
+        # Convert to measurements: map indexes to symbols.
+        dummy_noise_model = gtsam.noiseModel.Isotropic.Sigma(3, 1e-2)  # MFAS does not use this.
+        w_i1Ui2_measurements = self._binary_measurements_from_dict(w_i2Ui1_dict, w_i2Ui1_dict_tracks, dummy_noise_model)
 
-    def __get_initial_values(self, wTi_initial: List[Optional[PosePrior]]):
+        # Compute outlier weights using MFAS.
+        # TODO(ayush): parallelize this step.
+        outlier_weights: List[Dict[Tuple[int, int], float]] = []
```
ayushbaid, line 220: Just curious: can we make outlier_weights dict keyed on symbols instead of integers?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+            gtsam.Values containing initial translations (uses symbols for camera index).
         """
         initial = gtsam.Values()
         for i, wTi in enumerate(wTi_initial):
             if wTi is not None:
-                initial.insertPoint3(i, wTi.value.translation())
+                initial.insertPoint3(C(i), wTi.value.translation())
         return initial
 
+    def _select_tracks_for_averaging_by_measurements(
```
ayushbaid, line 271: It is not clear to me why we use the prefix `_by_measurements`

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+            tracks: List of all input tracks.
+            valid_cameras: Set of valid camera indices (these have direction measurements and valid rotations).
+            intrinsics: List of camera intrinsics.
+            measurements_per_camera: Number of track direction measurements that need to be observed by each camera.
+
+        Returns:
+            List of tracks to use for averaging.
+        """
+        filtered_tracks = []
+        valid_cameras_with_intrinsics = set([c for c in valid_cameras if intrinsics[c] is not None])
```
ayushbaid, line 296: Will we run into a scenario where we have a camera in `valid_cameras`, but not have intrinsics for it?

ayushbaid: We might not need to do this here and can just rely on `valid_cameras` input. Even if we do need to explicitly check for intrinsics, maybe do it before this function?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+        self,
+        tracks: List[SfmTrack2d],
+        valid_cameras: Set[int],
+        intrinsics: List[Optional[gtsfm_types.CALIBRATION_TYPE]],
+        measurements_per_camera=TRACKS_MEASUREMENTS_PER_CAMERA,
+    ) -> List[SfmTrack2d]:
+        """Removes bad tracks and selects the longest ones until all cameras see `measurements_per_camera` tracks.
+
+        Bad tracks are those that have fewer than 3 measurements from valid_cameras.
+        Sorts the remaining tracks in descending order, then selects the longest ones until all cameras see at least
```
ayushbaid, line 281: Are we sorting the tracks in this function?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+            tracks_subset.append(filtered_tracks[best_track_id])
+
+            # update the state variables: remaining_cover and improvement
+            improvement[best_track_id] = 0
+            for measurement in filtered_tracks[best_track_id].measurements:
+                # if this image just got covered the k'th time, it's done.
+                # lower the improvement for all tracks that see it
+                if remaining_cover[measurement.i] == 1:
+                    for t in camera_track_lookup[measurement.i]:
+                        improvement[t] = improvement[t] - 1 if improvement[t] > 0 else 0
```
ayushbaid, line 330: Nit: maybe move this step after the reduction in remaining cover, so that its more intuitive when checking against zero.

gtsfm/averaging/translation/averaging_1dsfm.py
ayushbaid: +1. There are be Nones in i2Ui1_measurements but not in gt_i2Ui1_measurements

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -233,114 +454,59 @@ def run_translation_averaging(
...
-    return conversion_utils.spherical_to_cartesian_directions(sampled_azimuth_elevation)
+        # Combine estimated global rotations and global translations to Pose(3) objects.
+        wTi_list = [
+            Pose3(wRi, wti) if wRi is not None and wti is not None else None for wRi, wti in zip(wRi_list, wti_list)
+        ]
+        return wTi_list, ta_metrics
 
 
-def _get_measurement_angle_errors(
+def get_measurement_angle_errors(
```
ayushbaid, line 505: I think we should move this function outside TA. Maybe the metric utils file?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -429,15 +596,28 @@ def _compute_metrics(
...
+            Relative translation directions between camera pairs, in world frame.
+            Set of camera indices for which we have valid rotations and measurements.
+    """
+
+    w_i2Ui1_dict = {}
+    valid_cameras: Set[int] = set()
     for (i1, i2), i2Ui1 in i2Ui1_dict.items():
         wRi2 = wRi_list[i2]
         if i2Ui1 is not None and wRi2 is not None:
-            # TODO: what if wRi2 is None, but wRi1 is not? Can we still transform.
```
ayushbaid, line 440: Why are we enforcing both cameras to have a global rotation. It makes sense to do it, but why do it now?

gtsfm/utils/sampling.py
```diff
@@ -39,3 +43,42 @@ def sample_points_on_plane(
...
+        List of sampled Unit3 directions.
+    """
+    if len(measurements) > max_kde_samples:
+        sampled_idx = np.random.choice(len(measurements), max_kde_samples, replace=False).tolist()
+        measurements_subset = [measurements[i] for i in sampled_idx]
+    else:
+        measurements_subset = measurements
+
+    measurements_spherical = conversion_utils.cartesian_to_spherical_directions(measurements_subset)
+    print(np.mean(measurements_spherical, axis=0))
```
johnwlambert, line 78: do we need the print?

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -233,114 +454,59 @@ def run_translation_averaging(
                 or ill-constrained system).
             A GtsfmMetricsGroup of 1DSfM metrics.
         """
-        logger.info("Running translation averaging on %d unit translations", len(i2Ui1_dict))
-        noise_model = gtsam.noiseModel.Isotropic.Sigma(NOISE_MODEL_DIMENSION, NOISE_MODEL_SIGMA)
-        if self._robust_measurement_noise:
-            huber_loss = gtsam.noiseModel.mEstimator.Huber.Create(HUBER_LOSS_K)
-            noise_model = gtsam.noiseModel.Robust.Create(huber_loss, noise_model)
+        logger.info("Running translation averaging on {} unit translations".format(len(i2Ui1_dict)))
```
johnwlambert, line 457: nit: prefer logger syntax with %d, to prevent unnecessary computation if logger level is set higher

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+                wti_values = algorithm.run(w_i1Ui2_measurements, 0.0, w_i1ti2_priors, wti_initial)
+            else:
+                wti_values = algorithm.run(w_i1Ui2_measurements, scale_factor)
+        except TypeError as e:
+            # TODO(akshay-krishnan): remove when no longer supporting gtsam versions before 4.2a7.
+            logger.error("TypeError: {}".format(str(e)))
+            algorithm = TranslationRecovery(w_i1Ui2_measurements)
+            wti_values = algorithm.run(scale_factor)
+
+        # transforming the result to the list of Point3
```
johnwlambert, line 420: nit: `Transforms the result to a list of Point3 objects.`

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+        valid_cameras_with_intrinsics = set([c for c in valid_cameras if intrinsics[c] is not None])
+        for track in tracks:
+            valid_cameras_track = track.select_for_cameras(camera_idxs=valid_cameras_with_intrinsics)
+            if valid_cameras_track.number_measurements() < MIN_TRACK_MEASUREMENTS_FOR_AVERAGING:
+                continue
+            filtered_tracks.append(valid_cameras_track)
+
+        tracks_subset = []
+
+        remaining_cover = {c: measurements_per_camera for c in valid_cameras_with_intrinsics}
```
johnwlambert, line 305: Could we add an inline comment explaining what `remaining_cover` means? didn't quite follow

gtsfm/averaging/translation/averaging_1dsfm.py
johnwlambert: I think `w_idx2_U_idx1` might be a bit too wordy. So I'm fine with re-using `i1` and `i2` here, with the help of comments wherever we use `i1` to mean `j` (landmark)

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -171,46 +166,272 @@ def _get_prior_measurements_in_world_frame(
...
+        inlier_w_i2Ui1_dict = {}
+        inlier_w_i2Ui1_dict_tracks = {}
+        inlier_cameras: Set[int] = set()
+        for (i1, i2) in w_i2Ui1_dict:
+            if (C(i2), C(i1)) in inliers:  # there is a flip in indices from w_i2Ui1_dict to inliers.
+                inlier_w_i2Ui1_dict[(i1, i2)] = w_i2Ui1_dict[(i1, i2)]
+                inlier_cameras.add(i1)
+                inlier_cameras.add(i2)
+
+        for (track_id, cam_id) in w_i2Ui1_dict_tracks:
```
johnwlambert, line 248: optional: we could also try something like
```python
for (j, i) in w_jUi_dict_tracks:
    if (C(i), L(j)) in inliers and i in inlier_cameras:
        inlier_w_jUi_dict_tracks[(j, i)] = w_jUi_dict_tracks[(j, i)]
```
to keep the `i` (camera index) and `j` (landmark index) notation we use elsewhere

gtsfm/data_association/data_assoc.py
```diff
@@ -25,9 +25,8 @@
 import gtsfm.utils.tracks as track_utils
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.common.image import Image
-from gtsfm.common.keypoints import Keypoints
 from gtsfm.common.pose_prior import PosePrior
-from gtsfm.data_association.dsf_tracks_estimator import DsfTracksEstimator
+from gtsfm.common.sfm_track import SfmTrack2d
```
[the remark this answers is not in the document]
akshay-krishnan: I created an issue for this

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: yeah good point, I removed the intrinsics check. If intrinsics is missing, something is very wrong, lets just crash then.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: because I had another function named _select_tracks_for_averaging before this. removed it now.

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: it is already. I dont know why there is an int type hint. are Symbols ints in python?

gtsfm/averaging/translation/averaging_1dsfm.py
akshay-krishnan: in theory if one of those measurements in None, the function can return a None. Do we want that?

akshay-krishnan:
Merging this, I think all comments have been addressed, thanks for the review!

PR #602, opened 2023-02-14
Fix TA metric computation failure when GT is not provided

akshay-krishnan:


gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -438,7 +438,7 @@ def run_translation_averaging(
         absolute_pose_priors: List[Optional[PosePrior]] = [],
         i2Ti1_priors: Dict[Tuple[int, int], PosePrior] = {},
         scale_factor: float = 1.0,
-        gt_wTi_list: Optional[List[Optional[Pose3]]] = None,
+        gt_wTi_list: List[Optional[Pose3]] = None,
```
travisdriver, line 441: Should this be `gt_wTi_list: List[Optional[Pose3]] = None`? If a function were to call `len(gt_wTi_list)` it would error out.

travisdriver (approved):
LGTM

PR #604, opened 2023-02-24
Loader delayed

stepanyanhayk:
In our efforts to run gtsfm on a cluster, we need to specify a worker/machine that would do the IO and share the data with other machines. To do this we need all our input to be delayed (so we can use dask to specify which worker should do the computation of the delayed task). This PR disallows all the direct access from `LoaderBase` to the images by creating dask computation graphs for the input.

### DIFF
--- a/gtsfm/runner/frontend_runner.py
+++ b/gtsfm/runner/frontend_runner.py
@@ -41,13 +41,11 @@ def run_frontend(
     image_pair_indices = loader.get_valid_pairs()
     camera_intrinsics_graph = loader.create_computation_graph_for_intrinsics()
     image_shapes_graph = loader.create_computation_graph_for_image_shapes()
-    
-    [image_shapes] = dask.compute(image_shapes_graph)
-    [camera_intrinsics] = dask.compute(camera_intrinsics_graph)
 
+    # TODO(stepanyanhayk) change function signature to only accept delayed objects
     (delayed_keypoints, delayed_putative_corr_idxs_dict,) = correspondence_generator.create_computation_graph(
         delayed_images=loader.create_computation_graph_for_images(),
-        image_shapes=image_shapes,
+        image_shapes=image_shapes_graph,
         image_pair_indices=image_pair_indices,
     )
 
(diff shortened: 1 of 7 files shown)


akshay-krishnan:
Nice work!

One problem with this design is that the base class assumes the derived classes do not read the images when computing the length of the dataset. Since you call len() everywhere. This is a fair assumption, but it would be nice to enforce it in the base class itself. Can we count the number of images in the dataset directory in the base loader itself? Does that work for all our loaders?

akshay-krishnan:
Please document the use of these functions in the loader base class. i.e in the base class description, we should mention that runners should only use the create_computation_graph methods when available, instead of the get_X methods.

akshay-krishnan (approved):

akshay-krishnan:
LGTM, modulo above comments.

ayushbaid:
> 

Yeah it is a decent assumption to make. Any loader implementation will have to work on getting the length in an efficient manner. Maybe add it to the docstring?

ayushbaid:
> Please document the use of these functions in the loader base class. i.e in the base class description, we should mention that runners should only use the create_computation_graph methods when available, instead of the get_X methods.

Can we make them private?

akshay-krishnan:
@stepanyanhayk please fix this unit test failure: 

```
___________________ TestFolderLoader.test_get_all_intrinsics ___________________

self = <test_olsson_loader.TestFolderLoader testMethod=test_get_all_intrinsics>

    def test_get_all_intrinsics(self) -> None:
        """Tests the graph for all intrinsics."""
    
>       all_intrinsics = self.loader.get_all_intrinsics()
E       AttributeError: 'OlssonLoader' object has no attribute 'get_all_intrinsics'

tests/loader/test_olsson_loader.py:133: AttributeError
=============================== warnings summary ===============================
```

gtsfm/runner/frontend_runner.py
```diff
@@ -39,12 +39,15 @@ def run_frontend(
...
+    camera_intrinsics_graph = loader.create_computation_graph_for_intrinsics()
+    image_shapes_graph = loader.create_computation_graph_for_image_shapes()
+    
+    [image_shapes] = dask.compute(image_shapes_graph)
+    [camera_intrinsics] = dask.compute(camera_intrinsics_graph)
 
     (delayed_keypoints, delayed_putative_corr_idxs_dict,) = correspondence_generator.create_computation_graph(
         delayed_images=loader.create_computation_graph_for_images(),
-        image_shapes=loader.get_image_shapes(),
+        image_shapes=image_shapes,
```
akshay-krishnan, line 50: can the correspondence generator not take delayed shapes?

stepanyanhayk: Ideally, we should always use delayed objects that are connected with reading/storing large files. Fixed it here and created a TODO to change the function signature in a new PR.

PR #606, opened 2023-02-28
Annotate loader

stepanyanhayk:


gtsfm/loader/loader_base.py
```diff
@@ -271,7 +273,10 @@ def create_computation_graph_for_images(self) -> List[Delayed]:
             list of delayed tasks for images.
         """
         N = len(self)
-        return [dask.delayed(self.get_image)(i) for i in range(N)]
+        annotatation = dask.annotate(workers=self.input_worker) if self.input_worker else dask.annotate()
```
akshay-krishnan, line 276: can we make this a class member? self.dask_annotation?

gtsfm/loader/loader_base.py
```diff
@@ -43,15 +43,17 @@ def get_ui_metadata() -> UiMetadata:
...
         """
         Args:
             max_resolution: integer representing maximum length of image's short side
                e.g. for 1080p (1920 x 1080), max_resolution would be 1080
+            input_worker: string representing ip address of the worker
         """
         if not isinstance(max_resolution, int):
             raise ValueError("Maximum image resolution must be an integer argument.")
         self._max_resolution = max_resolution
+        self.input_worker = input_worker
```
johnwlambert, line 56: nit: prefer leading underscore for private var.

gtsfm/loader/loader_base.py
```diff
@@ -43,15 +43,17 @@ def get_ui_metadata() -> UiMetadata:
             parent_plate="Loader and Retriever",
         )
 
-    def __init__(self, max_resolution: int = 1080) -> None:
+    def __init__(self, max_resolution: int = 1080, input_worker: str = None) -> None:
         """
         Args:
             max_resolution: integer representing maximum length of image's short side
                e.g. for 1080p (1920 x 1080), max_resolution would be 1080
+            input_worker: string representing ip address of the worker
```
johnwlambert, line 51: nit: can we add a period at the end of this line.

akshay-krishnan: Also, @stepanyanhayk you mentioned it needs the port address too? if so, could you add that to the docstring?

PR #607, opened 2023-02-28
Prototyping for using delayed objects for deep networks

ayushbaid:


akshay-krishnan:
could you update the code that uses detector_descriptor?

akshay-krishnan:
I think I get it now. In the base class, you load the model weights once, when the graph is still being built. And then pass these model weights as inputs to the delayed detect and describe method. I think that should work. Is the model object serializable?

ayushbaid:
> 

Yes thats the rough idea. I am a bit hesitant about naming the new arg model and not something general purpose.

I'll add a test that the return type of `get_model` function is serializable.

PR #608, opened 2023-03-09
Prototype for Superpoint detector which allows scatter

ayushbaid:
We want to optimize the inference/application of different GTSFM processes. 
Right now, we instantiate heavy objects every time in the main business logic function because we do not want to serialize the class's attributes again and again. 
This paradigm goes moves away from the OOP design. 
This will allow us to `scatter` the GTSFM processes to all nodes at the start, saving repeated serialization and deserialization


dellaert:
Can you explain a bit more?

gtsfm/new_detector.py
```diff
@@ -0,0 +1,101 @@
...
+        scores = model_results["scores"][0].detach().cpu().numpy()
+        keypoints = Keypoints(coordinates, scales=None, responses=scores)
+        # descriptors = model_results["descriptors"][0].detach().cpu().numpy().T
+
+        return keypoints
+
+
+@dask.delayed
+def detect(detector: NewDetector, image: Image) -> Keypoints:
+    return detector.apply(image)
```
ayushbaid, line 79: I plan to create functions like this for every module of GTSFM: detector, matcher, rotation averaging, etc. 

In the current design, we have `create_computation_graph` as a method of a detector class. In the `dask.delayed` call, we call another method of the class. This causes the whole class to be serialized and deserialized for every dask.delayed call. 
Curent code:
```
def create_computation_graph(self, image_graph: Delayed) -> Delayed:
        """Generates the computation graph for performing detection.
        Args:
            image_graph: computation graph for an image.
        Returns:
            Delayed task for detection on the input image.
        """
        return dask.delayed(self.detect)(image_graph)
```

In the current function, we delay the module, and pass it to the detect function. This gives us the potential to just serialize them once and scatter them to all the workers.

akshay-krishnan: Thanks Ayush. 
Do we need the function? Instead, can we make the model member a delayed object? So we could construct the class with a delayed (or future instance as you use it below) and reuse it in the method?

ayushbaid: I think the separation of dask related code into a function will be better. All the GTSFM processes will have no mention of dask at all. They just have to implement their API. I will make a new `dask_utils` which has one function for each process. What do you think?

gtsfm/new_detector.py
```diff
@@ -0,0 +1,101 @@
...
+            model_results = self._model({"image": image_tensor})
+        torch.cuda.empty_cache()
+
+        # Unpack results.
+        coordinates = model_results["keypoints"][0].detach().cpu().numpy()
+        scores = model_results["scores"][0].detach().cpu().numpy()
+        keypoints = Keypoints(coordinates, scales=None, responses=scores)
+        # descriptors = model_results["descriptors"][0].detach().cpu().numpy().T
+
+        return keypoints
```
travisdriver, line 74: Why have you implemented this as a detector only? This isn't saving any computation, as the descriptors are already being computed in the forward pass; they're just being thrown away here.

travisdriver:
I think it might be useful to do some runtime evaluations to verify that this implementation does indeed speed things up before you spend a significant amount of time refactoring all of the GTSfM processes.

gtsfm/new_detector.py
ayushbaid: This was just to show the new design proposal and get feedback.

ayushbaid:
> I think it might be useful to do some runtime evaluations to verify that this implementation does indeed speed things up before you spend a significant amount of time refactoring all of the GTSfM processes.

Yes. I will run experiments with the new and old design and time it.

PR #609, opened 2023-03-10
SSHCluster Runner

stepanyanhayk:
In this PR we are introducing an option to run gtsfm on a cluster using [SSHCluster](https://docs.dask.org/en/stable/deploying-ssh.html#dask.distributed.SSHCluster) class from dask. Command line takes the IP addresses of the workers as arguments. My current implementation assumes that the first worker (indexed at 0 in the list) will be responsible for the data input.

### DIFF
--- a/gtsfm/configs/cluster.yaml
+++ b/gtsfm/configs/cluster.yaml
@@ -1,3 +1,5 @@
+# replace by names of host machines, first worker is used for I/O
+
 workers:
   - host1.cc.gatech.edu
   - host2.cc.gatech.edu
--- a/gtsfm/configs/cluster.yaml
+++ b/gtsfm/configs/cluster.yaml
@@ -1,0 +1,4 @@
+workers:
+  - host1.cc.gatech.edu
+  - host2.cc.gatech.edu
+  - host3.cc.gatech.edu
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -138,7 +138,7 @@ class GtsfmRunnerBase:
             type=str,
             default=None,
             nargs="+",
-            help="list of worker ip addresses for the cluster"
+            help="list of worker ip addresses for the cluster, first worker is used as scheduler and should contain the dataset",
         )
         parser.add_argument(
             "--dashboard_address",
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -284,7 +284,7 @@ class GtsfmRunnerBase:
         with performance_report(filename="scene-optimizer-dask-report.html"):
             sfm_result = client.compute(delayed_sfm_result).result()
             future_io = client.compute(delayed_io)
-            io = [io_task.result() for io_task in future_io]
+            (*io, ) = [io_task.result() for io_task in future_io]
 
         assert isinstance(sfm_result, GtsfmData)
 
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -262,7 +262,9 @@ class GtsfmRunnerBase:
         )
 
         with performance_report(filename="correspondence-generator-dask-report.html"):
-            keypoints_list, putative_corr_idxs_dict = dask.compute(delayed_keypoints, delayed_putative_corr_idxs_dict)
+            future_keypoints = client.compute(delayed_keypoints)
+            keypoints_list = [keypoint.result() for keypoint in future_keypoints]
+            putative_corr_idxs_dict = client.compute(delayed_putative_corr_idxs_dict).result()
 
         delayed_sfm_result, delayed_io = self.scene_optimizer.create_computation_graph(
             keypoints_list=keypoints_list,
@@ -280,7 +282,9 @@ class GtsfmRunnerBase:
         )
 
         with performance_report(filename="scene-optimizer-dask-report.html"):
-            sfm_result, *io = dask.compute(delayed_sfm_result, *delayed_io)
+            sfm_result = client.compute(delayed_sfm_result).result()
+            io_future = client.compute(delayed_io)
+            io = [io_task.result() for io_task in io_future]
 
         assert isinstance(sfm_result, GtsfmData)
 
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -283,8 +283,8 @@ class GtsfmRunnerBase:
 
         with performance_report(filename="scene-optimizer-dask-report.html"):
             sfm_result = client.compute(delayed_sfm_result).result()
-            io_future = client.compute(delayed_io)
-            io = [io_task.result() for io_task in io_future]
+            future_io = client.compute(delayed_io)
+            io = [io_task.result() for io_task in future_io]
 
         assert isinstance(sfm_result, GtsfmData)
 
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -228,7 +228,10 @@ class GtsfmRunnerBase:
             cluster = SSHCluster(
                 [scheduler] + workers,
                 scheduler_options={"dashboard_address": self.parsed_args.dashboard_address},
-                worker_options={"n_workers": self.parsed_args.num_workers},
+                worker_options={
+                    "n_workers": self.parsed_args.num_workers,
+                    "nthreads": self.parsed_args.threads_per_worker
+                },
             )
             client = Client(cluster)
             self.loader._input_worker = list(client.scheduler_info()["workers"].keys())[0]
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -7,7 +7,7 @@ from pathlib import Path
 
 import dask
 import hydra
-from dask.distributed import Client, LocalCluster, performance_report
+from dask.distributed import Client, LocalCluster, SSHCluster, performance_report
 from hydra.utils import instantiate
 from omegaconf import OmegaConf
 
@@ -127,6 +127,19 @@ class GtsfmRunnerBase:
             default=None,
             help="tmp directory for dask workers, uses dask's default (/tmp) if not set",
         )
+        parser.add_argument(
+            "--run_cluster",
+            action="store_true",
+            default=False,
+            help="condition whether to run gtsfm on a cluster"
+        )
+        parser.add_argument(
+            "--cluster_workers",
+            default=None,
+            type=str,
+            nargs="+",
+            help="list of worker ip addresses for the cluster"
+        )
         return parser
 
     @abstractmethod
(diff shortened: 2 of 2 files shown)


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -202,10 +221,22 @@ def run(self) -> None:
...
-            n_workers=self.parsed_args.num_workers, threads_per_worker=self.parsed_args.threads_per_worker
-        )
+        # create dask cluster
+        if self.parsed_args.run_cluster:
+            workers = self.parsed_args.cluster_workers
+            scheduler = workers[0]
+            cluster = SSHCluster(
+                [scheduler] + workers,
+                scheduler_options={"dashboard_address": self.parsed_args.dashboard_address},
+                worker_options={"n_workers": self.parsed_args.num_workers},
```
akshay-krishnan, line 231: could you add threads per worker?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,25 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
+            action="store_true",
+            default=False,
+            help="condition whether to run gtsfm on a cluster"
+        )
+        parser.add_argument(
+            "--cluster_workers",
+            type=str,
+            default=None,
+            nargs="+",
+            help="list of worker ip addresses for the cluster"
```
akshay-krishnan, line 141: maybe add that first worker will be used as scheduler, and should contain the dataset?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,25 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
+        )
+        parser.add_argument(
+            "--dashboard_address",
```
akshay-krishnan, line 144: do you mean --dashboard_port?

akshay-krishnan (commented):
Can you test this on all datasets we run in the CI?

stepanyanhayk:
> Can you test this on all datasets we run in the CI? I realize this is a bit of work, but we want to be sure the cluster works for all dataset sizes/formats.

Sure, I'll test this on all the datasets we have!

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: wanted to keep consistency with dask's dashboard_address attribute, but I agree that dashboard_port is more user-friendly, addresses in the recent commit

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: addressed in the recent commit

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: addressed in the recent commit

akshay-krishnan:
Do we have a good reason for using dask.compute instead of client.compute now?

stepanyanhayk:
> Do we have a good reason for using dask.compute instead of client.compute now?

The difference between these functions is that dask.compute() performs computations synchronously while client.compute() performs asynchronously. I don't think we have a reason to stay with dask.compute. In fact, client.compute() could be better since it can submit multiple computations simultaneously (as they all are executed asynchronously).

akshay-krishnan:
> > Do we have a good reason for using dask.compute instead of client.compute now?
> 
> @akshay-krishnan the difference between these functions is that dask.compute() performs computations synchronously while client.compute() performs asynchronously. I don't think we have a reason to stay with dask.compute. In fact, client.compute() could be better since it can submit multiple computations simultaneously (as they all are executed asynchronously).

Right, so why are we adding more dask.compute() calls?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,26 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             default=None,
             help="tmp directory for dask workers, uses dask's default (/tmp) if not set",
         )
+        parser.add_argument(
+            "--run_cluster",
+            action="store_true",
+            default=False,
+            help="condition whether to run gtsfm on a cluster"
```
akshay-krishnan, line 134: nit: whether to run gtsfm on a cluster

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,26 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
+        )
+        parser.add_argument(
+            "--cluster_workers",
+            type=str,
+            default=None,
+            nargs="+",
+            help="list of worker ip addresses for the cluster,"
```
akshay-krishnan, line 141: nit: IP

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -202,10 +222,25 @@ def run(self) -> None:
...
+            cluster = SSHCluster(
+                [scheduler] + workers,
+                scheduler_options={"dashboard_address": self.parsed_args.dashboard_port},
+                worker_options={
+                    "n_workers": self.parsed_args.num_workers,
+                    "nthreads": self.parsed_args.threads_per_worker
+                },
+            )
+            client = Client(cluster)
+            self.loader._input_worker = list(client.scheduler_info()["workers"].keys())[0]
```
akshay-krishnan, line 238: add a comment on why we dont use `scheduler ` from above directly

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -244,8 +281,10 @@ def run(self) -> None:
             matching_regime=ImageMatchingRegime(self.parsed_args.matching_regime),
         )
 
-        with Client(cluster), performance_report(filename="scene-optimizer-dask-report.html"):
-            sfm_result, *io = dask.compute(delayed_sfm_result, *delayed_io)
+        with performance_report(filename="scene-optimizer-dask-report.html"):
+            sfm_result = client.compute(delayed_sfm_result).result()
+            future_io = client.compute(delayed_io)
+            (*io, ) = [io_task.result() for io_task in future_io]
```
akshay-krishnan, line 287: do io_tasks have any result?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,25 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
             help="tmp directory for dask workers, uses dask's default (/tmp) if not set",
         )
+        parser.add_argument(
+            "--run_cluster",
+            action="store_true",
+            default=False,
+            help="whether to run gtsfm on a cluster"
+        )
+        parser.add_argument(
+            "--cluster_config_name",
```
akshay-krishnan, line 137: just `--cluster_config` should do, technically its `--cluster_config_path`, but lets keep it short ?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,25 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
+            type=str,
+            default="cluster.yaml",
+            help="config listing IP worker addresses for the cluster,"
+            " first worker is used as scheduler and should contain the dataset",
+        )
+        parser.add_argument(
```
akshay-krishnan, line 143: surprised we did not have this already

gtsfm/configs/cluster.yaml
ayushbaid, line 1: I dont think we should add this file to the repo

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,25 @@ def construct_argparser(self) -> argparse.ArgumentParser:
...
+            default=False,
+            help="whether to run gtsfm on a cluster"
+        )
+        parser.add_argument(
+            "--cluster_config_name",
+            type=str,
+            default="cluster.yaml",
+            help="config listing IP worker addresses for the cluster,"
+            " first worker is used as scheduler and should contain the dataset",
+        )
```
ayushbaid, line 142: Can we run on a single machine using a yaml file too?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -127,6 +127,25 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             default=None,
             help="tmp directory for dask workers, uses dask's default (/tmp) if not set",
         )
+        parser.add_argument(
+            "--run_cluster",
+            action="store_true",
+            default=False,
+            help="whether to run gtsfm on a cluster"
+        )
```
ayushbaid, line 135: I think we can skip this arg and just make `--cluster_config_name` optional?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -202,10 +221,30 @@ def run(self) -> None:
...
+            client = Client(cluster)
+            # getting first worker's IP address and port to do IO
+            io_worker = list(client.scheduler_info()["workers"].keys())[0]
+            self.loader._input_worker = io_worker
+            self.scene_optimizer._output_worker = io_worker
+        else:
+            cluster = LocalCluster(
+                n_workers=self.parsed_args.num_workers, threads_per_worker=self.parsed_args.threads_per_worker
+            )
+            client = Client(cluster)
```
akshay-krishnan, line 247: do we not need to use the client and cluster variables anywhere?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -202,10 +221,30 @@ def run(self) -> None:
...
+                worker_options={
+                    "n_workers": self.parsed_args.num_workers,
+                    "nthreads": self.parsed_args.threads_per_worker,
+                },
+            )
+            client = Client(cluster)
+            # getting first worker's IP address and port to do IO
+            io_worker = list(client.scheduler_info()["workers"].keys())[0]
+            self.loader._input_worker = io_worker
+            self.scene_optimizer._output_worker = io_worker
```
ayushbaid, line 242: I will propose a new design where we just scatter the computation for detector/descriptor/matcher/verifier. You would not need to worry about pinning IO to workers in that case as the most of the stuff will happen outside dask.

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: we use the cluster variable only to create the client, but we do not need to use the client variable

gtsfm/configs/cluster.yaml
stepanyanhayk: I thought it would be helpful for users to have an example config so they can create one for their cluster.

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: Yes, I will be creating a config file to keep all the parameters (including the option to run on a single machine) in a separate PR.

gtsfm/runner/gtsfm_runner_base.py
stepanyanhayk: Scattering the computations would be great! I think it will also improve the runtime. I think we might still need IO workers as the data could be on one machine only and we need a way to restrict data reading/output tasks to the machine that has access to the data.

gtsfm/configs/cluster.yaml
akshay-krishnan: The way to specify workers is through the config. its unfair to expect users to know the format of the config. I know this file looks very trivial now, but the parameters will likely grow, and if not, we can remove it :)

gtsfm/runner/gtsfm_runner_base.py
akshay-krishnan: weird, alright

akshay-krishnan (approved):
LGTM. @ayushbaid has some comments that need addressing. regarding the redesign, my 2 cents is that we can merge this now, and change it in the redesign PR.

PR #613, opened 2023-03-17
save with absolute path

stepanyanhayk:
For successful runs on the cluster, we need our paths to be absolute, so the machine doing the output can understand the location of the file. This PR changes the relative path to an absolute path.

### DIFF
--- a/gtsfm/multi_view_optimizer.py
+++ b/gtsfm/multi_view_optimizer.py
@@ -7,6 +7,8 @@ from typing import Dict, List, Optional, Tuple
 import dask
 import numpy as np
 from dask.delayed import Delayed
+import os
+from pathlib import Path
 from gtsam import Pose3
 
 import gtsfm.common.types as gtsfm_types
@@ -23,6 +25,7 @@ from gtsfm.two_view_estimator import TwoViewEstimationReport
 from gtsfm.view_graph_estimator.view_graph_estimator_base import ViewGraphEstimatorBase
 from gtsfm.data_association.dsf_tracks_estimator import DsfTracksEstimator
 
+DEFAULT_OUTPUT_ROOT = Path(__file__).resolve().parent.parent
 
 class MultiViewOptimizer:
     def __init__(
@@ -39,6 +42,16 @@ class MultiViewOptimizer:
         self.data_association_module = data_association_module
         self.ba_optimizer = bundle_adjustment_module
         self._run_view_graph_estimator: bool = self.view_graph_estimator is not None
+        self._create_debugging_directories()
+
+    def _create_debugging_directories(self) -> None:
+        """Create output directories for GTSFM result debugging."""
+
+        self._debug_output_dir = DEFAULT_OUTPUT_ROOT / "debugging"
+        self._plot_cycle_consist_path = self._debug_output_dir / "cycle_consistency" / "plots"
+        
+        os.makedirs(self._debug_output_dir, exist_ok=True)
+        os.makedirs(self._plot_cycle_consist_path, exist_ok=True)
 
     def create_computation_graph(
         self,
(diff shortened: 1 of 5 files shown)


gtsfm/view_graph_estimator/view_graph_estimator_base.py
```diff
@@ -275,6 +278,12 @@ def create_computation_graph(
             - Dict of two_view_reports in the view graph
             - GtsfmMetricsGroup with the view graph estimation metrics
         """
+
+        # create debug directory for cycle_consistency
+        if debug_output_dir:
+            self._plot_cycle_consist_path = debug_output_dir / "cycle_consistency"
```
akshay-krishnan, line 284: why do we need this member variable?

akshay-krishnan (approved):

akshay-krishnan:
cc @johnwlambert

gtsfm/multi_view_optimizer.py
```diff
@@ -39,6 +43,13 @@ def __init__(
         self.data_association_module = data_association_module
         self.ba_optimizer = bundle_adjustment_module
         self._run_view_graph_estimator: bool = self.view_graph_estimator is not None
+        self._create_debug_directories()
+
+    def _create_debug_directories(self) -> None:
```
johnwlambert, line 48: Thanks for the PR, Hayk. Question — what’s the definition of a debug output, and how does it differ from other outputs of the system?

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -186,14 +198,15 @@ def __save_plots(
         plt.ylabel("Rotation error w.r.t GT")
         plt.axis("equal")
         plt.legend(loc="lower right")
-        plt.savefig(os.path.join("plots", f"gt_err_vs_{self._edge_error_aggregation_criterion}_agg_error.jpg"), dpi=400)
+        plt.savefig(
+            os.path.join(output_dir, f"gt_err_vs_{self._edge_error_aggregation_criterion}_agg_error.jpg"), dpi=400)
```
johnwlambert, line 202: Is this backwards compatible, ie is the same output location used now by default? Isn’t the output not going into a “plots” dir anymore?

gtsfm/multi_view_optimizer.py
```diff
@@ -23,6 +25,8 @@
 from gtsfm.view_graph_estimator.view_graph_estimator_base import ViewGraphEstimatorBase
 from gtsfm.data_association.dsf_tracks_estimator import DsfTracksEstimator
 
+DEFAULT_OUTPUT_ROOT = Path(__file__).resolve().parent.parent
```
johnwlambert, line 28: Could we instead pass the SceneOptimizer’s output_root property? We already have a default in the runner: https://github.com/borglab/gtsfm/blob/master/gtsfm/runner/gtsfm_runner_base.py#L142

johnwlambert:
> cc @johnwlambert

Left a few comments @akshay-krishnan @stepanyanhayk , thanks for working on this!

gtsfm/multi_view_optimizer.py
stepanyanhayk: Hi John, thanks for reviewing. Debug output is the output that we only use for debugging purposes. Other outputs e.g. `results` directory are later used for purposes like 3D visualization.

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
stepanyanhayk: The new output location is the "debug/cycle_consistency" directory instead of "plots". Do you think I should make it "debug/cycle_consistency/plots"?

gtsfm/multi_view_optimizer.py
stepanyanhayk: Addressed in the recent commit.

gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
akshay-krishnan: I suggested moving all intermediate results saved by gtsfm processes to the debug/ folder. I feel like these plots are less useful to end users but more useful to us for debugging.

akshay-krishnan (approved):

PR #614, opened 2023-03-21
annotate output

stepanyanhayk:
This PR allows us to specify a worker for the output tasks.


akshay-krishnan (approved):
Could you check CI produces correct results on a couple datasets before merging?

stepanyanhayk:
> Could you check CI produces correct results on a couple datasets before merging?

Done!

Issue #110, opened 2021-03-10
Optimize inference with deep models


ayushbaid:
The current style of storing the models as private variables of a class in `__init__`, and reuse them in functions. This leads to trouble as the scheduler gets locked up frequently and also fails for moderately sized datasets.

The current hack is to read the model from disk every single time, but its inefficient and makes the program slow.

Dask's recommended strategy is to wrap the model in delayed and pass it to the function each time. Will need to try that and properly benchmark compute.

johnwlambert:
https://examples.dask.org/machine-learning/torch-prediction.html

May need to add batches.

PR #617, opened 2023-04-02
Refactor to use Dask future APIs in frontend only

ayushbaid:
Summary of the  refactored design:
1. Each module will implement the `apply` function, which should compute the output products as well as the metrics. 
2. There will be no use of dask inside any module
3. The modules for which `apply` can be run in parallel (frontend) will be using Dask's futures. This logic will reside in `gtsfm_runner_base`


ayushbaid:
Some timing data from CI

| Dataset | This branch (mins) | Master (mins)|
| -------- | ------------------------ | ---------------------- |
| door-12 SIFT front end | 24 |  23 |
| door-12 deep front end | 61 | 58 |
| notre dame-20 deep front end | 91 | 51 | 

This does not look promising. Maybe the CI is not a good place to get benchmarking data?

ayushbaid:
Local runs indicate a similar performance degradation. So I guess the serialization and deserialization is more expensive than loading network weights from disk again. But to run GTSFM on a cluster, this might be a performance penalty we have to take. 

@johnwlambert @akshay-krishnan

ayushbaid:
This PR is one way of resolving #110

akshay-krishnan:
We could discuss this in more detail in a meeting. In the meantime, would it be worth benchmarking on the cluster? @stepanyanhayk could you help with that?

stepanyanhayk:
> 

Of course! I will run this on the cluster and share the benchmarks.

ayushbaid:
Thanks!

I expect performance to be even worse because of the network latency.

akshay-krishnan:
@ayushbaid  I wish we could profile specific components. For example, how long does detector-descriptor take on a single image on average when using the delayed vs the futures API? that would give us some more insight into what exactly causes the latency. this is definitely nontrivial on distributed workers, but have you tried on a single machine?

ayushbaid:
I benchmarked the detector in #608 and it showed promising results.

gtsfm/common/image.py
```diff
@@ -34,6 +34,10 @@ def width(self) -> int:
         """
         return self.value_array.shape[1]
 
+    @property
+    def shape(self) -> Tuple[int, int]:
```
akshay-krishnan, line 38: I added a shape property already, please pull from master

PR #618, opened 2023-04-02
DSF Track Estimation C++

stepanyanhayk:
This PR integrates gtsam's DSF track estimation into gtsfm. The main purpose of this change is to bring runtime speedup by using C++ instead of Python.


gtsfm/data_association/dsf_tracks_estimator.py
```diff
@@ -46,44 +46,28 @@ def run(self, matches_dict: Dict[Tuple[int, int], np.ndarray], keypoints_list: L
...
+            matches_map[gtsam.IndexPair(i1, i2)] = k_pairs
+        
+        # converting gtsfm Keypoints into gtsam Keypoints
+        keypoints_vector = gtsam.KeypointsVector()
+        for keypoint in keypoints_list:
+            keypoints_vector.append(
+                gtsam.gtsfm.Keypoints(keypoint.coordinates)
+            )
+        
+        tracks = gtsam.gtsfm.tracksFromPairwiseMatches(
```
johnwlambert, line 61: Thanks @stepanyanhayk! This looks good. Could you add a simple test for this DSF tracks estimator, using the example keypoints + matches from this test? https://github.com/borglab/gtsam/blob/develop/python/gtsam/tests/test_DsfTrackGenerator.py#L22-L35

I believe you can use it verbatim.

stepanyanhayk: Thanks for reviewing, John. Added a unittest for DSF track estimation in a new commit.

akshay-krishnan (commented):
A few questions: 
- What is the speedup youre getting? 
- Can we use the gtsam data structures in other places in gtsfm, so that we can reduce the spike in memory usage by this module, and the time to copy from one to another? I think this should be possible for at least some of them.

akshay-krishnan:
Could you verify the results in the CI? Make sure everything works without any accuracy loss?

johnwlambert:
Thanks for adding the unit test, @stepanyanhayk!

I believe CI is failing:
```
tests/data_association/test_dsf_tracks_estimator.py:113:1: E302 expected 2 blank lines, found 1
```
Do you mind taking a look, and reformatting?

gtsfm/data_association/dsf_tracks_estimator.py
johnwlambert: Marking as resolved, thanks!

johnwlambert:
Test and code looks great! Re @akshay-krishnan 's questions -- Hayk, could you attach a dashboard to this PR, showing the performance difference vs. master? Re benchmarking, you would see the greatest speedup on loftr, because it generates orders of magnitudes more tracks.

stepanyanhayk:
> 

Thanks a lot for the pointer, John. I will create the report today

akshay-krishnan:
@stepanyanhayk friendy ping on this

stepanyanhayk:
Sorry for the delay, @akshay-krishnan. These are the results I get from comparing the artifacts

image

image

image

akshay-krishnan:
would mainly look at data association metrics. 
# tracks computed with GT cameras EXCEEDS_REPOJ_ERROR, computed cameras: EXCEEDS_REPROJ_ERROR is regressing significantly, which means that we have more bad 2D tracks. 
Shouldnt this PR be a no-op in terms of quality? @johnwlambert ? Are there any logical changes?

akshay-krishnan (commented):
Nitpicky comment on comments: 
For example: "converting gtsfm Keypoints into gtsam Keypoints"
1. Punctuation -> "Converting gtsfm Keypoints into gtsam Keypoints."
2. Simple tense (like an instruction) -> "Convert gtsfm Keypoints into gtsam Keypoints."

Please refer to other comments in the file. I am personally flexible about these, but since its an open source project, I see some PRs fixing unrelated comments from other PRs. Better to write good comments to begin with :)

johnwlambert:
Interesting, thanks for the dashboard @stepanyanhayk! Yes, this change should be a no-op. To understand better if we're seeing stochasticity from gtsfm itself, vs. from the change, we could make the dashboard one more time, with a second run from the CI. I'll retrigger a new run also. Would you mind checking the visual quality on any of the datasets?

Which master branch PR were you comparing with? (could you share a link?)

stepanyanhayk:
> Interesting, thanks for the dashboard @stepanyanhayk! Yes, this change should be a no-op. To understand better if we're seeing stochasticity from gtsfm itself, vs. from the change, we could make the dashboard one more time, with a second run from the CI. I'll retrigger a new run also. Would you mind checking the visual quality on any of the datasets?
> 
> Which master branch PR were you comparing with? (could you share a link?)

I was comparing with [this](https://github.com/borglab/gtsfm/actions/runs/4587333955) master PR.

stepanyanhayk:
Hi @johnwlambert, @akshay-krishnan 

I have created a new dashboard with the recent run but it did not change the results, we still see the regression on the Data Association Metrics. 

image

In terms of visual quality, I do not see much difference between the master and cpp changes.

Master:
Screenshot 2023-04-30 at 5 08 36 PM
Screenshot 2023-04-30 at 5 08 29 PM

Branch:
Screenshot 2023-04-30 at 5 04 02 PM
Screenshot 2023-04-30 at 5 03 11 PM

To debug further I have created a small notebook comparing the results from DsfTracksEstimator.run() function calls on both master and cpp versions: only 72% of the resulting tracks are the same.

Screenshot 2023-04-30 at 5 35 03 PM

So, the CPP implementation and Python implementation do not output the same 2D tracks. What are your thoughts on this? Do you think the issue might be from the CPP side?

ayushbaid (approved):

akshay-krishnan:
@johnwlambert please update the dashboard once we can use your C++ fixes

johnwlambert:
BA Metrics:
Screen Shot 2023-09-04 at 9 59 14 PM

Screen Shot 2023-09-04 at 9 59 49 PM

PR #619, opened 2023-04-11
Parallelize triangulation

travisdriver:
Running on my machine with 8 workers for door-12, runtime goes from 1.24 minutes without parallelism to 1.03 with parallelism. I would expect larger datasets to have ven more significant runtime reductions.

### DIFF
--- a/gtsfm/data_association/data_assoc.py
+++ b/gtsfm/data_association/data_assoc.py
@@ -7,7 +7,7 @@ References:
 1. Richard I. Hartley and Peter Sturm. Triangulation. Computer Vision and Image Understanding, Vol. 68, No. 2,
    November, pp. 146–157, 1997
 
-Authors: Sushmita Warrier, Xiaolong Wu, John Lambert
+Authors: Sushmita Warrier, Xiaolong Wu, John Lambert, Travis Driver
 """
 import os
 from collections import Counter
@@ -73,6 +73,9 @@ class DataAssociation(GTSFMProcess):
         num_images: int,
         cameras: Dict[int, gtsfm_types.CAMERA_TYPE],
         tracks_2d: List[SfmTrack2d],
+        sfm_tracks: List[Optional[SfmTrack]],
+        avg_track_reproj_errors: List[Optional[float]],
+        triangulation_exit_codes: List[TriangulationExitCode],
         cameras_gt: List[Optional[gtsfm_types.CALIBRATION_TYPE]],
         relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
         images: Optional[List[Image]] = None,
@@ -99,9 +102,6 @@ class DataAssociation(GTSFMProcess):
         logger.debug("[Data association] input number of tracks: %s", len(tracks_2d))
         logger.debug("[Data association] input avg. track length: %s", np.mean(track_lengths_2d))
 
-        # Initialize 3D landmark for each track
-        point3d_initializer = Point3dInitializer(cameras, self.triangulation_options)
-
         # form GtsfmData object after triangulation
         triangulated_data = GtsfmData(num_images)
 
@@ -115,9 +115,12 @@ class DataAssociation(GTSFMProcess):
         exit_codes_wrt_computed: List[TriangulationExitCode] = []
         per_accepted_track_avg_errors = []
         per_rejected_track_avg_errors = []
-        for track_2d in tracks_2d:
+        assert len(tracks_2d) == len(sfm_tracks)
+        for j in range(len(tracks_2d)):
             # triangulate and filter based on reprojection error
-            sfm_track, avg_track_reproj_error, triangulation_exit_code = point3d_initializer.triangulate(track_2d)
+            sfm_track = sfm_tracks[j]
+            avg_track_reproj_error = avg_track_reproj_errors[j]
+            triangulation_exit_code = triangulation_exit_codes[j]
             exit_codes_wrt_computed.append(triangulation_exit_code)
             if triangulation_exit_code == TriangulationExitCode.CHEIRALITY_FAILURE:
                 continue
@@ -196,6 +199,27 @@ class DataAssociation(GTSFMProcess):
 
         return connected_data, data_assoc_metrics
 
+    def run_triangulation(
+        self,
+        cameras: Dict[int, gtsfm_types.CAMERA_TYPE],
+        tracks_2d: List[SfmTrack2d],
+    ) -> Tuple[List[Delayed], List[Delayed], List[Delayed]]:
+        # Initialize 3D landmark for each track
+        point3d_initializer = Point3dInitializer(cameras, self.triangulation_options)
+
+        # Loop through tracks and triangulate.
+        delayed_sfm_tracks, delayed_avg_track_reproj_errors, delayed_triangulation_exit_codes = [], [], []
+        for track_2d in tracks_2d:
+            # triangulate and filter based on reprojection error
+            (delayed_sfm_track, delayed_avg_track_reproj_error, delayed_triangulation_exit_code) = dask.delayed(
+                point3d_initializer.triangulate, nout=3
+            )(track_2d)
+            delayed_sfm_tracks.append(delayed_sfm_track)
+            delayed_avg_track_reproj_errors.append(delayed_avg_track_reproj_error)
+            delayed_triangulation_exit_codes.append(delayed_triangulation_exit_code)
+
+        return delayed_sfm_tracks, delayed_avg_track_reproj_errors, delayed_triangulation_exit_codes
+
     def create_computation_graph(
         self,
         num_images: int,
@@ -220,10 +244,20 @@ class DataAssociation(GTSFMProcess):
             data_assoc_metrics_graph: dictionary with different statistics about the data
                 association result
         """
+
+        # Triangulate 2D tracks.
+        sfm_tracks, avg_track_reproj_errors, triangulation_exit_codes = self.run_triangulation(
+            cameras=cameras.compute(), tracks_2d=tracks_2d.compute()
+        )
+
+        # Unpack results, create BA input, and compute metrics.
         ba_input_graph, data_assoc_metrics_graph = dask.delayed(self.run, nout=2)(
             num_images=num_images,
             cameras=cameras,
             tracks_2d=tracks_2d,
+            sfm_tracks=sfm_tracks,
+            avg_track_reproj_errors=avg_track_reproj_errors,
+            triangulation_exit_codes=triangulation_exit_codes,
             cameras_gt=cameras_gt,
             relative_pose_priors=relative_pose_priors,
             images=images_graph,
--- a/gtsfm/data_association/data_assoc.py
+++ b/gtsfm/data_association/data_assoc.py
@@ -68,7 +68,7 @@ class DataAssociation(GTSFMProcess):
         """Validate the track by checking its length."""
         return sfm_track is not None and sfm_track.numberMeasurements() >= self.min_track_len
 
-    def run(
+    def run_da(
         self,
         num_images: int,
         cameras: Dict[int, gtsfm_types.CAMERA_TYPE],
@@ -80,12 +80,15 @@ class DataAssociation(GTSFMProcess):
         relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
         images: Optional[List[Image]] = None,
     ) -> Tuple[GtsfmData, GtsfmMetricsGroup]:
-        """Perform the data association.
+        """Perform the data association and compute metrics.
 
         Args:
             num_images: Number of images in the scene.
             cameras: dictionary, with image index -> camera mapping.
             tracks_2d: list of 2D tracks.
+            sfm_tracks: List of triangulated tracks.
+            avg_track_repoj_errors: List of average reprojection errors per track.
+            triangulation_exit_codes: exit codes for each triangulation call.
             cameras_gt: list of GT cameras, to be used for benchmarking the tracks.
             images: a list of all images in scene (optional and only for track patch visualization)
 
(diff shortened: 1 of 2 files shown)


gtsfm/data_association/data_assoc.py
```diff
@@ -196,6 +202,44 @@ def run(
...
+            avg_track_repoj_errors: List of average reprojection errors per track.
+            triangulation_exit_codes: exit codes for each triangulation call.
+        """
+        # Initialize 3D landmark for each track
+        point3d_initializer = Point3dInitializer(cameras, self.triangulation_options)
+
+        # Loop through tracks and and generate delayed triangulation tasks.
+        triangulation_results = []
+        for track_2d in tracks_2d:
+            triangulation_results.append(dask.delayed(point3d_initializer.triangulate)(track_2d))
```
akshay-krishnan, line 229: can we do an nout=3 in dask.delayed, and fill in sfm_tracks, avg_track_reproj_errors, triangulation_exit_codes here instead of one more loop?

gtsfm/data_association/data_assoc.py
```diff
@@ -196,6 +202,44 @@ def run(
...
+
+        # Perform triangulation in parallel.
+        triangulation_results = dask.compute(*triangulation_results)
```
akshay-krishnan, line 232: do we need the compute? can we not return delayed objects?

akshay-krishnan (commented):
I think for a better speedup and to run on bigger datasets we will need to batch this. Instead of 1 delayed operation per track we would need 1 delayed operation for a batch of tracks (say 10-50). each delayed operation adds some latency/overhead on dask, and we want to make sure the added overhead is << time for each batch triangulation.

gtsfm/data_association/data_assoc.py
travisdriver: Surprisingly, separating the variables before calling `compute` is slower. The following code took 1.17 minutes:

```python
def run_triangulation(
        self,
        cameras: Dict[int, gtsfm_types.CAMERA_TYPE],
        tracks_2d: List[SfmTrack2d],
    ) -> Tuple[List[Delayed], List[Delayed], List[Delayed]]:
        """Performs triangulation of 2D tracks in parallel.

        Ref: https://docs.dask.org/en/stable/delayed-best-practices.html#compute-on-lots-of-computation-at-once

        Args:
            cameras: list of cameras wrapped up as Delayed.
            tracks_2d: list of tracks wrapped up as Delayed.

        Returns:
            sfm_tracks: List of triangulated tracks.
            avg_track_repoj_errors: List of average reprojection errors per track.
            triangulation_exit_codes: exit codes for each triangulation call.
        """
        # Initialize 3D landmark for each track
        point3d_initializer = Point3dInitializer(cameras, self.triangulation_options)

        # Loop through tracks and and generate delayed triangulation tasks.
        delayed_sfm_tracks, delayed_avg_track_reproj_errors, delayed_triangulation_exit_codes = [], [], []
        for track_2d in tracks_2d:
            delayed_sfm_track, delayed_avg_track_reproj_error, delayed_triangulation_exit_code = dask.delayed(
                point3d_initializer.triangulate, nout=3
            )(track_2d)
            delayed_sfm_tracks.append(delayed_sfm_track)
            delayed_avg_track_reproj_errors.append(delayed_avg_track_reproj_error)
            delayed_triangulation_exit_codes.append(delayed_triangulation_exit_code)

        # Perform triangulation in parallel.
        sfm_tracks, avg_track_reproj_errors, triangulation_exit_codes = dask.compute(
            delayed_sfm_tracks, delayed_avg_track_reproj_errors, delayed_triangulation_exit_codes
        )

        return sfm_tracks, avg_track_reproj_errors, triangulation_exit_codes
```

Maybe the scheduler isn't intelligent enough to realize the three inputs are from the same function call?

gtsfm/data_association/data_assoc.py
travisdriver: Returning delayed objects from delayed calls causes problems with Dask's collection API.

travisdriver:
> I think for a better speedup and to run on bigger datasets we will need to batch this. Instead of 1 delayed operation per track we would need 1 delayed operation for a batch of tracks (say 10-50). each delayed operation adds some latency/overhead on dask, and we want to make sure the added overhead is << time for each batch triangulation.

Great suggestion. I added the `MAX_DELAYED_TRIANGULATION_CALLS` variable to decide the batch size, i.e., `batch_size = int(np.ceil(len(tracks_2d) / MAX_DELAYED_TRIANGULATION_CALLS))`. Setting max calls to 1e4 runs in 1.02 minutes, and 1e3 and 1e2 max calls runs in 0.99 minutes. I kept the default value at 1e3.

gtsfm/data_association/data_assoc.py
```diff
@@ -196,6 +204,68 @@ def run(
...
+            cameras: list of cameras wrapped up as Delayed.
+            tracks_2d: list of tracks wrapped up as Delayed.
+
+        Returns:
+            sfm_tracks: List of triangulated tracks.
+            avg_track_repoj_errors: List of average reprojection errors per track.
+            triangulation_exit_codes: exit codes for each triangulation call.
+        """
+
+        def triangulate_batch(point3d_initializer: Point3dInitializer, tracks_2d: List[SfmTrack2d]):
```
johnwlambert, line 228: Thanks for adding this Travis. nit: could we add a return type hint here?

gtsfm/data_association/data_assoc.py
```diff
@@ -196,6 +204,68 @@ def run(
...
+        tracks_2d: List[SfmTrack2d],
+    ) -> Tuple[List[Delayed], List[Delayed], List[Delayed]]:
+        """Performs triangulation of 2D tracks in parallel.
+
+        Refs:
+        - https://docs.dask.org/en/stable/delayed-best-practices.html#compute-on-lots-of-computation-at-once
+        - https://docs.dask.org/en/stable/delayed-best-practices.html#avoid-too-many-tasks
+
+        Args:
+            cameras: list of cameras wrapped up as Delayed.
```
johnwlambert, line 219: nit: can we capitalize "List of cameras", since Google style dictates this

gtsfm/data_association/data_assoc.py
```diff
@@ -33,6 +33,8 @@
 
 logger = logger_utils.get_logger()
 
+MAX_DELAYED_TRIANGULATION_CALLS = 1e7
```
johnwlambert, line 36: nit: could we add a brief inline comment explaining this variable perhaps:
```python
# Heuristically set to limit the number of nodes in the Dask graph.
MAX_DELAYED_TRIANGULATION_CALLS = 1e7
```

gtsfm/data_association/data_assoc.py
```diff
@@ -196,6 +204,68 @@ def run(
 
         return connected_data, data_assoc_metrics
 
+    def run_triangulation(
```
johnwlambert, line 207: What's the difference between `run_triangulation()` and `run_da()`? is one parallelized, and the other is not?

Technically, neither does any data association, since we broke DSF into another module, so can we use a different name than `run_da()`? (I know the module is mis-named currently)

travisdriver: Technically it's still associating landmarks to keypoints. The class is also called DataAssociation, so we'd need to change the whole class name as well. Maybe we can change the naming conventions in another PR?

travisdriver:
@stepanyanhayk Would you mind running this on the cluster and letting us know whether or not it fixes the DA runtime issues?

gtsfm/data_association/data_assoc.py
travisdriver: Done

gtsfm/data_association/data_assoc.py
travisdriver: Done

gtsfm/data_association/data_assoc.py
travisdriver: Done

johnwlambert: Marking as resolved, thanks!

gtsfm/data_association/data_assoc.py
johnwlambert: Marking as resolved, thanks!

gtsfm/data_association/data_assoc.py
johnwlambert: I see, yeah I suppose it depends a bit on our definition of data association. Agreed it's orthogonal to this PR, but I still am a bit lost by the function naming, for the new function we're adding. Do we mean `run_da` vs. `run_da_batched`? or `run_triangulation` vs. `run_triangulation_batched`?

johnwlambert: Here's one definition of data association from Szeliski, that would suggest we rename this module:

In SFM, “data association” is the problem of determining correspondences, either between feature points or whole images. In the case of whole images, it can be seen as the validity of hypotheses that image pairs contain
sets of matching features corresponding to the same 3D points.

http://szeliski.org/papers/Roberts_DuplicateSfM_CVPR11.pdf

johnwlambert: (but of course module renaming can be for the future)

travisdriver: Batching is only enforced for triangulation when the `MAX_DELAYED_TRIANGULATION_CALLS` is violated. I think a simple if-else statement is sufficient as opposed to a separate run function.

I think the function naming is straightforward. `run_triangulation` triangulates the 2D tracks. `run_da` performs the DA step (or at least what we were previously referring to as DA), i.e., validating the triangulated tracks and building the input to BA.

Yes, data association is the process of determining correspondences. This module still plays a role in determining correspondences, as we filter correspondences based on the triangulation geometry. 

You're more than welcome to rename this module to whatever you'd like in a separate PR :)

gtsfm/data_association/data_assoc.py
```diff
@@ -115,9 +122,12 @@ def run(
         exit_codes_wrt_computed: List[TriangulationExitCode] = []
         per_accepted_track_avg_errors = []
         per_rejected_track_avg_errors = []
-        for track_2d in tracks_2d:
+        assert len(tracks_2d) == len(sfm_tracks)
+        for j in range(len(tracks_2d)):
```
johnwlambert, line 126: nit: since we don't triangulate anymore here, I believe the comment below needs to be updated

gtsfm/data_association/data_assoc.py
```diff
@@ -68,21 +72,27 @@ def __validate_track(self, sfm_track: Optional[SfmTrack]) -> bool:
...
         tracks_2d: List[SfmTrack2d],
+        sfm_tracks: List[Optional[SfmTrack]],
+        avg_track_reproj_errors: List[Optional[float]],
+        triangulation_exit_codes: List[TriangulationExitCode],
         cameras_gt: List[Optional[gtsfm_types.CALIBRATION_TYPE]],
         relative_pose_priors: Dict[Tuple[int, int], Optional[PosePrior]],
         images: Optional[List[Image]] = None,
     ) -> Tuple[GtsfmData, GtsfmMetricsGroup]:
-        """Perform the data association.
+        """Perform the data association and compute metrics.
```
johnwlambert, line 87: Since we don't triangulate here, and don't compute correspondences here, I propose we totally rename this method to `compute_da_metrics()`.

johnwlambert: (and edit the docstring also to remove the mention of "Perform the data association", since all we do is compute metrics here know if I understand correctly)

gtsfm/data_association/data_assoc.py
```diff
@@ -68,21 +72,27 @@ def __validate_track(self, sfm_track: Optional[SfmTrack]) -> bool:
...
 
         Args:
             num_images: Number of images in the scene.
             cameras: dictionary, with image index -> camera mapping.
             tracks_2d: list of 2D tracks.
+            sfm_tracks: List of triangulated tracks.
+            avg_track_repoj_errors: List of average reprojection errors per track.
```
johnwlambert, line 94: nit: typo on "reproj errors"

gtsfm/data_association/data_assoc.py
johnwlambert: Re-reading this, looks like we do also validate the tracks here, and compute a largest connected component. How about we consider renaming this as
```python
def validate_tracks_and_compute_metrics()
```
or
```python
def validate_triangulated_tracks_and_compute_metrics()
  """ """
```
or 
```python
def assemble_gtsfm_data_from_tracks()
```
How about this as a docstring:
  """Assembles GtsfmData, validates triangulated tracks, computes metrics.

Only the largest connected component of cameras, represented in 3d tracks, is retained.
"""

johnwlambert (approved):
Thanks Travis!

stepanyanhayk:
Hi @travisdriver,

Thanks for this PR. I run your changes against the master on the cluster and it is much faster!

Skydio-32, deep_front_end
- Master: 1.78m
- This branch: 1.21m

Also, here is a screenshot from the dask dashboard demonstrating that the tasks are indeed distributed between all the machines.

Screenshot 2023-04-14 at 12 42 45 AM

gtsfm/data_association/data_assoc.py
travisdriver: I think these are good suggestions, but again I think it makes more sense to save this for another PR in which you can refactor the module names and functions to your liking.

gtsfm/data_association/data_assoc.py
travisdriver: Done

gtsfm/data_association/data_assoc.py
travisdriver: Done

gtsfm/data_association/data_assoc.py
```diff
@@ -196,6 +204,68 @@ def run(
...
+                    dask.delayed(triangulate_batch)(point3d_initializer, tracks_2d[j : j + batch_size])
+                )
+
+        # Perform triangulation in parallel.
+        triangulation_results = dask.compute(*triangulation_results)
+
+        # Unpack results.
+        sfm_tracks, avg_track_reproj_errors, triangulation_exit_codes = [], [], []
+        if batch_size == 1:
+            for result in triangulation_results:
```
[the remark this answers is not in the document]
travisdriver: Done

akshay-krishnan (approved):
not sure if this is covered by unit tests, please check artifacts for at least one dataset before merging, just to be sure nothing breaks.

travisdriver:
> not sure if this is covered by unit tests, please check artifacts for at least one dataset before merging, just to be sure nothing breaks.

newplot (22)

[visual_comparison_dashboard.zip](https://github.com/borglab/gtsfm/files/11236653/visual_comparison_dashboard.zip)

akshay-krishnan:
Are all differences in that visualization above coming from randomness? Or did we change anything in this PR?

PR #603, opened 2023-02-23
Moved Retriever to config; default for AstroVision

travisdriver:
Moved Retriever to config and made default for AstroVision, as it works quite well. E.g., it's able to identify the loop closure between the fist and last images images in the `2011205_rc3` sequence without having to do exhaustive matching.

view_graph_estimator_output_topology

Screenshot from 2023-02-24 12-15-38

PR #623, opened 2023-04-14
Cleanup-only PR, no functional changes

akshay-krishnan:
A bunch of things to clean up from the recent submitted PRs: 
- type hints were not consistent since we changed some things to delayed and back. 
- removed some duplicate code in two-view estimator
- removed some unused arguments
- ground truth poses get saved in the same colmap format with the estimated tracks (for potential visualization)


travisdriver:
Would it be possible to hold off on merging this until I merge #619 and #603 (which I plan on doing today)? I'm afraid this will result in a lot of merge conflicts.

gtsfm/frontend/correspondence_generator/correspondence_generator_base.py
```diff
@@ -16,14 +16,14 @@ class CorrespondenceGeneratorBase:
     def create_computation_graph(
         self,
         delayed_images: List[Delayed],
-        image_shapes: List[Tuple[int, int]],
+        delayed_image_shapes: List[Delayed],
```
ayushbaid, line 19: Why make them delayed? I think we reverted to using non-delayed types during Hilti challenge.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -217,9 +216,7 @@ def run(self) -> None:
 
         # create dask cluster
         if self.parsed_args.cluster_config:
-            workers = OmegaConf.load(
-                os.path.join(
-                    self.parsed_args.output_root, "gtsfm", "configs", self.parsed_args.cluster_config))["workers"]
+            workers = OmegaConf.load(self.parsed_args.cluster_config)["workers"]
```
ayushbaid, line 219: Do we need to pass in the full path of the config file?

ayushbaid (approved):

ayushbaid:
Please enable mypy checks in CI if this PR resolves all of them.

akshay-krishnan:
> Please enable mypy checks in CI if this PR resolves all of them.

There are too many to be resolved in a single PR :) These are only some outrageous errors that are fixed.

gtsfm/runner/gtsfm_runner_base.py
akshay-krishnan: Isnt that a better choice? Using output_root did not make sense at all, that is where outputs are stored. we could use the root folder + "configs/", but maybe its better to just ask the user to provide the full path.

gtsfm/frontend/correspondence_generator/correspondence_generator_base.py
akshay-krishnan: they are already delayed, I am only fixing the type hints here :)

akshay-krishnan: the reason we made them delayed in a different PR was so that they could be accessed on a cluster. I dont think they need to be delayed though.. that would be on @stepanyanhayk  to check.

gtsfm/runner/run_scene_optimizer_astrovision.py
```diff
@@ -83,12 +83,10 @@ def run(self) -> None:
                 image_pair_indices=image_pair_indices,
                 image_graph=self.loader.create_computation_graph_for_images(),
                 all_intrinsics=self.loader.create_computation_graph_for_intrinsics(),
-                image_shapes=self.loader.create_computation_graph_for_image_shapes(),
                 relative_pose_priors=self.loader.get_relative_pose_priors(image_pair_indices),
                 absolute_pose_priors=self.loader.get_absolute_pose_priors(),
                 cameras_gt=self.loader.create_computation_graph_for_gt_cameras(),
                 gt_wTi_list=self.loader.get_gt_poses(),
-                gt_scene_mesh=gt_scene_trimesh_future,
```
travisdriver, line 91: Why did you remove this functionality?

travisdriver: Yes it is...

gtsfm/evaluation/metrics.py
```diff
@@ -90,8 +90,11 @@ def __init__(
         self._name = name
         if data is not None:
             # Cast to a numpy array
-            if isinstance(data, list) and all(isinstance(x, int) for x in data):
-                data = np.array(data, dtype=np.int_)
+            if isinstance(data, list):
```
johnwlambert, line 93: just curious, where was this issue showing up?

gtsfm/evaluation/metrics.py
```diff
@@ -310,7 +315,10 @@ def save_to_json(self, path: str) -> None:
         Args:
             path: path to json file.
         """
-        io.save_json_file(path, self.get_metrics_as_dict())
+        try:
+            io.save_json_file(path, self.get_metrics_as_dict())
```
johnwlambert, line 319: also curious, were some files not saving on some workers?

gtsfm/frontend/correspondence_generator/correspondence_generator_base.py
johnwlambert: Confirming that the image shapes to the CorrespondenceGenerator are currently delayed, according to this line: https://github.com/borglab/gtsfm/blob/master/gtsfm/runner/gtsfm_runner_base.py#L232

gtsfm/multi_view_optimizer.py
```diff
@@ -46,14 +45,14 @@ def create_computation_graph(
         self,
         images_graph: List[Delayed],
         num_images: int,
-        keypoints_graph: List[Delayed],
+        keypoints_graph: List[Keypoints],
```
johnwlambert, line 48: Confirming that indeed the type is `Keypoints` and not `Delayed`, since the `CorrespondenceGenerator` landed.

https://github.com/borglab/gtsfm/blob/master/gtsfm/scene_optimizer.py#L247

gtsfm/multi_view_optimizer.py
```diff
@@ -46,14 +45,14 @@ def create_computation_graph(
...
         i2Ri1_graph: Dict[Tuple[int, int], Delayed],
         i2Ui1_graph: Dict[Tuple[int, int], Delayed],
         v_corr_idxs_graph: Dict[Tuple[int, int], Delayed],
         all_intrinsics: List[Optional[gtsfm_types.CALIBRATION_TYPE]],
         absolute_pose_priors: List[Optional[PosePrior]],
         relative_pose_priors: Dict[Tuple[int, int], PosePrior],
-        two_view_reports_dict: Optional[Dict[Tuple[int, int], TwoViewEstimationReport]],
+        two_view_reports_dict: Optional[Dict[Tuple[int, int], Delayed]],
```
johnwlambert, line 55: Confirming that the type is indeed Delayed, per

https://github.com/borglab/gtsfm/blob/master/gtsfm/scene_optimizer.py#L254

These come out of the TwoViewEstimator as Delayed objects:
https://github.com/borglab/gtsfm/blob/master/gtsfm/two_view_estimator.py#L382

gtsfm/evaluation/metrics.py
johnwlambert: Conversion should already be automatic from None -> NaN if we are doing integer conversions for float:

```
>>> import numpy as np
>>> np.array([None, 1, 2, None])
array([None, 1, 2, None], dtype=object)
>>> x = np.array([None, 1, 2, None])
>>> y = np.array(x, dtype=np.float32)
>>> y
array([nan,  1.,  2., nan], dtype=float32)
```

but I guess we are using all ints here, and not casting to float32 with these int args?

gtsfm/view_graph_estimator/view_graph_estimator_base.py
```diff
@@ -163,6 +163,7 @@ def compute_metrics(
         calibrations: List[Cal3Bundler],
         two_view_reports: Dict[Tuple[int, int], TwoViewEstimationReport],
         view_graph_edges: List[Tuple[int, int]],
+        plots_output_dir: Optional[Path] = None,
```
johnwlambert, line 166: why note just do `plots_output_dir: Path = PLOT_BASE_PATH`, since line 191 effectively does this?

gtsfm/utils/io.py
```diff
@@ -210,7 +210,7 @@ def colmap2gtsfm(
                 sfmtrack.addMeasurement(image_id_to_idx[image_id], images[image_id].xys[point2d_idx])
             sfmtracks_gtsfm.append(sfmtrack)
 
-    return img_fnames, images_gtsfm, cameras_gtsfm, sfmtracks_gtsfm
+    return img_fnames, wTi_gtsfm, intrinsics_gtsfm, sfmtracks_gtsfm
```
johnwlambert, line 213: @akshay-krishnan could you update the two docstring `Returns` items also?

gtsfm/two_view_estimator.py
```diff
@@ -250,13 +250,18 @@ def run_2view(
...
         gt_wTi2: Optional[Pose3],
         gt_scene_mesh: Optional[Any] = None,
-    ) -> Tuple[Optional[Rot3], Optional[Unit3], np.ndarray, Dict[str, Optional[TwoViewEstimationReport]]]:
+    ) -> Tuple[
+        Optional[Rot3],
+        Optional[Unit3],
+        np.ndarray,
+        TwoViewEstimationReport,
+        TwoViewEstimationReport,
+        Optional[TwoViewEstimationReport],
```
johnwlambert, line 263: As far as I can tell, this last return arg should not be optional, and the docstring seems wrong in the ISP, since we never return a None report? https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/inlier_support_processor.py#L45

gtsfm/scene_optimizer.py
```diff
@@ -410,9 +406,35 @@ def save_visualizations(
...
 
+def get_gt_gtsfm_data(
+    cameras_gt: List[Optional[gtsfm_types.CAMERA_TYPE]],
+    ba_output: GtsfmData,
+) -> GtsfmData:
+    """Creates GtsfmData object with GT camera poses and estimated tracks.
+
+    Args:
+        gtsfm_data: GtsfmData object with estimated camera poses and tracks.
+        cameras_gt: list of GT cameras.
```
johnwlambert, line 417: nit: capitalization on "List of GT cameras"

gtsfm/scene_optimizer.py
```diff
@@ -410,9 +406,35 @@ def save_visualizations(
     return viz_graph_list
 
 
+def get_gt_gtsfm_data(
```
johnwlambert, line 409: gt_gtsfm sounds a bit weird, how about `get_ground_truth_gtsfm_data()`?

gtsfm/scene_optimizer.py
```diff
@@ -443,6 +465,18 @@ def save_gtsfm_data(
...
             )
         )
+
+        # Save the ground truth in the same format, for visualization.
+        # We use the estimated tracks here, with ground truth camera poses.
+        gt_gtsfm_data = dask.delayed(get_gt_gtsfm_data)(cameras_gt, ba_output_graph)
+        saving_graph_list.append(
+            dask.delayed(io_utils.export_model_as_colmap_text)(
+                gt_gtsfm_data,
+                image_graph,
```
johnwlambert, line 475: if possible, would be nice to have all 3 as keyword args, `gtsfm_data=gt_gtsfm_data, images=image_graph, ...`

https://github.com/borglab/gtsfm/blob/master/gtsfm/utils/io.py#L163-L165

gtsfm/scene_optimizer.py
johnwlambert: I wouldn't exactly call this GT, since the tracks can be noisy. I would call pure GT synthetic correspondences raycasted against mesh, w/ GT poses and GT intrinsics.

How about `get_gtsfm_data_w_est_tracks_and_gt_cameras()`?

gtsfm/scene_optimizer.py
```diff
@@ -410,9 +406,35 @@ def save_visualizations(
     return viz_graph_list
 
 
+def get_gt_gtsfm_data(
+    cameras_gt: List[Optional[gtsfm_types.CAMERA_TYPE]],
```
johnwlambert, line 410: the GT cameras should be all or nothing for each of our datasets (either for every image, or none of them)

gtsfm/scene_optimizer.py
```diff
@@ -367,17 +362,20 @@ def align_estimated_gtsfm_data(
         Updated ba_output GtsfmData object aligned to axes.
         Updated gt_pose_graph with GT poses aligned to axes.
     """
+    ba_input = ba_input.align_via_Sim3_to_poses(gt_wTi_list)
```
johnwlambert, line 365: from what I call, order of ops was very important for these alignment transforms. are we sure this still works out?

johnwlambert (approved):
LGTM

gtsfm/runner/run_scene_optimizer_astrovision.py
akshay-krishnan: Ive reverted this change, resolving

gtsfm/evaluation/metrics.py
akshay-krishnan: I dont think we want to cast all metrics to float. there are some metrics that are assumed to be `int` for good reason. for example: the track length histogram in the report relies on all numbers being ints.

gtsfm/evaluation/metrics.py
akshay-krishnan: I had problems with some metrics not being written to json because they could not be parsed. This was a way of not crashing the pipeline and knowing which metrics are causing the issue. I think its better than what we currently do, but also fine with not checking it in.

gtsfm/scene_optimizer.py
akshay-krishnan: yes, there are 2 alignments going on: 
1. aligning estimated poses to GT poses
2. aligning estimated poses so as to center the reconstruction

(1) was previously not happening for the saved gtsfm data. For the pose plots (which are not very usable for large datasets) (2) was happening before (1).

gtsfm/two_view_estimator.py
akshay-krishnan: thanks for catching that, it was annoying to have that optional around everywhere.

akshay-krishnan:
@johnwlambert some 3-4 unresolved comments left, could you let me know if those need any changes?

PR #624, opened 2023-04-18
Runner + Loader for the mobile brick dataset

akshay-krishnan:
The MobileBrick dataset has ground truth camera poses and aligned 3D models of objects of interest. The dataset has a "train" and "test" split where the train split contains synthetic images but the test set only contains images of real LEGOs with a 3D model aligned to the scene reconstruction. 

This PR has only been tested on the real sequences in the test split. Attached is an image and a gtsfm reconsutrction of the colosseum sequence. 
 
 

Screen Shot 2023-04-18 at 8 43 25 AM


gtsfm/loader/mobilebrick_loader.py
```diff
@@ -0,0 +1,135 @@
...
+            Image: the image at the query index.
+        """
+        if index < 0 or index >= len(self):
+            raise IndexError(f"Image index {index} is invalid")
+
+        # Read in image.
+        img = io_utils.load_image(self._image_paths[index])
+        return Image(value_array=img.value_array, exif_data=img.exif_data, file_name=img.file_name)
+
+    def get_camera_intrinsics_full_res(self, index: int) -> Cal3Bundler:
```
johnwlambert, line 95: Thanks for the PR, Akshay!

I wonder if we could add a flag or config option to infer intrinsics from exif or image size, rather than always using GT ones?

gtsfm/loader/mobilebrick_loader.py
```diff
@@ -0,0 +1,135 @@
...
+            raise IndexError(f"Image index {index} is invalid")
+
+        wTi = self._wTi[index]
+        return wTi
+
+    def is_valid_pair(self, idx1: int, idx2: int) -> bool:
+        """Checks if (idx1, idx2) is a valid pair. idx1 < idx2 is required.
+
+        Args:
+            idx1: first index of the pair.
```
johnwlambert, line 129: Small nit: Google style would say first word should be capitalized, throughout this file :-)

“idx1: First index of the pair.”

gtsfm/loader/mobilebrick_loader.py
```diff
@@ -0,0 +1,135 @@
...
+            image_path = os.path.join(self._image_dir, f"{i:06d}.jpg")
+            self._image_paths.append(image_path)
+
+        # Load GT intrinsics
+        self._intrinsics_dir = os.path.join(data_dir, "intrinsic")
+        self._intrinsics = []
+        for i in range(self._num_images):
+            intrinsics_file = os.path.join(self._intrinsics_dir, f"{i:06d}.txt")
+            K = np.loadtxt(intrinsics_file)
+            self._intrinsics.append(Cal3Bundler((K[0, 0] + K[1, 1]) / 2, 0, 0, K[0, 2], K[1, 2]))
```
johnwlambert, line 53: Could we add a small unit test with loading two images?

gtsfm/loader/mobilebrick_loader.py
akshay-krishnan: Yes, I would also want to test it in that setting first. Waiting on @duyanwei 's PR to be merged so I could potentially use some utils from there.

gtsfm/loader/mobilebrick_loader.py
akshay-krishnan: copilot should get better at this :(

gtsfm/loader/mobilebrick_loader.py
akshay-krishnan: Added unit tests, let me know if 5 images are too many.

gtsfm/loader/mobilebrick_loader.py
akshay-krishnan: Done

gtsfm/common/image.py
```diff
@@ -34,6 +34,13 @@ def width(self) -> int:
         """
         return self.value_array.shape[1]
 
+    @property
+    def shape(self) -> tuple:
```
johnwlambert, line 38: nit: `Tuple[int, int]` or `tuple[int,int]`?

gtsfm/common/image.py
```diff
@@ -34,6 +34,13 @@ def width(self) -> int:
...
+        """
+        The shape of the image (H, W, C).
```
johnwlambert, line 40: nit: can we fit it on one line? `"""The shape of the image (H,W)."""`

Btw we don't use channels elsewhere when describing the shape, do we need it here? 
https://github.com/borglab/gtsfm/blob/581f84ab605e99302a9c6425dc67e38f82b70660/gtsfm/loader/loader_base.py#L219

gtsfm/loader/mobilebrick_loader.py
johnwlambert: gentle ping on this, now that his PR is landed :  -)

gtsfm/loader/mobilebrick_loader.py
```diff
@@ -0,0 +1,135 @@
...
+        """
+        return self._num_images
+
+    def get_image_full_res(self, index: int) -> Image:
+        """Get the image at the given index, at full resolution.
+
+        Args:
+            index: The index to fetch.
+
+        Raises:
```
johnwlambert, line 82: nit: usually `Raises` would come after the `Returns`: https://google.github.io/styleguide/pyguide.html#244-decision

gtsfm/loader/mobilebrick_loader.py
johnwlambert: LGTM

gtsfm/loader/mobilebrick_loader.py
```diff
@@ -0,0 +1,135 @@
...
+            image_path = os.path.join(self._image_dir, f"{i:06d}.jpg")
+            self._image_paths.append(image_path)
+
+        # Load GT intrinsics
+        self._intrinsics_dir = os.path.join(data_dir, "intrinsic")
+        self._intrinsics = []
+        for i in range(self._num_images):
+            intrinsics_file = os.path.join(self._intrinsics_dir, f"{i:06d}.txt")
+            K = np.loadtxt(intrinsics_file)
+            self._intrinsics.append(Cal3Bundler((K[0, 0] + K[1, 1]) / 2, 0, 0, K[0, 2], K[1, 2]))
```
johnwlambert, line 53: nit: keyword args on the Cal3Bundler args could make this a little more readable to me, but up to you.

gtsfm/common/image.py
johnwlambert: Actually I think it's fine to include channels, since it just parallels what the numpy attribute .shape field would have said.

johnwlambert (approved):
LGTM, thanks!

gtsfm/common/image.py
akshay-krishnan: yeah, would not want to restrict an `Image` class to 3 channel images :)

gtsfm/loader/mobilebrick_loader.py
akshay-krishnan: added a flag to support approximate intrinsics. from my experiments, this significantly impacts results. The default value of 1.2 for the focal length factor from colmap provides results that are not usable. The value Ive chosen (0.8) is closer to ground truth, but I have not done any more tuning there. we should find an algorithmic way to solve this problem.

travisdriver:
Could you post some pictures of the final reconstructions when you get the chance

gtsfm/loader/mobilebrick_loader.py
johnwlambert: Interesting, thanks for adding this.

gtsfm/loader/mobilebrick_loader.py
```diff
@@ -0,0 +1,152 @@
...
+        Raises:
+            IndexError: If an out-of-bounds image index is requested.
+        """
+        if index < 0 or index >= len(self):
+            raise IndexError(f"Image index {index} is invalid")
+
+        if self._intrinsics:
+            return self._intrinsics[index]
+        else:
+            # 0.8 is better than the default factor of 1.2 for this dataset, but it has not been fully tuned.
```
johnwlambert, line 119: nit: could be helpful to add a TODO to use a system-wide default, instead of per-dataset options.

johnwlambert (approved):
Thanks! Feel free to merge

gtsfm/loader/mobilebrick_loader.py
akshay-krishnan: The system-wide default is 1.2, but it would not work for all datasets. unless we change the verifier or refine intrinsics in BA. 
Are you suggesting adding a TODO to update this when we find a way to use a system-wide default?

johnwlambert: Yup, exactly

akshay-krishnan:
> Could you post some pictures of the final reconstructions when you get the chance

@travisdriver I updated the PR description

PR #629, opened 2023-04-25
Fix netvlad output dir

akshay-krishnan:
NetVLAD writes plots to a different output directory "plots" in the repo root, but outputs are saved in a different path, based on --output_dir. 
The metrics later look inside "--output_dir" for NetVLAD's plots, and crashes when it cannot find them there.


johnwlambert (approved):
LGTM

PR #635, opened 2023-05-03
Use the GT intrinsics instead of initialized intrinsics for metrics

akshay-krishnan:
The two-view estimator was using the intrinsics used for initializing SfM estimates as the GT estimates. This PR switches to loading and using the GT intrinsics themselves. 

This makes no difference for all datasets besides MobileBrick since we use GT intrinsics as initialization for others. All other loaders can be updated to separate the two following this PR, in one of two ways: 
-  If they use GT intrinsics (eg: from colmap) to initialize, this must be avoided
-  If data from EXIF is used as ground truth, it could optionally be replaced by "better" intrinsics (eg: from colmap). 

This PR also fixes a bug that was using the inlier ratio wrt estimated model before BA as the inlier ratio wrt estimated model after BA. It would be worth checking the inlier ratio wrt estimated model for other datasets after this PR. For a MobileBrick sequence, I notice a change in inlier ratio pre-BA of 85% to an inlier ratio post-BA of 55%.

### DIFF
--- a/gtsfm/loader/loader_base.py
+++ b/gtsfm/loader/loader_base.py
@@ -64,7 +64,7 @@ class LoaderBase(GTSFMProcess):
         Note: length should be found without loading images into memory.
 
         Returns:
-            the number of images.
+            The number of images.
         """
 
     # ignored-abstractmethod
@@ -73,13 +73,13 @@ class LoaderBase(GTSFMProcess):
         """Get the image at the given index, at full resolution.
 
         Args:
-            index: the index to fetch.
+            index: The index to fetch.
 
         Raises:
-            IndexError: if an out-of-bounds image index is requested.
+            IndexError: If an out-of-bounds image index is requested.
 
         Returns:
-            Image: the image at the query index.
+            Image: The image at the query index.
         """
 
     # ignored-abstractmethod
@@ -88,10 +88,10 @@ class LoaderBase(GTSFMProcess):
         """Get the camera intrinsics at the given index, valid for a full-resolution image.
 
         Args:
-            the index to fetch.
+            index: The index to fetch.
 
         Returns:
-            intrinsics for the given camera.
+            Intrinsics for the given camera.
         """
 
     # ignored-abstractmethod
@@ -100,10 +100,10 @@ class LoaderBase(GTSFMProcess):
         """Get the camera pose (in world coordinates) at the given index.
 
         Args:
-            index: the index to fetch.
+            index: The index to fetch.
 
         Returns:
-            the camera pose w_P_index.
+            The camera pose w_P_index.
         """
 
     # TODO: Rename this to get_gt_camera.
@@ -111,7 +111,7 @@ class LoaderBase(GTSFMProcess):
         """Gets the GT camera at the given index.
 
         Args:
-            index: the index to fetch.
+            index: The index to fetch.
 
         Returns:
             Camera object with intrinsics and extrinsics, if they exist.
@@ -132,11 +132,11 @@ class LoaderBase(GTSFMProcess):
         Note: All inherited classes should call this super method to enforce this check.
 
         Args:
-            idx1: first index of the pair.
-            idx2: second index of the pair.
+            idx1: First index of the pair.
+            idx2: Second index of the pair.
 
         Returns:
-            validation result.
+            Validation result.
         """
         return idx1 < idx2
 
@@ -147,13 +147,13 @@ class LoaderBase(GTSFMProcess):
         Each loader implementation should set a `_max_resolution` attribute.
 
         Args:
-            index: the index to fetch.
+            index: The index to fetch.
 
         Raises:
-            IndexError: if an out-of-bounds image index is requested.
+            IndexError: If an out-of-bounds image index is requested.
 
         Returns:
-            Image: the image at the query index. It will be resized to satisfy the maximum
+            Image: The image at the query index. It will be resized to satisfy the maximum
                 allowed loader image resolution if the full-resolution images for a dataset
                 are too large.
         """
@@ -184,6 +184,18 @@ class LoaderBase(GTSFMProcess):
     def __rescale_intrinsics(
         self, intrinsics_full_res: gtsfm_types.CALIBRATION_TYPE, image_index: int
     ) -> gtsfm_types.CALIBRATION_TYPE:
+        """Rescale the intrinsics to match the image resolution.
+
+        Reads the image from disk to determine the scaling factor.
+
+        Args:
+            intrinsics_full_res: Intrinsics for the given camera at full resolution.
+            image_index: The index to fetch.
+
+        Returns:
+            Rescaled intrinsics for the given camera at the desired resolution.
+        """
+
         if intrinsics_full_res.fx() <= 0:
             raise RuntimeError("Focal length must be positive.")
 
@@ -210,10 +222,10 @@ class LoaderBase(GTSFMProcess):
         Each loader implementation should set a `_max_resolution` attribute.
 
         Args:
-            the index to fetch.
+            index: The index to fetch.
 
         Returns:
-            intrinsics for the given camera.
+            Intrinsics for the given camera.
         """
         intrinsics_full_res = self.get_camera_intrinsics_full_res(index)
         if intrinsics_full_res is None:
@@ -228,10 +240,10 @@ class LoaderBase(GTSFMProcess):
         be overridden by subclasses to return a superior ground truth intrinsics.
 
         Args:
-            the index to fetch.
+            index: The index to fetch.
 
         Returns:
-            intrinsics for the given camera.
+            Intrinsics for the given camera.
         """
         return self.get_camera_intrinsics_full_res(index)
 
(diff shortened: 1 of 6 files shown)


gtsfm/loader/loader_base.py
```diff
@@ -196,25 +219,42 @@ def get_camera_intrinsics(self, index: int) -> Optional[gtsfm_types.CALIBRATION_
...
+        """Get the GT camera intrinsics at the given index, for a possibly resized image.
+
+        Determine how the camera intrinsics and images should be jointly rescaled based on desired img. resolution.
+        Each loader implementation should set a `_max_resolution` attribute.
+
+        The GT camera intrinsics are the same as the camera intrinsics by default, but can be overridden by defining
+        get_gt_camera_intrinsics_full_res method in derived classes.
+
+        Args:
+            the index to fetch.
```
johnwlambert, line 248: nit: google style : - )

```python
Args:
  index: The index to fetch.
```

gtsfm/loader/loader_base.py
```diff
@@ -196,25 +219,42 @@ def get_camera_intrinsics(self, index: int) -> Optional[gtsfm_types.CALIBRATION_
...
+        """
+        return self.get_camera_intrinsics_full_res(index)
+
+    def get_gt_camera_intrinsics(self, index: int) -> Optional[gtsfm_types.CALIBRATION_TYPE]:
+        """Get the GT camera intrinsics at the given index, for a possibly resized image.
+
+        Determine how the camera intrinsics and images should be jointly rescaled based on desired img. resolution.
+        Each loader implementation should set a `_max_resolution` attribute.
+
+        The GT camera intrinsics are the same as the camera intrinsics by default, but can be overridden by defining
```
johnwlambert, line 244: this sentence is a little weird: "The GT camera intrinsics are the same as the camera intrinsics by default,". maybe we mean "...are the same as the camera instrinsics returned by the loader in general for certain loader implementations, such as the COLMAP loader?"

gtsfm/loader/loader_base.py
```diff
@@ -196,25 +219,42 @@ def get_camera_intrinsics(self, index: int) -> Optional[gtsfm_types.CALIBRATION_
...
-            k2=0.0,
-            u0=intrinsics_full_res.px() * scale_u,
-            v0=intrinsics_full_res.py() * scale_v,
-        )
-        return rescaled_intrinsics
+        By default, this is implemented to return the same value as `get_camera_intrinsics_full_res`. However, this can
+        be overridden by subclasses to return a superior ground truth intrinsics.
+
+        Args:
+            the index to fetch.
```
johnwlambert, line 231: nit: Google style on this docstring : - )

tests/loader/test_mobilebrick_loader.py
```diff
@@ -50,8 +50,8 @@ def test_get_image_full_res(self):
         with self.assertRaises(IndexError):
             self.loader.get_image_full_res(5)
 
-    def test_get_camera_intrinsics_full_res(self):
-        """Test the camera intrinsics at a given index."""
+    def test_get_gt_camera_intrinsics_full_res(self):
```
johnwlambert, line 53: we aren't calling `get_gt_camera_intrinsics_full_res()` in this fn though, right?

Bc we don't pass `use_gt_intrinsics: bool = True` in `setUp()`?

gtsfm/two_view_estimator.py
```diff
@@ -263,81 +325,56 @@ def run_2view(
...
+                pre_ba_v_corr_idxs,
                 camera_intrinsics_i1,
                 camera_intrinsics_i2,
                 pre_ba_i2Ri1,
                 pre_ba_i2Ui1,
                 i2Ti1_prior,
             )
+            post_ba_inlier_ratio_wrt_estimate = float(len(post_ba_v_corr_idxs)) / len(putative_corr_idxs)
+            post_ba_report = self.__get_2view_report_from_results(
+                post_ba_i2Ri1,
```
johnwlambert, line 363: nit: can we use keywords args since we got so many args rn?

gtsfm/two_view_estimator.py
```diff
@@ -215,6 +215,68 @@ def bundle_adjust(
...
+            # if we have the expected GT data, evaluate the computed relative pose
+            R_error_deg, U_error_deg = compute_relative_pose_metrics(
+                i2Ri1_computed, i2Ui1_computed, gt_camera_i1.pose(), gt_camera_i2.pose()
+            )
+            # TODO: add support for other camera models
+            if isinstance(gt_camera_i1, PinholeCameraCal3Bundler) and isinstance(
+                gt_camera_i2, PinholeCameraCal3Bundler
+            ):
+                inlier_mask_wrt_gt, reproj_error_wrt_gt = metric_utils.compute_correspondence_metrics(
+                    keypoints_i1,
```
johnwlambert, line 258: nit: can we use keyword args here since we're changing the keyword order and bc we got so many args rn here?

gtsfm/two_view_estimator.py
```diff
@@ -215,6 +215,68 @@ def bundle_adjust(
...
+        gt_camera_i1: Optional[gtsfm_types.CAMERA_TYPE],
+        gt_camera_i2: Optional[gtsfm_types.CAMERA_TYPE],
+        gt_scene_mesh: Optional[Any],
+    ) -> TwoViewEstimationReport:
+        """Generate a TwoViewEstimationReport from the results of the two-view estimation.
+
+        Currently metrics wrt GT camera only supports cases where gt_camera is PinholeCameraCal3Bundler.
+
+        Args:
+            i2Ri1_computed: computed relative rotation.
```
johnwlambert, line 235: nit: Capitalization for google style

gtsfm/two_view_estimator.py
```diff
@@ -263,81 +325,56 @@ def run_2view(
...
                 keypoints_i2,
-                verified_corr_idxs,
+                pre_ba_v_corr_idxs,
                 camera_intrinsics_i1,
                 camera_intrinsics_i2,
                 pre_ba_i2Ri1,
                 pre_ba_i2Ui1,
                 i2Ti1_prior,
             )
+            post_ba_inlier_ratio_wrt_estimate = float(len(post_ba_v_corr_idxs)) / len(putative_corr_idxs)
```
johnwlambert, line 361: nice catch!

gtsfm/two_view_estimator.py
```diff
@@ -215,6 +215,68 @@ def bundle_adjust(
 
         return i2Ti1_optimized.rotation(), Unit3(i2Ti1_optimized.translation()), valid_corr_idxs
 
+    def __get_2view_report_from_results(
```
johnwlambert, line 218: nice refactor / de-dup : - )

johnwlambert (commented):
Looks good, a few random nits on my end

gtsfm/two_view_estimator.py
akshay-krishnan: copilot would write in lowercase, probably because all other comments in this file used lowercase, Ive updated them all.

gtsfm/loader/loader_base.py
akshay-krishnan: again, looks like copilot follows the norm in the file :)

akshay-krishnan:
the inlier ratio change has some downstream effects. 

As expected, all inlier ratios reduce significantly post BA: 
image

fewer image-pairs make it past the inlier support processor: 
image

The ones that did not make it past seem to be outliers in view graph estimation (reduced number of outliers here):
image

from BA and averaging metrics, astrovision seems to improve a little bit, but palace degrades noticeably. The 200%+ change for palace in rotation error is from 0.09 to 0.3. Dont know why this happens, it seems to need the edges we lost? 
image

gtsfm/loader/loader_base.py
akshay-krishnan: "default" was about the base class implementation, I reworded a bit

johnwlambert:
Interesting. @akshay-krishnan  If this change leads to a regressions, we could just plug in the single pre_* var where the post_* var is and keep all the other changes as is, with a TODO to investigate later?

akshay-krishnan:
@johnwlambert Im happy to do any investigation before submitting this PR. It feels wrong to use the ratio from pre_ba_ when we are using the correspondences post_ba_. 
I can check if there's any obvious tuning we can do for palace to avoid that regression. But if not, this might be a regression we have to take (at least temporarily) for the purpose of being correct?

akshay-krishnan:
But if you think I should do that in a separate PR and merge this now, that works too.

johnwlambert:
I think since there are many other changes in the PR, better to replace that one var, verify no regression, and then merge this now, before we end up with some merge conflicts. I think correctness is measured by the end metrics alone :-) thanks for your work on this and for spotting this!

akshay-krishnan:
@johnwlambert merging this PR, Ive reverted to using the original pre_ba_ ratio for post_ba_.

johnwlambert:
@akshay-krishnan thanks for the update — do you mind creating the dashboard one more time to make sure this solved the regression?

akshay-krishnan:
Post-BA metrics
image

johnwlambert (approved):
Thanks!

PR #640, opened 2023-06-15
Triangulation reprojection error filtering before 2-view BA

travisdriver:
We were not filtering matches by the reprojection error before performing 2-view bundle adjustment, as there was a separate triangulation method defined in the `TwoViewEstimator`.


travisdriver:
Does not have much of an impact for the CI datasets. The "regressions" are on the order of 0.01. I think it should still be something be check for, since the fidelity of the two-view pose estimates is very important. Also, Theia and COLMAP do it.

Results for 10 pixel threshold for all cases:

newplot (2)

[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/11763888/visual_comparison_dashboard.html.zip)

akshay-krishnan:
Theoretically speaking, verified correspondences lie on the same plane and should therefore have low triangulation error. Any error might come from the decomposition of the E/F matrix into (R,t), but that does not mean we should discard the correspondence. 

If there is no significant difference, we should try to keep the pipeline simpler. But we can include this step if there is an empirical improvement.

travisdriver:
> Theoretically speaking, verified correspondences lie on the same plane and should therefore have low triangulation error. Any error might come from the decomposition of the E/F matrix into (R,t), but that does not mean we should discard the correspondence.
> 
> If there is no significant difference, we should try to keep the pipeline simpler. But we can include this step if there is an empirical improvement.

I think it would be a good idea to have to option to turn this on, e.g., for verifying epipolar-guided matches. I can keep it off by default.

gtsfm/common/gtsfm_data.py
```diff
@@ -165,6 +180,7 @@ def add_track(self, track: SfmTrack) -> bool:
             i, _ = track.measurement(j)
 
             if i not in self._cameras:
+                # TODO (travisdriver): Should we throw an error here?
```
akshay-krishnan, line 183: TODO: should we maybe keep the track without this measurement?

gtsfm/two_view_estimator.py
```diff
@@ -160,56 +104,54 @@ def bundle_adjust(
...
         start_time = timeit.default_timer()
-        # TODO: add flag to switch between verified and putative correspondences.
-        triangulated_tracks, triangulated_indices = self.triangulate_two_view_correspondences(
-            camera_i1=camera_i1,
-            camera_i2=camera_i2,
-            keypoints_i1=keypoints_i1,
-            keypoints_i2=keypoints_i2,
-            corr_idxs=verified_corr_idxs,
-        )
+        for j, (idx1, idx2) in enumerate(verified_corr_idxs):
```
akshay-krishnan, line 126: I have a slight preference to having this in a helper like it previously was, but upto you.

akshay-krishnan (approved):
unrelated, but @ayushbaid why doesnt gtsfm data allow adding tracks for which a subset of cameras could be in the struct? 

@travisdriver please disable the filtering post triangulation by default before submitting, we could enable it if it makes a difference.

gtsfm/common/gtsfm_data.py
travisdriver: If the track is referencing a measurement that doesn't exist, I think we should throw an error instead of adding the corrupted track. But I'm open to see what others think @ayushbaid @johnwlambert

.gitignore
```diff
@@ -160,7 +160,7 @@ plots/
 results/
 result_metrics/
 visual_comparison_dashboard.html
```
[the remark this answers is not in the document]
travisdriver: done

gtsfm/two_view_estimator.py
travisdriver: done

travisdriver:
> unrelated, but @ayushbaid why doesnt gtsfm data allow adding tracks for which a subset of cameras could be in the struct?
> 
> @travisdriver please disable the filtering post triangulation by default before submitting, we could enable it if it makes a difference.

I made the reprojection error threshold `np.inf` by default and don't explicitly set a threshold for the TwoViewEstimator.

travisdriver:
Final comparison dashboard: [visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/11777274/visual_comparison_dashboard.html.zip)

PR #593, opened 2022-11-17
Utility script for producing inputs to Nerfstudio

womackj1:
After doing a GTSfM reconstruction on the set1_lund_door dataset, you can produce Nerfstudio inputs with the following command:

`python3 prepare_nerfstudio.py --results_path=../../results --camera_model=perspective --images_dir=../../tests/data/set1_lund_door/images/
`

Then, if you have Nerfstudio setup, you can train a nerf with a command like:

`ns-train nerfacto --data nerfstudio-input/ --vis viewer --viewer.websocket-port=7008`

Screen Shot 2022-11-16 at 9 01 29 PM

PR #642, opened 2023-06-15
Feature/nerfstudio gtsfm rebase

travisdriver:
Rebase #593.


travisdriver:
Not sure what's happening with the failed cache test. It passed on my local machine and this PR does not make any changes to the cache.

gtsfm/utils/io.py
```diff
@@ -250,11 +249,12 @@ def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
             k2 = 0
             calibrations.append(Cal3Bundler(fx, k1, k2, u0, v0))
         elif model == "RADIAL":
-            _, _, img_w, img_h, fx, u0, v0, k1, k2 = cam_params[:9]
-            img_w, img_h, fx, u0, v0, k1, k2 = (
+            _, _, img_w, img_h, fx, fy, u0, v0, k1, k2 = cam_params[:10]
```
akshay-krishnan, line 252: was this a bug?

gtsfm/utils/prepare_nerfstudio.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+CAMERA_MODELS = {
+    "perspective": CameraModel.OPENCV,
+    "fisheye": CameraModel.OPENCV_FISHEYE,
+}
+
+
+def colmap_to_json(cameras_path: Path, images_path: Path, output_dir: Path, camera_model: CameraModel) -> int:
+    """Converts COLMAP's cameras.bin and images.bin to a JSON file.
```
akshay-krishnan, line 38: the code seems to read the .txt files, not the .bin ones?

gtsfm/utils/prepare_nerfstudio.py
```diff
@@ -0,0 +1,162 @@
...
+    OPENCV_FISHEYE = "OPENCV_FISHEYE"
+
+
+CAMERA_MODELS = {
+    "perspective": CameraModel.OPENCV,
+    "fisheye": CameraModel.OPENCV_FISHEYE,
+}
+
+
+def colmap_to_json(cameras_path: Path, images_path: Path, output_dir: Path, camera_model: CameraModel) -> int:
```
akshay-krishnan, line 37: is this code borrowed form nerfstudio? a pointer would be nice!

akshay-krishnan: https://github.com/nerfstudio-project/nerfstudio/blob/1066f8f4330f69363e24984ddf8dc2e7299a23ff/nerfstudio/process_data/colmap_utils.py#L382

gtsfm/utils/prepare_nerfstudio.py
```diff
@@ -0,0 +1,162 @@
...
+        image_path = os.path.join(images_dir, image_filename)
+        image = io.load_image(image_path)
+        resized_image = images.resize_image(image, image_height, image_width)
+        resized_image_path = os.path.join(resized_images_dir, image_filename)
+        io.save_image(resized_image, resized_image_path)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--results_path", required=True, help="Relative or absolute path to /gtsfm/results/")
```
akshay-krishnan, line 132: nit: /gtsfm/results -> GTSFM results?

akshay-krishnan: is this the output_dir we use when running gtsfm?

gtsfm/utils/prepare_nerfstudio.py
```diff
@@ -0,0 +1,162 @@
...
+    parser.add_argument(
```
akshay-krishnan, line 133: a default would be nice here I think - "perspective".

gtsfm/utils/prepare_nerfstudio.py
```diff
@@ -0,0 +1,162 @@
...
+    cameras_file = os.path.join(results_path, "ba_output", "cameras.txt")
+    images_file = os.path.join(results_path, "ba_output", "images.txt")
+    nerfstudio_input_dir = os.path.join(results_path, "nerfstudio-input")
+    if not os.path.exists(nerfstudio_input_dir):
+        os.makedirs(nerfstudio_input_dir)
+    image_width, image_height = colmap_to_json(
+        cameras_file, images_file, nerfstudio_input_dir, camera_model=CAMERA_MODELS[args.camera_model]
+    )
+
+    # Resize images (as GTSfM does) to match intrinsics in transforms.json
```
akshay-krishnan, line 157: the size in transforms.json comes from the images.txt and cameras.txt that we generate in gtsfm, so its basically the downsizing that gtsfm does?

thirdparty/nerfstudio/colmap_utils.py
```diff
@@ -0,0 +1,438 @@
+"""Source: Copied from Nerfstudio.
```
akshay-krishnan, line 1: nit: "forked" from Nerfstudio, which itself forks from COLMAP :)

akshay-krishnan (commented):
Looks almost good. I think we want to make sure this runs with the latest nerfstudio. Also need to add some documentation on how to run nerfsudio with the outputs this script generates.

gtsfm/utils/prepare_nerfstudio.py
travisdriver: Yes

gtsfm/utils/prepare_nerfstudio.py
travisdriver: Yes

gtsfm/utils/io.py
travisdriver: No we just assumed fx == fy, but NerfStudio requires fx and fy

thirdparty/nerfstudio/colmap_utils.py
travisdriver: Do we actually need this file? I thought this file was already checked in

akshay-krishnan:
Seems to work as is, I made some minor changes above too. 

Here are some nerfstudio renders for the mobilebrick dataset (I took these before NeRF training finished, so not the best)

nerf1
nerf2

akshay-krishnan:
cc @dellaert

akshay-krishnan:
Thanks @womackj1 for putting this together!

gtsfm/utils/io.py
johnwlambert: @akshay-krishnan this doesn't correspond to the colmap "Radial" format afaik

https://github.com/colmap/colmap/blob/5c1d58a085920c41cf2d9c892da8041b1fc0c86d/src/colmap/sensor/models.h#L225-L231

gtsfm/utils/io.py
```diff
@@ -301,15 +302,19 @@ def write_cameras(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) ->
...
             k2 = calibration.k2()
+            p1 = 0.0
+            p2 = 0.0
+            # missing tangential distortion coefficients
 
             image_height = images[i].height
             image_width = images[i].width
 
-            f.write(f"{i} {camera_model} {image_width} {image_height} {fx} {u0} {v0} {k1} {k2}\n")
+            f.write(f"{i} {camera_model} {image_width} {image_height} {fx} {fy} {u0} {v0} {k1} {k2} {p1} {p2}\n")
```
johnwlambert, line 317: same note here -- this doesn't correspond to the colmap "Radial" format afaik

https://github.com/colmap/colmap/blob/5c1d58a085920c41cf2d9c892da8041b1fc0c86d/src/colmap/sensor/models.h#L225-L231

johnwlambert: (these are OPENCV models)

https://github.com/colmap/colmap/blob/5c1d58a085920c41cf2d9c892da8041b1fc0c86d/src/colmap/sensor/models.h#L239-L247

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,165 @@
...
+    cameras_path: Path, images_path: Path, output_dir: Path, camera_model: CameraModel
+) -> Tuple[int, int]:
+    """Converts COLMAP's cameras.bin and images.bin to a JSON file.
+    Args:
+        cameras_path: Path to the cameras.txt file.
+        images_path: Path to the images.txt file.
+        output_dir: Path to the output directory.
+        camera_model: Camera model used.
+    Returns:
+        The number of registered images.
```
johnwlambert, line 47: return arg should be Tuple of width/height, right?

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,165 @@
...
+
+    out["frames"] = frames
+
+    with open(os.path.join(output_dir, "transforms.json"), "w", encoding="utf-8") as f:
+        json.dump(out, f, indent=4)
+
+    return cameras[1].width, cameras[1].height
+
+
+def resize_and_save_images_for_nerfstudio(images_dir, image_width, image_height, resized_images_dir):
```
johnwlambert, line 115: nit: type hint : - )

gtsfm/utils/io.py
```diff
@@ -271,8 +275,11 @@ def read_cameras_txt(fpath: str) -> Optional[List[Cal3Bundler]]:
                 float(k2),
             )
             calibrations.append(Cal3Bundler(fx, k1, k2, u0, v0))
-
+            if return_img_dims:
+                img_dims.append((img_h, img_w))
     assert len(calibrations) == num_cams
+    if return_img_dims:
+        return calibrations, img_dims
```
johnwlambert, line 282: nit: Can we always return a tuple of calibrations and img_dims? I've found that a variable number of return arguments can lead to some bugs and I find the signature a bit hard to read.

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+from gtsam import Rot3
+
+from gtsfm.utils import images
+from gtsfm.utils import io as io_utils
+
+
+class CameraModel(Enum):
+    """Enum for camera types."""
+
+    OPENCV = "OPENCV"
```
johnwlambert, line 26: just curious, would it be cleaner to import these from https://github.com/borglab/gtsfm/blob/master/thirdparty/colmap/scripts/python/read_write_model.py#L55-L67?

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def colmap_to_json(data_dir: str, output_dir: str, camera_model: CameraModel) -> Tuple[int, int]:
+    """Converts GTSfM's cameras.txt to to a JSON file, saves it in output_dir.
+    Args:
+        cameras_path: Path to the cameras.txt file.
+        images_path: Path to the images.txt file.
+        output_dir: Path to the output directory.
+        camera_model: Camera model used.
+    Returns:
```
johnwlambert, line 43: nit: newlines between L37 and L38, and between L42 and L43. Can we note in the return args section "Image dimensions as a tuple (width, height)."

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+    if camera_model == CameraModel.OPENCV:
+        out.update(
+            {
+                "k1": float(camera_params.k1()),
+                "k2": float(camera_params.k2()),
+                "p1": 0.0,
+                "p2": 0.0,
+            }
+        )
+    if camera_model == CameraModel.OPENCV_FISHEYE:
```
johnwlambert, line 89: should this be `elif` instead of `if`?

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+def resize_and_save_images_for_nerfstudio(
+    images_dir: str, image_width: int, image_height: int, resized_images_dir: str
+) -> None:
+    """Resizes images to the size used in GTSfM and saves for nerfstudio.
+    Args:
+        images_dir: path to the original images.
+        images_width: int width of resized images
+        images_height: int height of resized images.
+        resized_images_dir: path to directory where the resized images will be saved.
+
```
johnwlambert, line 116: nit: extra newline here

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+    return img_w, img_h
+
+
+def resize_and_save_images_for_nerfstudio(
+    images_dir: str, image_width: int, image_height: int, resized_images_dir: str
+) -> None:
+    """Resizes images to the size used in GTSfM and saves for nerfstudio.
+    Args:
+        images_dir: path to the original images.
```
johnwlambert, line 112: nit: capitalization 
```
    images_dir: Path to the original images.
```

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+    image_width, image_height = colmap_to_json(
+        data_dir, nerfstudio_input_dir, camera_model=CAMERA_MODELS[args.camera_model]
+    )
+
+    # Resize images (as GTSfM does) to match intrinsics in transforms.json
+    gtsfm_images_dir = args.images_dir
+    nerfstudio_images_dir = os.path.join(nerfstudio_input_dir, "images")
+    if not os.path.exists(nerfstudio_images_dir):
+        os.makedirs(nerfstudio_images_dir)
+    resize_and_save_images_for_nerfstudio(gtsfm_images_dir, image_width, image_height, nerfstudio_images_dir)
```
johnwlambert, line 160: just curious, would it be simpler to just scale the intrinsics, instead of resaving the images on disk, especially if there are 100s of them?

scripts/prepare_nerfstudio.py
```diff
@@ -0,0 +1,160 @@
...
+        }
+        frames.append(frame)
+
+    # Only supports single intrinsics datasets.
+    camera_params = calibrations[0]
+    img_h, img_w = img_dims[0]
+
+    out = {
+        "fl_x": float(camera_params.fx()),
+        "fl_y": float(camera_params.fy()),
```
johnwlambert, line 72: would `fy()` exist for a radial camera calibration?

johnwlambert (approved):
Thanks!

scripts/prepare_nerfstudio.py
akshay-krishnan: yeah it is defined in the Cal3 base class, Cal3Bundler uses the same value for fx and fy

scripts/prepare_nerfstudio.py
akshay-krishnan: Nerfstudio expects the image directory to be in the same directory as their json file, so we have to copy anyway. But you make a good point. It might be better to use the full res images to train nerfs, so I made this default. 
Now we scale intrinsics and copy full res images by default, but there is a flag to use gtsfm's resolution if anyone wants to.

johnwlambert: Thanks!

gtsfm/utils/io.py
johnwlambert: Marking as resolved, thanks!

scripts/prepare_nerfstudio.py
johnwlambert: resolved, thanks

scripts/prepare_nerfstudio.py
johnwlambert: resolved, thanks

johnwlambert (approved):
LGTM

btw I think CI fails
```
FAILED tests/utils/test_io_utils.py::TestIoUtils::test_read_cameras_txt - IndexError: list index out of range
```

womackj1:
> Thanks @womackj1 for putting this together!

Thanks, I hope progress with GTSfM is going well!

PR #643, opened 2023-06-16
Increase number of CI cores to 8

travisdriver:


travisdriver:
It works with a couple of caveats (see .github/workflow/benchmark-wildcat.yml).

.github/scripts/execute_single_benchmark_wildcat.sh
```diff
@@ -0,0 +1,64 @@
+#!/bin/bash
+
+# Datasets are stored in ~/sfm_datasets in order to reduce runtimes by not 
+# redownloading each (very large) dataset every CI run. Any new datasets must be
+# downloaded and stored in ~/sfm_datasets before running this action.
+DATASET_PREFIX=/home/tdriver6/sfm_datasets
```
akshay-krishnan, line 6: its not good practice to commit/merge paths. is there no other way of making this work?

.github/scripts/execute_single_benchmark_wildcat.sh
```diff
@@ -0,0 +1,64 @@
...
+echo "Share intrinsics for all images? ${SHARE_INTRINSICS}"
+
+# Setup the command line arg if intrinsics are to be shared
+if [ "$SHARE_INTRINSICS" == "true" ]; then
+  export SHARE_INTRINSICS_ARG="--share_intrinsics"
+else
+  export SHARE_INTRINSICS_ARG=""
+fi
+
+echo "Share intrinsics CLI argument: ${SHARE_INTRINSICS_ARG}"
```
akshay-krishnan, line 33: redundant logging?

.github/scripts/execute_single_benchmark_wildcat.sh
```diff
@@ -0,0 +1,64 @@
...
+    --colmap_files_dirpath $DATASET_PREFIX/$COLMAP_FILES_DIRPATH \
+    --config_name ${CONFIG_NAME}.yaml \
+    --max_frame_lookahead $MAX_FRAME_LOOKAHEAD \
+    --max_resolution ${MAX_RESOLUTION} \
+    ${SHARE_INTRINSICS_ARG} \
+    --correspondence_generator_config_name loftr.yaml \
+    --num_workers 1 \
+    --mvs_off
+
+elif [ "$LOADER_NAME" == "astrovision" ]; then
```
akshay-krishnan, line 57: not supporting 1dsfm?

.github/workflows/benchmark-wildcat.yml
```diff
@@ -0,0 +1,116 @@
...
+      - name: Cache frontend
+        uses: actions/cache@v3
+        env:
+          # Increase this value to reset cache
+          CACHE_NUMBER_FRONTEND: 0
+        with:
+          path: cache
+          key: ${{ matrix.config_dataset_info[0] }}-${{ matrix.config_dataset_info[1] }}-${{ matrix.config_dataset_info[2] }}-${{ matrix.config_dataset_info[6] }}-${{ env.CACHE_NUMBER_FRONTEND }}
+      ##########################################################################
+      # Notes (travisdriver):
```
akshay-krishnan, line 51: could you make this less verbose (and not in first person)?

.github/workflows/benchmark.yml
```diff
@@ -1,6 +1,6 @@
 name: Benchmark GTSFM on select datasets using SIFT and Deep front-ends
 
-on: [pull_request, workflow_dispatch]
```
akshay-krishnan, line 3: why are we removing this?

gtsfm/data_association/dsf_tracks_estimator.py
```diff
@@ -42,17 +42,17 @@ def run(self, matches_dict: Dict[Tuple[int, int], np.ndarray], keypoints_list: L
             list of all valid SfmTrack2d generated by the matches.
         """
         # check to ensure dimensions of coordinates are correct
-        dims_valid = all([kps.coordinates.ndim == 2 for kps in keypoints_list])
-        if not dims_valid:
-            raise Exception("Dimensions for Keypoint coordinates incorrect. Array needs to be 2D")
+        # dims_valid = all([kps.coordinates.ndim == 2 for kps in keypoints_list])
```
akshay-krishnan, line 45: why is this removed?

akshay-krishnan (commented):
why are we checking in the comparison dashboard .zip?

akshay-krishnan:
why are we changing to JointNetVladSequential in the PR?

.github/scripts/execute_single_benchmark_wildcat.sh
travisdriver: This is the actual path on the machine. It won't be different for different user

.github/scripts/execute_single_benchmark_wildcat.sh
travisdriver: good catch, I'll remove it from the other one as well

.github/scripts/execute_single_benchmark_wildcat.sh
travisdriver: I'll add Gendarmenmarkt in another PR

.github/workflows/benchmark-wildcat.yml
travisdriver: I could just add these comments to the PR and take them out of the bash file?

.github/workflows/benchmark.yml
travisdriver: Just for testing to avoid spawning a bunch of actions everytime I commit, will add back before merge

gtsfm/data_association/dsf_tracks_estimator.py
travisdriver: This doesn't allow for no keypoints. This was just a hack, but I could have it ignore None but check the others

travisdriver:
> why are we changing to JointNetVladSequential in this PR?

Just testing on Skydio, will move everything to a separate config if I find something that works well

duyanwei:
> It works with a couple of caveats (see .github/workflow/benchmark-wildcat.yml).
> 
> Next I will add Gendarmenmarkt. @duyanwei Could you post the settings you've been using in your trials?

Sure, I will post it once the exif PR is finalized.

akshay-krishnan:
Does this PR move all CI tasks to the self-hosted runner, even for these small dataset benchmarks?

travisdriver:
> Does this PR move all CI tasks to the self-hosted runner, even for these small dataset benchmarks?

No, only Skydio-501 for now. Then some of the 1DSfM datasets later.

.github/scripts/execute_single_benchmark_wildcat.sh
```diff
@@ -0,0 +1,62 @@
...
+
+# Setup the command line arg if intrinsics are to be shared
+if [ "$SHARE_INTRINSICS" == "true" ]; then
+  export SHARE_INTRINSICS_ARG="--share_intrinsics"
+else
+  export SHARE_INTRINSICS_ARG=""
+fi
+
+# Run GTSFM on the dataset.
+if [ "$LOADER_NAME" == "olsson-loader" ]; then
```
akshay-krishnan, line 34: so if this is only for skydio-501, why are do we need the other loaders here?

.github/workflows/benchmark-wildcat.yml
```diff
@@ -0,0 +1,116 @@
+name: Benchmark GTSFM on select datasets on the wildcat machine
+
+################################################################################
+# Please only run this action after making sure that the shoter benchmarks
```
akshay-krishnan, line 4: nit: typo shorter

.github/workflows/benchmark-wildcat.yml
```diff
@@ -0,0 +1,116 @@
...
+# (defined in benchmark.yml) all run correctly and show no major regressions.
+#
+# Please be conscious of multiple queued actions, as they run sychronously on
```
akshay-krishnan, line 7: why do we have these instructions here if the workflow dispatches automatically on pull requests?

.github/workflows/benchmark-wildcat.yml
akshay-krishnan: Yes, could you replace lines 53-65 with `TODO: setup micromamba/miniconda to update env automatically.`

.github/workflows/benchmark-wildcat.yml
```diff
@@ -0,0 +1,116 @@
...
+      #     init-shell: >-
+      #       bash
+      #     cache-environment: true
+      #     post-cleanup: "all"
+      - name: Download third-party weights
+        run: |
+          ./download_model_weights.sh
+      - name: Prepare dataset
+        run: |
+          echo "Datasets are stored in ~/sfm_datasets in order to reduce runtimes by not"
```
akshay-krishnan, line 86: I guess we need the full path here too?

.github/workflows/benchmark.yml
akshay-krishnan: approval pending this important change.

gtsfm/data_association/dsf_tracks_estimator.py
akshay-krishnan: did you come across a case where we had no keypoints in an image? how is that handled? are Keypoints None? Or is it an empty matrix? 

`all([kps.coordinates.ndim == 2 for kps in keypoints_list if kps is not None])`

gtsfm/two_view_estimator.py
```diff
@@ -293,7 +293,7 @@ def run_2view(
         )
 
         # Optionally, do two-view bundle adjustment
-        if self._bundle_adjust_2view:
+        if self._bundle_adjust_2view and len(pre_ba_v_corr_idxs) >= self.processor._min_num_inliers_est_model:
```
akshay-krishnan, line 296: I'd prefer this to a be logged (at least to debug) rather than a silent skip: 
```
if len(pre_ba_v_corr_idxs) < self.processor._min_num_inliers_est_model: 
  logger.info("Insufficient verified correspondences for 2-view BA, skipping")
```

thirdparty/hloc/download_netvlad.py
```diff
@@ -0,0 +1,25 @@
+import subprocess
+import logging
```
akshay-krishnan, line 2: import order, please fix

akshay-krishnan (approved):
please move the config changes to a new one, if these are specific to skydio-501. Approved modulo pending comments.

.github/scripts/execute_single_benchmark_wildcat.sh
travisdriver: I plan to add a 1DSfM and AstroVision dataset next.

gtsfm/data_association/dsf_tracks_estimator.py
travisdriver: Yes, I had a couple cases when using LoFTR. Keypoints.coordinates is None. I'll add this.

.github/scripts/execute_single_benchmark_wildcat.sh
travisdriver: I don't think there is without restructuring how we run the benchmarks. Either way, I think this is fine for now.

travisdriver:
Some notes on running the self-hosted runner with micromamba/miniconda to update the environment for each CI run:

A persistent micromamba `gtsfm-v1` environment is maintained on the wildcat machine. Therefore, the environment will need to be updated manually if any changes are made to the dependencies.

I was not able to get the `setup-miniconda` or `setup-micromamba` to work. Specifically, `setup-miniconda` always hung on "Solving environment..." (maybe I was being impatient), and `setup-micromamba` kept throwing an error because it tried to reinstall micromamba for each CI run, but there was already an existing installation (because the machine persists unlike the GitHub-hosted runners).

I think the current solution is fine for now. I wasn't able to find much on how others have set up self-hosted runners, so this might even be the only way to do it.

gtsfm/two_view_estimator.py
travisdriver: Why do we need to log this? Either way it will just get thrown out by the InlierSupportProcessor. This just avoids the extra BA step on something that would otherwise be thrown away.

thirdparty/hloc/download_netvlad.py
travisdriver: done

.github/workflows/benchmark-wildcat.yml
travisdriver: done

.github/workflows/benchmark-wildcat.yml
travisdriver: fixed

gtsfm/data_association/dsf_tracks_estimator.py
travisdriver: added

.github/workflows/benchmark-wildcat.yml
travisdriver: done

.github/scripts/execute_single_benchmark_wildcat.sh
travisdriver: done

duyanwei:
Hi @travisdriver, the exif PR has been merged. And here is the command I'm using to run 1DSFM dataset:
```bash
python gtsfm/runner/run_scene_optimizer_1dsfm.py --config_name deep_front_end.yaml --dataset_root /data/1dsfm/Gendarmenmarkt --max_frame_lookahead 500 --num_workers 1 --threads_per_worker 8 --max_resolution 640 --mvs_off
```

travisdriver:
I decided to go with the `min_track_length = 3` solution since the poses were more accurate and the point cloud was less noisy, albiet with 7 less reconstructed cameras (i.e., 489 as opposed to 496). We can change it later on if we would like.

Screen Shot 2023-06-20 at 2 49 22 PM

Screen Shot 2023-06-20 at 2 50 05 PM

PR #641, opened 2023-06-15
Minimum triangulation angle

travisdriver:
Rebases #581.

PR #644, opened 2023-06-17
Minimum Triangulation Angle

travisdriver:
Rebases #641.

### DIFF
--- a/gtsfm/data_association/point3d_initializer.py
+++ b/gtsfm/data_association/point3d_initializer.py
@@ -81,7 +81,7 @@ class TriangulationOptions(NamedTuple):
 
     mode: TriangulationSamplingMode
     reproj_error_threshold: float = np.inf  # defaults to no filtering unless specified
-    min_triangulation_angle: float = 1.0
+    min_triangulation_angle: float = 0.0
 
     # RANSAC parameters
     min_inlier_ratio: float = 0.1
--- a/gtsfm/data_association/point3d_initializer.py
+++ b/gtsfm/data_association/point3d_initializer.py
@@ -81,7 +81,7 @@ class TriangulationOptions(NamedTuple):
 
     mode: TriangulationSamplingMode
     reproj_error_threshold: float = np.inf  # defaults to no filtering unless specified
-    min_triangulation_angle: float = 3.0
+    min_triangulation_angle: float = 1.0
 
     # RANSAC parameters
     min_inlier_ratio: float = 0.1
--- a/gtsfm/data_association/point3d_initializer.py
+++ b/gtsfm/data_association/point3d_initializer.py
@@ -42,7 +42,7 @@ class TriangulationExitCode(Enum):
     INLIERS_UNDERCONSTRAINED = 2  # insufficent number of inlier measurements
     POSES_UNDERCONSTRAINED = 3  # insufficent number of estimated camera poses
     EXCEEDS_REPROJ_THRESH = 4  # estimated 3d point exceeds reprojection threshold
-    LOW_TRIANGULATION_ANGLE = 5  # estimated 3d point exceeds reprojection threshold
+    LOW_TRIANGULATION_ANGLE = 5  # maximum triangulation angle lower than threshold
 
 
 class TriangulationSamplingMode(str, Enum):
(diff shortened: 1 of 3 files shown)


ayushbaid (commented):
Any metrics for this change?

travisdriver:
> Any metrics for this change?

Makes everything much much worse, but this is just on the small datasets. I suspect it will improve performance for the large datasets.

travisdriver:
Even with a 1 degree threshold there are some significant regressions... I'll set the threshold to 0 for now and do more testing on the larger datasets once we merge #643. 

newplot (7)

[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/11795590/visual_comparison_dashboard.html.zip)

gtsfm/data_association/point3d_initializer.py
```diff
@@ -67,8 +69,9 @@ class TriangulationOptions(NamedTuple):
     See the following slides for a derivation of the #req'd samples: http://www.cse.psu.edu/~rtc12/CSE486/lecture15.pdf
 
     Args:
-        reproj_error_threshold: the maximum reprojection error allowed.
         mode: triangulation mode, which dictates whether or not to use robust estimation.
+        reproj_error_threshold: the maximum reprojection error allowed.
+        min_triangulation_angle: the minimum angle (in degrees) subtended at the triangulated point from 2 cameras.
```
akshay-krishnan, line 74: mention that this is a threshold, and that tracks for which the max angle at any two cameras < threshold will be rejected.

gtsfm/utils/tracks.py
```diff
@@ -74,3 +77,28 @@ def classify_tracks3d_with_gt_cameras(
         tracks_2d.append(SfmTrack2d(measurements))
 
     return classify_tracks2d_with_gt_cameras(tracks_2d, cameras_gt, reproj_error_thresh_px)
+
+
+def get_triangulation_angle(track3d: SfmTrack, cameras: Dict[int, gtsfm_types.CAMERA_TYPE]) -> float:
```
akshay-krishnan, line 82: I think we want the `max` reflected in the name somehow `get_max_triangulation_angle` ?

gtsfm/utils/tracks.py
```diff
@@ -74,3 +77,28 @@ def classify_tracks3d_with_gt_cameras(
...
+    """Get the angle subtended by the cameras at the 3D landmark of the track.
```
akshay-krishnan, line 83: add "in degrees", ideally might be good to have that in the name as well.

gtsfm/data_association/point3d_initializer.py
```diff
@@ -41,6 +42,7 @@ class TriangulationExitCode(Enum):
     INLIERS_UNDERCONSTRAINED = 2  # insufficent number of inlier measurements
     POSES_UNDERCONSTRAINED = 3  # insufficent number of estimated camera poses
     EXCEEDS_REPROJ_THRESH = 4  # estimated 3d point exceeds reprojection threshold
+    LOW_TRIANGULATION_ANGLE = 5  # estimated 3d point exceeds reprojection threshold
```
akshay-krishnan, line 45: update the comment after copying it? :)

akshay-krishnan (approved):

gtsfm/data_association/point3d_initializer.py
travisdriver: Oopsies

akshay-krishnan:
> Even with a 1 degree threshold there are some significant regressions... I'll set the threshold to 0 for now and do more testing on the larger datasets once we merge #643.

It might help to add a "triangulation_angle" metric to point3d initializer, just like we have track length. I suspect this might be removing a lot of the length 2 tracks that we found to be useful?

gtsfm/utils/tracks.py
```diff
@@ -74,3 +77,28 @@ def classify_tracks3d_with_gt_cameras(
...
+    camera_ind: List[int] = []
+    for k in range(track3d.numberMeasurements()):
+        i, _ = track3d.measurement(k)
+        camera_ind.append(i)
+
+    angles: List[float] = []
+    for i1, i2 in itertools.combinations(camera_ind, 2):
+        ray1 = track3d.point3() - cameras[i1].pose().translation()
+        ray2 = track3d.point3() - cameras[i2].pose().translation()
+        angles.append(np.rad2deg(np.arccos(np.dot(ray1, ray2) / (np.linalg.norm(ray1) * np.linalg.norm(ray2)))))
```
johnwlambert, line 102: I believe we have an existing function that shares similar functionality -- https://github.com/borglab/gtsfm/blob/bf58ea2d48711d7047ffb195fa053a1b81fe270f/gtsfm/densify/mvs_utils.py#L53

Could we re-use that?

PR #646, opened 2023-06-17
Remove default-factor-initialized focal length

duyanwei:
#### Fix 
Remove default-factor-initialized focal length value.

#### How
- Remove the default focal length value from the original API `get_intrinsics_from_exif()`.
- Move it to a new API `get_intrinsics()`. (A better method name is needed.)
- OneDSFM loader default only reads images with exif.

### DIFF
--- a/gtsfm/loader/one_d_sfm_loader.py
+++ b/gtsfm/loader/one_d_sfm_loader.py
@@ -17,9 +17,6 @@ from gtsfm.loader.loader_base import LoaderBase
 
 logger = logger_utils.get_logger()
 
-# Focal length is initialized to 1.2 * largest dimension of image if EXIF data is not available.
-NO_EXIF_DEFAULT_FOCAL_LENGTH_FACTOR = 1.2
-
 
 class OneDSFMLoader(LoaderBase):
     """Loader for datasets used in 1DSFM and Colmap papers.
@@ -41,7 +38,7 @@ class OneDSFMLoader(LoaderBase):
         folder: str,
         image_extension: str = "jpg",
         max_resolution: int = 640,
-        enable_no_exif: bool = False,
+        default_focal_length_factor: float = 0.0,
     ) -> None:
         """Initializes to load from a specified folder on disk.
 
@@ -52,15 +49,16 @@ class OneDSFMLoader(LoaderBase):
                 the smaller of the height/width of the image. e.g. for 1080p (1920 x 1080),
                 max_resolution would be 1080. If the image resolution max(height, width) is
                 greater than the max_resolution, it will be downsampled to match the max_resolution.
-            enable_no_exif: flag to whether to read images without exif.
+            default_focal_length_factor: focal length is initialized to default_focal_length_factor * largest dimension
+            of image if EXIF data is not available. Non-positive value means exif required.
         """
         super().__init__(max_resolution=max_resolution)
-        self._enable_no_exif = enable_no_exif
+        self._default_focal_length_factor = default_focal_length_factor
 
         # Fetch all the file names in /images folder.
         search_path = os.path.join(folder, "images", f"*.{image_extension}")
 
-        if not self._enable_no_exif:
+        if self._default_focal_length_factor > 0.0:
             self._image_paths = glob.glob(search_path)
         else:
             (self._image_paths, num_all_imgs) = self.get_images_with_exif(search_path)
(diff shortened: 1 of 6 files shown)


gtsfm/loader/one_d_sfm_loader.py
```diff
@@ -51,18 +52,45 @@ def __init__(
...
+            self._image_paths = glob.glob(search_path)
+        else:
+            (self._image_paths, num_all_imgs) = self.get_images_with_exif(search_path)
+            logger.info("Read %d images with exif out of %d in total.", len(self._image_paths), num_all_imgs)
 
+        self._num_imgs = len(self._image_paths)
         if self._num_imgs == 0:
             raise RuntimeError(f"Loader could not find any images with the specified file extension in {search_path}")
 
+    def get_images_with_exif(self, search_path: str) -> Tuple[List[str], int]:
```
travisdriver, line 73: I think this makes more sense in the base loader (it's not necessarily specific to 1dsfm).

gtsfm/loader/one_d_sfm_loader.py
```diff
@@ -101,9 +129,14 @@ def get_camera_intrinsics_full_res(self, index: int) -> Optional[Cal3Bundler]:
...
-        intrinsics = io_utils.load_image(self._image_paths[index]).get_intrinsics_from_exif(
-            default_focal_length_factor=NO_EXIF_DEFAULT_FOCAL_LENGTH_FACTOR
-        )
+        # TODO(yanwei) It is probably fine to only use the second method, as
+        # the images have been filtered before in consturctor.
+        if not self._enable_no_exif:
+            intrinsics = io_utils.load_image(self._image_paths[index]).get_intrinsics_from_exif()
+        else:
+            intrinsics = io_utils.load_image(self._image_paths[index]).get_intrinsics(
+                default_focal_length_factor=NO_EXIF_DEFAULT_FOCAL_LENGTH_FACTOR
```
travisdriver, line 138: I agree with your TODO that I don't think its necessary to check the exif flag again

gtsfm/loader/one_d_sfm_loader.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         folder: str,
         image_extension: str = "jpg",
         max_resolution: int = 640,
+        enable_no_exif: bool = False,
```
travisdriver, line 44: Instead of the `enable_no_exif` and also a globally defined default focal length, I think it would make sense to have the default focal length as an attribute, and if it's None, then that should be interpreted as requiring EXIF.

gtsfm/loader/one_d_sfm_loader.py
duyanwei: Done.

gtsfm/loader/one_d_sfm_loader.py
duyanwei: Done.

gtsfm/loader/one_d_sfm_loader.py
duyanwei: Done.
I kept it as float type and only non-positive value requires exif.  In my opinion it makes no sense to pass None to `get_intrinsics` in order to process images with no exif. Please help review when you have time. Thanks.

travisdriver: I would prefer None over a negative number. It's more explicit, since focal lengths can actually be negative.

tests/common/test_image.py
```diff
@@ -91,7 +91,7 @@ def test_get_intrinsics_from_exif_focal_length(self, mock_init, mock_lookup):
             "FocalPlaneResolutionUnit": 2,
         }
         image = Image(np.random.randint(low=0, high=255, size=(im_h, im_w, 3)), exif_data)
-        computed_intrinsics = image.get_intrinsics_from_exif(default_focal_length_factor=DEFAULT_FOCAL_LENGTH_FACTOR)
+        computed_intrinsics = image.get_intrinsics(default_focal_length_factor=DEFAULT_FOCAL_LENGTH_FACTOR)
```
akshay-krishnan, line 94: it looks like exif_data is being passed to the image class, so arent these tests checking for it? do we need focal length factor then?

akshay-krishnan (approved):

gtsfm/loader/one_d_sfm_loader.py
akshay-krishnan: I actually preferred `enable_no_exif` flag. A user is not going to know what the preferred value for the focal length factor is. It makes sense to have a reasonable default for it. And another flag to decide whether or not to use it.

akshay-krishnan: without `enable_no_exif`, please add some documentation on what values are recommended.

travisdriver: I'm fine either way

duyanwei: Thanks for the comments. Yeah, using a negative value to do the check is definitely a bad idea. I'll fix it.

tests/common/test_image.py
duyanwei: This case tests the intrinsics are from exif data when it is available. It does not matter if the factor is passed or not. Is that what you are asking?

akshay-krishnan: yes, so why not just use the get_intrinsics_from_exif function without the default?

akshay-krishnan: and maybe add another unit test where we dont pass exif data and call get_intrinsics with a default focal length factor?

duyanwei: There was already one at line20.
Anyway, I will re-arrange the test cases, making sure all the cases are covered.

duyanwei: Hello @akshay-krishnan , I've refined the test cases a bit. Now the first five tests `get_intrinsics_from_exif()` method: `no exif`, `no tags` and reading focal length from three different tags.  The following two tests `get_intrinsics()` method: `no exif` and `with exif`.

PR #647, opened 2023-06-20
Fix to avoid dividing by zero when computing post_ba_inlier_ratio

duyanwei:


travisdriver:
This is already addressed in #643

PR #651, opened 2023-06-29
Refactor Triangulation Reprojection Error

travisdriver:
When we triangulate with RANSAC, we choose the inliers based on the reprojection errors from a track triangulated with _two cameras_. Then, we return the inliers, triangulate again with _all of the cameras_, and recompute all of the reprojection errors.

This means that the recomputed reprojection errors could be _larger_ than those used to choose the best inlier track, and then it can be rejected with the EXCEEDS_REPROJ_THRESHOLD exit flag. I don't think this is consistent.

Therefore, we should either retriangulate the point during RANSAC with all of the identified inlier cameras so that the result is consistent after exiting RANSAC, or we should not recompute the reprojection errors after RANSAC. I think the latter makes more sense.

### DIFF
--- a/gtsfm/data_association/point3d_initializer.py
+++ b/gtsfm/data_association/point3d_initializer.py
@@ -132,7 +132,7 @@ class Point3dInitializer:
             CameraSetCal3Bundler if isinstance(sample_camera, PinholeCameraCal3Bundler) else CameraSetCal3Fisheye
         )
 
-    def execute_ransac_variant(self, track_2d: SfmTrack2d) -> np.ndarray:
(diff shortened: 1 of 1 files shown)


travisdriver:
Not a huge increase in performance, but I think that's expected since our reprojection error thresholds are relatively large. Those can be tuned in another PR. Also running on the Wildcat CI to see how it affects Skydio-501.

[visual_comparison_dashboard.html.zip](https://github.com/borglab/gtsfm/files/11919708/visual_comparison_dashboard.html.zip)

newplot (8)

travisdriver:
newplot (24)

PR #653, opened 2023-07-04
Add tanks and temples dataset loader

johnwlambert:
Add image and mesh loader for scenes from the Tanks & Temple dataset.


gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
```diff
@@ -0,0 +1,105 @@
...
+            colmap_ply_fpath=colmap_ply_fpath,
+        )
+
+        
+        mesh = loader.reconstruct_mesh()
+        # TODO(jolambert): File Open3d bug to add pickle support for TriangleMesh.
+        open3d_mesh_path = tempfile.NamedTemporaryFile(suffix='.obj').name
+        open3d.io.write_triangle_mesh(filename=open3d_mesh_path, mesh=mesh)
+
+        loader_future = client.scatter(loader, broadcast=False)
```
ayushbaid, line 84: We could just avoid futures for this file I guess?

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
ayushbaid, line 1: I am wondering if we can just generate the matches apriori and use something like a cache to just load the matches from the disk.

travisdriver: I was also thinking this would be the best approach, as this is more for debugging and will likely only be used by us.

gtsfm/configs/synthetic_front_end.yaml
```diff
@@ -0,0 +1,81 @@
+# Default deep front-end configuration.
```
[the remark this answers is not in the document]
johnwlambert: Updated now, thanks for catching that.

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
johnwlambert: I see -- if we wish to use the Tanks and Temples loader to get GT poses to measure pose error, what's the difference between computing matches offline vs. online?

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
johnwlambert: Hmm wouldn't we lose all parallelization benefits then?

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
```diff
@@ -0,0 +1,107 @@
+"""Correspondence generator that utilizes direct matching of keypoints across an image pair, without descriptors.
```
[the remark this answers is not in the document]
johnwlambert: Updated, thanks for catching that.

gtsfm/configs/synthetic_front_end.yaml
```diff
@@ -0,0 +1,97 @@
...
+  pose_angular_error_thresh: 5 # degrees
+
+  # retriever:
+  #   _target_: gtsfm.retriever.sequential_retriever.SequentialRetriever
+  #   max_frame_lookahead: 4
+
+  # retriever:
+  #   _target_: gtsfm.retriever.netvlad_retriever.NetVLADRetriever
+  #   num_matched: 50
+  #   min_score: 0.3
```
travisdriver, line 17: Remove

gtsfm/configs/synthetic_front_end.yaml
```diff
@@ -0,0 +1,97 @@
...
+    min_score: 0.2
+    max_frame_lookahead: 3
+
+
+  correspondence_generator:
+    _target_: gtsfm.frontend.correspondence_generator.synthetic_correspondence_generator.SyntheticCorrespondenceGenerator
+    #dataset_root: /Users/johnlambert/Downloads/Tanks_and_Temples_Barn_410
+    dataset_root: /home/runner/work/gtsfm/gtsfm/Tanks_and_Temples_Barn_410
+    # dataset_root: /home/tdriver6/sfm_datasets/Tanks_and_Temples_Barn_410
+    scene_name: Barn
```
travisdriver, line 31: Remove comments lines, also what's the role of this dataset_root variable? Is this new?

gtsfm/runner/run_scene_optimizer_synthetic_tanks_and_temples.py
```diff
@@ -0,0 +1,62 @@
...
+from gtsfm.runner.gtsfm_runner_base import GtsfmRunnerBase
+
+logger = logger_utils.get_logger()
+
+
+_TANKS_AND_TEMPLES_RESOLUTION_PX = 1080
+
+
+class GtsfmRunnerSyntheticTanksAndTemplesLoader(GtsfmRunnerBase):
+    tag = "GTSFM with LiDAR scans, COLMAP camera poses, and image names stored in Tanks and Temples format"
```
travisdriver, line 20: Why does it have to just be tanks and temples? Would this run with any dataset with a GT mesh?

tests/data/tanks_and_temples_barn/Barn_COLMAP_SfM.log
travisdriver, line 1: Do we need to check in this file?

tests/loader/test_tanks_and_temples_loader.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+class TanksAndTemplesLoaderTest(unittest.TestCase):
+    def setUp(self) -> None:
+        scene_name = "Barn"  # 'Truck'
+
+        # Uncomment lines below to run locally.
+        # _TEST_DATA_ROOT = Path('/Users/johnlambert/Downloads/Tanks_and_Temples_Barn_410')
+        # lidar_ply_fpath = _TEST_DATA_ROOT / f'{scene_name}.ply'
+        # colmap_ply_fpath = _TEST_DATA_ROOT / f'{scene_name}_COLMAP.ply'
```
travisdriver, line 31: Why is there a difference between how it works locally and on the CI? Is this done in any other tests?

tests/loader/test_tanks_and_temples_loader.py
```diff
@@ -0,0 +1,201 @@
...
+        #         DIRECTION_ANGULAR_ERROR_DEG_THRESHOLD,
+        #     )
+
+    def test_mesh(self) -> None:
+        # Skip this test in the CI, and only uncomment it to run it locally, since it requires PLY.
+        return
+        mesh = self.loader.reconstruct_mesh()
+        num_sampled_3d_points = 20000
+        pcd = mesh.sample_points_uniformly(number_of_points=num_sampled_3d_points)
+        # pcd = mesh.sample_points_poisson_disk(number_of_points=num_sampled_3d_points, pcl=pcd)
```
travisdriver, line 159: remove

.github/scripts/execute_single_benchmark_self_hosted.sh
```diff
@@ -17,8 +17,24 @@ SHARE_INTRINSICS=$7
...
 
+## Download the Tanks & Temples dataset. ################
+WGET_URL1=https://github.com/johnwlambert/gtsfm-datasets-mirror/releases/download/tanks-and-temples-barn/Tanks_and_Temples_Barn_410.zip
+ZIP_FNAME=Tanks_and_Temples_Barn_410.zip
+echo "Downloading ${ZIP_FNAME} with WGET"
+wget -O $DATASET_PREFIX/Tanks_and_Temples_Barn_410.zip $WGET_URL1
+unzip -qq $DATASET_PREFIX/$ZIP_FNAME -d $DATASET_PREFIX
+pwd $DATASET_PREFIX
+ls -ltrh $DATASET_PREFIX
+ls -ltrh $DATASET_PREFIX/$DATASET_ROOT
```
ayushbaid, line 33: Why do we ls?

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
```diff
@@ -0,0 +1,119 @@
...
+        dataset_root = self._dataset_root
+        scene_name = self._scene_name
+
+        img_dir = f"{dataset_root}/{scene_name}"
+        poses_fpath = f"{dataset_root}/{scene_name}_COLMAP_SfM.log"
+        lidar_ply_fpath = f"{dataset_root}/{scene_name}.ply"
+        colmap_ply_fpath = f"{dataset_root}/{scene_name}_COLMAP.ply"
+        ply_alignment_fpath = f"{dataset_root}/{scene_name}_trans.txt"
+        bounding_polyhedron_json_fpath = f"{dataset_root}/{scene_name}.json"
+        loader = TanksAndTemplesLoader(
```
ayushbaid, line 69: this will have to be pinned to the machine with the input worker, right? Hence I was suggesting we create the correspondences on disk beforehand. But fine with it given the time constraint.

travisdriver:
The Flake8 checks are failing

gtsfm/two_view_estimator.py
```diff
@@ -228,6 +228,7 @@ def __get_2view_report_from_results(
             R_error_deg, U_error_deg = compute_relative_pose_metrics(
                 i2Ri1_computed, i2Ui1_computed, gt_camera_i1.pose(), gt_camera_i2.pose()
             )
+            print(f"Two view estimator: [R error deg] {R_error_deg}, [U error deg] {U_error_deg}")
```
ayushbaid, line 231: Why add this?

ayushbaid (approved):

tests/data/tanks_and_temples_barn/Barn_COLMAP_SfM.log
johnwlambert: Yes I think we should included it, this is a small ground truth file used for the unit tests. Tanks and Temples uses this file format to provide GT (i.e. COLMAP) poses.

gtsfm/two_view_estimator.py
johnwlambert: Removed now, thanks.

.github/scripts/execute_single_benchmark_self_hosted.sh
johnwlambert: Removed, now, thanks.

tests/loader/test_tanks_and_temples_loader.py
johnwlambert: Good point, removed now. This really was more of an "integration test", so I moved it to a visualization script for Tanks and Temples (`gtsfm/visualization/visualize_scene_tanks_and_temples.py`).

tests/data/tanks_and_temples_barn/Barn_COLMAP_SfM.log
johnwlambert: Resolving, since this file was checked in as minimal test data in https://github.com/borglab/gtsfm/pull/750.

gtsfm/configs/synthetic_front_end.yaml
johnwlambert: Since the `CorrespondenceGeneratorBase` class and its child classes don't have access to the `loader`, to avoid modifying all the other APIs, for this special case we need to instantiate a loader, so we need data paths to do so (to access loader functionality like creating meshes etc.)

gtsfm/configs/synthetic_front_end.yaml
```diff
@@ -0,0 +1,102 @@
...
+      _target_: gtsfm.frontend.cacher.global_descriptor_cacher.GlobalDescriptorCacher
+      global_descriptor_obj:
+        _target_: gtsfm.frontend.global_descriptor.netvlad_global_descriptor.NetVLADGlobalDescriptor
+    retriever:
+      _target_: gtsfm.retriever.joint_netvlad_sequential_retriever.JointNetVLADSequentialRetriever
+      num_matched: 2
+      min_score: 0.2
+      max_frame_lookahead: 3
+
+    # retriever:
```
ayushbaid, line 22: Remove?

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
johnwlambert: @ayushbaid @travisdriver wanted to revisit this -- I would prefer to keep as-is and make this code self-contained so that we only need to run one command, instead of having to make two new scripts (one to generate all the correspondences beforehand, and then making a new one to accept saved correspondences)

gtsfm/configs/synthetic_front_end.yaml
```diff
@@ -0,0 +1,102 @@
...
+    #   _target_: gtsfm.retriever.netvlad_retriever.NetVLADRetriever
+    #   num_matched: 50
+    #   min_score: 0.3
+
+  correspondence_generator:
+    _target_: gtsfm.frontend.correspondence_generator.synthetic_correspondence_generator.SyntheticCorrespondenceGenerator
+    #dataset_root: /Users/johnlambert/Downloads/Tanks_and_Temples_Barn_410
+    #dataset_root: /usr/local/gtsfm-data/Tanks_and_Temples_Barn_410
+    dataset_root: /home/runner/work/gtsfm/gtsfm/Tanks_and_Temples_Barn_410 # Path for CI.
```
ayushbaid, line 35: I'd prefer this being loaded from the dataset root passed from the terminal, but its not an easy fix. So we should be fine with it.

gtsfm/frontend/correspondence_generator/synthetic_correspondence_generator.py
ayushbaid: Yes, but since we are just loading from disk, we would not lose much in terms of time I think

ayushbaid (approved):

gtsfm/runner/run_scene_optimizer_synthetic_tanks_and_temples.py
johnwlambert: It could be for any dataset w/ a GT mesh. I've added a TODO for that.

gtsfm/configs/synthetic_front_end.yaml
johnwlambert: Removed, thanks.

PR #658, opened 2023-07-16
Add correspondence augmenter

ayushbaid:
Correspondence augmenter fits in between the two-view estimator and the multi-view estimator. It uses the pose estimates from the 2-view estimator and runs a different correspondence generator to augmend correspondences

### DIFF
--- a/gtsfm/frontend/correspondence_augmenter.py
+++ b/gtsfm/frontend/correspondence_augmenter.py
@@ -137,10 +137,13 @@ class CorrespondenceAugmenter:
         results: Dict[Tuple[int, int], TWO_VIEW_OUTPUT] = {}
         for (i1, i2), two_view_output in two_view_outputs.items():
             new_corr_idxs = verified_corr_idxs_dict[(i1, i2)]
-            new_corr_idxs[:, 0] += keypoint_idx_offsets[i1]
-            new_corr_idxs[:, 1] += keypoint_idx_offsets[i2]
+            if len(new_corr_idxs) > 0:
+                new_corr_idxs[:, 0] += keypoint_idx_offsets[i1]
+                new_corr_idxs[:, 1] += keypoint_idx_offsets[i2]
 
-            augmented_v_corr_idxs = np.append(two_view_output[2], new_corr_idxs, axis=0)
+                augmented_v_corr_idxs = np.append(two_view_output[2], new_corr_idxs, axis=0)
+            else:
+                augmented_v_corr_idxs = two_view_output[2]
 
             two_view_reports = two_view_output[3]
             two_view_report_before_augmentation = two_view_reports[TWO_VIEW_REPORT_TAG.POST_ISP]
--- a/gtsfm/frontend/correspondence_augmenter.py
+++ b/gtsfm/frontend/correspondence_augmenter.py
@@ -83,6 +83,9 @@ class CorrespondenceAugmenter:
             triangulation_options: TriangulationOptions,
             ba_optimizer: BundleAdjustmentOptimizer,
         ) -> np.ndarray:
+            if i2Ri1 is None or i2Ui1 is None:
+                return np.array([], dtype=np.uint64)
+
             i2Ti1_prior = PosePrior(
                 value=Pose3(i2Ri1, i2Ui1.point3()),
                 covariance=RELATIVE_POSE_PRIOR_SIGMA,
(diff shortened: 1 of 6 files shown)


johnwlambert:
Any preliminary results if this helps, eg. lightglue + loftr?

gtsfm/frontend/two_view_ba_utils.py
```diff
@@ -0,0 +1,131 @@
...
+        )
+        track, _, _ = point3d_initializer.triangulate(track2d)
+        if track is not None:
+            triangulated_indices.append(j)
+            triangulated_tracks.append(track)
+
+    return triangulated_indices, triangulated_tracks
+
+
+def bundle_adjust(
```
johnwlambert, line 59: Thanks for the PR @ayushbaid. QQ -- don't we already have code for 2-view BA here? https://github.com/borglab/gtsfm/blob/master/gtsfm/two_view_estimator.py#L123

Didn't quite follow the difference with the code here

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -252,6 +258,20 @@ def run(self) -> GtsfmData:
                 gt_scene_mesh=self.loader.get_gt_scene_trimesh(),
             )
 
+            if self.scene_optimizer.correspondence_augmenter:
```
johnwlambert, line 261: @ayushbaid would it be cleaner to get the two set of correspondences, and merge their keypoints with the KeypointAggregator, and then run the two-view estimator once? Don't we only want to estimate (R,t) once with all the correspondences?

gtsfm/frontend/two_view_ba_utils.py
johnwlambert: Oh I think I see now, we are just moving the other code to this new file.

gtsfm/frontend/correspondence_augmenter.py
```diff
@@ -0,0 +1,179 @@
...
+        for keypoints_original, keypoints_new in zip(keypoints_list, keypoints_to_augment_list):
+            keypoints_augmented, offset = self._augment_keypoints(
+                keypoints_original=keypoints_original, keypoints_new=keypoints_new
+            )
+            augmented_keypoints_list.append(keypoints_augmented)
+            keypoint_idx_offsets.append(offset)
+
+        # Merge the corr_idxs
+        results: Dict[Tuple[int, int], TWO_VIEW_OUTPUT] = {}
+        for (i1, i2), two_view_output in two_view_outputs.items():
```
johnwlambert, line 138: Could we share the code block here with the code here? https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/correspondence_generator/keypoint_aggregator/keypoint_aggregator_unique.py#L43-L56 Is it identical?

gtsfm/frontend/correspondence_augmenter.py
```diff
@@ -0,0 +1,179 @@
...
+        gt_scene_mesh: Optional[Any],
+    ) -> Tuple[List[Keypoints], Dict[Tuple[int, int], TWO_VIEW_OUTPUT]]:
+        (
+            keypoints_to_augment_list,
+            putative_corr_idxs_to_augment,
+        ) = self._correspondence_generator.generate_correspondences(
+            client=client, images=images, image_pairs=list(two_view_outputs.keys())
+        )
+
+        def verify_putative_correspondences(
```
johnwlambert, line 75: Just curious, have you checked if we get similar accuracy from 

1. Merge unverified, and estimate 2-view w/ verification once.
2. Match and verify separately, then merge? 

My understanding was that https://www.kaggle.com/competitions/image-matching-challenge-2023/discussion/417407 did (1) not (2)
@travisdriver fyi

gtsfm/frontend/two_view_ba_utils.py
ayushbaid: Yes. It helps me reuse it here. Although I am wondering if I should just do triangulation in correspondence augmenter and skip BA.

gtsfm/runner/gtsfm_runner_base.py
ayushbaid: I thought about that approach too, but I am wondering if having _lower_ quality correspondences before 2-view estimator will result in poor pose estimates. What do you think?

gtsfm/frontend/correspondence_augmenter.py
ayushbaid: Yes I saw that too. Let me work on approach 1 too then.

travisdriver: Yes, I think (1) makes more sense

ayushbaid:
Closing as #663 seems like a simple and preferred approach.

PR #659, opened 2023-07-16
Add feature to retry connecting to the cluster on failures

stepanyanhayk:
Because of network connectivity issues, sometimes connecting to the cluster can fail from the first attempt. This PR introduces a feature to retry connecting to the cluster if it fails on the first attempt. The number of times it will try to reconnect depends on the parsed argument, which is set to 3 as default.

### DIFF
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -213,15 +213,16 @@ class GtsfmRunnerBase:
                 logger.info(f"Using {workers} as workers")
                 try:
                     cluster = SSHCluster(
-                            [scheduler] + workers,
-                            scheduler_options={"dashboard_address": self.parsed_args.dashboard_port},
-                            worker_options={
-                                "n_workers": self.parsed_args.num_workers,
-                                "nthreads": self.parsed_args.threads_per_worker,
-                            },
-                        )
+                        [scheduler] + workers,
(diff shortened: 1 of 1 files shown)


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -199,14 +205,25 @@ def run(self) -> GtsfmData:
...
+                        scheduler_options={"dashboard_address": self.parsed_args.dashboard_port},
+                        worker_options={
+                            "n_workers": self.parsed_args.num_workers,
+                            "nthreads": self.parsed_args.threads_per_worker,
+                        },
+                    )
+                    connected = True
+                except Exception as e:
+                    logger.info(f"Worker failed to start: {str(e)}")
+                    retry_count += 1
```
akshay-krishnan, line 226: Could you move this entire logic to a separate `setup_ssh_cluster_with_retries` function?

akshay-krishnan (commented):
thanks!

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: @stepanyanhayk lgtm except I agree with Akshay, this code would be cleaner if we decomposed it into a helper

stepanyanhayk: Done.

akshay-krishnan (approved):

johnwlambert (approved):

PR #665, opened 2023-07-24
generalize self-hosted runner to other machines, change data location

akshay-krishnan:
updates data location to a common /usr/local/gtsfm-data which all users can access. 
also changed the instance from wildcat to eagle, but this does not matter, we can potentially run on any machine.


johnwlambert (approved):
Thanks! Which data is under `/usr/local/gtsfm-data` now?

akshay-krishnan:
just skydio501 for now

akshay-krishnan:
testing it here: https://github.com/borglab/gtsfm/actions/runs/5641337570

travisdriver:
Why was this merged before making sure that the CI ran successfully :/

PR #669, opened 2023-07-25
Support configuring Shonan uncertainty from config

johnwlambert:
Add support to allow configuring Shonan uncertainty from a config.

### DIFF
--- a/gtsfm/averaging/rotation/shonan.py
+++ b/gtsfm/averaging/rotation/shonan.py
@@ -28,19 +28,25 @@ import gtsfm.utils.logger as logger_utils
 from gtsfm.averaging.rotation.rotation_averaging_base import RotationAveragingBase
 from gtsfm.common.pose_prior import PosePrior
 
-TWOVIEW_ROTATION_SIGMA = 1
 POSE3_DOF = 6
 
 logger = logger_utils.get_logger()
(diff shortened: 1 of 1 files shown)


travisdriver:
Could you add the uncertainty to the config file?

awayfromdog1:
> Could you add the uncertainty to the config file?

Thanks for the review @travisdriver. In a follow up PR we will be able to tune this from config.

I’d prefer to keep some default value here for now in case people forget to specify it in the config.

johnwlambert:
> Could you add the uncertainty to the config file?

Thanks for the review @travisdriver. In a follow up PR we will be able to tune this from config.

I’d prefer to keep some default value here for now in case people forget to specify it in the config

akshay-krishnan (approved):
would be nice to have it in config, can be done in another PR

travisdriver:
> > Could you add the uncertainty to the config file?
> 
> 
> 
> Thanks for the review @travisdriver. In a follow up PR we will be able to tune this from config.
> 
> 
> 
> I’d prefer to keep some default value here for now in case people forget to specify it in the config

You can keep a default value in the constructor, I was just saying it would be nice to expose it in the config since I thought that's what the PR title implied.

PR #670, opened 2023-07-25
Moved NetVLAD weight download to bash

travisdriver:


johnwlambert:
@travisdriver can we close this since we landed the weight bash download change in https://github.com/borglab/gtsfm/pull/674/files?

PR #678, opened 2023-07-31
Fix tests and docstring after removing image extension from olsson loader

ayushbaid:


akshay-krishnan (approved):
not sure how this caused the memory increase you mentioned, but thanks for fixing!

PR #684, opened 2023-08-04
Add utility function for reading and visualizing scene data in binary format.

johnwlambert:
Run:
```
python gtsfm/visualization/view_scene.py --output_dir tanks_temples-2023-08-03-colmap-results/Barn/sparse/0 --file_format bin
```
See:
Screen Shot 2023-08-03 at 11 55 24 PM

### DIFF
--- a/gtsfm/utils/io.py
+++ b/gtsfm/utils/io.py
@@ -22,6 +22,7 @@ import gtsfm.utils.images as image_utils
 import gtsfm.utils.logger as logger_utils
 import gtsfm.utils.reprojection as reproj_utils
 import gtsfm.visualization.open3d_vis_utils as open3d_vis_utils
+import thirdparty.colmap.scripts.python.read_write_model as colmap_io
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.common.image import Image
 from gtsfm.common.sfm_track import SfmTrack2d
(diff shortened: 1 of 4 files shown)


gtsfm/utils/io.py
```diff
@@ -450,13 +457,40 @@ def read_points_txt(fpath: str) -> Tuple[Optional[np.ndarray], Optional[np.ndarr
...
+    Returns:
+        5-tuple of:
+            wTi_list: List of camera poses for each image.
+            img_fnames: List of image file names, for each image.
+            calibrations: Calibration object for each camera.
+            point_cloud: Float array of shape (N,3) representing per-point x/y/z coordinates.
+            rgb: Uint8 array of shape (N,3) representing per-point colors.
+    """
+    if file_format == "txt":
+        points_fpath = f"{data_dir}/points3D.txt"
```
akshay-krishnan, line 481: maybe just check which format exists in the directory? do we need another parameter here?

johnwlambert: Good suggestion, updated now to automatically infer file format of scene data based on file existence.

gtsfm/utils/io.py
```diff
@@ -450,13 +457,49 @@ def read_points_txt(fpath: str) -> Tuple[Optional[np.ndarray], Optional[np.ndarr
...
+        images_fpath = f"{data_dir}/images.txt"
+        cameras_fpath = f"{data_dir}/cameras.txt"
+        wTi_list, img_fnames = read_images_txt(images_fpath)
+        calibrations = read_cameras_txt(cameras_fpath)
+        point_cloud, rgb = read_points_txt(points_fpath)
+
+    elif file_format == "bin":
+        cameras, images, points3d = colmap_io.read_model(path=data_dir, ext=".bin")
```
akshay-krishnan, line 498: could we use this for .txt too and do the `colmap2gtsfm` ? if not, maybe add a comment/TODO why?

akshay-krishnan (approved):

gtsfm/utils/io.py
johnwlambert: Good point, I'll add a TODO in a follow-up.

gtsfm/utils/io.py
johnwlambert: (Marking as resolved).

PR #686, opened 2023-08-05
Allow configurable memory limit per worker

johnwlambert:


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -237,7 +240,9 @@ def run(self) -> GtsfmData:
             self.scene_optimizer._output_worker = io_worker
         else:
             cluster = LocalCluster(
-                n_workers=self.parsed_args.num_workers, threads_per_worker=self.parsed_args.threads_per_worker
+                n_workers=self.parsed_args.num_workers,
```
akshay-krishnan, line 243: could you add this to the SSHCluster above too? if it supports the arg?

johnwlambert: Haven't had a chance to test it with SSHCluster, so would prefer to defer that to a follow up PR if possible. It unblocks translation averaging, though, on a single machine

ayushbaid (approved):

PR #693, opened 2023-08-12
Add metrics to measure Translation Averaging execution time

johnwlambert:


gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -486,9 +490,19 @@ def run_translation_averaging(
...
-        ta_metrics = compute_metrics(set(w_i2Ui1_dict_inliers.keys()), i2Ui1_dict, wRi_list, wti_list, gt_wTi_list)
+        ta_metrics = compute_metrics(
+            set(w_i2Ui1_dict_inliers.keys()),
+            i2Ui1_dict,
+            wRi_list,
+            wti_list,
+            gt_wTi_list,
+            inlier_computation_duration_sec,
+            lm_duration_sec,
+        )
```
akshay-krishnan, line 506: compute metrics as before, simply use ta_metrics.add_metric(GtsfmMetric('my_time_sec', time)) to add the time. I dont think it needs to go into the function that does the evaluation.

PR #694, opened 2023-08-12
Add metrics to measure timing for rotation averaging, translation averaging, and bundle adjustment

johnwlambert:


gtsfm/bundle/bundle_adjustment.py
```diff
@@ -285,6 +289,48 @@ def __cameras_to_model(
 
         return sorted(list(cameras))
 
+    def run_ba_at_threshold(
```
akshay-krishnan, line 292: nit: not sure if 'threshold' is important enough to be in the name, since its optional.

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -412,6 +423,63 @@ def evaluate(
 
         return ba_metrics
 
+    def _run_ba_with_profiling(
```
akshay-krishnan, line 426: nit: _run_ba_and_evaluate() might be a better name, since we compute other metrics too.

akshay-krishnan (approved):

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Updated the name now to `run_ba_stage_with_filtering()`.

gtsfm/bundle/bundle_adjustment.py
johnwlambert: Good idea, updated now to `_run_ba_and_evaluate()`

PR #695, opened 2023-08-13
remove the image_extensions flag (almost) completely

akshay-krishnan:
It was only being used in one_d_sfm loader. 

Now they use use a util to find all images in a directory. 

I also updated the CI benchmark to not pass this flag. It is however still a part of the artifact name, because changing this could break our offline scripts to parse any artifacts. It is also tricky (though very doable) to change the indices of the remaining fields from the "dataset matrix config" used in the CI.

### DIFF
--- a/gtsfm/bundle/bundle_adjustment.py
+++ b/gtsfm/bundle/bundle_adjustment.py
@@ -297,7 +297,32 @@ class BundleAdjustmentOptimizer:
         reproj_error_thresh: Optional[float],
         verbose: bool = True,
     ) -> Tuple[GtsfmData, GtsfmData, List[bool], float]:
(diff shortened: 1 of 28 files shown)


akshay-krishnan:
my changes from another branch creeped in, lets wait until that is merged

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -485,6 +478,63 @@ def evaluate(
 
         return ba_metrics
 
+    def _run_ba_with_profiling(
```
ayushbaid, line 481: Where do we use this function?

ayushbaid (commented):
I'll review this PR after the other one gets merged

ayushbaid (approved):
LGTM!

PR #698, opened 2023-08-15
Use unified config in CI and configure num_matched via CLI

johnwlambert:


akshay-krishnan (approved):
did you check our CI benchmarks for the unified config?

PR #700, opened 2023-08-26
Measure view-graph estimation runtime

johnwlambert:


gtsfm/view_graph_estimator/cycle_consistent_rotation_estimator.py
```diff
@@ -138,7 +140,14 @@ def run(
...
 
-        logger.info("Found %d consistent rel. rotations from %d original edges.", len(valid_edges), len(input_edges))
+        end_time = time.time()
+        duration_sec = end_time - start_time
+        logger.info(
+            "Found %d consistent rel. rotations from %d original edges in %.2f sec.",
+            len(valid_edges),
+            len(input_edges),
+            duration_sec,
+        )
```
akshay-krishnan, line 150: could you add a metric? this will drown in all the noise.

johnwlambert: I think we may not need one, but we could add in a follow-up PR. Logged stats indicate this part is <1 sec for all runs.

johnwlambert: (I believe the logged stat info is sufficient for now)

akshay-krishnan: interesting, thanks!

akshay-krishnan (approved):

PR #701, opened 2023-08-26
Prevent heartbeat timeout

johnwlambert:
Dask will otherwise kill workers if they dont heartbeat for more than 30 mins.


ayushbaid (commented):
I am curious we do it here instead of gtsfm runner

akshay-krishnan:
> I am curious we do it here instead of gtsfm runner

+1, could you move it to the runner? I think there are other dask config settings there too.

johnwlambert:
@akshay-krishnan @ayushbaid I moved it to `gtsfm/runner/gtsfm_runner_base.py` now, thanks.

akshay-krishnan (approved):

PR #706, opened 2023-09-11
convert 3d_viz flag to false

stepanyanhayk:
`save_sfm_data_viz` function is one of the most time-consuming functions on the cluster. This PR changes default value of 3d_viz flag from True to False.

image


johnwlambert (approved):
Thanks!

PR #710, opened 2023-09-14
Measure GtsfmData I/O time and remove redundant I/O ops

johnwlambert:


gtsfm/scene_optimizer.py
```diff
@@ -362,55 +364,54 @@ def get_gtsfm_data_with_gt_cameras_and_est_tracks(
...
+        images=images,
+        save_dir=os.path.join(output_dir, "ba_output_gt"),
+    )
 
-    return saving_graph_list
+    end_time = time.time()
+    duration_sec = end_time - start_time
+    logger.info("GtsfmData I/O took %.2f sec.", duration_sec)
+    # Delete old version of React results directory.
+    shutil.rmtree(REACT_RESULTS_PATH)
```
akshay-krishnan, line 412: why do this delete?

gtsfm/scene_optimizer.py
```diff
@@ -362,55 +364,54 @@ def get_gtsfm_data_with_gt_cameras_and_est_tracks(
...
+    # Save a duplicate copy of the directory in REACT_RESULTS_PATH.
+    shutil.copytree(src=results_path, dst=REACT_RESULTS_PATH)
```
akshay-krishnan, line 414: could you include this copying in the I/O time above as well?

scripts/collect_results.py
```diff
@@ -136,6 +162,55 @@ def main(user_root: Path, output_fpath: str) -> None:
     save_table_to_tsv(table, headers, output_fpath)
 
 
+def _make_runtime_pie_chart(experiment_roots: Sequence[Path]) -> None:
```
akshay-krishnan, line 165: nice!

akshay-krishnan (approved):

gtsfm/scene_optimizer.py
johnwlambert: Turns out copytree only works if the directory doesn't exist, so a delete before is necessary

gtsfm/scene_optimizer.py
johnwlambert: Good catch, updated now to include this React I/O time.

PR #711, opened 2023-09-14
Use average coordinate of merged keypoints during NMS

johnwlambert:


gtsfm/frontend/correspondence_generator/keypoint_aggregator/keypoint_aggregator_dedup.py
```diff
@@ -51,11 +51,16 @@ def append_unique_keypoints(
...
             is_duplicate = np.any(diff_norms <= self.nms_merge_radius)
             if len(per_image_kpt_coordinates[i]) > 0 and is_duplicate:
                 self.duplicates_found += 1
-                i_indices[k] = np.argmin(diff_norms)
+                img_global_kpt_idx = np.argmin(diff_norms)
+                i_indices[k] = img_global_kpt_idx
+                # Modify global keypoint coordinate to be set to average value of merged detections, instead of
+                # using the first identified coordinate.
+                updated_uv = np.mean([per_image_kpt_coordinates[i][img_global_kpt_idx], uv], axis=0)
+                per_image_kpt_coordinates[i][img_global_kpt_idx] = updated_uv
```
travisdriver, line 62: Shouldn't you take the mean _after_ you've determined which keypoints are to be merged?

gtsfm/frontend/correspondence_generator/keypoint_aggregator/keypoint_aggregator_dedup.py
```diff
@@ -45,11 +51,16 @@ def append_unique_keypoints(
 
         for k, uv in enumerate(keypoints.coordinates):
             diff_norms = np.linalg.norm(per_image_kpt_coordinates[i] - uv, axis=1)
-            # TODO(johnwlambert,ayushbaid): test loosening threshold below to some epsilon.
-            is_identical = np.any(diff_norms == 0)
-            if len(per_image_kpt_coordinates[i]) > 0 and is_identical:
+            is_duplicate = np.any(diff_norms <= self.nms_merge_radius)
+            if len(per_image_kpt_coordinates[i]) > 0 and is_duplicate:
```
akshay-krishnan, line 55: have you tested with len(per_image_kpt_coordinates[i]) == 0? does the `<= ` work with empty arrrays? 
Also np.any should be false in that case, so you would not need that condition at all?

PR #716, opened 2023-09-18
Add back option to save 2-view correspondence visualizations

johnwlambert:


ayushbaid (approved):

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -320,6 +321,30 @@ def run(self) -> GtsfmData:
             two_view_estimation_duration_sec = time.time() - two_view_estimation_start_time
 
         i2Ri1_dict, i2Ui1_dict, v_corr_idxs_dict, two_view_reports_dict = unzip_two_view_results(two_view_results_dict)
+
+        # annotation = dask.annotate(workers=self._output_worker) if self._output_worker else dask.annotate()
```
akshay-krishnan, line 325: I dont think the annotation works with the futures API. Besides if this code is in the runner, it should always run on the scheduler machine. so annotation can be removed?

akshay-krishnan (approved):

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Thanks, removed now.

PR #717, opened 2023-09-18
Parallel mfas

travisdriver:


gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -209,20 +238,49 @@ def compute_inliers(
...
+                outlier_weights.append(
+                    dask.delayed(MFASWrapper)(MFAS, future_w_i2Ui1_dict, future_w_iUj_dict_tracks, direction)
+                )
+        else:
+            for j in range(0, len(projection_directions), batch_size):
+                outlier_weights.append(
+                    dask.delayed(MFASWrapper)(
+                        MFAS, future_w_i2Ui1_dict, future_w_iUj_dict_tracks, projection_directions[j : j + batch_size]
+                    )
+                )
```
johnwlambert, line 275: Looks good @travisdriver. I think CI is failing though?

```
FAILED tests/averaging/translation/test_averaging_1dsfm.py::TestTranslationAveraging1DSFM::test_circle_all_edges - ValueError: not enough values to unpack (expected 2, got 0)
```

gtsfm/averaging/translation/averaging_1dsfm.py
```diff
@@ -214,24 +258,41 @@ def compute_inliers(
...
+        else:
+            for j in range(0, len(projection_directions), batch_size):
+                batched_outlier_weights.append(
+                    dask.delayed(MFASWrapper)(
+                        MFAS, future_w_i2Ui1_dict, future_w_iUj_dict_tracks, projection_directions[j : j + batch_size]
+                    )
+                )
+
+        # Compute outlier weights in parallel.
+        batched_outlier_weights = dask.compute(*batched_outlier_weights)
```
ayushbaid, line 285: This might be a dask anti-pattern. https://docs.dask.org/en/latest/delayed-best-practices.html

We should not be calling compute repeatedly. We to have a function which unzips and chains the batch results

ayushbaid (changes requested):

gtsfm/averaging/translation/averaging_1dsfm.py
travisdriver: Could you elaborate a little more? I'm only calling compute once.

travisdriver: Could you remove this since we agreed that the architecture is good?

ayushbaid: Removed.

ayushbaid (approved):

johnwlambert (approved):

PR #727, opened 2023-10-04
Run multiple trials of (num_matched, max_frame_lookahead), select trial by mean track len

johnwlambert:


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -289,6 +272,84 @@ def run(self) -> GtsfmData:
...
+            COMBINATIONS = [
+                (0, 5),
+                (0, 10),
+                (5, 0),
+                (5, 5),
+                (5, 10),
+                (10, 0),
+                (10, 5),
+                (10, 10),
+            ]
```
travisdriver, line 289: I think allowing the user to put in a list of `num_matched` or `max_frame_lookahead` values would be cleaner (similar to what we do for multi-step BA), as opposed to hardcoding the combinations.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -150,6 +152,12 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             default=3,
             help="Number of times to retry cluster connection if it fails.",
         )
+        parser.add_argument(
+            "--run_low_quality",
```
ayushbaid, line 156: I would recommend changing the name to something like `run_single_trial` or `skip_trials` with low quality mentioned in the description.

gtsfm/runner/gtsfm_runner_base.py
ayushbaid: I like this idea, but inputting a large list will be troublesome I think.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -289,6 +272,84 @@ def run(self) -> GtsfmData:
             process_graph_generator.is_image_correspondence = True
         process_graph_generator.save_graph()
 
+        if self.parsed_args.run_low_quality:
+            # TODO(johnwlambert): Check if these are None, otherwise don't override.
+            COMBINATIONS = [(self.parsed_args.num_matched, self.parsed_args.max_frame_lookahead)]
+        else:
+            # Run multiple trials at different (num_matched, max_frame_lookahead) settings.
+            COMBINATIONS = [
```
ayushbaid, line 280: We should dynamically create combinations around the user input of num_matched and max frame lookahead too?

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -289,6 +272,84 @@ def run(self) -> GtsfmData:
...
+            ]
+
+        experiment_roots = []
+        output_root = self.scene_optimizer.output_root
+        for (num_matched, max_frame_lookahead) in COMBINATIONS:
+
+            experiment_root = output_root / f"num_matched{num_matched}_maxframelookahead{max_frame_lookahead}"
+            experiment_roots.append(experiment_root)
+            self.scene_optimizer.output_root = experiment_root
+            print(f"Run and save to {self.scene_optimizer.output_root}")
```
ayushbaid, line 298: Please replace it with logging

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -289,6 +272,84 @@ def run(self) -> GtsfmData:
...
+                (5, 5),
+                (5, 10),
+                (10, 0),
+                (10, 5),
+                (10, 10),
+            ]
+
+        experiment_roots = []
+        output_root = self.scene_optimizer.output_root
+        for (num_matched, max_frame_lookahead) in COMBINATIONS:
```
ayushbaid, line 293: There are retrievers which work with only one of num_matched/max_frame_lookahead? We throw value errors which might be wrong. Example for `ImageMatchingRegime.RETRIEVAL`

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -289,6 +272,84 @@ def run(self) -> GtsfmData:
...
+            dirpath = Path(experiment_root) / "result_metrics"
+            json_fname = "bundle_adjustment_metrics.json"
+            metric_name = "3d_track_lengths_filtered"
+            section_name = Path(json_fname).stem
+
+            json_data = io_utils.read_json_file(f"{dirpath}/{json_fname}")[section_name]
+            mean_track_length = json_data[metric_name]["summary"]["mean"]
+            mean_track_lengths.append(mean_track_length)
+
+        print("Mean track lengths: ", [np.round(length, 3) for length in mean_track_lengths])
```
ayushbaid, line 339: Replace with logging

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -289,6 +272,84 @@ def run(self) -> GtsfmData:
...
+        best_idx = np.argmax(mean_track_lengths)
+        best_experiment_root = experiment_roots[best_idx]
+        print("Selected: index=", best_idx, best_experiment_root)
```
ayushbaid, line 342: Replace with logging

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Updated, thanks.

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Updated, thanks.

gtsfm/runner/gtsfm_runner_base.py
johnwlambert: Updated, thanks.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -150,6 +152,12 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             default=3,
             help="Number of times to retry cluster connection if it fails.",
         )
+        parser.add_argument(
+            "--run_single_trial",
```
akshay-krishnan, line 156: might be better to default to running a single trial while we are still in this development phase. the most frequent usage of gtsfm right now  (by us) involves running a single trial.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -301,6 +274,105 @@ def run(self) -> GtsfmData:
...
+            # Run multiple trials at different (num_matched, max_frame_lookahead) settings.
+            COMBINATIONS = [
+                (0, 5),
+                (0, 10),
+                (5, 0),
+                (5, 5),
+                (5, 10),
+                (10, 0),
+                (10, 5),
+                (10, 10),
```
akshay-krishnan, line 290: potentially useful to make the max values here configurable, not needed in this PR though

PR #731, opened 2023-10-10
Add relative rotation metric on Rotation Averaging module

johnwlambert:


gtsfm/utils/geometry_comparisons.py
```diff
@@ -263,17 +263,41 @@ def compare_global_poses(
...
+        wRi1: Global rotation of camera i1.
+
+    Returns:
+        Angle between two-view measurement and synthetic relative rotation in degrees.
+    """
+    if i2Ri1 is None or wRi2 is None or wRi1 is None:
+        return None
+
+    # Synthetic measurement.
+    i2Ri1_synthetic = wRi2.between(wRi1)
```
akshay-krishnan, line 286: I dont think we need to call this "synthetic measurement", as it comes from the estimate. It is basically the consistency between the estimate and the measurement.

gtsfm/utils/metrics.py
```diff
@@ -233,6 +233,27 @@ def compute_rotation_angle_metric(wRi_list: List[Optional[Rot3]], gt_wRi_list: L
...
+        wRi_list: List of estimated camera global rotations.
+
+    Returns:
+        A GtsfmMetric for the relative rotation angle errors, in degrees.
+    """
+    angles: List[Optional[float]] = []
+    for i1, i2 in i2Ri1_dict:
+        i2Ri1 = i2Ri1_dict[(i1, i2)]
+        angles.append(
+            comp_utils.compute_rotation_angle_measurement_consistency(i2Ri1, wRi2=wRi_list[i2], wRi1=wRi_list[i1])
```
akshay-krishnan, line 252: can we not do the .between() here? do we need that extra function?

akshay-krishnan (approved):

gtsfm/utils/metrics.py
johnwlambert: Good point, we didn't need the extra function. I've removed it now, and consolidate the 3-4 lines of code into the caller directly.

gtsfm/utils/geometry_comparisons.py
johnwlambert: I think it gets a little confusing though, since the measurement is also an estimate (it's an estimated measurement).

So I think the dichotomy has to be what kind of 2-view measurement is it?
(1) estimated by system directly
(2) synthetically derived by other system outputs

johnwlambert: @akshay-krishnan I updated the name to `i2Ri1_derived`, hopefully this is clearer now

akshay-krishnan:
sounds good. feel free to merge. 
Just curious, are we supposed to read this metric as a measure of noise in the measurements or the estimate error? Do we already have a metric where we do this, but using ground truth "measurements"?

PR #739, opened 2023-11-14
2 stage filtering in ViewGraphEstimator

johnwlambert:


gtsfm/multi_view_optimizer.py
```diff
@@ -42,6 +46,10 @@ def __init__(
         self.ba_optimizer = bundle_adjustment_module
         self._run_view_graph_estimator: bool = self.view_graph_estimator is not None
```
ayushbaid, line 47: We should use the flag for stage 2 too?

gtsfm/multi_view_optimizer.py
```diff
@@ -112,6 +120,21 @@ def create_computation_graph(
                 two_view_reports_dict,
                 debug_output_dir,
             )
+            (
+                viewgraph_i2Ri1_graph,
+                viewgraph_i2Ui1_graph,
+                viewgraph_v_corr_idxs_graph,
+                viewgraph_two_view_reports_graph,
+                viewgraph_estimation_metrics,
```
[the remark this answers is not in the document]
johnwlambert: IMHO this will be overly verbose. We already get visualizations saved for each of the two stages, currently, and I think the combined performance is what matters most.

gtsfm/multi_view_optimizer.py
johnwlambert: @ayushbaid did you mean to add the MEDIAN_EDGE_ERROR stage to the config also?

ayushbaid: I meant to make it optional. If we do not want to run stage 2 we can easily skip it. 

But if we get good performance across the board and there is no reason to not use it, I am fine.

ayushbaid (approved):

PR #740, opened 2023-11-15
Initialize NetVLAD in constructor

travisdriver:


johnwlambert (approved):
thanks! looks good but have we benchmarked its runtime without retrieval cacher yet? let's do that before merging

PR #749, opened 2023-11-24
Check for indeterminate linear system in 2-view BA

johnwlambert:


ayushbaid (approved):
We are not changing the config?

PR #750, opened 2023-11-24
Add minimal loader for Tanks & Temples datasets

johnwlambert:


johnwlambert:
LGTM'd by @akshay-krishnan

PR #773, opened 2024-01-25
Add ETH3D Benchmarks

travisdriver:


scripts/eth3d_benchmark.sh
```diff
@@ -0,0 +1,113 @@
...
+				fi
+
+				echo "Dataset: ${dataset}"
+				echo "Num matched: ${num_matched}"
+				echo "Max frame lookahead: ${max_frame_lookahead}"
+				echo "Correspondence Generator: ${correspondence_generator_config_name}"
+				echo "Num workers: ${num_workers}"
+
+				images_dir=/home/tdriver6/dev/eth3d_datasets/${dataset}_dslr_undistorted/${dataset}/images
+                                colmap_files_dirpath=/home/tdriver6/dev/eth3d_datasets/${dataset}_dslr_undistorted/${dataset}/dslr_calibration_undistorted
```
akshay-krishnan, line 88: can we use an environment variable for `/home/tdriver6/dev/eth3d_datasets/` ? 
Also if youre using the cluster machines please move data to the shared data location?

akshay-krishnan (approved):

scripts/eth3d_benchmark.sh
```diff
@@ -0,0 +1,113 @@
...
+					num_workers=1
+				fi
+
+				echo "Dataset: ${dataset}"
+				echo "Num matched: ${num_matched}"
+				echo "Max frame lookahead: ${max_frame_lookahead}"
+				echo "Correspondence Generator: ${correspondence_generator_config_name}"
+				echo "Num workers: ${num_workers}"
+
+				images_dir=/home/tdriver6/dev/eth3d_datasets/${dataset}_dslr_undistorted/${dataset}/images
```
johnwlambert, line 87: thanks for adding this. Could you add an inline comment in the line above that would show the original download URL?

gtsfm/utils/io.py
```diff
@@ -290,6 +290,16 @@ def read_cameras_txt(
                 float(k1),
                 float(k2),
             )
+        elif model == "PINHOLE":
```
johnwlambert, line 293: could be nice to include an inline comment here that
# The 'PINHOLE' model is defined as a 4-parameter camera model: fx, fy, cx, cy
# It is defined in https://github.com/colmap/colmap/blob/main/src/colmap/sensor/models.h

gtsfm/two_view_estimator_cacher.py
```diff
@@ -50,6 +50,8 @@ def __generate_cache_key(
         self, keypoints_i1: Keypoints, keypoints_i2: Keypoints, putative_corr_idxs: np.ndarray
     ) -> str:
         """Generates a cache key according to keypoint coordinates and putative correspondence indices."""
+        if putative_corr_idxs.size == 0:  # catch no correspondences
+            return cache_utils.generate_hash_for_numpy_array(np.array([]))
```
johnwlambert, line 54: could we add a brief comment explaining why our existing logic won't handle this case? Could we add a test also?

scripts/eth3d_benchmark.sh
```diff
@@ -0,0 +1,113 @@
+# Script to launch jobs over various datasets & front-ends.
```
johnwlambert, line 1: this script looks good. I might mention ETH3D explicitly in the comment here, e.g.
```
# Script to launch jobs over various ETH3D datasets & front-ends.
# See https://www.eth3d.net/data/schoeps2017cvpr.pdf for more details.
```

scripts/eth3d_benchmark.sh
```diff
@@ -0,0 +1,113 @@
...
+
+USER_ROOT=$1
+CLUSTER_CONFIG=$2
+
+now=$(date +"%Y%m%d_%H%M%S")
+
+
+datasets=(
+	office
```
johnwlambert, line 10: can we sort these in alphabetical order?

scripts/eth3d_benchmark.sh
```diff
@@ -0,0 +1,113 @@
+# Script to launch jobs over various datasets & front-ends.
+
+USER_ROOT=$1
+CLUSTER_CONFIG=$2
+
+now=$(date +"%Y%m%d_%H%M%S")
+
+
+datasets=(
```
johnwlambert, line 9: Could be helpful to add a comment:
```
# Includes all "high-resolution multi-view" datasets from 'training' split (i.e. w/ public GT data)
# See https://www.eth3d.net/datasets for more information.
```

gtsfm/utils/io.py
```diff
@@ -517,28 +527,17 @@ def read_scene_data_from_colmap_format(
...
-        calibrations, img_dims = read_cameras_txt(cameras_fpath)
-        point_cloud, rgb = read_points_txt(points_fpath)
-
-    elif file_format == "bin":
-        cameras, images, points3d = colmap_io.read_model(path=data_dir, ext=".bin")
-        img_fnames, wTi_list, calibrations, _, point_cloud, rgb, img_dims = colmap2gtsfm(
-            cameras, images, points3d, load_sfmtracks=False
-        )
+    cameras, images, points3d = colmap_io.read_model(path=data_dir, ext=file_format)
+    img_fnames, wTi_list, calibrations, _, point_cloud, rgb, img_dims = colmap2gtsfm(
```
johnwlambert, line 538: can we put this change in a separate PR?

I don't think this new function is equivalent to the old one, since the other one provides images in a lexicographically sorted order: https://github.com/borglab/gtsfm/blob/master/gtsfm/utils/io.py#L378

the .bin format does not do that, and it will create issues for any retrieval

johnwlambert:
CI is failing, its fixed at HEAD

gtsfm/utils/io.py
travisdriver: `colmap2gtsfm` now lexigraphically sorts by the filenames as before

scripts/eth3d_benchmark.sh
travisdriver: fixed

gtsfm/utils/io.py
johnwlambert: Gentle bump on this

scripts/eth3d_benchmark.sh
```diff
@@ -0,0 +1,114 @@
...
+					num_workers=1
+				fi
+
+				echo "Dataset: ${dataset}"
+				echo "Num matched: ${num_matched}"
+				echo "Max frame lookahead: ${max_frame_lookahead}"
+				echo "Correspondence Generator: ${correspondence_generator_config_name}"
+				echo "Num workers: ${num_workers}"
+
+				images_dir=/home/tdriver6/Downloads/eth3d_datasets/${dataset}_dslr_undistorted/${dataset}/images
```
johnwlambert, line 88: could you move these hard-coded paths to the top, and could we use an uppercase variable name?

scripts/eth3d_benchmark.sh
travisdriver: added

johnwlambert (approved):

scripts/eth3d_benchmark.sh
johnwlambert: Marking as resolved, thanks.

PR #776, opened 2024-01-28
Use `align_robust` from Colmap to align reconstructions

ayushbaid:
`align_robust` offers a robust alignment mechanism to align a reconstruction to list of camera positions (no orientation), which looks to work better than GTSFM's logic

### DIFF
--- a/gtsfm/evaluation/compare_colmap_outputs.py
+++ b/gtsfm/evaluation/compare_colmap_outputs.py
@@ -7,13 +7,18 @@ import argparse
 import os
 from typing import Dict, List
 
-from gtsam import Pose3
+import numpy as np
+import pycolmap
+from gtsam import Pose3, Similarity3
+from scipy.spatial.transform import Rotation
 
 import gtsfm.runner.gtsfm_runner_base as runner_base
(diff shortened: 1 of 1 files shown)


travisdriver (approved):
LGTM

travisdriver:
> `align_robust` offers a robust alignment mechanism to align a reconstruction to list of camera positions (no orientation), which looks to work better than GTSFM's logic

Does this mean it doesn't fit to orientation?

ayushbaid:
> > `align_robust` offers a robust alignment mechanism to align a reconstruction to list of camera positions (no orientation), which looks to work better than GTSFM's logic
> 
> Does this mean it doesn't use orientation as all to compute the Sim(3)?

Yep looks like that.

ayushbaid:
> > > `align_robust` offers a robust alignment mechanism to align a reconstruction to list of camera positions (no orientation), which looks to work better than GTSFM's logic
> > 
> > 
> > Does this mean it doesn't use orientation as all to compute the Sim(3)?
> 
> Yep looks like that.

Atleast not absolute wTis. Maybe it uses relative rotations

PR #777, opened 2024-02-01
Initialize shonan using minimum spanning tree

ayushbaid:


gtsfm/averaging/rotation/shonan.py
```diff
@@ -35,6 +37,36 @@
...
+    """Sample a random rotation by generating a sample from the 4d unit sphere."""
+    q = np.random.randn(4) * 0.03
+    # make unit-length quaternion
+    q /= np.linalg.norm(q)
+    qw, qx, qy, qz = q
+    R = Rot3(qw, qx, qy, qz)
+    return R
+
+
+def initialize_global_rotations_using_mst(num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Rot3]) -> List[Rot3]:
```
johnwlambert, line 50: Thanks for adding!

could we put this function in a small adjacent helper file? This file is getting a bit big. Can we add a test in a simple 3 or 4 edge example?

ayushbaid:
I cannot make sense of the results. For datasets where GTSFM performs poorly, we do good. However, we see some regression on other datasets.


rotation_avergaging_result

ba_result

gtsfm/averaging/rotation/shonan.py
```diff
@@ -99,7 +102,7 @@ def get_isotropic_noise_model_sigma(covariance: np.ndarray) -> float:
         return between_factors
 
     def _run_with_consecutive_ordering(
-        self, num_connected_nodes: int, between_factors: BetweenFactorPose3s
+        self, num_connected_nodes: int, between_factors: BetweenFactorPose3s, initial=Optional[Values]
```
johnwlambert, line 105: is Optional[Values] meant as a type annotation or as a default value?

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,38 @@
...
+def random_rotation() -> Rot3:
+    """Sample a random rotation by generating a sample from the 4d unit sphere."""
+    q = np.random.randn(4) * 0.03
+    # make unit-length quaternion
+    q /= np.linalg.norm(q)
+    qw, qx, qy, qz = q
+    return Rot3(qw, qx, qy, qz)
+
+
+def initialize_global_rotations_using_mst(num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Rot3]) -> List[Rot3]:
```
johnwlambert, line 21: nit: can we add a full docstring explaining the Args and Returns?

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,38 @@
...
+    # Create a graph from the relative rotations dictionary
+    graph = nx.Graph()
+    for i1, i2 in i2Ri1_dict.keys():
+        # TODO: use inlier count as weight
+        graph.add_edge(i1, i2, weight=1)
```
johnwlambert, line 26: without a weight, it's not a MST, but just a ST. Is this intended?

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,38 @@
...
+
+    # Compute the Minimum Spanning Tree (MST)
+    mst = nx.minimum_spanning_tree(graph)
+
+    wRis = [random_rotation() for _ in range(num_images)]
```
johnwlambert, line 31: QQ -- why initialize with random values? We can compute a true MST with real values

```
        edges = i2Ri1_dict.keys()

        # Find the largest connected component.
        cc_nodes = graph_utils.get_nodes_in_largest_connected_component(edges)
        cc_nodes = sorted(cc_nodes)

        wRi_list = [None] * num_images
        # Choose origin node.
        origin_node = cc_nodes[0]
        wRi_list[origin_node] = Rot3()

        G = nx.Graph()
        G.add_edges_from(edges)

        # Ignore 0th node, as we already set its global pose as the origin.
        for dst_node in cc_nodes[1:]:
            # Determine the path to this node from the origin. ordered from [origin_node,...,dst_node]
            path = nx.shortest_path(G, source=origin_node, target=dst_node)

            wRi = Rot3()
            for i1, i2 in zip(path[:-1], path[1:]):
                # NOTE: i1, i2 may not be in sorted order here. May need to reverse ordering.
                if i1 < i2:
                    i1Ri2 = i2Ri1_dict[(i1, i2)].inverse()
                else:
                    i1Ri2 = i2Ri1_dict[(i2, i1)]

                # wRi = wR0 * 0R1
                wRi = wRi.compose(i1Ri2)

            wRi_list[dst_node] = wRi

        return wRi_list
```

gtsfm/utils/rotation.py
johnwlambert, line 1: nit: would spanning_tree be a better name for this file?

gtsfm/utils/rotation.py
ayushbaid: Yes right now I do not pipe in any additional information into Shonan. Eventually we can use some form of weight here.

gtsfm/utils/rotation.py
ayushbaid: I added a random initialization for some nodes which might not be in the MST.

ayushbaid: But I see that the `minimum_spanning_tree` returns a forest in that case, and I can use that.

ayushbaid: > QQ -- why initialize with random values? We can compute a true ST (non-MST) with real values
> 
> ```
>         edges = i2Ri1_dict.keys()
> 
>         # Find the largest connected component.
>         cc_nodes = graph_utils.get_nodes_in_largest_connected_component(edges)
>         cc_nodes = sorted(cc_nodes)
> 
>         wRi_list = [None] * num_images
>         # Choose origin node.
>         origin_node = cc_nodes[0]
>         wRi_list[origin_node] = Rot3()
> 
>         G = nx.Graph()
>         G.add_edges_from(edges)
> 
>         # Ignore 0th node, as we already set its global pose as the origin.
>         for dst_node in cc_nodes[1:]:
>             # Determine the path to this node from the origin. ordered from [origin_node,...,dst_node]
>             path = nx.shortest_path(G, source=origin_node, target=dst_node)
> 
>             wRi = Rot3()
>             for i1, i2 in zip(path[:-1], path[1:]):
>                 # NOTE: i1, i2 may not be in sorted order here. May need to reverse ordering.
>                 if i1 < i2:
>                     i1Ri2 = i2Ri1_dict[(i1, i2)].inverse()
>                 else:
>                     i1Ri2 = i2Ri1_dict[(i2, i1)]
> 
>                 # wRi = wR0 * 0R1
>                 wRi = wRi.compose(i1Ri2)
> 
>             wRi_list[dst_node] = wRi
> 
>         return wRi_list
> ```

Your suggested code will lead to `None` for camera which are not in the largest connected component. I chose to leave them as random instead.

gtsfm/utils/rotation.py
ayushbaid: I indented to keep all rotations related util functions here. I feel this is not as generic to be named a spanning tree right now because the args are rotations and not a generic type.

johnwlambert:
@ayushbaid CI appears to still be failing:

```
=========================== short test summary info ============================
FAILED tests/averaging/rotation/test_shonan.py::TestShonanRotationAveraging::test_circle_all_edges - AssertionError: False is not true
====== 1 failed, 435 passed, 7 skipped, 21 warnings in 982.10s (0:16:22) =======
```

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,52 @@
...
+    return Rot3(qw, qx, qy, qz)
+
+
+def initialize_global_rotations_using_mst(
+    num_images: int, i2Ri1_dict: Dict[Tuple[int, int], Rot3], edge_weights: Dict[Tuple[int, int], int]
+) -> List[Rot3]:
+    """Initialize rotations using minimum spanning tree (weighted by number of correspondences)
+
+    Args:
+        num_images: number of images in the scene.
```
johnwlambert, line 28: nit: capitalization "Number of images..."

gtsfm/utils/rotation.py
johnwlambert: @ayushbaid But we prune to the largest CC before calling Shonan:

https://github.com/borglab/gtsfm/blob/master/gtsfm/multi_view_optimizer.py#L125

So I don't think we would ever encounter such a case.

gtsfm/averaging/rotation/shonan.py
johnwlambert: Marking as resolved, thanks for splitting off into the other file.

gtsfm/averaging/rotation/shonan.py
johnwlambert: Marking as resolved, thanks for updating.

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,52 @@
...
+        i2Ri1_dict: dictionary of relative rotations (i1, i2): i2Ri1
+        edge_weights: weight of the edges (i1, i2). All edges in i2Ri1 must have an edge weight
```
johnwlambert, line 30: nit: missing period / punctuation on each line

gtsfm/utils/rotation.py
johnwlambert: Marking as resolved, thanks for adding.

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,52 @@
...
+
+    assert nx.is_connected(graph), "Relative rotation graph is not connected"
+
+    # Compute the Minimum Spanning Tree (MST)
+    mst = nx.minimum_spanning_tree(graph)
+
+    wRis = [Rot3() for _ in range(num_images)]
+    for i1, i2 in sorted(mst.edges):
+        if (i1, i2) in i2Ri1_dict:
+            wRis[i2] = wRis[i1] * i2Ri1_dict[(i1, i2)].inverse()
```
johnwlambert, line 48: I think this would be a bit more readable if you made an intermediate var i1Ri2

```
                # NOTE: i1, i2 may not be in sorted order here. May need to reverse ordering.
                if i1 < i2:
                    i1Ri2 = i2Ri1_dict[(i1, i2)].inverse()
                else:
                    i1Ri2 = i2Ri1_dict[(i2, i1)]
                wRis[i2] = wRis[i1] * i1Ri2 
```

gtsfm/utils/rotation.py
johnwlambert: I updated.

gtsfm/utils/rotation.py
johnwlambert: I updated.

gtsfm/utils/rotation.py
johnwlambert: I updated.

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,54 @@
...
+"""
+
+from typing import Dict, List, Tuple
+
+import networkx as nx
+import numpy as np
+from gtsam import Rot3
+
+
+def random_rotation() -> Rot3:
```
johnwlambert, line 13: @ayushbaid appears to be unused, right?

tests/utils/test_rotation_utils.py
```diff
@@ -0,0 +1,33 @@
...
+import gtsfm.utils.geometry_comparisons as geometry_comparisons
+import gtsfm.utils.rotation as rotation_util
+import tests.data.sample_poses as sample_poses
+
+ROTATION_ANGLE_ERROR_THRESHOLD_DEG = 2
+
+
+class TestRotationUtil(unittest.TestCase):
+    def test_mst_initialization(self):
+        """Test for 4 poses in a circle, with a pose connected all others."""
```
johnwlambert, line 17: nit: typo "connected to all others"?

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,54 @@
...
+    graph = nx.Graph()
+    for i1, i2 in i2Ri1_dict.keys():
+        graph.add_edge(i1, i2, weight=edge_weights[(i1, i2)])
+
+    assert nx.is_connected(graph), "Relative rotation graph is not connected"
+
+    # Compute the Minimum Spanning Tree (MST)
+    mst = nx.minimum_spanning_tree(graph)
+
+    wRis = [Rot3() for _ in range(num_images)]
```
johnwlambert, line 36: I think this code will only work if the poses are given in an ordered line

tests/utils/test_rotation_utils.py
```diff
@@ -0,0 +1,148 @@
...
+        wRi_list_computed = rotation_util.initialize_global_rotations_using_mst(
+            num_images,
+            i2Ri1_dict,
+            edge_weights={(i1, i2): (i1 + i2) * 100 for i1, i2 in i2Ri1_dict.keys()},
+        )
+
+        wRi_list_euler_deg_est = [np.rad2deg(wRi.roll()) for wRi in wRi_list_computed]
+        assert np.allclose(wRi_list_euler_deg_est, wRi_list_euler_deg_expected)
+
+    def test_greedily_construct_st_mixed_order_chain(self) -> None:
```
johnwlambert, line 132: @ayushbaid I added a test that fails with the current implementation

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,100 @@
...
+            i1Ri2 = i2Ri1_dict[(i1, i2)].inverse()
+        else:
+            i1Ri2 = i2Ri1_dict[(i2, i1)]
+        wRis[i2] = wRis[i1] * i1Ri2
+
+    return wRis
+
+
+
+def initialize_mst(
```
johnwlambert, line 49: @ayushbaid maybe we should consider Travis' implementation here?

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,127 @@
...
+        path = nx.shortest_path(G, source=origin_node, target=dst_node)
+
+        # Chain relative rotations w.r.t. origin node. Initialize as identity Rot3 w.r.t origin node `i1`.
+        wRi1 = Rot3()
+        for i1, i2 in zip(path[:-1], path[1:]):
+            # NOTE: i1, i2 may not be in sorted order here. May need to reverse ordering.
+            if i1 < i2:
+                i1Ri2 = i2Ri1_dict[(i1, i2)].inverse()
+            else:
+                i1Ri2 = i2Ri1_dict[(i2, i1)]
```
ayushbaid, line 57: This is not guaranteed, right?

tests/utils/test_rotation_utils.py
ayushbaid: Do we need to change the lookup logic? Or do you have a fix in mind?

johnwlambert: I added a fix -- traversing the shortest path in the MST between the origin and destination, and chaining the poses together

gtsfm/utils/rotation.py
```diff
@@ -0,0 +1,127 @@
...
+            # Path order is (origin -> ... -> i1 -> i2 -> ... -> dst_node). Set `i2` to be new `i1`.
+            wRi1 = wRi1 * i1Ri2
+
+        wRi_list[dst_node] = wRi1
+
+    return wRi_list
+
+
+# def initialize_mst(
```
akshay-krishnan, line 66: is this not used? if not, please remove.

tests/averaging/rotation/test_shonan.py
```diff
@@ -34,11 +37,17 @@ def __execute_test(self, i2Ri1_input: Dict[Tuple[int, int], Rot3], wRi_expected:
...
 
         Args:
-            i2Ri1_input: relative rotations, which are input to the algorithm.
-            wRi_expected: expected global rotations.
+            i2Ri1_input: Relative rotations, which are input to the algorithm.
+            wRi_expected: Expected global rotations.
         """
         i1Ti2_priors: Dict[Tuple[int, int], PosePrior] = {}
-        wRi_computed = self.obj.run_rotation_averaging(len(wRi_expected), i2Ri1_input, i1Ti2_priors)
+        two_view_estimation_reports = {
```
akshay-krishnan, line 44: could you add another test, keeping this one unchanged, (for default behavior that does not need two view reports)?

tests/averaging/rotation/test_shonan.py
```diff
@@ -71,43 +80,54 @@ def test_panorama(self):
...
-                covariance=np.eye(6) * 1e-5,
-                type=PosePriorType.SOFT_CONSTRAINT,
-            )
-        }
-
-        wRi_computed = self.obj.run_rotation_averaging(len(expected_wRi_list), i2Ri1_dict, i1Ti2_priors)
-        self.assertTrue(
-            geometry_comparisons.compare_rotations(wRi_computed, expected_wRi_list, ROTATION_ANGLE_ERROR_THRESHOLD_DEG)
-        )
+    # def test_simple_with_prior(self):
```
akshay-krishnan, line 103: why is this being removed?

akshay-krishnan (commented):
High level comments: 

- I think a better API to is pass "edge_weights" to the averaging modules, not the correspondences. The edge weights can be computed in the verifier for instance. Not a blocker for this PR though. 

- I think this initialization should be made optional, with a flag to change it. This would be a blocker for me :)

gtsfm/utils/rotation.py
ayushbaid: Removed

gtsfm/utils/geometry_comparisons.py
```diff
@@ -55,6 +56,7 @@ def compare_rotations(
     relative_rotations_angles = np.array(
         [compute_relative_rotation_angle(aRi, aRi_) for (aRi, aRi_) in zip(aRi_list, aRi_list_)], dtype=np.float32
     )
+    print(relative_rotations_angles)
```
ayushbaid, line 59: Needs to be removed

PR #779, opened 2024-03-06
Add script to jointly visualize GT and estimated scene

johnwlambert:


ayushbaid:
Can we not add GT as an optional arg in the current `view_scene` script?

johnwlambert:
> Can we not add GT as an optional arg in the current `view_scene` script?

Good idea, that's cleaner. Updated now

gtsfm/visualization/view_scene.py
```diff
@@ -63,7 +64,35 @@ def view_scene(args: argparse.Namespace) -> None:
...
+
+        if len(gt_calibrations) == 1:
+            gt_calibrations = gt_calibrations * len(img_fnames)
+
+        for i in range(len(wTi_list_gt)):
+            wTi_list_gt[i] = zcwTw.compose(wTi_list_gt[i])
+
+        # Align the poses.
+        n = min(len(wTi_list), len(wTi_list_gt))
+        wTi_aligned_list, rSe = alignment_utils.align_poses_sim3_ignore_missing(wTi_list_gt[:n], wTi_list[:n])
```
ayushbaid, line 80: How would this work if the missing cameras are towards the start of the sequence.

ayushbaid (approved):

gtsfm/visualization/view_scene.py
johnwlambert: @ayushbaid good catch, fixed now

gtsfm/visualization/view_scene.py
```diff
@@ -63,7 +69,73 @@ def view_scene(args: argparse.Namespace) -> None:
...
 
-    draw_scene_open3d(point_cloud, rgb, wTi_list, calibrations, args)
+    wTi_list_gt, gt_calibrations = _load_common_gt_poses(args, img_fnames)
+
+    if args.gt_olsson_dir is not None or args.gt_colmap_dir is not None:
+        if len(gt_calibrations) == 1:
+            gt_calibrations = gt_calibrations * len(img_fnames)
+
+        for i in range(len(wTi_list_gt)):
+            wTi_list_gt[i] = zcwTw.compose(wTi_list_gt[i])
```
ayushbaid, line 79: Why are we applying this transform to GT poses, but not the points?

ayushbaid (approved):
LGTM

gtsfm/visualization/view_scene.py
johnwlambert: Because I've set it up to only visualize GT poses, not GT points. Otherwise gets a bit too crowded in addition to the reconstructed points and reconstructed poses.

PR #781, opened 2024-03-10
add more unit tests on robust Sim(3) alignment

johnwlambert:


tests/utils/test_alignment_utils.py
```diff
@@ -257,6 +257,91 @@ def add_dummy_measurements_to_track(track: SfmTrack) -> SfmTrack:
...
+        bTi_list = copy.deepcopy(aTi_list)
+
+        aligned_bTi_list_est, aSb = alignment_utils.align_poses_sim3_ignore_missing(aTi_list, bTi_list)
+
+        for aTi, aTi_ in zip(aTi_list, aligned_bTi_list_est):
+            assert np.allclose(aTi.rotation().matrix(), aTi_.rotation().matrix(), atol=1e-3)
+            assert np.allclose(aTi.translation(), aTi_.translation(), atol=1e-3)
+
+
+    def test_ransac_align_poses_sim3_ignore_two_missing_estimated_poses(self) -> None:
```
ayushbaid, line 301: We should also cover the case where the missing cameras are different between the two inputs.

ayushbaid (approved):
LGTM. Thanks for adding this.

tests/utils/test_alignment_utils.py
johnwlambert: That's a good idea, let's add in a follow-up.

PR #799, opened 2024-05-14
Reduce Shonan pMin to 3

ayushbaid:
The paper https://arxiv.org/pdf/2008.02737 has experiments suggesting that Shonan should converge on SO(3) for a significant amount of time


ayushbaid:
Time improvement across the board, and no change in error.
image

akshay-krishnan (approved):

PR #803, opened 2024-05-24
Restore changes that were reverted 

akshay-krishnan:
I should not have reverted these changes in https://github.com/borglab/gtsfm/commit/6abf345522a215e25d34c0198bee0ae87eb856fa. 

The intent was to revert the previous merge commit, but that also reverts all the previous commits that the merge commit consists of (duh!).


akshay-krishnan:
alternatively, I just reverted the revert commit in #804 . this PR can be ignored.

PR #805, opened 2024-05-28
Bump up shonan optimality threshold param

ayushbaid:
This commit passes in the optimality treshold explicitly from GTSFM: a value of `-1e-5` compared to `-1e-4` default value.
I tried using 0 and `1e-5` (value used in paper for large datasets) but even the unit tests are taking up a lot of time with higher thresholds.


travisdriver:
Do you have any results showing this is better?

ayushbaid:
Not much change at `-1e-5` threshold
ra

PR #806, opened 2024-06-11
Use square root of number of inliers as weights

ayushbaid:


travisdriver:
Does this work better?

ayushbaid:
Results are a mixed bag. We should not merge this. 

ra
ba

PR #813, opened 2024-10-14
Refactoring/Adding Documentation

travisdriver:
I'm trying to include some documentation on how we modularize the Structure-from-Motion problem to make it easier for user and contributors to get up-to-speed on GTSfM.

I tried to emulate Nerfstudio's documentation structure, but I'm open to suggestions (and contributions :)).

### DIFF
--- a/assets/CORRESPONDENCE_GENERATOR.md
+++ b/assets/CORRESPONDENCE_GENERATOR.md
@@ -1,0 +1,171 @@
+# Correspondence Generator
+
+![Alt text](gtsfm-overview-correspondence-generator.svg?raw=true)
+
+## What is a Correspondence Generator?
+
+The Correspondence Generator is responsible for taking in putative image pairs from the [`ImagePairsGenerator`](https://github.com/borglab/gtsfm/blob/master/gtsfm/retriever/image_pairs_generator.py) and returning keypoints for each image and correspondences between each specified image pair. Correspondence generation is implemented by the [`CorrespondenceGeneratorBase`](https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/correspondence_generator/correspondence_generator_base.py) class defined below.
+
+```python
+class CorrespondenceGeneratorBase:
+    """Base class for correspondence generators."""
+
+    @abstractmethod
+    def generate_correspondences(
+        self,
+        client: Client,
+        images: List[Future],
+        image_pairs: List[Tuple[int, int]],
+    ) -> Tuple[List[Keypoints], Dict[Tuple[int, int], np.ndarray]]:
+        """Apply the correspondence generator to generate putative correspondences.
+
+        Args:
+            client: Dask client, used to execute the front-end as futures.
+            images: List of all images, as futures.
+            image_pairs: Indices of the pairs of images to estimate two-view pose and correspondences.
+
+        Returns:
+            List of keypoints, one entry for each input images.
+            Putative correspondence as indices of keypoints, for pairs of images.
+        """
+```
+
+## Types of Correspondence Generators
+
+We provide support for two correspondence generation paradigms: [feature extraction _then_ matching](#feature-detection-and-description-then-matching) and [_detector-free_ matching](#detector-free-matching). The supported Correspondence Generator paradigms are detailed below.
+
+
+### Feature Detection and Description _then_ Matching
+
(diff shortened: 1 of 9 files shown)


akshay-krishnan:
It would be best to do this with multiple PRs, so we can first try to merge this and I will work on a multi-view optimizer page in parallel.

assets/IMAGE_PAIRS_GENERATOR.md
```diff
@@ -0,0 +1,115 @@
...
+
+Global descriptor modules are implemented following the [`GlobalDescriptorBase`](https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/global_descriptor/global_descriptor_base.py) class and must be wrapped using a corresponding [`RetrieverBase`](https://github.com/borglab/gtsfm/blob/master/gtsfm/retriever/retriever_base.py) implementation, where the global descriptor module takes in individual images and outputs their corresponding descriptor and the retriever module takes these descriptors descriptors and computes the image pair similarity scores and outputs the putative image pairs based on a specified threshold (see [`NetVLADGlobalDescriptor`](https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/global_descriptor/netvlad_global_descriptor.py) and [`NetVLADRetriever`](https://github.com/borglab/gtsfm/blob/master/gtsfm/retriever/netvlad_retriever.py)).
+
+```python
+class RetrieverBase(GTSFMProcess):
```
akshay-krishnan, line 62: we should explain what a Retreiver is if this is needed.

assets/CORRESPONDENCE_GENERATOR.md
```diff
@@ -0,0 +1,177 @@
...
+
+- [Loader](assets/LOADER.md)
+- [Image Pairs Generator](assets/IMAGE_PAIRS_GENERATOR.md)
+- **Correspondence Generator**
+- [Two View Estimator](assets/TWO_VIEW_ESTIMATOR.md)
+- [Multiview Optimizer](assets/MULTIVIEW_OPTIMIZER.md)
+
+## What is a Correspondence Generator?
+
+The Correspondence Generator is responsible for taking in putative image pairs from the [`ImagePairsGenerator`](https://github.com/borglab/gtsfm/blob/master/gtsfm/retriever/image_pairs_generator.py) and returning keypoints for each image and correspondences between each specified image pair. Correspondence generation is implemented by the [`CorrespondenceGeneratorBase`](https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/correspondence_generator/correspondence_generator_base.py) class defined below.
```
akshay-krishnan, line 13: nit: returning keypoints for each image and (_their_ / _keypoint_)  correspondences between each specified image pair.

assets/IMAGE_PAIRS_GENERATOR.md
```diff
@@ -0,0 +1,115 @@
...
+
+- [Loader](assets/LOADER.md)
+- **Image Pairs Generator**
+- [Correspondence Generator](assets/CORRESPONDENCE_GENERATOR.md)
+- [Two View Estimator](assets/TWO_VIEW_ESTIMATOR.md)
+- [Multiview Optimizer](assets/MULTIVIEW_OPTIMIZER.md)
+
+## What is an Image Pair Generator?
+
+The Image Pair Generator takes in images from the Loader and outputs putative image pairs for correspondence generation. Image pair generation is implemented by the [`ImagePairsGenerator`](https://github.com/borglab/gtsfm/blob/master/gtsfm/retriever/image_pairs_generator.py) class defined below, which wraps a specific [`Retriever`](https://github.com/borglab/gtsfm/blob/master/gtsfm/retriever/retriever_base.py), and, optionally, a [`GlobalDescriptor`](https://github.com/borglab/gtsfm/blob/master/gtsfm/frontend/global_descriptor/global_descriptor_base.py).
```
akshay-krishnan, line 13: It would be good to know what we assume from a reader of this documentation. It seems like they would already need some knowledge of SfM. If not, terms like "putative image pairs" are unclear. What makes a pair? (answer: view overlap, potential for keypoint correspondences).

assets/IMAGE_PAIRS_GENERATOR.md
```diff
@@ -0,0 +1,115 @@
...
+            descriptors = client.gather(descriptor_futures)
+
+        return self._retriever.get_image_pairs(
+            global_descriptors=descriptors, image_fnames=image_fnames, plots_output_dir=plots_output_dir
+        )
+```
+
+## Global Image Descriptors
+
+Global desriptors work similar to local feature desriptors except that these methods generate a single descriptor for each image. Distances between these global image descriptors can then be used as a metric for the expected "matchability" of the image pairs during the correspondence generation phase, where a threshold can be used to reject potentially dissimilar image pairs before conducting correspondence generation. This reduces the likelihood of matching image pairs with little to no overlap that could cause erroneous correspondences to be inserted into the optimization process in the back-end while also significantly reducing the runtimes as compared to exhaustive matching.
```
akshay-krishnan, line 57: nit: I dont see much of a similarity here, the descriptors are actually very different. To the point that I would say "unlike feature descriptors that learn a local descriptor for each image patch/pixel, global descriptors generate a single descriptor for each image". The difference seems more important that any similarity (which would only be in the model architecture).

assets/gtsfm-overview.svg
akshay-krishnan, line 1: a small but siginificant change: rotation averaging should occur before translation averaging.

akshay-krishnan: this should be changed for all files.

PR #816, opened 2024-10-15
Add Essential Matrix Optimization

travisdriver:
Add essential matrix optimization similar to what's presented in this paper: https://openaccess.thecvf.com/content_iccv_2015/papers/Sweeney_Optimizing_the_Viewing_ICCV_2015_paper.pdf.

This is what's also used by GLOMAP.


travisdriver:
Very promising results on South Building

# Without E Optimization

Screenshot from 2024-10-16 15-12-43

# With E Optimization

Screenshot from 2024-10-16 15-12-59

travisdriver:
Better results across the board

newplot (31)

[visual_comparison_dashboard.zip](https://github.com/user-attachments/files/17406327/visual_comparison_dashboard.zip)

akshay-krishnan:
among non-rc3 medium sized datasets, it seems clearly better for south building, but worse for gerrard hall and palace of fine arts. we should focus on these medium sized and other non-CI large datasets.

gtsfm/two_view_estimator.py
```diff
@@ -360,6 +371,67 @@ def run_2view(
...
+        keypoints: keypoints for each images.
+        use_robust_loss: whether to use Huber loss.
+
+    Returns:
+        Optimized relative rotations and directions.
+    """
+    # Create GTSAM objects.
+    graph = gtsam.NonlinearFactorGraph()
+    initial = gtsam.Values()
+    noise_model=gtsam.noiseModel.Isotropic.Sigma(1, 1e-2)
```
akshay-krishnan, line 401: probably is too high, considering this is used on calibrated keypoints?

gtsfm/two_view_estimator.py
```diff
@@ -318,16 +319,26 @@ def run_2view(
...
+            #     keypoints_i1,
+            #     keypoints_i2,
+            #     pre_ba_v_corr_idxs,
+            #     camera_intrinsics_i1,
+            #     camera_intrinsics_i2,
+            #     pre_ba_i2Ri1,
+            #     pre_ba_i2Ui1,
+            #     i2Ti1_prior,
+            # )
+            post_ba_i2Ri1, post_ba_i2Ui1 = optimize_essential_matrix(
```
akshay-krishnan, line 332: maybe use an option to enable this? you could do: 
```
if len(pre_ba_v_corr_idxs) >= self.processor._min_num_inliers_est_model:
  if self._bundle_adjust: 
     ...
  elif self._optimize_essential_matrix: 
     ...
```

gtsfm/two_view_estimator.py
```diff
@@ -360,6 +371,67 @@ def run_2view(
...
+        huber_loss = gtsam.noiseModel.mEstimator.Huber.Create(1.345)
+        noise_model = gtsam.noiseModel.Robust.Create(huber_loss, noise_model)
+
+    # Add essential matrix factors.
+    mkps1 = keypoints_i1.coordinates[corr_idxs_i1i2[:, 0]]
+    mkps2 = keypoints_i2.coordinates[corr_idxs_i1i2[:, 1]]
+    initial.insert(0, gtsam.EssentialMatrix(i2Ri1, i2Ui1))
+    for kp1, kp2 in zip(mkps1, mkps2):
+        graph.add(
+            gtsam.EssentialMatrixFactor(
```
akshay-krishnan, line 412: we should probably use the newer implementations @dellaert added. I think some bugs with the essential matrix were fixed too?

PR #820, opened 2025-01-26
Bugfix: retain edges that were not part of a cycle in cycle_consistent_rotation_estimator.py

akshay-krishnan:


akshay-krishnan:
newplot (90)
BA metrics before and after the change:
- we estimate more cameras (this is an important metric!)
- since we have more cameras, pose AUC at higher ranges goes up
- rotation and translation errors seem to increase slightly for some datasets (these are already in the < 1 deg range)
- there is a drop in # highly accurately estimated cameras for some datasets (worsens for 3, improves for 2, others no change)

akshay-krishnan:
Not including these edges can cause problems like in this example. Here I show camera 71 (connected to all cameras on the left), and camera 73 (connected to all cameras on the right), but we miss camera 72, because it had edges to 71 and 73, and there was not edge between 71 and 73 themselves. We therefore miss out on the loop closure constraint, and there is duplicate structure

pre_cam_71
pre_cam_73

akshay-krishnan:
post this change, we dont lose camera 72, and the duplicate is noticably less, although still exists

post_cam_72

dellaert:
Cool.
What does “cycle-consistent” mean then, though? Just trying to understand, the PR clearly leads to better results.

akshay-krishnan:
@dellaert Heres my reasoning: 
All edges are equally likely to be outliers. For edges that are in cycles, we can identify outliers by looking at the cylic rotation error. For edges that are not part of a cycle, there is no such evidence. I dont think we should assume edges that are not part of a cycle are likely to be outliers. In some cases (such as the example above), these edges connect cameras that do not have other edges, or subgraphs that are more densely connected, so discarding them can cause the graph to break.

dellaert (approved):
LGTM!

PR #823, opened 2025-02-12
Use all landmarks in the translation averaging problem

akshay-krishnan:
This significantly improves estimated pose accuracy, and provides more 3D tracks. 

We disable running the MFAS outlier rejection stage. This saves a lot of runtime, although we spend some of those savings on an increased optimization time for TA and BA.


akshay-krishnan:
newplot - 2025-02-10T230652 733

gtsfm/configs/colmap_front_end.yaml
```diff
@@ -68,7 +68,8 @@ SceneOptimizer:
       _target_: gtsfm.averaging.translation.averaging_1dsfm.TranslationAveraging1DSFM
       robust_measurement_noise: True
       projection_sampling_method: SAMPLE_INPUT_MEASUREMENTS
-      reject_outliers: True
+      reject_outliers: False
```
dellaert, line 71: Should this be: reject_outliers_with_MFAS?

dellaert (approved):
1 small comment and two requests but could also merge as is. CI fails but I have not looked why.

PR #825, opened 2025-02-14
Improve README: Add Colab Example, Subtitles, and Clarifications

tiantianxiangshang629:
This PR enhances the README file by:
- Adding a Colab example for easier experimentation.
- Introducing subtitles for better readability.
- Removing or improving confusing content to make instructions clearer.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ Before running reconstruction, if you intend to use modules with pre-trained wei
 ./download_model_weights.sh
 ```  
 
-## Running SfM  
+### Running SfM  
 
 To process a dataset containing only an **image directory and EXIF metadata**, ensure your dataset follows this structure:  
 
@@ -88,8 +88,6 @@ To explore all available options and configurations, run:
 ```bash
 python gtsfm/runner/run_scene_optimizer_olssonloader.py -h
 ```  
-
-### Running GTSFM with the Deep Front-End  
 
 For example, if you want to use the **Deep Front-End (recommended)** on the `"door"` dataset, run:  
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -50,62 +50,95 @@ pip install -e .
...
 
 ```bash
 ./download_model_weights.sh
-```
+```  
+
+### Running SfM  
 
-To run SfM with a dataset with only an image directory and EXIF, with image file names ending with "jpg", please create the following file structure like
+To process a dataset containing only an **image directory and EXIF metadata**, ensure your dataset follows this structure:
```
akshay-krishnan, line 68: the EXIF metadata is optional, but only the 1dsfm loader supports that. 

Maybe we can add a line that says something like: 
If you are using a dataset without EXIF, use the one_d_sfm_loader: 

<example_command>

?

README.md
```diff
@@ -50,62 +50,95 @@ pip install -e .
...
 
-Currently we require EXIF data embedded into your images (or you can provide ground truth intrinsics in the expected format for an Olsson dataset, or COLMAP-exported text data, etc.)
+You can monitor the distributed computation using the [Dask dashboard](http://localhost:8787/status).  
+**Note:** The dashboard will only display activity while tasks are actively running.  
 
-If you would like to compare GTSfM output with COLMAP output, please run:
+### Required Image Metadata  
 
-```python
+Currently, we require **EXIF data** embedded into your images. Alternatively, you can provide:
```
akshay-krishnan, line 103: see my note above.

dellaert (approved):
I agree with comments by @akshay-krishnan but I think the PR is a net improvement *as is*. It would be good to have a follow-up PR where the loaders are explained.

PR #826, opened 2025-02-14
Update linux environment file

akshay-krishnan:
The current environment file seems broken since kornia needs more recent python versions (we use python 3.8). 
Current versions of pytorch also need python>=3.9. 
Pytorch has stopped supporting conda releases, so Ive moved it to pip.


dellaert (approved):

akshay-krishnan:
CI is now restored!! 🏁

PR #832, opened 2025-03-25
Add graph partitioning functionality for SfM pipeline

liuzongyue6:
# Add Graph Partitioning Functionality for SfM Pipeline

This PR introduces a graph partitioning system to GTSFM that allows the SfM pipeline to process large datasets more efficiently by dividing the image graph into manageable subgraphs.

## Changes

- Added `GraphPartitionerBase` abstract class that defines the interface for all graph partitioners
- Implemented `SinglePartition` as a baseline partitioner that keeps all image pairs in a single partition
- Created utility functions in `subgraph_utils.py` for handling subgraph operations
- Added test `test_graph_partitioner` for the partitioning functionality
- Updated `gtsfm_runner_base` to support graph partitioning

## Testing

The implementation has been ran through pytest as
```
pytest tests --cov=sfm \
--ignore tests/repro_tests \
--ignore tests/loader/test_argoverse_dataset_loader.py \
--ignore tests/runner/test_frontend_runner.py
```
Result as

```
======================================= short test summary info =======================================
FAILED tests/frontend/correspondence_generator/test_colmap_correspondence_generator.py::TestColmapCorrespondenceGenerator::test_data - AttributeError: 'pycolmap.Database' object has no attribute 'read_image_with_name'
================== 1 failed, 446 passed, 7 skipped, 8 warnings in 505.88s (0:08:25) ===================


```

PR #834, opened 2025-03-30
Fix flake8 errors in runner_base

akshay-krishnan:
#832 introduced some flake8 failures which this PR tries to fix.


akshay-krishnan:
Unit tests are passing

PR #838, opened 2025-04-20
Fix: Support 'inlier_mask' in LoRANSAC for pycolmap>=3.11

tiantianxiangshang629:
This PR updates the LoRANSAC verifier (`gtsfm/frontend/verifier/loransac.py`) to support both `"inliers"` and `"inlier_mask"` keys from the result dictionary returned by pycolmap.

In pycolmap versions <= 3.10, the RANSAC verification functions returned a list of inlier indices (`"inliers"`).  
Starting from pycolmap 3.11, the output has been updated to return a boolean mask array (`"inlier_mask"`) instead of indices.

To ensure GTSFM remains compatible with both older and newer pycolmap versions, we dynamically detect which key is present and convert accordingly.

Note: pycolmap has been deprecated, and the Python bindings have been moved back into the main COLMAP repository.  
This behavior change is documented in the COLMAP repository in [PR #2923](https://github.com/colmap/colmap/pull/2923/files).

### DIFF
--- a/gtsfm/frontend/verifier/loransac.py
+++ b/gtsfm/frontend/verifier/loransac.py
@@ -86,7 +86,11 @@ class LoRansac(VerifierBase):
         camera_i2: pycolmap.Camera = pycolmap_utils.get_pycolmap_camera(camera_intrinsics_i2)
 
         result_dict = pycolmap.essential_matrix_estimation(
-            uv_i1, uv_i2, camera_i1, camera_i2, self._ransac_options,
+            uv_i1,
+            uv_i2,
+            camera_i1,
+            camera_i2,
+            self._ransac_options,
         )
         return result_dict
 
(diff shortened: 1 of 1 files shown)


gtsfm/frontend/verifier/loransac.py
```diff
@@ -86,7 +86,11 @@ def __estimate_essential_matrix(
         camera_i2: pycolmap.Camera = pycolmap_utils.get_pycolmap_camera(camera_intrinsics_i2)
 
         result_dict = pycolmap.essential_matrix_estimation(
-            uv_i1, uv_i2, camera_i1, camera_i2, self._ransac_options,
+            uv_i1,
```
akshay-krishnan, line 89: is your line width set to 120? this formatting change should ideally not happen. you can also just revert it.

akshay-krishnan (approved):
thanks for fixing!

gtsfm/frontend/verifier/loransac.py
tiantianxiangshang629: yes I ran with 120
```
black --line-length 120
```

tiantianxiangshang629: I tried to revert it, but when I do, it automatically reformats.

PR #840, opened 2025-04-20
Add BinaryTreePartition class for hierarchical image-pair partitioning

tiantianxiangshang629:
This PR introduces a new BinaryTreePartition class to the GTSFM graph partitioning module. The partitioner supports hierarchical binary tree-based splitting of image-pair graphs, enabling configurable partition depths for scalable and efficient scene processing.

Key Features:

- Implements BinaryTreePartition as a subclass of GraphPartitionerBase.
- Partitions image pairs into 2^max_depth leaf partitions using METIS-based symbolic ordering.
- Constructs a binary tree structure with configurable maximum depth.

### DIFF
--- a/gtsfm/graph_partitioner/binary_tree_partition.py
+++ b/gtsfm/graph_partitioner/binary_tree_partition.py
@@ -186,12 +186,12 @@ class BinaryTreePartition(GraphPartitionerBase):
         leaf_idx_counter = [0]  # mutable counter to track leaf index
         node_to_idx = dict()
 
-        def dfs(n: BinaryTreeNode) -> List[Dict]:
-            if n.is_leaf():
+        def dfs(node: BinaryTreeNode) -> List[Dict]:
+            if node.is_leaf():
                 idx = leaf_idx_counter[0]
                 leaf_idx_counter[0] += 1
-                node_to_idx[n] = idx
-                exclusive_keys = set(n.keys)
+                node_to_idx[node] = idx
+                exclusive_keys = set(node.keys)
                 return [
                     {
                         "exclusive_keys": [gtsam.Symbol(u).index() for u in exclusive_keys],
@@ -203,19 +203,19 @@ class BinaryTreePartition(GraphPartitionerBase):
                     }
                 ]
 
-            left_part = dfs(n.left)
-            right_part = dfs(n.right)
+            left_part = dfs(node.left)
+            right_part = dfs(node.right)
 
-            if n.left.is_leaf() and n.right.is_leaf():
-                left_keys = set(n.left.keys)
-                right_keys = set(n.right.keys)
+            if node.left.is_leaf() and node.right.is_leaf():
+                left_keys = set(node.left.keys)
+                right_keys = set(node.right.keys)
                 shared_edges = [
                     (gtsam.Symbol(u).index(), gtsam.Symbol(v).index())
                     for u, v in nx_graph.edges()
                     if (u in left_keys and v in right_keys) or (u in right_keys and v in left_keys)
                 ]
-                i = node_to_idx[n.left]
-                j = node_to_idx[n.right]
+                i = node_to_idx[node.left]
+                j = node_to_idx[node.right]
                 shared_edge_map[(i, j)] = shared_edges
 
             return left_part + right_part
--- a/gtsfm/graph_partitioner/binary_tree_partition.py
+++ b/gtsfm/graph_partitioner/binary_tree_partition.py
@@ -1,0 +1,196 @@
+"""Implementation of a binary tree graph partitioner.
+
+This partitioner recursively partitions image pair graphs into a binary tree
+structure up to a specified depth, using METIS-based ordering. Leaf nodes
+represent explicit image keys and associated edge groupings.
+
+Authors: Shicong Ma
+"""
+
+from typing import Dict, List, Tuple
+
+import gtsam
+import networkx as nx
+from gtsam import SymbolicFactorGraph
+
+import gtsfm.utils.logger as logger_utils
+from gtsfm.graph_partitioner.graph_partitioner_base import GraphPartitionerBase
+
+logger = logger_utils.get_logger()
+
+
+class BinaryTreeNode:
+    """Node class for a binary tree representing partitioned sets of image keys."""
+
+    def __init__(self, keys: List[int], depth: int):
+        """
+        Initialize a BinaryTreeNode.
+
+        Args:
+            keys: Image indices at this node (only populated at leaf level).
+            depth: Depth level in the binary tree.
+        """
+        self.keys = keys  # Only at leaves
+        self.left = None
+        self.right = None
+        self.depth = depth
+
+    def is_leaf(self) -> bool:
+        """Check whether this node is a leaf node."""
+        return self.left is None and self.right is None
+
+
+class BinaryTreePartition(GraphPartitionerBase):
+    """Graph partitioner that uses a binary tree to recursively divide image pairs."""
+
+    def __init__(self, max_depth: int = 2):
+        """
+        Initialize the BinaryTreePartition object.
+
+        Args:
+            max_depth: Maximum depth of the binary tree; results in 2^depth partitions.
+        """
+        super().__init__(process_name="BinaryTreePartition")
+        self.max_depth = max_depth
+
+    def partition_image_pairs(self, image_pairs: List[Tuple[int, int]]) -> List[List[Tuple[int, int]]]:
+        """Partition image pairs into subgroups using a binary tree.
+
+        Args:
+            image_pairs: List of image index pairs (i, j), where i < j.
+
+        Returns:
+            A list of image pair subsets, one for each leaf in the binary tree.
+        """
+        if not image_pairs:
+            logger.warning("No image pairs provided for partitioning.")
+            return []
+
+        symbol_graph, _, nx_graph = self._build_graphs(image_pairs)
+        ordering = gtsam.Ordering.MetisSymbolicFactorGraph(symbol_graph)
+        binary_tree_root_node = self._build_binary_partition(ordering)
+
+        num_leaves = 2**self.max_depth
+        image_pairs_per_partition = [[] for _ in range(num_leaves)]
+
+        partition_details = self._compute_leaf_partition_details(binary_tree_root_node, nx_graph)
+
+        logger.info(f"BinaryTreePartition: partitioned into {len(partition_details)} leaf nodes.")
+
+        for i in range(num_leaves):
+            edges_explicit = partition_details[i].get("edges_within_explicit", [])
+            edges_shared = partition_details[i].get("edges_with_shared", [])
+            image_pairs_per_partition[i] = edges_explicit + edges_shared
+
+        for i, part in enumerate(partition_details):
+            explicit_keys = part.get("explicit_keys", [])
+            edges_within = part.get("edges_within_explicit", [])
+            edges_shared = part.get("edges_with_shared", [])
+
+            logger.info(
+                f"Partition {i}:\n"
+                f"  Explicit Image Keys that only exist within the current partition "
+                f"({len(explicit_keys)}): {sorted(explicit_keys)}\n"
+                f"  Internal Edges ({len(edges_within)}): {edges_within}\n"
+                f"  Shared Edges   ({len(edges_shared)}): {edges_shared}\n"
+            )
+
+        return image_pairs_per_partition
+
+    def _build_graphs(self, image_pairs: List[Tuple[int, int]]) -> Tuple[SymbolicFactorGraph, List[int], nx.Graph]:
+        """Construct GTSAM and NetworkX graphs from image pairs.
+
+        Args:
+            image_pairs: List of image index pairs.
+
+        Returns:
+            A tuple of (SymbolicFactorGraph, list of keys, NetworkX graph).
+        """
+        sfg = gtsam.SymbolicFactorGraph()
+        nxg = nx.Graph()
+        keys = set()
+
+        for i, j in image_pairs:
+            key_i = gtsam.symbol("x", i)
+            key_j = gtsam.symbol("x", j)
+            keys.add(key_i)
+            keys.add(key_j)
+
+            sfg.push_factor(key_i, key_j)
+            nxg.add_edge(key_i, key_j)
+
+        return sfg, list(keys), nxg
+
+    def _build_binary_partition(self, ordering: gtsam.Ordering) -> BinaryTreeNode:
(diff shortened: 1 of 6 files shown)


akshay-krishnan:
@tiantianxiangshang629 status of this PR?

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,129 @@
...
+        self.max_depth = max_depth
+
+    def partition_image_pairs(self, image_pairs: List[Tuple[int, int]]) -> List[List[Tuple[int, int]]]:
+        if not image_pairs:
+            logger.warning("No image pairs provided for partitioning.")
+            return []
+
+        symbol_graph, _, nx_graph = self._build_graphs(image_pairs)
+        ordering = gtsam.Ordering.MetisSymbolicFactorGraph(symbol_graph)
+        self.root = self._build_binary_partition(ordering)
```
akshay-krishnan: we try to be functional, so it would be preferable to not make this a class member, and pass it to methods that need it.

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,129 @@
...
+
+        num_leaves = 2**self.max_depth
+        partition_details = [{} for _ in range(num_leaves)]
+        image_pair_partitions = [[] for _ in range(num_leaves)]
+
+        self._collect_leaf_partitions(self.root, nx_graph, partition_details, leaf_idx=[0])
```
akshay-krishnan: functionality again, would be nice to just create partition_details in that method and return it.

akshay-krishnan: also, a more informative name would be nice.

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,129 @@
...
+            logger.warning("No image pairs provided for partitioning.")
+            return []
+
+        symbol_graph, _, nx_graph = self._build_graphs(image_pairs)
+        ordering = gtsam.Ordering.MetisSymbolicFactorGraph(symbol_graph)
+        self.root = self._build_binary_partition(ordering)
+
+        num_leaves = 2**self.max_depth
+        partition_details = [{} for _ in range(num_leaves)]
+        image_pair_partitions = [[] for _ in range(num_leaves)]
```
akshay-krishnan: nitpick, image_pairs_per_partition would be a nice name here.

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,129 @@
...
+            image_pair_partitions[i] = edges_explicit + edges_shared
+
+        for i, part in enumerate(partition_details):
+            explicit_keys = part.get("explicit_keys", [])
+            edges_within = part.get("edges_within_explicit", [])
+            edges_shared = part.get("edges_with_shared", [])
+
+            logger.info(
+                f"Partition {i}:\n"
+                f"  Explicit Keys ({len(explicit_keys)}): {sorted(explicit_keys)}\n"
```
akshay-krishnan: The meaning of "explicit keys" is probably unclear to the user.

gtsfm/graph_partitioner/binary_tree_partition_test.py
akshay-krishnan, line 1: we start test names with test_ and put it inside the tests directory, in the corresponding folder. youll find other tests there.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -157,7 +162,7 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             "--graph_partitioner",
             type=str,
             default="single",
-            choices=["single", "other_partitioner_types"],
+            choices=["single", "binary_tree_eight_partitions", "other_partitioner_types"],
```
akshay-krishnan: we should not build the number of partitions into the name, it would just be a config parameter.

akshay-krishnan: actually this argument should not be needed at all. we should just be able to specify the class name in the yaml file. same way we specify classes like Ransac. This is still a TODO. If you want to change that in this PR, feel free to, but you dont have to.

gtsfm/scene_optimizer.py
```diff
@@ -90,7 +89,6 @@ def __init__(
         self._pose_angular_error_thresh = pose_angular_error_thresh
         self.output_root = Path(output_root)
         self._output_worker = output_worker
-        self._create_output_directories()
```
akshay-krishnan, line 93: why is this removed?

gtsfm/utils/io.py
```diff
@@ -223,10 +224,16 @@ def colmap2gtsfm(
         elif camera_model_name == "RADIAL":
             f, cx, cy, k1, k2 = cameras[img.camera_id].params[:5]
             fx = f
+        elif camera_model_name == "OPENCV":
+            fx, fy, cx, cy, k1, k2, p1, p2 = cameras[img.camera_id].params[:8]
         else:
             raise ValueError(f"Unsupported COLMAP camera type: {camera_model_name}")
 
-        intrinsics_gtsfm.append(Cal3Bundler(fx, k1, k2, cx, cy))
+        if camera_model_name == "OPENCV":
```
akshay-krishnan: were you able to run gtsfm with Cal3DS2? I think it was not as good as Cal3Bundler when we have used it before.

gtsfm/scene_optimizer.py
```diff
@@ -30,16 +30,15 @@
...
 from gtsfm.multi_view_optimizer import MultiViewOptimizer
 from gtsfm.retriever.image_pairs_generator import ImagePairsGenerator
 from gtsfm.retriever.retriever_base import ImageMatchingRegime
-from gtsfm.two_view_estimator import (
-    POST_ISP_REPORT_TAG,
-    VIEWGRAPH_REPORT_TAG,
-    TwoViewEstimationReport,
-    TwoViewEstimator,
-)
+from gtsfm.two_view_estimator import (POST_ISP_REPORT_TAG,
```
akshay-krishnan: there seems to be some weird formatting here, might get flagged by the CI.

akshay-krishnan (commented):
High level comment: we need docstrings. 
At different levels: document level, class level and method/function level. 
Please take a look at other gtsfm files.

gtsfm/utils/io.py
tiantianxiangshang629: Yes, I was. The reason for using it is that Cal3Bundler does not accept those distortion arguments from OPENCV.

gtsfm/scene_optimizer.py
tiantianxiangshang629: I implemented a new create_output_directories function to create the partition result directories.

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: Done

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: Done

gtsfm/graph_partitioner/binary_tree_partition_test.py
tiantianxiangshang629: Done

gtsfm/scene_optimizer.py
tiantianxiangshang629: Done

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: Done

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: Done

gtsfm/runner/gtsfm_runner_base.py
tiantianxiangshang629: I updated the scene optimizer to read the partition config from the yaml file.

gtsfm/utils/io.py
travisdriver: How did you get it to run with Cal3DS2 without adding Cal3DS2 support to the bundle adjustment module? Only Cal3Bundler and Cal3Fisheye are supported I believe.

travisdriver: Could you remove this since it's unclear whether Cal3DS2 is actually used in BA? I will open another PR this week that added distortion model compatibility.

tiantianxiangshang629: Sorry I forgot to remove this. Done.

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,196 @@
...
+
+    def __init__(self, max_depth: int = 2):
+        """
+        Initialize the BinaryTreePartition object.
+
+        Args:
+            max_depth: Maximum depth of the binary tree; results in 2^depth partitions.
+        """
+        super().__init__(process_name="BinaryTreePartition")
+        self.max_depth = max_depth
```
akshay-krishnan: does it make sense to have a more user-friendly option here? for ex: number of cameras per cluster is something a user could tune, but no idea how that maps to max_depth.

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,196 @@
...
+            nx_graph: NetworkX graph built from image pairs.
+
+        Returns:
+            A list of dictionaries containing partition details per leaf node.
+        """
+        if node.is_leaf():
+            explicit_keys = set(node.keys)
+            return [
+                {
+                    "explicit_keys": [gtsam.Symbol(u).index() for u in explicit_keys],
```
akshay-krishnan: not sure what the term "explicit" means in this context. do you mean keys that dont occur in other clusters?

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,196 @@
...
+        """
+        ordered_keys = [ordering.at(i) for i in range(ordering.size())]
+
+        def split(keys: List[int], depth: int) -> BinaryTreeNode:
+            if depth == self.max_depth:
+                return BinaryTreeNode(keys, depth)
+
+            mid = len(keys) // 2
+            left_node = split(keys[:mid], depth + 1)
+            right_node = split(keys[mid:], depth + 1)
```
akshay-krishnan, line 165: so how does this result in overlapping nodes?

gtsfm/graph_partitioner/binary_tree_partition.py
```diff
@@ -0,0 +1,196 @@
...
+        image_pairs_per_partition = [[] for _ in range(num_leaves)]
+
+        partition_details = self._compute_leaf_partition_details(binary_tree_root_node, nx_graph)
+
+        logger.info(f"BinaryTreePartition: partitioned into {len(partition_details)} leaf nodes.")
+
+        for i in range(num_leaves):
+            edges_explicit = partition_details[i].get("edges_within_explicit", [])
+            edges_shared = partition_details[i].get("edges_with_shared", [])
+            image_pairs_per_partition[i] = edges_explicit + edges_shared
```
akshay-krishnan: so do shared edges have both nodes in this partition? if not, that should not be included in the returned result. We should maybe change the API to return shared edges and "explicit" edges separately.

akshay-krishnan: I expect shared edges to be useful for merging. So if we do return them, we should do a dict[tuple[int, int], list[edges]] where the dict keys are indices of the partition pairs that share those edges.

tiantianxiangshang629: done

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: It doesn't — the split() function ensures disjoint sets per leaf. However, edges across children introduce overlap — not in keys, but in edge connectivity.

tiantianxiangshang629:
> looking almost good. we likely need to change the graph_partitioner API to return intra-partition (prefer this to the word explicit) and inter-partition edges separately.
> 
> It looks like the partitions dont share nodes? So we dont currently handle merging such partitions.

Thanks for the feedback! Handling partition merging is definitely important, but it was outside the scope of this PR, which focuses on partitioning and exposing intra- and inter-partition edges. I agree we should revisit the merging logic — possibly as a follow-up PR.

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: Change explicit to exclusive

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: done

akshay-krishnan:
> > looking almost good. we likely need to change the graph_partitioner API to return intra-partition (prefer this to the word explicit) and inter-partition edges separately.
> > It looks like the partitions dont share nodes? So we dont currently handle merging such partitions.
> 
> Thanks for the feedback! Handling partition merging is definitely important, but it was outside the scope of this PR, which focuses on partitioning and exposing intra- and inter-partition edges. I agree we should revisit the merging logic — possibly as a follow-up PR.

I definitely did not mean that we should handle it in this PR, just that it should be documented. It would be good to know what kind of partitions the class produces to a user.

gtsfm/graph_partitioner/binary_tree_partition.py
akshay-krishnan: makes sense, it should maybe return the edges too then?

akshay-krishnan (approved):

tiantianxiangshang629:
In this PR, I’ve computed and stored them in self.inter_partition_edges_map, and exposed them via the get_inter_partition_edges() API. 

Totally understand your point on documentation, I have clarified in the docstring what kind of partitions this class produces.

gtsfm/graph_partitioner/binary_tree_partition.py
tiantianxiangshang629: Yes, both intra_partition_edges and inter_partition_edges are computed.
Each leaf partition includes its own intra_partition_edges, which are returned directly via partition_image_pairs(), while inter_partition_edges between sibling partitions are available via get_inter_partition_edges().

PR #845, opened 2025-06-17
Integrating mast3r as a correspondence generator

akshay-krishnan:
This PR integrates mast3r as a correspondence generator within gtsfm. We now add mast3r as a third-party submodule dependency, and also inherit some requirements from it in environment.yml. 

As a correspondence generator, we only use the keypoints and matches from mast3r and ignore the relate pose. Unlike other dense correspondence estimators in gtsfm, we do not use the "keypoint aggregation" step, as this is already done in the mast3r formulation, i.e matches arise from a unique (albeit low-resolution) grid on each image, which defines the keypoints.


environment_linux.yml
```diff
@@ -18,7 +18,7 @@ dependencies:
   - asyncssh
   - python-graphviz
   # core functionality and APIs
-  - matplotlib>=3.5
+  - matplotlib
```
travisdriver: Why this change?

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,341 @@
...
+
+    Returns:
+        The resized image.
+    """
+    max_size = max(img.height, img.width)
+    ratio = float(long_edge_size) / max_size
+    new_height = int(img.height * ratio)
+    new_width = int(img.width * ratio)
+    new_image = image_utils.resize_image(img, new_height, new_width)
+    return new_image
```
travisdriver: Do we not have a function for this already? If not, maybe we should move this to image utils

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,341 @@
...
+            idx=img_id,
+            instance=str(img_id),
+        ),
+        crop_start,
+        scale_factor,
+    )
+
+
+def symmetric_inference(model, img1, img2, device):
+    """Performs symmetric inference using the MASt3R model on a pair of images.
```
travisdriver: Could you add a slightly more detailed description of what symmetric inference is?

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,341 @@
...
+    max_size = max(img.height, img.width)
+    ratio = float(long_edge_size) / max_size
+    new_height = int(img.height * ratio)
+    new_width = int(img.width * ratio)
+    new_image = image_utils.resize_image(img, new_height, new_width)
+    return new_image
+
+
+def preprocess_image(image: Image, img_id: int, device) -> Tuple[Dict[str, Any], Tuple[int, int], float]:
+    """Preprocesses image for MASt3R model input, including resizing and normalization.
```
travisdriver: Does it do something other than resize and normalization?

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,341 @@
...
+
+    # decoder 1-2
+    res11, res21 = decoder(feat1, feat2, pos1, pos2, shape1, shape2)
+    # decoder 2-1
+    res22, res12 = decoder(feat2, feat1, pos2, pos1, shape2, shape1)
+
+    return (res11, res21, res22, res12)
+
+
+def extract_correspondences(feats, qonfs, subsample=8, device=None, ptmap_key="pred_desc"):
```
travisdriver: Maybe this should be a static method of the class if no other classes use it?

travisdriver: Same with functions above this

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,341 @@
...
+                image_idx: The index of the image.
+                keypoints: The new keypoints to add.
+                keypoints_idx: The original indices of the new keypoints in the feature grid.
+            """
+            existing_idx = indices_for_image.get(image_idx, np.array([], dtype=np.int64))
+            indices_for_image[image_idx], unique_idx = np.unique(
+                np.concatenate([existing_idx, keypoints_idx]), return_index=True
+            )
+            existing_keypoints = keypoints_for_image.get(image_idx, np.array([], dtype=np.float32).reshape(0, 2))
+            keypoints_for_image[image_idx] = np.vstack([existing_keypoints, keypoints])[unique_idx]
```
travisdriver, line 123: Would it be clearer to make this not a nested function?

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,341 @@
...
+                    descs, qonfs, device=device, subsample=subsample
+                )
+                conf_score = (C11.mean() * C12.mean() * C21.mean() * C22.mean()).sqrt().sqrt()
+                # TODO: use the matching score to filter correspondences.
+                matching_score = (float(conf_score), float(scores.sum()), len(scores))
+
+                matches_im1 = ((matches_im1 + np.array(crop1)) + 0.5) / scale1
+                matches_im2 = ((matches_im2 + np.array(crop2)) + 0.5) / scale2
+
+            return (matches_im1, matches_im2, idx1, idx2)
```
travisdriver: I'm a little worried about the readability of this function with two nested functions

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: I dont think we have a function for resizing acc to longest edge. Sure, moved it.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: updated

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: not necessarily. I basically copied these functions from mast3r codebase (so theyre not super readable). I can make them staticmethods in the gtsfm class, but I wanted to keep that class uncluttered. To some degree, readers dont need to fully understand what `extract_correspondences` does, but can still understand the class at a high level. But sure, it can be a staticmethod if you prefer that.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: fair, I moved one of them outside. The other is designed to be a nested function (it changes those dicts inside it). I can force it to be non-nested if you prefer, but I dont think it hurts readability much as is.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: It is designed to be a nested function (it changes those dicts inside it). I can force it to be non-nested if you prefer, but I dont think it hurts readability much as is.

the downside of being a non-nested function is that it becomes less obvious  that it is NOT functional and changes the dicts (so hurts readability IMO).

environment_linux.yml
akshay-krishnan: good call, restored.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
travisdriver: Still think this would be useful

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
travisdriver: I think I would prefer it to be a static class method since it's only valid for the Mast3rCorrespondenceGenerator class. I would also prefix it with an underscore to signify it's for internal use only.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
travisdriver: That's makes sense

environment_linux.yml
```diff
@@ -28,12 +28,12 @@ dependencies:
...
   - scipy
   - hydra-core
-  # # 3rd party algorithms for different modules
+  # 3rd party algorithms for different modules
   - kornia>=0.7.3
   - pycolmap
   # io
   - h5py
-  - plotly>=5.0
+  - plotly
```
travisdriver: Why this change?

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
```diff
@@ -0,0 +1,332 @@
...
+            descs, qonfs, device=device, subsample=subsample
+        )
+        conf_score = (C11.mean() * C12.mean() * C21.mean() * C22.mean()).sqrt().sqrt()
+        # TODO: use the matching score to filter correspondences.
+        matching_score = (float(conf_score), float(scores.sum()), len(scores))
+
+        matches_im1 = ((matches_im1 + np.array(crop1)) + 0.5) / scale1
+        matches_im2 = ((matches_im2 + np.array(crop2)) + 0.5) / scale2
+
+    return (matches_im1, matches_im2, idx1, idx2)
```
travisdriver: Would also prefer this to be a static method

travisdriver (approved):
LGTM modulo some small outstanding comments! Please make sure the CI errors aren't due to changes in this PR before merging.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: The use of a static method from within the class's methods is a bit weird, but alright, I dont have too strong an opinion against it. Typically its meant for external use.

gtsfm/frontend/correspondence_generator/mast3r_correspondence_generator.py
akshay-krishnan: The use of a static method from within the class's methods is a bit weird, but alright, I dont have too strong an opinion against it. Typically static methods are also meant for external use.

akshay-krishnan:
@dellaert my unit test seems to be failing due to what looks like a huggingface version issue (it did pass before my last commit). should not block you from trying it out.

PR #847, opened 2025-06-29
Formatting fixes only

akshay-krishnan:
To fix the currently failing CI format checker.

### DIFF
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -12,11 +12,8 @@ import numpy as np
 
 from dask import config as dask_config
 from dask.distributed import Client, LocalCluster, SSHCluster, performance_report
-from gtsam import (
-    Rot3,
-    Unit3,
-    Pose3,
-)
+from gtsam import Rot3, Pose3, Unit3
+
 from hydra.utils import instantiate
 from omegaconf import OmegaConf
 
@@ -51,7 +48,7 @@ class GtsfmRunnerBase:
     def tag(self):
         pass
 
-    def __init__(self, override_args = None) -> None:
+    def __init__(self, override_args=None) -> None:
         argparser: argparse.ArgumentParser = self.construct_argparser()
         self.parsed_args: argparse.Namespace = argparser.parse_args(args=override_args)
         if self.parsed_args.dask_tmpdir:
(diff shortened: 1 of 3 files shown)


gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -5,15 +5,19 @@
 import time
 from abc import abstractmethod, abstractproperty
 from pathlib import Path
-from typing import Any, Dict, List, Tuple, Set
+from typing import Any, Dict, List, Tuple
```
dellaert: These are no longer needed in later versions of python...

gtsfm/utils/merging.py
```diff
@@ -2,9 +2,19 @@
 
 Authors: Richi Dubey
 """
+
 import numpy as np
-from typing import Any, Dict, List, Tuple, Set
-from gtsam import Rot3, Unit3, NonlinearFactorGraph, Values, Symbol, noiseModel, PriorFactorPose3, LevenbergMarquardtOptimizer, Pose3
+from typing import Dict, List, Tuple, Set
```
dellaert: same, you could just use list, dict, etc...

dellaert (approved):

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -5,15 +5,18 @@
...
 import numpy as np
 
 from dask import config as dask_config
 from dask.distributed import Client, LocalCluster, SSHCluster, performance_report
-from gtsam import Rot3, Unit3, NonlinearFactorGraph, Values, Symbol, noiseModel, PriorFactorPose3, LevenbergMarquardtOptimizer, Pose3
+from gtsam import (
+    Rot3,
+    Unit3,
+    Pose3,
+)
```
travisdriver: Move to one line

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -48,7 +51,7 @@ class GtsfmRunnerBase:
     def tag(self):
         pass
 
-    def __init__(self, override_args: Any = None) -> None:
+    def __init__(self, override_args = None) -> None:
```
travisdriver: Why remove the type hint here?

akshay-krishnan: I just removed the dependency on typing. "Any" is not very useful anyway.

travisdriver: I still think it's better than nothing

PR #849, opened 2025-07-09
Fix dependency versions

akshay-krishnan:
PR to pin the versions of some dependencies, hopefully this fixes the CI.

### DIFF
--- a/environment_linux.yml
+++ b/environment_linux.yml
@@ -30,7 +30,7 @@ dependencies:
   - hydra-core
   # # 3rd party algorithms for different modules
   - kornia>=0.7.3
-  - pycolmap
+  - pycolmap==3.11.1
   # io
   - h5py
   - plotly>=5.0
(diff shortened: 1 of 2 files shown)


dellaert:
CI fails with
```
E   AttributeError: 'gtsam.gtsam.Cal3_S2' object has no attribute 'compose'
```

and indeed, I don't know what such a method would do :-)

akshay-krishnan:
> LGTM, CI failure is probably unrelated to this PR

It comes from migrating to the newer gtsam version. Apparently `createPoses` used to do something like `createCameras(cal)`, but now is more true to its name: https://github.com/borglab/gtsam/commit/26a3728d8068e50049a4d70cf5d3cc243d57c8f9

 gtsfm uses it (in surprisingly many tests!), expecting cameras in return. So I'll create a createCameras wrapper and use that instead.

PR #851, opened 2025-07-23
Add Support for Cal3DS2

travisdriver:
- Adds support for Cal3DS2 and Cal3_S2. 
- Fixed a bug where [if you resized the input image the calibration would default back to Cal3Bundler and ignore distortion](https://github.com/borglab/gtsfm/blob/1def45f50e8a9501ada3d79cfbe54151612a2b04/gtsfm/loader/loader_base.py#L218).
- Now uses original Gerrard Hall dataset with OPENCV model

### DIFF
--- a/gtsfm/utils/io.py
+++ b/gtsfm/utils/io.py
@@ -15,7 +15,7 @@ import h5py
 import numpy as np
 import open3d
 import simplejson as json
-from gtsam import Cal3Bundler, Cal3_S2, Cal3DS2, Point3, Pose3, Rot3, SfmTrack
+from gtsam import Cal3Bundler, Point3, Pose3, Rot3, SfmTrack
 from PIL import Image as PILImage
 from PIL.ExifTags import GPSTAGS, TAGS
 
@@ -313,9 +313,6 @@ def write_cameras(gtsfm_data: GtsfmData, images: List[Image], save_dir: str) -> 
     os.makedirs(save_dir, exist_ok=True)
 
     # TODO: handle shared intrinsics
-    # Assumes all camera models have five intrinsic parameters
-    camera_model = "RADIAL"
-
     file_path = os.path.join(save_dir, "cameras.txt")
     with open(file_path, "w") as f:
         f.write("# Camera list with one line of data per camera:\n")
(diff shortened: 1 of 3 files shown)


gtsfm/data_association/point3d_initializer.py
```diff
@@ -136,9 +135,7 @@ def __init__(self, track_camera_dict: Dict[int, gtsfm_types.CAMERA_TYPE], option
             raise ValueError("No camera positions were estimated, so triangulation is not feasible.")
 
         sample_camera = list(self.track_camera_dict.values())[0]
-        self._camera_set_class = (
-            CameraSetCal3Bundler if isinstance(sample_camera, PinholeCameraCal3Bundler) else CameraSetCal3Fisheye
-        )
+        self._camera_set_class = gtsfm_types.get_camera_set_class_for_calibration(sample_camera.calibration())
```
akshay-krishnan, line 138: does not need to be done in this PR, but I wonder if the latest version of gtsam supports set(CameraClass) using the native python set. This is possible for things like vector now, so hopefully for sets too?

gtsfm/utils/io.py
```diff
@@ -209,24 +209,35 @@ def colmap2gtsfm(
...
+        elif camera_model_name == "OPENCV":
+            # See https://github.com/colmap/colmap/blob/1f6812e333a1e4b2ef56aa74e2c3873e4e3a40cd/src/colmap/sensor/models.h#L241  # noqa: E501
+            fx, fy, cx, cy, k1, k2, p1, p2 = cameras[img.camera_id].params
+        else:
+            raise ValueError(f"Unsupported COLMAP camera type: {camera_model_name}")
+
+        if camera_model_name in ["SIMPLE_RADIAL", "RADIAL"]:
+            intrinsics_gtsfm.append(Cal3Bundler(f, k1, k2, cx, cy))
+        elif camera_model_name in ["PINHOLE"]:
+            # TODO(travisdriver): Use Cal3_S2 instead.
```
akshay-krishnan: so remove the TODO now?

akshay-krishnan (approved):

travisdriver:
newplot

[visual_comparison_dashboard.zip](https://github.com/user-attachments/files/21620384/visual_comparison_dashboard.zip)

gtsfm/utils/io.py
travisdriver: Done

PR #853, opened 2025-08-07
Gaussian Splatting Implementation that works on Mac

dellaert:
Merged #852 into this branch and then made changes to make it work on Mac.
Trick to make it work with Hydra: dummy class on Mac:
```python
if sys.platform == "darwin":
    class GaussianSplatting:
        def __init__(self, *args, **kwargs):
            """ Gaussian Splatting is not supported on Mac (darwin).
                Please run on Linux with CUDA and gsplat installed."""
            pass
else:
    imports...
    class GaussianSplatting:
    ...
```

Also made a change to display
```
🚀 Dask Dashboard available at: http://127.0.0.1:8787/status
```

when running :-)

### DIFF
--- a/environment_mac.yml
+++ b/environment_mac.yml
@@ -39,6 +39,7 @@ dependencies:
   - pyparsing>=3.0.9
   # testing
   - parameterized
+  - wget
   - pip:
       - torch
       - torchvision>=0.20.0
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -99,6 +99,12 @@ class GtsfmRunnerBase:
             help="Override flag for retriever (choose from among gtsfm/configs/retriever).",
         )
         parser.add_argument(
+            "--gaussian_splatting_config_name",
+            type=str,
+            default="base_gs",
+            help="Override flag for your own gaussian splatting implementation.",
+        )
+        parser.add_argument(
             "--max_resolution",
             type=int,
             default=760,
@@ -120,7 +126,8 @@ class GtsfmRunnerBase:
         parser.add_argument(
             "--share_intrinsics", action="store_true", help="Shares the intrinsics between all the cameras."
         )
-        parser.add_argument("--mvs_off", action="store_true", help="Turn off dense MVS reconstruction")
+        parser.add_argument("--run_mvs", action="store_true", help="Run dense MVS reconstruction")
+        parser.add_argument("--run_gs", action="store_true", help="Run Gaussian Splatting")
         parser.add_argument(
             "--output_root",
             type=str,
@@ -209,6 +216,15 @@ class GtsfmRunnerBase:
                 logger.info("\n\nRetriever override: " + OmegaConf.to_yaml(retriever_cfg))
                 scene_optimizer.image_pairs_generator._retriever = instantiate(retriever_cfg.retriever)
 
+        # Override gaussian splatting
+        if self.parsed_args.gaussian_splatting_config_name is not None:
+            with hydra.initialize_config_module(config_module="gtsfm.configs.gaussian_splatting"):
+                gs_cfg = hydra.compose(
+                    config_name=self.parsed_args.gaussian_splatting_config_name,
+                )
+                logger.info("\n\nGaussian Splatting override: " + OmegaConf.to_yaml(gs_cfg))
+                scene_optimizer.gaussian_splatting_optimizer = instantiate(gs_cfg.gaussian_splatting_optimizer)
+
         if self.parsed_args.max_frame_lookahead is not None:
             if scene_optimizer.image_pairs_generator._retriever._matching_regime in [
                 ImageMatchingRegime.SEQUENTIAL,
@@ -245,8 +261,11 @@ class GtsfmRunnerBase:
                     f"{scene_optimizer.image_pairs_generator._retriever._matching_regime}"
                 )
 
-        if self.parsed_args.mvs_off:
+        if not self.parsed_args.run_mvs:
             scene_optimizer.run_dense_optimizer = False
+        
+        if not self.parsed_args.run_gs:
+            scene_optimizer.run_gaussian_splatting_optimizer = False
 
         logger.info("\n\nSceneOptimizer: " + str(scene_optimizer))
         return scene_optimizer
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -10,7 +10,8 @@ import dask
 import hydra
 import numpy as np
 from dask import config as dask_config
-from dask.distributed import Client, LocalCluster, SSHCluster, performance_report
+from dask.distributed import (Client, LocalCluster, SSHCluster,
+                              performance_report)
 from gtsam import Pose3, Rot3, Unit3
 from hydra.utils import instantiate
 from omegaconf import OmegaConf
@@ -23,12 +24,14 @@ import gtsfm.utils.viz as viz_utils
 from gtsfm import two_view_estimator
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
-from gtsfm.frontend.correspondence_generator.image_correspondence_generator import ImageCorrespondenceGenerator
+from gtsfm.frontend.correspondence_generator.image_correspondence_generator import \
+    ImageCorrespondenceGenerator
 from gtsfm.graph_partitioner.graph_partitioner_base import GraphPartitionerBase
 from gtsfm.loader.loader_base import LoaderBase
 from gtsfm.retriever.retriever_base import ImageMatchingRegime
 from gtsfm.scene_optimizer import SceneOptimizer
-from gtsfm.two_view_estimator import TWO_VIEW_OUTPUT, TwoViewEstimationReport, run_two_view_estimator_as_futures
+from gtsfm.two_view_estimator import (TWO_VIEW_OUTPUT, TwoViewEstimationReport,
+                                      run_two_view_estimator_as_futures)
 from gtsfm.ui.process_graph_generator import ProcessGraphGenerator
 from gtsfm.utils.subgraph_utils import group_results_by_subgraph
 
@@ -324,6 +327,9 @@ class GtsfmRunnerBase:
             cluster = LocalCluster(**local_cluster_kwargs)
             client = Client(cluster)
 
+        # Display Dask dashboard URL before processing starts
+        print(f"\n🚀 Dask Dashboard available at: {client.dashboard_link}")
+        
         # Create process graph.
         process_graph_generator = ProcessGraphGenerator()
         if isinstance(self.scene_optimizer.correspondence_generator, ImageCorrespondenceGenerator):
--- a/gtsfm/runner/gtsfm_runner_base.py
+++ b/gtsfm/runner/gtsfm_runner_base.py
@@ -10,8 +10,7 @@ import dask
 import hydra
 import numpy as np
 from dask import config as dask_config
-from dask.distributed import (Client, LocalCluster, SSHCluster,
-                              performance_report)
+from dask.distributed import Client, LocalCluster, SSHCluster, performance_report
 from gtsam import Pose3, Rot3, Unit3
 from hydra.utils import instantiate
 from omegaconf import OmegaConf
@@ -24,14 +23,12 @@ import gtsfm.utils.viz as viz_utils
 from gtsfm import two_view_estimator
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.evaluation.metrics import GtsfmMetric, GtsfmMetricsGroup
-from gtsfm.frontend.correspondence_generator.image_correspondence_generator import \
-    ImageCorrespondenceGenerator
+from gtsfm.frontend.correspondence_generator.image_correspondence_generator import ImageCorrespondenceGenerator
 from gtsfm.graph_partitioner.graph_partitioner_base import GraphPartitionerBase
 from gtsfm.loader.loader_base import LoaderBase
 from gtsfm.retriever.retriever_base import ImageMatchingRegime
 from gtsfm.scene_optimizer import SceneOptimizer
-from gtsfm.two_view_estimator import (TWO_VIEW_OUTPUT, TwoViewEstimationReport,
-                                      run_two_view_estimator_as_futures)
+from gtsfm.two_view_estimator import TWO_VIEW_OUTPUT, TwoViewEstimationReport, run_two_view_estimator_as_futures
 from gtsfm.ui.process_graph_generator import ProcessGraphGenerator
 from gtsfm.utils.subgraph_utils import group_results_by_subgraph
 
@@ -266,7 +263,7 @@ class GtsfmRunnerBase:
 
         if not self.parsed_args.run_mvs:
             scene_optimizer.run_dense_optimizer = False
-        
+
         if not self.parsed_args.run_gs:
             scene_optimizer.run_gaussian_splatting_optimizer = False
 
@@ -329,7 +326,7 @@ class GtsfmRunnerBase:
 
         # Display Dask dashboard URL before processing starts
         print(f"\n🚀 Dask Dashboard available at: {client.dashboard_link}")
-        
+
         # Create process graph.
         process_graph_generator = ProcessGraphGenerator()
         if isinstance(self.scene_optimizer.correspondence_generator, ImageCorrespondenceGenerator):
--- a/gtsfm/scene_optimizer.py
+++ b/gtsfm/scene_optimizer.py
@@ -25,26 +25,23 @@ import gtsfm.utils.ellipsoid as ellipsoid_utils
 import gtsfm.utils.io as io_utils
 import gtsfm.utils.logger as logger_utils
 import gtsfm.utils.viz as viz_utils
-import gtsfm.utils.splat as splat_utils
-import gtsfm.splat.rendering as gtsfm_rendering
 from gtsfm.common.gtsfm_data import GtsfmData
 from gtsfm.common.image import Image
 from gtsfm.common.keypoints import Keypoints
 from gtsfm.common.pose_prior import PosePrior
 from gtsfm.densify.mvs_base import MVSBase
-from gtsfm.splat.gs_base import GSBase
-from gtsfm.frontend.correspondence_generator.correspondence_generator_base import CorrespondenceGeneratorBase
+from gtsfm.frontend.correspondence_generator.correspondence_generator_base import \
+    CorrespondenceGeneratorBase
 from gtsfm.graph_partitioner.graph_partitioner_base import GraphPartitionerBase
 from gtsfm.graph_partitioner.single_partition import SinglePartition
 from gtsfm.multi_view_optimizer import MultiViewOptimizer
 from gtsfm.retriever.image_pairs_generator import ImagePairsGenerator
 from gtsfm.retriever.retriever_base import ImageMatchingRegime
-from gtsfm.two_view_estimator import (
-    POST_ISP_REPORT_TAG,
-    VIEWGRAPH_REPORT_TAG,
-    TwoViewEstimationReport,
-    TwoViewEstimator,
-)
+from gtsfm.splat.gs_base import GSBase
+from gtsfm.two_view_estimator import (POST_ISP_REPORT_TAG,
+                                      VIEWGRAPH_REPORT_TAG,
+                                      TwoViewEstimationReport,
+                                      TwoViewEstimator)
 
 matplotlib.use("Agg")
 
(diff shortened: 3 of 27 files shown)


environment_mac.yml
```diff
@@ -39,6 +39,7 @@ dependencies:
   - pyparsing>=3.0.9
   # testing
   - parameterized
+  - wget
   - pip:
       - torch
```
akshay-krishnan, line 44: do we need the versions pinned for the mac env too?

gtsfm/runner/gtsfm_runner_base.py
akshay-krishnan, line 1: it looks like your formatter is not properly configured? A lot of things seem to be reformatted, which they shouldnt be. we use line-width 120.

gtsfm/runner/gtsfm_runner_base.py
```diff
@@ -245,8 +264,11 @@ def construct_scene_optimizer(self) -> SceneOptimizer:
                     f"{scene_optimizer.image_pairs_generator._retriever._matching_regime}"
                 )
 
-        if self.parsed_args.mvs_off:
+        if not self.parsed_args.run_mvs:
             scene_optimizer.run_dense_optimizer = False
```
akshay-krishnan, line 265: these options inside scene_optimizer should also be default False. And then you could do, 

if self.parsed_args.run_mvs: 
    scene_optimizer.run_dense_optimizer = True

gtsfm/scene_optimizer.py
akshay-krishnan, line 1: formatting inconsistent for this file as well

gtsfm/scene_optimizer.py
```diff
@@ -281,6 +290,33 @@ def create_computation_graph(
                 metrics_graph_list.append(densify_metrics_graph)
             if downsampling_metrics_graph is not None:
                 metrics_graph_list.append(downsampling_metrics_graph)
+        
+        if self.run_gaussian_splatting_optimizer and self.gaussian_splatting_optimizer is not None:
+            import gtsfm.splat.rendering as gtsfm_rendering
+            import gtsfm.utils.splat as splat_utils
```
akshay-krishnan: we just import everything at the top. I dont think the environment is configured to skip dependencies if users are not splatting, so might as well import them all at once.

gtsfm/scene_optimizer.py
```diff
@@ -281,6 +290,33 @@ def create_computation_graph(
...
+            img_dict_graph = dask.delayed(get_image_dictionary)(images)
```
akshay-krishnan, line 298: If this function doing some I/O? If so, you will need a decorator to do this. we restrict all I/O to happen on specific workers on dask.

gtsfm/scene_optimizer.py
```diff
@@ -281,6 +290,33 @@ def create_computation_graph(
...
+                        splats=splats_graph
+                    )
+                )
+                delayed_results.append(
+                    dask.delayed(gtsfm_rendering.generate_interpolated_video)(
+                        images_graph = img_dict_graph, 
+                        sfm_result_graph = ba_output_graph,
+                        cfg_result_graph = cfg_graph,
+                        splats_graph=splats_graph,
+                        video_path = self._interp_video_save_fpath,
```
akshay-krishnan: nit: would rename that variable to `video_fpath`. Path could also mean the camera trajectory.

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+    # --- Rendering ---
+    fps: int = 30
+    num_frames: int = 10
+
+if sys.platform == "darwin":
+    class GaussianSplatting:
+        def __init__(self, *args, **kwargs):
+            """ Gaussian Splatting is not supported on Mac (darwin).
+                Please run on Linux with CUDA and gsplat installed."""
+            pass
```
akshay-krishnan, line 96: @dellaert  do we need this here? or in whatever file imports this class? I guess this is not directly imported when its used (because we use hydra?). Its just not ideal to have everything indented.

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+            if use_sfm_init:
+                logger.info("Initializing from SfM points.")
+                points = torch.from_numpy(self.full_dataset.points).float()
+                sh0_values = (torch.from_numpy(self.full_dataset.point_colors).float() - 0.5) / 0.28209479177387814
+                features_dc = torch.nn.Parameter(sh0_values)
+            else:
+                if init_type == "sfm":
+                    logger.info("Warning: 'sfm' init chosen but no points found. Falling back to 'random'.")
+                logger.info(f"Initializing with {init_num_pts} random points.")
+                random_scale = 10.0
```
akshay-krishnan: is that commonly used? maybe add a reference to wherever you found this? 
(seems a bit too high, but Ive never played with it) 

would make a global variable for this too

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+                sh_degree: degree for the spherical harmonics
+                batch_size: batch size for the setup
+            """
+            
+            use_sfm_init = (init_type == "sfm" and self.full_dataset.points is not None)
+            
+            if use_sfm_init:
+                logger.info("Initializing from SfM points.")
+                points = torch.from_numpy(self.full_dataset.points).float()
+                sh0_values = (torch.from_numpy(self.full_dataset.point_colors).float() - 0.5) / 0.28209479177387814
```
akshay-krishnan: we usually make a global variable like 

`_SH0_NORMALIZATION_FACTOR = 0.28209479177387814` 

and then use it. optionally could add a comment on where the number is from.

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+                features_dc = torch.nn.Parameter(sh0_values)
+            else:
+                if init_type == "sfm":
+                    logger.info("Warning: 'sfm' init chosen but no points found. Falling back to 'random'.")
+                logger.info(f"Initializing with {init_num_pts} random points.")
+                random_scale = 10.0
+                points = torch.nn.Parameter((torch.rand((init_num_pts, 3)) - 0.5) * random_scale)
+                features_dc = torch.nn.Parameter(torch.rand(init_num_pts, 3))
```
akshay-krishnan, line 189: these are just 0-1? looks like when you initialize from sfm points, they do take on negative values.

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+            self.optimizer_class = torch.optim.Adam
+            
+            self.optimizers = {
+                name: self.optimizer_class(
+                    [{"params": self.splats[name], "lr": lr * math.sqrt(batch_size), "name": name}],
+                    eps=1e-15 / math.sqrt(batch_size),
+                    betas=(1 - batch_size * (1 - 0.9), 1 - batch_size * (1 - 0.999)),
+                )
+                for name, _, lr in params
+            }
```
akshay-krishnan: we use a functional style, so all member variables are created in the constructor and never changed. 

I dont think `self.splats` and `self.optimizers` need to be member variables (do they?) as they are input-data dependent. Simply return them as results from this function and use them where they are to be used.

akshay-krishnan: Also all member variables need to be serializable for dask. 
Since you dont customize self.optimizer_class right now, that does not need to be a member variable.

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+            self,
+            camtoworlds: Tensor,
+            Ks: Tensor,
+            width: int,
+            height: int,
+            render_mode: str,
+            **kwargs,
+        ) -> Tuple[Tensor, Tensor, Dict]:
+
+            means = self.splats["means"]
```
akshay-krishnan: pass the splats as inputs please?

gtsfm/splat/gaussian_splatting.py
```diff
@@ -0,0 +1,406 @@
...
+            Args:
+                images_graph: A dictionary mapping indices to Image objects.
+                sfm_result_graph: A GtsfmData object containing poses, points, etc.
+
+            Returns:
+                splats: 3D Gaussian Splats defining the entire scene
+                cfg: Config file with training parameters
+            """
+
+            self.full_dataset = GaussianSplattingData(images_graph, sfm_result_graph)
```
akshay-krishnan: same comment for this, we dont modify `self` outside the constructor.

gtsfm/splat/gs_data.py
```diff
@@ -0,0 +1,221 @@
...
+        self.actual_img_dims = [(images[i].height, images[i].width) for gs_i, i in self._gaussiansplatting_idx_to_camera_idx.items()]
+        
+        self._intrinsics = [self._sfm_result.get_camera(i).calibration().K() for gs_i, i in self._gaussiansplatting_idx_to_camera_idx.items()]
+        
+        self._dataparser_outputs = self._generate_dataparser_outputs_from_gtsfm_data(self._sfm_result, self._images)
+
+        self._camtoworlds = self._dataparser_outputs['cameras']['camera_to_worlds'].numpy()
+        
+        self._scene_scale = self._dataparser_outputs['dataparser_scale']
+        self.transform_matrix = self._dataparser_outputs['transform_matrix']
```
akshay-krishnan: nit, we name transformation matrices as w_T_cam, which transforms from camera frame to world frame (right side of T is the source frame and the left side the target).

akshay-krishnan: you could also use something like "camtoworld", though ideally not.

gtsfm/splat/rendering.py
```diff
@@ -0,0 +1,153 @@
...
+    def interp(points, n, k, s):
+        """Runs multidimensional B-spline interpolation on the input points."""
+        sh = points.shape
+        pts = np.reshape(points, (sh[0], -1))
+        k = min(k, sh[0] - 1)
+        tck, _ = scipy.interpolate.splprep(pts.T, k=k, s=s)
+        u = np.linspace(0, 1, n, endpoint=False)
+        new_points = np.array(scipy.interpolate.splev(u, tck))
+        new_points = np.reshape(new_points.T, (n, sh[1], sh[2]))
+        return new_points
```
akshay-krishnan: move the functions outside this function. rename `interp` to `b_spline_interpolate`

gtsfm/splat/rendering.py
```diff
@@ -0,0 +1,153 @@
...
+            far_plane=cfg.far_plane,
+            render_mode="RGB+ED",
+        )  # [1, H, W, 4]
+        colors = torch.clamp(renders[..., 0:3], 0.0, 1.0)  # [1, H, W, 3]
+        # depths = renders[..., 3:4]  # [1, H, W, 1]
+        # depths = (depths - depths.min()) / (depths.max() - depths.min())
+        # canvas_list = [colors, depths.repeat(1, 1, 1, 3)]
+
+        # write images
+        # canvas = torch.cat(canvas_list, dim=2).squeeze(0).cpu().detach().numpy()
```
akshay-krishnan: maybe write the depths too? does not need to be done in this PR, but since you have the code here already, why not uncomment it?

scripts/gaussian_splatting/custom_dataloader.py
```diff
@@ -0,0 +1,362 @@
...
+from typing import Any, Dict, Literal, Tuple, Optional
+import cv2
+import numpy as np
+import torch
+import torch.nn.functional as F
+from gtsfm.utils.io import read_scene_data_from_colmap_format
+
+
+# mimics the rotation_matrix_between function from https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/cameras/camera_utils.py
+def get_rotation_matrix_from_two_vectors(vec1: torch.Tensor, vec2: torch.Tensor) -> torch.Tensor:
```
akshay-krishnan: these functions are defined in the utils, why are they repeated here?

scripts/gaussian_splatting/custom_dataloader.py
```diff
@@ -0,0 +1,362 @@
...
+        ]
+    )
+    theta = torch.acos(torch.clip(torch.dot(a, b), -1, 1))
+
+    # Rodrigues rotation formula. https://en.wikipedia.org/wiki/Rodrigues%27_rotation_formula
+    return torch.eye(3) + torch.sin(theta) * skew_sym_mat + (1 - torch.cos(theta)) * (skew_sym_mat @ skew_sym_mat)
+
+
+# this function is taken from https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/cameras/camera_utils.py and has reduced functionality
+def auto_orient_and_center_poses(
```
akshay-krishnan: also repeated function

scripts/gaussian_splatting/custom_dataloader.py
```diff
@@ -0,0 +1,362 @@
...
+    else:
+        raise ValueError(f"Unknown orientation method: {method}")
+    
+    transform = torch.cat([R, R @ -translation[..., None]], dim=-1)
+    poses_new = transform.to(poses.device) @ poses
+
+    return poses_new, transform
+
+# follows _undistort_image function from https://github.com/nerfstudio-project/nerfstudio/blob/main/nerfstudio/data/utils/dataloaders.py
+def _undistort_image(
```
akshay-krishnan: also repeated function

gtsfm/scene_optimizer.py
hkhanuja: I think Frank made this change to avoid errors while running on Mac, but need to confirm with him.

gtsfm/scene_optimizer.py
hkhanuja: The _get_image_dictionary_ was previously defined as 
def get_image_dictionary(image_list: List[Image]) -> Dict[int, Image]:
    """Convert a list of images to the MVS input format."""
    img_dict = {i: img for i, img in enumerate(image_list)}
    return img_dict
and used in mvs as well. It doesn't do any I/O operations.

gtsfm/runner/gtsfm_runner_base.py
hkhanuja: The _run_dense_optimizer_ in the scene optimizer is set as _self.run_dense_optimizer = self.dense_multiview_optimizer is not None_ and in most config yaml files _dense_multiview_optimizer_ target class is defined and in some it is commented out (so None basically), meaning by default it gets set to True mostly, that is why just changing this line won't work, will need to change the complete flow. Is that what we should do?

gtsfm/scene_optimizer.py
hkhanuja: okay will make that change

gtsfm/splat/gaussian_splatting.py
hkhanuja: I am not sure what you mean by commonly used here?

gtsfm/splat/gaussian_splatting.py
hkhanuja: okay will do that. The global variables declaration happens in the same file they are used or somewhere else?

scripts/gaussian_splatting/custom_dataloader.py
hkhanuja: yes this is the scripts folder, we discussed in the previous meeting to let it be like this and mention in the README that it is not maintained and is just for the purposes of introduction and/or experimentation.

gtsfm/splat/rendering.py
hkhanuja: does the depth video need to be a separate video or a side by side with the original?

gtsfm/splat/rendering.py
hkhanuja: I mentioned in a comment above the main function that is taken from _gsplat_ directly, they define it like that and that is the reason I did not change it. Should I do that?

gtsfm/splat/gs_data.py
hkhanuja: so this matrix basically is the matrix from the _auto_orient_and_center_poses_ function in the utils/splat.py file. It orients and centers the poses and then same transformation is applied to the sfm points. Not sure what the naming convention should be here

gtsfm/splat/gaussian_splatting.py
hkhanuja: okay, yes they are input-data dependent, it was just easier for me then.

gtsfm/scene_optimizer.py
hkhanuja: i can make the necessary changes to my vs code, so that it maintains consistency for future code as well. (will ask for those changes in the meeting)

gtsfm/splat/gaussian_splatting.py
hkhanuja: need a bit more explanation on what is meant here?

hkhanuja: The random initialization was taken from splatfacto (nerfstudio), I will have to see the mean values, but now when I think about it we can remove the random part altogether, because for us GS always runs after sparse recontruction. What do you think?

gtsfm/runner/gtsfm_runner_base.py
akshay-krishnan: one way to solve that would be override the main_cfg before instantiating scene_optimizer. 

The other is what youve done, but its not gtsfm style to modify class members :) maybe we can pass it as an argument to create_computation_graph, but I dont feel too strongly about it.

gtsfm/utils/splat.py
```diff
@@ -0,0 +1,248 @@
+"""Utility functions for 3D Gaussian Splatting.
```
akshay-krishnan, line 1: we typically unit test such util functions. would be good to add some tests.

gtsfm/utils/splat.py
```diff
@@ -0,0 +1,248 @@
...
+
+from gsplat import export_splats
+import torch
+import numpy as np
+from typing import Tuple, Literal
+import cv2
+import random
+from sklearn.neighbors import NearestNeighbors
+import math
+from gtsfm.utils import logger as logger_utils
```
akshay-krishnan: imports need to be sorted alphabetically

gtsfm/splat/gs_data.py
akshay-krishnan: if its rotation and translating, it is an SE3, so T works. could be something like gsplat_T_gtsfm, I'll leave the exact frame names to you.

gtsfm/scene_optimizer.py
akshay-krishnan: We should maybe add a comment that its an intentional exception to the norm to support mac.

gtsfm/splat/rendering.py
akshay-krishnan: yes please

gtsfm/splat/rendering.py
akshay-krishnan: what do you mean original?

scripts/gaussian_splatting/custom_dataloader.py
akshay-krishnan: yeah, i didnt leave too many comments for it, its okay, but good to just import these functions if theyre the same.

gtsfm/splat/rendering.py
hkhanuja: by original i mean the RGB video

gtsfm/splat/gs_data.py
hkhanuja: okay

gtsfm/runner/gtsfm_runner_base.py
hkhanuja: It was setup this way with MVS before as well, so we can leave it like this but if you do not feel as strongly about it

scripts/gaussian_splatting/custom_dataloader.py
hkhanuja: should I have a separate utils folder in the scripts or should I import from the utils/splat file

gtsfm/utils/splat.py
hkhanuja: all the functions are taken from nerfstudio with proper references in the comments above the function.

gtsfm/splat/gs_data.py
hkhanuja: the self.transformation_matrix does not convert from one coordinate system to another, it changes the orientation in the same coordinate system, thats why I think it should be left as is.

gtsfm/utils/splat.py
akshay-krishnan: sure, but tests help make sure that they dont break in the future when someone changes something. and right now none of the GS stuff has any tests, so could break after you graduate :)

gtsfm/splat/gaussian_splatting.py
akshay-krishnan: yeah, i cant remember what I meant here either

gtsfm/splat/rendering.py
```diff
@@ -0,0 +1,156 @@
...
+    """Converts from pose matrices to (position, lookat, up) format."""
+    pos = poses[:, :3, -1]
+    lookat = poses[:, :3, -1] - dist * poses[:, :3, 2]
+    up = poses[:, :3, -1] + dist * poses[:, :3, 1]
+    return np.stack([pos, lookat, up], 1)
+
+
+# Helper function for generate_interpolated_path function
+def points_to_poses(points):
+    """Converts from (position, lookat, up) format to pose matrices."""
```
akshay-krishnan: we need args and returns sections in the docstrings of all methods.

hkhanuja: added

gtsfm/utils/splat.py
hkhanuja: added

dellaert:
Woohoo ! Pls. delete the branch if you no longer need it.

PR #854, opened 2025-08-29
Save Cal3Bundler as RADIAL

travisdriver:
Saves Cal3Bundler as RADIAL instead of SIMPLE_RADIAL in `gtsfm_calibration_to_colmap_camera` to be consistent with `colmap_camera_to_gtsam_calibration`.

### DIFF
--- a/gtsfm/utils/pycolmap_utils.py
+++ b/gtsfm/utils/pycolmap_utils.py
@@ -97,7 +97,7 @@ def gtsfm_calibration_to_colmap_camera(
     """
     if isinstance(calibration, gtsam.Cal3Bundler):
         return ColmapCamera(
-            model="SIMPLE_RADIAL",
+            model="RADIAL",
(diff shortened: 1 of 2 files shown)


gtsfm/utils/pycolmap_utils.py
```diff
@@ -97,7 +97,7 @@ def gtsfm_calibration_to_colmap_camera(
     """
     if isinstance(calibration, gtsam.Cal3Bundler):
         return ColmapCamera(
-            model="SIMPLE_RADIAL",
+            model="RADIAL",
```
hkhanuja, line 100: yes, this was the only change needed. 👍

hkhanuja (approved):
seems good to me

dellaert:
Ah, excellent. I’ll just fixed that in #860 so I’ll cherry pick this into that PR.

PR #855, opened 2025-10-03
Fix some warnings/mypy errors

dellaert:
- Added line length args to black/flake8
- fixed some warnings
- ignored more files
- improved README
- Gentle warnings if weights or checkpoints not found, before exiting


dellaert:
@akshay-krishnan feel free to merge if you like it

.vscode/settings.json
```diff
@@ -1,12 +1,12 @@
 {
     "editor.formatOnSave": true,
     "editor.defaultFormatter": "ms-python.black-formatter",
+    "black-formatter.args": ["--line-length", "120"],
```
akshay-krishnan, line 4: does this work for you? for me it needed the setting in pyproject.toml. 
https://github.com/borglab/gtsfm/blob/master/pyproject.toml

gtsfm/densify/mvs_patchmatchnet.py
```diff
@@ -119,6 +119,13 @@ def densify(
 
         # Check if cuda devices are available, and load the pretrained model
         #   the pretrained checkpoint should be pre-downloaded using gtsfm/download_model_weights.sh
+        if not PATCHMATCHNET_WEIGHTS_PATH.exists():
+            logger.error(
+                "PatchmatchNet weights not found at %s. Please run 'bash download_model_weights.sh' from the repo root.",
+                PATCHMATCHNET_WEIGHTS_PATH,
+            )
+            exit(1)
```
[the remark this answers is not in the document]
akshay-krishnan: agree, not sure what exit would do to any downstream I/O.

.vscode/settings.json
dellaert: It did. I guess they agree so we can keep both

PR #859, opened 2025-10-05
Add splat transformation unit tests

hkhanuja:
Pull Request to add utility function and unit tests to transform a Gaussian Splat from one coordinate frame to another, using gtsam.Similarity3


tests/utils/test_splat_utils.py
```diff
@@ -86,6 +94,58 @@ def test_get_viewmat(self):
...
+        np.testing.assert_allclose(gauss_B_3["scales"], expected_scale_3, atol=1e-7)
+
+        # --- Test 4: Combined Transformation ---
+        R_4 = Rotation.from_euler("xyz", [30, 45, 60], degrees=True).as_matrix()
+        rot4 = gtsam.Rot3(R_4)
+        trans4 = gtsam.Point3(-1, 0.5, 1)
+        scale4 = 0.5
+        sim3_4 = gtsam.Similarity3(rot4, trans4, scale4)
+        gauss_B_4 = transform_gaussian(base_gaussian, sim3_4)
+        expected_mean_4 = scale4 * (R_4 @ base_gaussian["means"] + trans4)
```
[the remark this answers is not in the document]
dellaert: Not sure copilot knows about Eade, but check to make sure.

gtsfm/utils/splat.py
```diff
@@ -263,3 +264,29 @@ def get_viewmat(camera_to_world: torch.Tensor) -> torch.Tensor:
...
     viewmat[:, 3, 3] = 1.0
     return viewmat
+
+
+def transform_gaussian(gaussianA, sim3_B_A):
+    """
+    Transforms a Gaussian Splat from one coordinate system to another using gtsam.Similarity3
+    Args:
+        gaussianA (dict): A dictionary representing the Gaussian in coordinate system A.
+        sim3_B_A (gtsfm.Similarity3): The transformation from coordinate system A to B.
```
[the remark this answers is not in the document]
dellaert: copilot is correct

tests/utils/test_splat_utils.py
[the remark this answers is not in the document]
hkhanuja: this changes how point transformation actually happens in Similarity3.
p′=s(Rp+t)

.vscode/settings.json
```diff
@@ -71,25 +73,34 @@
         "keypoint",
         "keypoints",
         "Khanuja",
+        "kneighbors",
```
dellaert, line 77: Hmmm. A library?

.vscode/settings.json
```diff
@@ -71,25 +73,34 @@
...
+        "ndarray",
         "nerfstudio",
         "netvlad",
         "Olsson",
         "olssonloader",
         "partitioner",
-        "splatfacto",
         "Patchmatch",
         "patchmatchnet",
+        "quats",
```
dellaert: Hmmm. A library?

gtsfm/utils/splat.py
```diff
@@ -10,6 +10,7 @@
 import cv2
 import numpy as np
 import torch
+from scipy.spatial.transform import Rotation
```
dellaert: Why not use gtsam?

gtsfm/utils/splat.py
```diff
@@ -263,3 +264,29 @@ def get_viewmat(camera_to_world: torch.Tensor) -> torch.Tensor:
     viewmat[:, :3, 3:4] = T_inv
     viewmat[:, 3, 3] = 1.0
     return viewmat
+
+
+def transform_gaussian(gaussianA, sim3_B_A):
```
dellaert: add actual type annotations for mypy.

gtsfm/utils/splat.py
dellaert: OK. I feel pretty strongly about naming conventions here. we use: a Sim3 transforms *from* A *to* B, as in $aP = aSb \cdot bP$, with $aP$ a point in A and $bP$ a point in $b$, we use $aSb$. B goes in, A comes out...

gtsfm/utils/splat.py
```diff
@@ -263,3 +264,29 @@ def get_viewmat(camera_to_world: torch.Tensor) -> torch.Tensor:
...
+    Args:
+        gaussianA (dict): A dictionary representing the Gaussian in coordinate system A.
+        sim3_B_A (gtsfm.Similarity3): The transformation from coordinate system A to B.
+    Returns:
+        gaussianB: A dictionary representing the Gaussian in coordinate system B.
+    """
+    meanA = gaussianA["means"]
+    meanB = sim3_B_A.transformFrom(meanA)
+
+    rotationBA = Rotation.from_matrix(sim3_B_A.rotation().matrix())
```
dellaert: Please read an use gtsam code :-) I'm sure you can get a gtsam.Rot3, and that then supports *, without converting to matrix. You can use gtsam.Quaternion, but make sure to check the convention. Here is Rot3 constructor
```
Rot3::Rot3(const gtsam::Quaternion& q) : rot_(q.toRotationMatrix()) {
}
```
And gtsam.Quaternion is a typedef to [Eigen's Quaternion](https://libeigen.gitlab.io/eigen/docs-nightly/classEigen_1_1Quaternion.html), which has many* constructors. Assuming `gaussianA["quats"])).as_quat()` is a vector, it needs to be a 4D vector expression representing quaternion coefficients in the order [x, y, z, w].

gtsfm/utils/splat.py
```diff
@@ -263,3 +264,29 @@ def get_viewmat(camera_to_world: torch.Tensor) -> torch.Tensor:
...
+    rotationB = (rotationBA * Rotation.from_quat(gaussianA["quats"])).as_quat()
+
+    scaleB = sim3_B_A.scale() * gaussianA["scales"]
+
+    # we only update the means, quats and scales (which both result in covariance) as opacity and color do not change.
```
dellaert: Ah, why is that? Suppose my other frame is upside down?

tests/utils/test_splat_utils.py
```diff
@@ -5,18 +5,26 @@
 
 import unittest
 
+import gtsam
+import numpy as np
 import torch
+from scipy.spatial.transform import Rotation
 
 from gtsfm.utils.splat import (
```
dellaert: The style we try to adhere to does
```
from gtsfm.utils import splat`
```
and then uses `splat.transform_gaussian`

tests/utils/test_splat_utils.py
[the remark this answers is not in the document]
dellaert: I think you're correct. I told you the AI has strong biases against our convention :-(

.vscode/settings.json
hkhanuja: yes function, from sklearn.neighbors library

.vscode/settings.json
hkhanuja: quats is my variable name for quaternions, it is how rotations are defined for gsplat's rasterization function.

gtsfm/utils/splat.py
hkhanuja: will have to check if gtsam is compatible with quaternions, since the rotations for the gaussians are defined in that format, I need to convert them to matrix format and then after transformation,, convert them back.,

hkhanuja: I see that you have mentioned that in the comment below, I will go through it

gtsfm/utils/splat.py
hkhanuja: yes, will go through the gtsam code.

gtsfm/utils/splat.py
hkhanuja: okay

tests/utils/test_splat_utils.py
hkhanuja: okay

gtsfm/utils/splat.py
hkhanuja: since I am going from A to B: _bP = bSa . aP_ then the naming should be _bSa_?

gtsfm/utils/splat.py
dellaert: Please rename then :-) e.g., quaternion (if it's a single one)
And remove from settings.json.

gtsfm/utils/splat.py
hkhanuja: The opacity will definitely not change, it is a property of the Gaussian, not of the coordinate space. The color is represented through spherical harmonics and the coefficients are also intrinsic to the Gaussian; the final color is dependent on the viewing direction with respect to the Gaussian. So when the other frame (frame in a different partition) is transformed to the first partition space and the Gaussian is also transformed to it, the viewing direction from the pose to the camera remains the same and hence the color remains the same.

gtsfm/utils/splat.py
dellaert: Yes, indeed

gtsfm/utils/splat.py
dellaert: You’re absolutely right. But, a Gaussian has a mean and a covariance. If the mean and covariance are $aP$ and $aCa$. Then I think
```
bP = bSa * aP
bRa = bSa.rotation().matrix()
bCb = bRa * aCa * bRa.transpose() * bSa.scale()
```
Not entirely sure about the scale

dellaert: Note a covariance is denoted `aCa` as is quadratic form

gtsfm/utils/splat.py
```diff
@@ -263,3 +264,29 @@ def get_viewmat(camera_to_world: torch.Tensor) -> torch.Tensor:
...
+        gaussianA (dict): A dictionary representing the Gaussian in coordinate system A.
+        sim3_B_A (gtsam.Similarity3): The transformation from coordinate system A to B.
+    Returns:
+        gaussianB: A dictionary representing the Gaussian in coordinate system B.
+    """
+    meanA = gaussianA["means"]
+    meanB = sim3_B_A.transformFrom(meanA)
+
+    rotationBA = Rotation.from_matrix(sim3_B_A.rotation().matrix())
+    rotationB = (rotationBA * Rotation.from_quat(gaussianA["quats"])).as_quat()
```
dellaert: I guess this might be right but, naming convention says `bRa`. I’m a bit surprised gaussians are stored with just rotations and a scale and not the full covariance.. are there no “flattish” Gaussians?

.vscode/settings.json
dellaert: Please rename it

gtsfm/utils/splat.py
hkhanuja: yes I agree a Gaussian has a mean and covariance, but in Gaussian Splatting, they never explicitly deal with covariances because optimizing them is problematic. Excerpt from the paper:
`Covariance matrices have physical meaning only when
 they are positive semi-definite. For our optimization of all our
parameters, we use gradient descent that cannot be easily constrained
 to produce such valid matrices, and update steps and gradients can
 very easily create invalid covariance matrices.
 As a result, we opted for a more intuitive, yet equivalently expressive representation for optimization. The covariance matrix Σ
 of a 3D Gaussian is analogous to describing the configuration of an
 ellipsoid. Given a scaling matrix 𝑆 and rotation matrix 𝑅, we can
 find the corresponding Σ:
 Σ =𝑅𝑆𝑆^𝑇𝑅^𝑇
 (6)
 To allow independent optimization of both factors, we store them
 separately: a 3D vector 𝑠 for scaling and a quaternion 𝑞 to represent
 rotation.`

gtsfm/utils/splat.py
hkhanuja: yes, in the original paper the authors decided to not store full covariances, I have attached the excerpt from the paper in the comment above. will change the naming convention.

dellaert: Ah, yes, I missed that the scales are a 3D vector

dellaert (approved):
Awesome!!

For a next PR: I do hope you're not storing Gaussians as 10,000 python dictionaries, so a next PR might be: how do you make this work in batch :-)

Make sure you install flake8 and format with back, CI said this:
```
Run set -e
gtsfm/utils/splat.py:2[9](https://github.com/borglab/gtsfm/actions/runs/18262583231/job/51992574624?pr=859#step:5:10)2:121: E501 line too long (124 > 120 characters)
```

dellaert:
Obv, please fix CI before merging...

PR #861, opened 2025-10-05
ImagePairs

dellaert:
- While trying to create a test for merging the geometry, I was struck by how much we use `List[Tuple[int,int]]`, and for different things. This adds an `ImagePairs` type in `gtsfm.types` that is used everywhere we talk about a set of view graph edges. I'm open to changing the name and the location where it is defined.
- also moved two partitioning tests to their correct location


akshay-krishnan:
@dellaert I may have preferred a name like `ImagePairIndices` as these are not images themselves. But its longer.

PR #888, opened 2025-10-07
Fix gtsam.Cal3 type hint to use CALIBRATION_TYPE

liuzongyue6:
gtsfm/utils/pycolmap_utils.py
Fix gtsam.Cal3 type hint to use CALIBRATION_TYPE

- Replace gtsam.Cal3 with CALIBRATION_TYPE in gtsfm_calibration_to_colmap_camera function
- Fixes AttributeError: module 'gtsam' has no attribute 'Cal3'
- CALIBRATION_TYPE is the proper Union type for GTSAM calibration objects


dellaert:
Hmmm. What version of gtsam are you using?

dellaert:
upgrade to 4.3a0 if not there already

liuzongyue6:
You are right, after `pip install --no-cache-dir gtsam==4.3a0` The type issue is resolved
But it looks like break the dask distribute
I will look into this tomorrow.

PR #889, opened 2025-10-07
Fix BA with partitioning

dellaert:
Partitioning Fix:
- used `VisibilityGraph` as part of refactor
- added gtsfm/configs/unified_binary.yaml
- fixed issue with camera initialization (index 0 might not be in partition) - Big assist from Claude 4!

Smaller issues:
- added  🚀 emoji to timing messages (and 🔥 for final one)
- refactored noise model usage
- fixed loggers in 3 modules
- Made Lightglue imports on demand


dellaert:
I would split this if we had a bit more time before CVPR

dellaert:
@liuzongyue6 take a look at my changes in "TVE" :wink:

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -3,28 +3,22 @@
...
-from gtsam import (
-    BetweenFactorPose3,
-    NonlinearFactorGraph,
-    PinholeCameraCal3Fisheye,
-    PriorFactorPose3,
-    SfmTrack,
-    Values,
-    symbol_shorthand,
-)
+from gtsam import BetweenFactorPose3, NonlinearFactorGraph, PinholeCameraCal3Fisheye, PriorFactorPose3, SfmTrack, Values
```
akshay-krishnan, line 15: that looks too long for 120 chars...

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -160,7 +155,7 @@ def __pose_priors(
         self,
         absolute_pose_priors: List[Optional[PosePrior]],
         initial_data: GtsfmData,
-        camera_for_origin: gtsfm_types.CAMERA_TYPE,
+        first_valid_camera_idx: int,
```
akshay-krishnan, line 158: nit: camera_for_origin does not need to be the first_valid_camera from this method's perspective.

akshay-krishnan (approved):

gtsfm/bundle/bundle_adjustment.py
dellaert: Nope, works *just* right :-)

gtsfm/bundle/bundle_adjustment.py
dellaert: Agreed. Might fix in future PR

PR #893, opened 2025-10-07
more efficient reprojection errors

dellaert:
- eliminated copy/paste with iterators (Claude!)
- used nanmean
- added tests with failures (nans)
- fixed typing issues with helpers

Interesting to see effect (if any) on CI


dellaert:
Looks like zero impact :-)

PR #896, opened 2025-10-08
Feature/demo logging and scatter pattern

liuzongyue6:
## Files Changed
- `scripts/local_scheduler_db_remote_worker_demo.py`: switch to client.scatter to test broadcast
- no interference to other functions

## Related Issues
- `broadcast = True`, square function called 
`distributed.worker - ERROR - Could not find data: ['int-8d3e19c54d3fb6b7ee03843326025e51', 'int-d3395e15f605bc35ab1bac6341a285e2', 'int-a86a0b566184e3a052b5fa888c540f88', 'int-aba9cdc96c49f4e4e8eba1b557fcd037', 'int-ace910284a9b4d9b877cf9f8e4fd9770', 'int-28e585bdc153903d7b74053a025d1e47', 'int .ect` 
- `broadcast = False`, square function run through and **multi-workers used** 

### Next Step
- Figure out why square set up, with broadcast = False, could distribute tasks to multiple workeers
- Figure out why broadcast = True, cause no data found. Suspect it is related to ssh tunneling

### DIFF
--- a/scripts/local_scheduler_db_remote_worker_demo.py
+++ b/scripts/local_scheduler_db_remote_worker_demo.py
@@ -25,15 +25,19 @@ Author: Zongyue Liu
 """
 
 import atexit
+import datetime
+import json
+import logging
 import os
+import psycopg2
 import signal
 import socket
 import subprocess
 import time
+import traceback
+
 from typing import Any, Dict, List
 from functools import partial
-import datetime
-import psycopg2
 
 import dask.distributed
 import yaml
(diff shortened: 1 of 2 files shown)


dellaert (approved):
LGTM. Maybe use logger instead of print in second script as well unless you have a reason.

PS, cancel the CI for things that don;t affect the benchmarjs. Maybe add the scipts folder in the excluded paths in ci.yml

dellaert:
Added "[skip benchmarks]" as these are just scripts. No need to produce CO(2) ♨️

PR #898, opened 2025-10-08
Test with dataclass

dellaert:
...instead of TWO_VIEW_OUTPUT tuple


gtsfm/frontend/correspondence_generator/image_correspondence_generator.py
```diff
@@ -96,7 +96,7 @@ def generate_correspondences_and_estimate_two_view(
         gt_cameras: List[Optional[CAMERA_TYPE]],
         gt_scene_mesh: Optional[Any],
         two_view_estimator: TwoViewEstimator,
-    ) -> Tuple[List[Keypoints], Dict[Tuple[int, int], TWO_VIEW_OUTPUT]]:
+    ) -> Tuple[List[Keypoints], Dict[Tuple[int, int], TwoViewResult]]:
```
akshay-krishnan, line 99: could be done in a future PR. Ideally I think we need the List[Keypoints] to be a dict[int, Keypoints] ?

gtsfm/frontend/two_view_result_lifecycle.md
```diff
@@ -0,0 +1,293 @@
...
+    i2Ui1: Optional[Unit3]                          # Relative unit translation (final)  
+    v_corr_idxs: np.ndarray                         # Verified correspondence indices (final)
+    pre_ba_report: Optional[TwoViewEstimationReport]  # Before bundle adjustment
+    post_ba_report: Optional[TwoViewEstimationReport] # After bundle adjustment  
+    post_isp_report: Optional[TwoViewEstimationReport] # After inlier support processing
+```
+
+## Data Flow Overview
+
+```mermaid
```
akshay-krishnan, line 22: The graph is super useful, and a good idea, but some details here could be improved (some of these are currently misleading): 
- putative correspondences are also per-pair. so why dont we have putative_corr 1, putative_corr 2 ... putative_corr N, the same way as two-view result? 
- GtsfmRunnerBase actually also starts correspondence generation. Maybe you meant that it gathers the two-view results, but it also spawns the futures to begin with. I would skip this entirely and directly feed the annotated graph to scene optimizer. 
- The logic inside MultiViewOptimizer is a bit oversimplified (which is okay, but just want to flag). RotationAveraging, TranslationAveraging, DataAssociation etc are depicted as parallel branches but they are actually sequential. There is an outlier rejection step before rotation and translation averaging which select which of the two-view results are to be used downstream.

tests/data_association/test_data_association.py
```diff
@@ -84,11 +84,9 @@ def generate_noisy_2d_measurements(
     return keypoints_list, img_idxs, cameras
 
 
-def get_2d_tracks(
-    corr_idxs_dict: Dict[Tuple[int, int], np.ndarray], keypoints_list: List[Keypoints]
-) -> List[SfmTrack2d]:
+def get_2d_tracks(correspondences: AnnotatedGraph[np.ndarray], keypoints_list: List[Keypoints]) -> List[SfmTrack2d]:
     tracks_estimator = DsfTracksEstimator()
-    return tracks_estimator.run(corr_idxs_dict, keypoints_list)
+    return tracks_estimator.run(correspondences, keypoints_list)
```
akshay-krishnan, line 89: the idxs informs that these are actually indices. Correspondences could also be represented as [x1, y1, x2, y2], but in our case they are indices to the keypoints. 

Im okay with the simplification, but just letting you know why something like `correspondence_idxs` would be a good name too.

gtsfm/two_view_estimator.py
```diff
@@ -40,14 +41,33 @@
...
-    Optional[Rot3],
-    Optional[Unit3],
-    np.ndarray,
-    TwoViewEstimationReport,
-    TwoViewEstimationReport,
-    TwoViewEstimationReport,
-]
+
+@dataclasses.dataclass
+class TwoViewResult:
```
akshay-krishnan, line 46: potentially could move this to a more low-level file (types.py maybe?) two-view estimator is quite high-level and defining types here can cause cyclic import issues if we were to import it in some `utils` file (eg. io.py)

akshay-krishnan: I have run into cyclic import issues with TwoViewEstimationReport when doing some refactors.

.github/workflows/ci.yml
```diff
@@ -96,7 +96,7 @@ jobs:
         uses: actions/cache@v4
         env:
           # Increase this value to reset cache
-          CACHE_NUMBER_FRONTEND: 0
+          CACHE_NUMBER_FRONTEND: 1
```
akshay-krishnan, line 99: update before merging?

tests/data_association/test_data_association.py
dellaert: Agreed. I'd rather opt for something easy to read. The type can be inferred by the IDE

.github/workflows/ci.yml
dellaert: As we discussed, will stay "version 1"

gtsfm/two_view_estimator.py
dellaert: agreed, will fix

gtsfm/frontend/correspondence_generator/image_correspondence_generator.py
dellaert: The key challenge is that this change affects the entire pipeline from correspondence generation through to track estimation, so it needs to be done comprehensively to avoid breaking the system. Do we have a wish list of things in some issue/project?

gtsfm/frontend/two_view_result_lifecycle.md
dellaert: Fixed

PR #1094, opened 2026-03-27
Global Positioner

kathirgounder:
Trying to vibe code and express Glomap's Global Positioner step with GTSAM to get some practice with factor graphs

Akshay had already implemented the BilinearAngleTranslationFactors in GTSAM and wired it in the TranslationRecovery class, but we were not combining Translation Averaging + Triangulation into one step like GLOMAP does. This PR replaces those two steps with the GlobalPositioner and initializes BA that way. Initial Experiments on Door and Palace show BA doing a lot less work thanks to the better initialization and way faster runtimes since we don't need triangulation. Trying to queue up a full run of benchmarks to compare performance


dellaert:
closed?

PR #1100, opened 2026-04-01
Add Global Positioner module 

kathirgounder:
!_Implemented a standalone Global Positioner Class in GTSAM to match SOTA Global SFM and now bringing those changes into GTSFM, this was fun :)_!

GTSAM PR: https://github.com/borglab/gtsam/pull/2495

Akshay had already implemented the BilinearAngleTranslationFactors in GTSAM and wired it in the TranslationRecovery class, but we were not combining Translation Averaging + Triangulation into one step like GLOMAP does. I scrolled through Slack for a couple of years and it seems like the furthest we went was we tried jointly estimating Translations and 3D Tracks, but we kept camera-camera constraints and threw away the estimated 3D points, and ran the triangulation step again to regenerate 3D points to initialize BA (data_association in gtsfm). Glomap's insight was that we didn't need to do manual triangulation after Translation Averaging anymore thanks to the GlobalPositioner

This PR replaces Translation Averaging + Triangulation completely with the GlobalPositioner and initializes BA that way. Here's a small table on our benchmarks. We've also reached parity with the SOTA on tough IMC datasets like British Museum and Brussels. 

image

**Some videos showing the GlobalPositioner in action:**

_Palace of Fine Arts:_

https://github.com/user-attachments/assets/adea7d6c-9608-41b7-94f9-fcbcf6c061fc

_Brussels:_

https://github.com/user-attachments/assets/887343a1-aac9-4384-9132-636237f9524f

_South Building:_

https://github.com/user-attachments/assets/9df769f1-15a1-488a-b476-163ffa9462fb

_Gerrard Hall:_

https://github.com/user-attachments/assets/a67aa9cf-7ddb-4b94-be28-2db4347b33e5

_Door:_

https://github.com/user-attachments/assets/f1c50c11-c5a8-4f90-8e62-9d63d4bb711b


dellaert:
Could you include the comparison with baseline in the PR comment?

kathirgounder:
> Could you include the comparison with baseline in the PR comment?

Hi Frank @dellaert, just added a screenshot of the metrics comparison and some videos.

The existing visualize_benchmark_comparison.py compares the same configs across two different CI runs (master vs branch) by matching artifact names, it wasn't built to compare two different configs and code paths within the same run since they have different artifact names and the script pairs by name (unified sift vs unified sift_global_positioner - both are identical besides the fact one does TA + Triangulation verus GlobalPositioning), the global positioner is a clean swap / drop in that replaces the two stages and I added separate configs for it and added it to the ci.yml, so I had claude write a quick script to pull down all the metrics and just compare the results between the two from the same CI Run.

dellaert:
Amazing! @akshay-krishnan whaddayathink? I’ll leave you to approve…

kathirgounder:
> Amazing! @akshay-krishnan whaddayathink? I’ll leave you to approve…

Will try to clean up and add more documentation on some of the things I added (Viz Script and optimizer logging) and also discuss with Akshay how to go about updating the default CI strategy before trying to merge this in

kathirgounder:
Hey @dellaert @akshay-krishnan !

I was potentially thinking of adding a separate GTSAM class for the GlobalPositioner instead of using the TranslationRecovery class. The semantics are a bit misleading in comparison to how I am using it to express the Global Positioning Optimization. TranslationRecovery's documentation, variable names, and API all describe camera-to-camera translation directions. The current interface does not indicate anywhere that tracks are supported and assumes all measurements are camera-camera. 

Our GTSFM code passes C(cam_idx) and L(track_idx) as keys, computes world-frame bearings from feature tracks, and maps results back to cameras+landmarks, all in Python wrapper code that a potential GTSAM user would have to reverse-engineer from our application to understand the GP use case. I think a dedicated class would make the camera+landmark formulation cleaner. 

Additionally TranslationRecovery's addPrior method is designed specifically for Translation Averaging, where it tries fix the 4 degrees of freedom required for TA by pinning key1 to the origin and key2 to default scale * measured(). In GP, edges are camera-to-landmark, so key2 is a landmark. The scale prior pins the entire reconstruction's scale to one depth estimate for one landmark rather than establishing a baseline. In our tests the optimizer overcomes this since the prior is soft, but to match Glomaps GP formulation we would only add a prior to the first camera. 

I think this will make both the GTSFM code and GTSAM code more explicit and separate. GTSAM already has the factor (BilinearAngleTranslationFactor) and the solver machinery, what's missing is the solve pipeline that makes GP as
easy to use and expressed as a first class citizen like ShonanAveraging and TranslationRecovery are for their respective problems! 

I also wanted to learn how to contribute to GTSAM :)

gtsfm/multi_view_optimizer.py
```diff
@@ -185,32 +188,42 @@ def create_computation_graph(
...
-        ba_input_graph, data_assoc_metrics_graph = self.data_association_module.create_computation_graph(
-            num_images,
-            init_cameras_graph,
-            ta_inlier_tracks_2d_graph,
-            cameras_gt,
-            pose_priors_graph,
-            images,
-        )
+
+        if self.global_positioner is not None:
```
akshay-krishnan, line 193: this can be done in a follow-up PR. I would prefer global positioner and translation averaging + triangulation to inherit from a common base class. That way you wont need this if-else here entirely.

gtsfm/global_positioner/global_positioner.py
```diff
@@ -0,0 +1,565 @@
+"""Global positioning of cameras and 3D points via bearing constraints.
```
akshay-krishnan, line 1: how much of this is duplicated code in other parts of gtsfm? can we move such functions to any utils?

gtsfm/global_positioner/global_positioner.py
```diff
@@ -0,0 +1,565 @@
...
+DEFAULT_MAX_REPROJ_ERROR = 5.0      # Pixels — for filtering output tracks
+DEFAULT_MAX_TRACKS = 0              # 0 = no cap (C++ solver is fast enough)
+
+
+# ──────────────────────────────────────────────────────────────────────────────
+# Bearing computation
+# ──────────────────────────────────────────────────────────────────────────────
+
+
+def compute_world_bearings(
```
akshay-krishnan, line 85: nit: we dont use the term bearing in gtsfm code - its "direction" or "translation direction"

akshay-krishnan (commented):
ultimately we are using the translationrecovery class here as well. 

It would be cleaner to re-use the existing averaging_1dsfm code as much as possible. feel free to refactor that class to support this "global positioner" variant if needed.

akshay-krishnan:
visualization is very cool! based on what Im seeing it should be fine to stop iterating sooner, not much happens after the first ~10 iterations. There is also a global rotation error guage freedom, we probably need to strengther the prior, likely inside gtsam code.

akshay-krishnan:
CI seems to be failing due to some formatting errors.

nantonzhang (approved):

PR #1101, opened 2026-04-03
BA updates - save covariances, grid size -> patch size, more config options

akshay-krishnan:
- Add functionality to compute and save the covariances using gtsam.Marginals. This is only for debugging, not for use in the algorithm. 
- Min track length pre-BA should be 3. 
- Do not BA cameras that have < 20 tracks. What do we we do with them? retain the pre-BA camera poses as these could be important to merge children upstream. Should we also retain the tracks? discarding them now, could revisit later if that causes problems. 
- change from grid-size to patch size for filtering tracks. grid-size is unclear, perhaps this was using recangular patches before? 
- expose some more options to customize BA, eg. relative cost tolerance.

### DIFF
--- a/gtsfm/bundle/bundle_adjustment.py
+++ b/gtsfm/bundle/bundle_adjustment.py
@@ -75,6 +75,9 @@ class BundleAdjustmentOptions:
     calibration_prior_dist_sigma: float | Sequence[float] = 0.1
     calibration_prior_pp_sigma: float = 1e-5
     robust_noise_basin: float = 1.345
+    min_tracks_per_camera: int = 20
+    compute_pose_covariances: bool = False
+    optimizer_relative_cost_tol: float = 1e-3
 
     def to_optimizer(self, **overrides) -> "BundleAdjustmentOptimizer":
         """Construct a :class:`BundleAdjustmentOptimizer` from these options.
@@ -98,6 +101,9 @@ class BundleAdjustmentOptions:
             calibration_prior_dist_sigma=self.calibration_prior_dist_sigma,
             calibration_prior_pp_sigma=self.calibration_prior_pp_sigma,
             robust_noise_basin=self.robust_noise_basin,
+            min_tracks_per_camera=self.min_tracks_per_camera,
+            compute_pose_covariances=self.compute_pose_covariances,
+            optimizer_relative_cost_tol=self.optimizer_relative_cost_tol,
         )
         kwargs.update(overrides)
         return BundleAdjustmentOptimizer(**kwargs)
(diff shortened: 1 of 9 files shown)


gtsfm/bundle/bundle_adjustment.py
```diff
@@ -556,16 +587,26 @@ def run_simple_ba(
...
-            logger.warning("Skipping bundle adjustment because all cameras are without tracks.")
-            return initial_data, 0.0
-        optimized_data, _, final_error = self.__optimize_and_recover(
-            initial_data, graph, self._ordering_type if not cameras_without_tracks else "COLAMD"
+        graph = self.__construct_simple_factor_graph(cameras_to_model, initial_data, robust_noise_basin)
+        optimized_data, result_values, final_error = self.__optimize_and_recover(
+            initial_data, graph, self._ordering_type
         )
+        if self._compute_pose_covariances:
+            try:
```
hkhanuja, line 602: nothing wrong here, but just confused that we are taking the covariance matrices from the cameras_to_model and then set it in optimized data which is the recovered after optimization so should have those inherently?

hkhanuja:
Just thinking if it should be called number of tracks per camera or number of measurements per camera? and I think if we have 2 cameras with less than 20 measurements after all pre-ba filtering, we shouldn't remove those tracks because those tracks might have measurements in other cameras (maybe helpful in merging ahead)

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: are you suggesting moving this to __optimize_and_recover ?

hkhanuja: no I am just a little confused on what this if block does

akshay-krishnan:
> Just thinking if it should be called number of tracks per camera or number of measurements per camera? and I think if we have 2 cameras with less than 20 measurements after all pre-ba filtering, we shouldn't remove those tracks because those tracks might have measurements in other cameras (maybe helpful in merging ahead)

we dont remove the tracks, we remove the camera (which removes the measurements during BA). tracks_per_camera or measurements_per_camera is the same since it is defined for a particular camera (tracks == measurements). while measurements is more accurate, tracks is a simpler name.

gtsfm/configs/vggt_megaloc_phototourism.yaml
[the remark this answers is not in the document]
kathirgounder: use_pose_prior_all_cameras: bool = False
    use_pose_prior_first_camera: bool = False <- are the two options in BundleAdjustmentOptimizer

gtsfm/bundle/bundle_adjustment.py
akshay-krishnan: marginals computes the covariance matrices based on the graph. we update them in the gtsfm_data result. alternative can provide the covariance matrices when the gtsfm_data is created. but that needs a bit more plumbing.

akshay-krishnan:
newplot

I see a large bump on south building - but this is unexpected. Smaller datasets show a minor degradation - but these are in the < 1 deg error already, so can be ignored.

hkhanuja (approved):

PR #1105, opened 2026-04-16
patch mvo code to read updated configs

nantonzhang:


gtsfm/cluster_optimizer/cluster_mvo.py
```diff
@@ -68,6 +73,15 @@ def __init__(
...
         self._pose_angular_error_thresh = pose_angular_error_thresh
+        # These legacy cluster-merging options are still present in several Hydra configs,
+        # but the leaf-only MVO optimizer does not consume them directly.
+        _ = (
+            drop_outlier_after_camera_merging,
+            drop_child_if_merging_fail,
+            drop_camera_with_no_track,
+            plot_reprojection_histograms,
+            run_bundle_adjustment_on_parent,
+        )
```
akshay-krishnan, line 84: why is this code needed?

akshay-krishnan: the fix is to update the configs.

PR #1109, opened 2026-04-26
Frontend changes

hkhanuja:
This PR makes a significant number of changes to the frontend:
1. Change the focal length calibration by adding `view_graph_calibration.py`
2. Adds PoseLib (https://github.com/PoseLib/PoseLib) to compute R and t 
3. Switches to using ColmapSift features (4 new config files)
4. Makes the view graph more denser
5. Fixes small issues in stage BA so now in multistage BA, previous BA outputs are used.
6. Colmap loader uses 1.2 or `_default_focal_length_factor `if `_use_gt_intrinsics `is set as False.

Run with command

`./run --config_name megaloc_sift_single --loader colmap --dataset_dir benchmarks/phototourism/british_museum/ --output_root <output_folder> --worker_memory_limit "48GB" +loader.colmap_files_subdir=sparse loader.use_gt_intrinsics=false`

### DIFF
--- a/gtsfm/bundle/bundle_adjustment.py
+++ b/gtsfm/bundle/bundle_adjustment.py
@@ -660,7 +660,7 @@ class BundleAdjustmentOptimizer:
             valid_mask[track_idx] = is_valid
         return optimized_data, filtered_result, valid_mask, final_error
 
-    def _run_ba_stages(
+    def run_ba(
         self,
         initial_data: GtsfmData,
         absolute_pose_priors: List[Optional[PosePrior]],
@@ -716,35 +716,6 @@ class BundleAdjustmentOptimizer:
                 )
 
         return optimized_data, filtered_result, cumulative_valid_mask, step_times  # type: ignore
-
-    def run_ba(
-        self,
-        initial_data: GtsfmData,
-        absolute_pose_priors: List[Optional[PosePrior]],
-        relative_pose_priors: Dict[Tuple[int, int], PosePrior],
-        verbose: bool = True,
-    ) -> Tuple[GtsfmData, GtsfmData, List[bool]]:
-        """Runs bundle adjustment by forming a factor graph and optimizing it using Levenberg–Marquardt optimization.
-
-        Args:
-            initial_data: Initialized cameras, tracks w/ their 3d landmark from triangulation.
-            absolute_pose_priors: Priors to be used on cameras.
-            relative_pose_priors: Priors on the pose between two cameras.
-            verbose: Boolean flag to print out additional info for debugging.
-
-        Results:
-            Optimized camera poses, 3D point w/ tracks, and error metrics, aligned to GT (if provided).
-            Optimized camera poses after filtering landmarks (and cameras with no remaining landmarks).
-            Valid mask as a list of booleans, indicating for each input track whether it was below the re-projection
-                threshold.
-        """
-        optimized_data, filtered_result, valid_mask, _ = self._run_ba_stages(
-            initial_data=initial_data,
-            absolute_pose_priors=absolute_pose_priors,
-            relative_pose_priors=relative_pose_priors,
-            verbose=verbose,
-        )
-        return optimized_data, filtered_result, valid_mask
 
     def _run_ba_and_evaluate(
         self,
--- a/gtsfm/bundle/bundle_adjustment.py
+++ b/gtsfm/bundle/bundle_adjustment.py
@@ -477,23 +477,25 @@ class BundleAdjustmentOptimizer:
 
     def __optimize_and_recover(
         self, initial_data: GtsfmData, graph: NonlinearFactorGraph, ordering_type: str
-    ) -> Tuple[GtsfmData, Values, float]:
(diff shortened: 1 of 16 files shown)


gtsfm/view_graph_estimator/view_graph_calibration.py
```diff
@@ -0,0 +1,266 @@
...
+    num_rejected = 0
+
+    for var_idx, cam_idx in enumerate(sorted_cameras):
+        old_focal = initial_focals[var_idx]
+        new_focal = optimized_focals[var_idx]
+        ratio = new_focal / old_focal if old_focal > 0 else 0
+
+        if min_focal_ratio <= ratio <= max_focal_ratio:
+            old_K = initial_intrinsics[cam_idx]
+            if isinstance(old_K, Cal3Bundler):
```
akshay-krishnan, line 227: should support other camera models as well.

gtsfm/cluster_merging.py
```diff
@@ -316,6 +316,17 @@ class MergedNodeResult:
     pre_ba_metrics: GtsfmMetricsGroup
 
 
+@dataclass(frozen=True)
+class MergedNodeSummary:
```
akshay-krishnan, line 320: dont we have these metrics somewhere already?

gtsfm/multi_view_optimizer.py
```diff
@@ -29,12 +32,16 @@
 from gtsfm.products.one_view_data import OneViewData
 from gtsfm.products.two_view_result import TwoViewResult
 from gtsfm.products.visibility_graph import AnnotatedGraph
+from gtsfm.utils import verification as verification_utils
 from gtsfm.view_graph_estimator.cycle_consistent_rotation_estimator import (
     CycleConsistentRotationViewGraphEstimator,
     EdgeErrorAggregationCriterion,
 )
+from gtsfm.view_graph_estimator.view_graph_calibration import calibrate_view_graph
```
akshay-krishnan: we typically import the module not the method

gtsfm/multi_view_optimizer.py
```diff
@@ -253,6 +289,134 @@ def create_computation_graph(
...
+    should reflect that final state.
+    """
+    synced_reports: AnnotatedGraph[TwoViewEstimationReport] = {}
+
+    for edge, report in two_view_reports.items():
+        if report is None or edge not in i2Ri1_dict or edge not in i2Ui1_dict or edge not in v_corr_idxs_dict:
+            continue
+
+        v_corr_idxs = v_corr_idxs_dict[edge]
+        synced_reports[edge] = dataclasses.replace(
```
akshay-krishnan, line 327: you could add new fields instead of replacing.

gtsfm/multi_view_optimizer.py
```diff
@@ -253,6 +289,134 @@ def create_computation_graph(
...
+            v_corr_idxs=v_corr_idxs,
+            num_inliers_est_model=v_corr_idxs.shape[0],
+            i2Ri1=i2Ri1_dict[edge],
+            i2Ui1=i2Ui1_dict[edge],
+        )
+
+    return synced_reports
+
+
+def reestimate_relative_poses(
```
akshay-krishnan, line 338: we should reuse the verifier for this instead of duplicating this code.

gtsfm/two_view_estimator.py
```diff
@@ -93,8 +93,8 @@ def __init__(
             robust_noise_basin=1.345,
             use_karcher_mean_factor=False,
             use_pose_prior_first_camera=True,
-            calibration_prior_focal_sigma=1e-5,
-            calibration_prior_dist_sigma=1e-5,
+            calibration_prior_focal_sigma=20.0,
```
akshay-krishnan: i think I noticed some regressions when i did that. are you sure about this change?

kathirgounder: I think we need to set this based on whether we're using ground truth intrinsics or not,  right ?

kathirgounder: Tight sigma if using ground truth and loose if not

gtsfm/bundle/bundle_adjustment.py
```diff
@@ -670,24 +738,13 @@ def run_ba(
...
                 threshold.
         """
-        num_ba_steps = len(self._reproj_error_thresholds)
-        for step, reproj_error_thresh in enumerate(self._reproj_error_thresholds):
-            # Use intermediate result as initial condition for next step.
-            (optimized_data, filtered_result, valid_mask, final_error) = self.run_ba_stage_with_filtering(
-                initial_data,
-                absolute_pose_priors,
-                relative_pose_priors,
-                reproj_error_thresh,
```
akshay-krishnan, line 750: this functions feels redundant now. so why not rename _run_ba_stages to run_ba and delete this?

gtsfm/two_view_estimator.py
akshay-krishnan: yeah it should be a constructor argument.

kathirgounder (approved):

PR #1110, opened 2026-04-27
New class to run VGGT with tracks from gtsfm frontend

akshay-krishnan:


nantonzhang (approved):
LGTM, discussed offline

PR #1116, opened 2026-06-23
VGGT Omega implementation

hkhanuja:
This PR adds the VGGT-Omega backend to the GTSfM pipeline.
Since the VGGT-Omega weights are protected under a license and cannot be distributed, the user needs to get access to the model and then download the weights using their `hf_token `using `scripts/download_weights.sh --hf_token <your hf_token>`

Since the VGGT-Omega poses can be considered accurate, we add a factor to the prior camera poses. (can be experimented with)
We also set the camera type to Cal3DS2 to incorporate different focal lengths (fx and fy) and include distortion parameters.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -13,3 +13,6 @@
 [submodule "thirdparty/FastVGGT"]
 	path = thirdparty/FastVGGT
 	url = https://github.com/mystorm16/FastVGGT.git
+[submodule "thirdparty/vggt-omega"]
+	path = thirdparty/vggt-omega
+	url = https://github.com/facebookresearch/vggt-omega.git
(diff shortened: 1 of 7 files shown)


akshay-krishnan:
we should not need a new cluster_optimizer pipeline: gtsfm/cluster_optimizer/cluster_vggt_omega_with_frontend.py

we are adding a new model, which is a new geometry_transformer. this should be supported in the existing cluster_optimizer pipeline. if this needs some changes to the API of the geometry transformer class, or the existing cluster optimizer pipeline, that can be done. 

let me know if that makes sense? If there is a logical reason why these cannot share the same cluster optimizer pipeline, please explain that here.

PR #1121, opened 2026-08-13
Add first-class GTSFM Studio workflow

suhanpark:
## Summary

Introduce a packaged, browser-based GTSFM Studio that launches with `gtsfm run` and provides a FastAPI-backed React workspace for configuring, running, monitoring, and inspecting reconstructions.

## Why

The existing workflow requires users to know repository paths and command-line arguments. This creates a first-class package entry point and a discoverable UI while retaining the terminal execution path.

## User impact

- Launch the local workspace with `gtsfm run`
- Choose datasets, reconstruction models, splat implementations, and detected hardware in the UI
- Upload images or select maintained sample data
- Monitor jobs and inspect/download reconstruction results in the integrated 3D viewer
- Use VGGT as the default reconstruction model

## Validation

- `python -m pytest tests/visualization/test_runtime.py tests/ui/test_optional_process_graph.py -q` (36 passed)
- `pnpm build`

## Review notes

This is PR 1 of a stacked draft series. Later drafts isolate the reusable Modal runtime, workspace-control improvements, and live Gaussian visualization.

PR #1122, opened 2026-08-13
Publish reusable Modal runtime image

suhanpark:
## Summary

Make the remote-VM runtime reproducible and reusable through a prebuilt Docker image published manually to Docker Hub.

## Why

Building the full CUDA, PyTorch, Open3D, and GTSFM dependency stack during every Modal deployment is slow and error-prone. A prebuilt runtime reduces repeated downloads and makes deployment behavior easier to inspect.

## Changes

- Add a maintained Modal CUDA runtime Dockerfile
- Add the manual Docker Hub publishing helper
- Preserve CUDA package indexes and system compression dependencies
- Install dependencies in the runtime virtual environment
- Keep future Modal Volume mount paths empty during image extension
- Document the image publication workflow

## Validation

- Python compilation of `visualization/modal_app.py` and `visualization/modal_deployment.py`
- `git diff --check`

## Stack

Depends on #1121. This draft targets `studio/package-cli` so the review contains only the runtime-image changes. It should be retargeted to `master` after #1121 merges.


visualization/modal_deployment.py
```diff
@@ -28,7 +28,7 @@
     "H200",
     "B200",
 }
-DEFAULT_MODAL_RUNTIME_IMAGE = "ghcr.io/borglab/gtsfm-modal-runtime:latest"
+DEFAULT_MODAL_RUNTIME_IMAGE = "docker.io/su071301/gtsfm-modal-runtime:firstclass"
```
akshay-krishnan, line 31: what is this path for? why are we changing from a borglab path to a su071301 path?

suhanpark: it shouldnt be. im removing the custom prebuilt-image setup and using Modal’s direct build path. This registry reference is no longer needed.

PR #1128, opened 2026-08-15
Auto-detect dataset formats and expand samples

suhanpark:
## Summary
Detect uploaded and sample dataset layouts, then configure the matching loader automatically.

## Changes
- add structural detection for MobileBrick, Tanks and Temples, Hilti, YFCC IMB, Argoverse, Olsson, AstroVision, COLMAP, and generic image collections
- resolve detected paths relative to portable uploaded roots
- expand the curated GitHub sample catalog from 3 to 9 scenes and correct the Crane Mast image count
- add an **Auto-detect (recommended)** format option with a detection explanation in the UI
- cover the detector, uploads, sample catalog, and runtime integration with tests

## Why
GTSFM sample datasets use different directory layouts. Requiring users to know the exact loader and path arguments made the Studio workflow unnecessarily error-prone.

## Validation
- 47 focused visualization tests passed
- staged React/TypeScript frontend build passed
- `git diff --check` passed

## Stack
Depends on #1127 and targets `studio/runtime-reliability` so this remains a focused, reviewable change.


visualization/datasets.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+IMAGE_SUFFIXES = {".avif", ".bmp", ".heic", ".heif", ".jpeg", ".jpg", ".png", ".tif", ".tiff", ".webp"}
+
+
+def _relative_files(root: Path) -> list[Path]:
+    return [path.relative_to(root) for path in root.rglob("*") if path.is_file()]
+
+
+def detect_dataset_format(root: Path) -> dict[str, Any]:
```
akshay-krishnan, line 16: what is this for? I dont think we should depend on folder structure, and should pipe the config through which contains the name of the loader.

PR #1129, opened 2026-08-15
Add preset-aware Hydra configuration UI

suhanpark:
## Summary
Expose GTSFM's Hydra configuration safely through Studio, with controls that follow the selected reconstruction and Gaussian-splatting presets.

## Changes
- discover configuration schemas dynamically across 37 root presets without importing PyTorch during workspace startup
- expose 35 safe Gaussian-splatting parameters and preset-specific reconstruction fields
- load the schema lazily in the background so the main UI renders immediately
- hide internal paths and secret-bearing settings
- validate typed overrides on both the client and server before launching GTSFM
- add JSON configuration import with a generated template, nested or dotted keys, API aliases, and clear validation feedback
- pass validated reconstruction and Gaussian overrides through to the runner
- organize these controls in a collapsible **Advanced Configuration** section

## Why
The command line already supports powerful Hydra presets and overrides, but users had to know their exact names and types. This adds a guided UI while preserving Hydra as the source of truth.

## Validation
- 48 focused runtime and shutdown tests passed
- React/TypeScript production build passed
- Black, isort, and flake8 passed on changed Python files
- `git diff --check` passed

## Stack
Depends on the dataset auto-detection draft and targets `studio/dataset-autodetect` to keep review scope focused.


visualization/runtime.py
```diff
@@ -611,6 +613,221 @@ def configuration_schema() -> dict[str, Any]:
...
+        elif isinstance(child, (bool, int, float, str)) and child != "???":
+            fields.append(_hydra_field_descriptor(path, child, _hydra_field_group(path)))
+    return fields
+
+
+def _composed_model_configuration(config_name: str, loader: str, overrides: list[str] | None = None) -> dict[str, Any]:
+    """Compose a root preset exactly as the runner does, without importing Hydra at startup."""
+
+    import hydra
+    from omegaconf import OmegaConf
```
akshay-krishnan, line 774: all imports should be at the top, unless they are GS rendering related.

gtsfm/runner.py
```diff
@@ -232,6 +231,12 @@ def construct_argparser(self) -> argparse.ArgumentParser:
             default=None,
             help="Override the Gaussian Splatting training iteration count.",
         )
+        parser.add_argument(
+            "--gaussian_splatting_override",
```
akshay-krishnan, line 235: why do we call this override? we should follow similar naming as done for the other modules (like correspondence_generator?)

visualization/runtime.py
suhanpark: got it, moved

gtsfm/runner.py
suhanpark: it’s called override because it applies individual key=value changes to the selected Gaussian splatting preset without modifying the YAML file. Preset selection already uses gaussian_splatting_config_name, matching correspondence_generator_config_name. 

we can name this something like `—gaussian_splatting_config_option` if it's more descriptive

PR #1132, opened 2026-08-31
Add torch two-way matcher; harden TwoWayMatcher against degenerate inputs

kathirgounder:
## Summary

Adds `TorchTwoWayMatcher`, a GPU-capable mutual-nearest-neighbour matcher with Lowe's ratio test that produces the same `(N, 2)` match-index contract as `TwoWayMatcher`, and hardens `TwoWayMatcher` against degenerate inputs (empty descriptor sets, a single keypoint) that previously raised.

This is one of the components behind the verified-view-graph pipeline used for the ECCV'26 SFM-DL paper; it is standalone and has no dependency on the rest of that work.

## Behavior

- `frontend/matcher/torch_twoway_matcher.py` (new): torch implementation, CUDA when available with CPU fallback. Drop-in via config: `matcher_obj._target_: gtsfm.frontend.matcher.torch_twoway_matcher.TorchTwoWayMatcher`.
- `frontend/matcher/twoway_matcher.py`: degenerate-input handling — returns an empty match array instead of raising.

## Tests

- `tests/test_torch_twoway_matcher.py`: parity with `TwoWayMatcher` on synthetic descriptors.
- `tests/test_twoway_matcher_degenerate.py`: empty / single-keypoint inputs for both matchers.

## Review focus

Ratio-test semantics and index ordering match the existing matcher exactly; the degenerate-input paths.

### DIFF
--- a/gtsfm/frontend/matcher/twoway_matcher.py
+++ b/gtsfm/frontend/matcher/twoway_matcher.py
@@ -104,6 +104,11 @@ class TwoWayMatcher(MatcherBase):
     def __perform_matching(self, descriptors_1: np.ndarray, descriptors_2: np.ndarray) -> np.ndarray:
         """Run the core logic for matching.
 
+        NOTE: cv2's BFMatcher parallelizes internally over the GLOBAL OpenCV thread pool, which
+        defaults to all cores — with N dask workers that is N full-node thread pools thrashing
+        each other (same oversubscription bug as pycolmap SIFT num_threads=-1, fixed 818f7b4f).
+        Parallelism comes from dask workers; each matcher task must stay single-threaded.
+
         Args:
             descriptors_1: descriptors for the 1st image.
             descriptors_2: descriptors for the 2nd image.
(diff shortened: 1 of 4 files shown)


gtsfm/frontend/matcher/twoway_matcher.py
```diff
@@ -134,7 +141,12 @@ def __perform_oneway_matching(self, descriptors_1: np.ndarray, descriptors_2: np
 
         if self._ratio_test_threshold is not None:
             all_matches = opencv_matcher.knnMatch(descriptors_1, descriptors_2, k=2)
-            matches = [m1 for m1, m2 in all_matches if m1.distance <= self._ratio_test_threshold * m2.distance]
+            # knnMatch(k=2) yields <2 candidates per query when the train side has <2 descriptors
+            # (near-featureless internet images) — those queries cannot pass a ratio test; skip them.
+            matches = [
```
akshay-krishnan, line 146: why is this changed from the older m1, m2?

akshay-krishnan (approved):

PR #1133, opened 2026-08-31
Integrate VGGT-Omega as a geometry transformer

kathirgounder:
## Summary

Integrates **VGGT-Omega** as a geometry transformer, runnable through the existing `ClusterVGGTWithFrontend` via the injected `geometry_transformer` — so a VGGT vs. VGGT-Omega comparison differs only in the per-cluster geometry model. This is the omega arm used for the ECCV'26 SFM-DL paper results; the verified-view-graph pipeline it was paired with lands separately on `mapping-vggt`.

## What changes

- **`frontend/vggt_omega_geometry_transformer.py`** (new): model loading from the `thirdparty/vggt-omega` submodule, omega's own image preprocessing (`load_image_batch_vggt_omega_loader`: 512 px, 16-patch aligned, aspect-ratio cropped) and `predict` returning the unified `GeometryTransformerOutput`.
- **`frontend/geometry_transformer.py`**: new `GeometryTransformer.load_image_batch` hook — image preprocessing now *follows the transformer* (default = VGGT loader, omega overrides). The per-pixel depth lookup downstream indexes the model's depth map via the `original_coords` returned here, so the loader must match the model. No-op for all VGGT configs.
- **`cluster_optimizer/cluster_vggt.py`, `cluster_vggt_with_frontend.py`**: `_load_vggt_inputs` dispatches to `transformer.load_image_batch`; a new shared `_model_loading_plan` helper replaces the duplicated model-caching blocks in both constructors. Behavior-identical for VGGT, and it closes a real footgun: a transformer without a `config` previously (a) crashed the constructor and, once guarded, (b) inherited the default `("default_vggt_loader", …)` cache key — which makes `_resolve_vggt_model` load a cached **VGGT** model and hand it to omega's `predict`. Config-less transformers now get no shared cache key and manage their own model.
- **`__repr__`** of both optimizers tolerates config-less transformers (class name enters the cluster cache key).
- **`configs/vggt_omega_sift_frontend_megaloc_phototourism.yaml`** (new): identical to `vggt_sift_frontend_megaloc_phototourism.yaml` except the geometry transformer — a controlled A/B.
- **`.gitmodules`** + `thirdparty/vggt-omega` submodule; **`scripts/download_model_weights.sh`** gains `--hf_token` for the gated omega checkpoint (and `mkdir -p` fixes).
- `utils/torch.py`: docstring fix.

## Ops / licensing note

VGGT-Omega weights are **gated and CC-BY-NC-4.0** (non-redistributable) and inference is **CUDA-only**. Nothing is downloaded by default; omega imports are only triggered when the omega transformer is instantiated, so all VGGT paths and tests are unaffected.

```bash
git submodule update --init thirdparty/vggt-omega
bash scripts/download_model_weights.sh --hf_token <token>
```

## Tests

- `tests/frontend/test_vggt_omega_geometry_transformer.py` (new, CPU-only, skips without the submodule): 16-patch alignment in both modes, mixed-aspect batch padding, `original_coords` mapping loader pixels into the padded omega frame, and the transformer's mode normalisation.
- Existing `tests/frontend/test_cluster_vggt.py` / matcher suites unchanged and green.

## Review focus

`_model_loading_plan` (the cache-key policy) and the `load_image_batch` hook. Everything else is additive.

### DIFF
--- a/gtsfm/cluster_optimizer/cluster_vggt.py
+++ b/gtsfm/cluster_optimizer/cluster_vggt.py
@@ -112,12 +112,22 @@ def _load_vggt_inputs(
     indices: list[int],
     mode: str,
     *,
+    transformer=None,
     save_processed_image: bool = False,
     output_root: Optional[str] = None,
     image_names: Optional[tuple[str, ...]] = None,
 ):
-    """Load and preprocess a batch of images for VGGT."""
-    image_batch, original_coords = load_image_batch_vggt_loader(loader, indices, mode=mode)
+    """Load and preprocess a batch of images for the geometry model.
+
+    Preprocessing follows the geometry transformer: VGGT and VGGT-Omega differ in resolution, patch
+    alignment and cropping, and the per-pixel depth lookup downstream indexes the model's depth map via
+    the ``original_coords`` returned here — so the loader must match the model. ``transformer=None``
+    keeps the legacy VGGT loader.
+    """
+    if transformer is not None:
+        image_batch, original_coords = transformer.load_image_batch(loader, indices, mode=mode)
+    else:
+        image_batch, original_coords = load_image_batch_vggt_loader(loader, indices, mode=mode)
     if not save_processed_image or output_root is None or image_names is None:
         return image_batch, original_coords
     if len(image_names) != image_batch.shape[0]:
@@ -156,6 +166,39 @@ def _load_vggt_inputs(
         comments="",
     )
     return image_batch, original_coords
+
+
+def _model_loading_plan(
+    transformer: Any, weights_path: Optional[Path], model_cache_key: Hashable | bool | None
+) -> tuple[dict[str, Any], Hashable | None]:
+    """Decide how a cluster optimizer's geometry model is loaded on workers.
+
+    Transformers that expose a ``config`` (the VGGT family) are loaded through the shared VGGT loader
+    and cached under a key derived from the weights path and model kwargs. Transformers without a
+    ``config`` (e.g. VGGT-Omega) manage their own model inside ``predict``: they get no loader kwargs
+    and NO shared cache key — a cache key would make ``_resolve_vggt_model`` hand them a cached *VGGT*
+    model.
+
+    Returns:
+        ``(loader_kwargs, cache_key)`` to store on the optimizer.
+    """
+    config = getattr(transformer, "config", None)
+    if config is None:
+        return {}, None
+
+    loader_kwargs: dict[str, Any] = {}
+    if weights_path is not None:
+        loader_kwargs["weights_path"] = weights_path
+    model_kwargs = config.model_ctor_kwargs
+    if model_kwargs:
+        loader_kwargs["model_kwargs"] = model_kwargs
+
+    if model_cache_key is False:
+        return loader_kwargs, None
+    if model_cache_key is None:
+        kwargs_key = tuple(sorted((k, repr(v)) for k, v in model_kwargs.items())) if model_kwargs else None
+        return loader_kwargs, ("default_vggt_loader", weights_path, kwargs_key)
+    return loader_kwargs, model_cache_key
 
 
 def _resolve_vggt_model(cache_key: Hashable | None, loader_kwargs: dict[str, Any] | None) -> Any | None:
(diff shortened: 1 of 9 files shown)


akshay-krishnan:
is this an alternative to https://github.com/borglab/gtsfm/pull/1116 ?

kathirgounder:
> is this an alternative to #1116 ?

@akshay-krishnan Yeah

gtsfm/cluster_optimizer/cluster_vggt.py
```diff
@@ -112,12 +112,22 @@ def _load_vggt_inputs(
...
-    image_batch, original_coords = load_image_batch_vggt_loader(loader, indices, mode=mode)
+    """Load and preprocess a batch of images for the geometry model.
+
+    Preprocessing follows the geometry transformer: VGGT and VGGT-Omega differ in resolution, patch
+    alignment and cropping, and the per-pixel depth lookup downstream indexes the model's depth map via
+    the ``original_coords`` returned here — so the loader must match the model. ``transformer=None``
+    keeps the legacy VGGT loader.
+    """
+    if transformer is not None:
+        image_batch, original_coords = transformer.load_image_batch(loader, indices, mode=mode)
```
akshay-krishnan, line 128: why is image loading code starting to live inside the transformer? how different is it from the load_image_batch_vggt_loader that we use? the original_coords logic was already brittle, I dont think we should duplicate that.

akshay-krishnan, line 128: It would be nice to refactor original_coords into a dataclass with explicit fields. the use of indices is a source for potential bugs.

gtsfm/cluster_optimizer/cluster_vggt.py
```diff
@@ -158,6 +168,39 @@ def _load_vggt_inputs(
...
+    loader_kwargs: dict[str, Any] = {}
+    if weights_path is not None:
+        loader_kwargs["weights_path"] = weights_path
+    model_kwargs = config.model_ctor_kwargs
+    if model_kwargs:
+        loader_kwargs["model_kwargs"] = model_kwargs
+
+    if model_cache_key is False:
+        return loader_kwargs, None
+    if model_cache_key is None:
```
akshay-krishnan, line 198: different behavior for False and None is not ideal, can we make this more explicit somehow? might need another argument for the `False` .

gtsfm/frontend/vggt_omega_geometry_transformer.py
```diff
@@ -0,0 +1,347 @@
...
+from vggt_omega.utils.pose_enc import encoding_to_camera
+
+# ---------------------------------------------------------------------------
+# Image loading
+# ---------------------------------------------------------------------------
+
+DEFAULT_FIXED_RESOLUTION_VGGT_OMEGA = 512
+
+
+def load_image_batch_vggt_omega_loader(
```
akshay-krishnan, line 89: as mentioned in another comment, we should not be duplicating this.

gtsfm/frontend/vggt_omega_geometry_transformer.py
```diff
@@ -0,0 +1,347 @@
...
+        )
+    return checkpoint
+
+
+def load_model(
+    device: torch.device,
+):
+    """Load the VGGT Omega model weights on the requested device."""
+    if device.type != "cuda":
+        raise RuntimeError("VGGT-Omega requires CUDA.")
```
akshay-krishnan, line 226: does it?

akshay-krishnan (approved):
can be merged once the loading function is deduplicated.

gtsfm/cluster_optimizer/cluster_vggt.py
kathirgounder: Preprocessing is part of the model contract — resolution, patch grid (518/14 vs 512/16), and omega's aspect-ratio clamp are what each model was trained with — and the depth lookup downstream indexes the model's output frame through `original_coords`, so the loader and the model have to move together; the hook keeps them paired (default = the VGGT loader, no-op for every VGGT config).

On the duplication: agreed, and fixed in 86e20d56. The brittle part is now single-sourced — each loader only expresses its per-image resize/crop policy and reports an `ImagePlacement` (a small dataclass: left/top/scaled_w/scaled_h); batch padding and the `original_coords` row construction happen in exactly one place, `assemble_image_batch` in geometry_transformer.py. To make the refactor provably behavior-identical I captured 12 golden cases from the **pre-refactor** loaders (crop/pad/balanced/max_size × uniform/mixed/extreme-aspect batches) into `tests/frontend/test_image_batch_loaders.py`; both loaders reproduce them exactly after the refactor.

gtsfm/frontend/vggt_omega_geometry_transformer.py
kathirgounder: Deduplicated in 86e20d56 — see the thread on cluster_vggt.py: shared `assemble_image_batch` + `ImagePlacement`, with the pre-refactor behavior pinned by golden tests.

gtsfm/cluster_optimizer/cluster_vggt.py
kathirgounder: Half of this landed with the dedup in 86e20d56: the producer side now goes through the `ImagePlacement` dataclass with named fields, and the packed (N, 6) row layout is documented and constructed in a single function. Replacing the tensor with a dataclass on the **consumer** side touches multi_view_tracker, cluster_vggt_with_frontend, loader_base and the telemetry dump (and the tensor crosses the Dask boundary), so I'd rather do that as its own focused PR — happy to take it as an immediate follow-up.

gtsfm/cluster_optimizer/cluster_vggt.py
kathirgounder: Done in 86e20d56 — `model_cache_key` is now `Hashable | None` only, and the opt-out is an explicit `use_model_cache: bool = True` constructor argument on both optimizers (passing a bool as the key raises with a migration hint). The `False` arm was only reachable from hand-written code — one test, which is migrated.

gtsfm/frontend/vggt_omega_geometry_transformer.py
kathirgounder: Good catch — no, not strictly. Upstream's `forward()` autocasts with `device_type="cuda"`, and on a CUDA-less machine that degrades to a no-op (verified on current torch: `torch.cuda.is_bf16_supported()` returns False rather than raising, and the autocast context just warns and disables). So the model does run on CPU in full fp32 — merely orders of magnitude slower at ~1B params. Replaced the hard `raise` with a warning in 86e20d56.

PR #1134, opened 2026-08-31
De-nest the cluster frontend (no nested Dask clients); memory-bounded two-view helpers

kathirgounder:
## What

The per-cluster frontend (`ClusterMVO._run_correspondence_generator` / `_run_two_view_estimation`, inherited by `ClusterVGGTWithFrontend`) runs inside a Dask task and then opened a `worker_client()` to submit and gather per-image / per-pair sub-tasks. That tasks-launching-tasks pattern keeps every feature future and every `TwoViewResult` of the cluster resident on one worker during the nested gather, and it is what fell over on large Phototourism scenes on a single RAM-limited node.

This PR removes the nesting and adds the memory-bounded two-view primitives the upcoming global-verification PR builds on:

1. **Inline correspondence generation.** `CorrespondenceGeneratorBase.generate_correspondences_inline(images, visibility_graph)` — same contract as `generate_correspondences`, over materialized images, plain loops, no client. Implemented for all five generators so no config regresses:
   - `DetDesc`, `Image` (LoFTR): loops over the detector/matcher.
   - `Colmap`: the synchronous body *is* the inline variant; the Dask entry point becomes `gather + inline`.
   - `MASt3R`: the aggregation is factored into `_aggregate`, shared by both entry points.
   - `Synthetic`: scene preparation factored into `_prepare_scene`, shared by both entry points.
2. **Inline two-view.** `create_two_view_results_inline` (full results — what the cluster frontend consumes) and `create_v_corr_idxs_inline`, which drops each `TwoViewResult` (three reports + putative indices) the moment its `v_corr_idxs` is extracted, so memory is bounded by the output rather than by the number of pairs. `create_v_corr_idxs_futures` chunks the latter across the worker pool: shared inputs are scattered **once**, each as a single blob (`client.scatter([obj])[0]` — a bare scatter of a list/dict explodes it into per-element futures), chunks run `pure=False` (`run_2view` has cacher side effects), and only the small per-chunk dicts are gathered via `as_completed`.
3. **`ClusterMVO.create_frontend_graph`** passes the image futures directly as task inputs — Dask materializes them as an ordinary dependency — instead of re-creating `Future(key=...)` handles inside a nested client.

Outputs are identical (pinned by the tests below). `run_2view` still executes per pair, so `TwoViewEstimatorCacher` hits/writes are unchanged. No new flags.

## Who uses what

- `generate_correspondences_inline` + `create_two_view_results_inline`: the cluster frontend, from this PR on.
- `create_v_corr_idxs_inline` / `_futures`: the next PR in the series (global view-graph verification: verify the full retrieval graph *before* partitioning; only `v_corr_idxs` is consumed there). They land here with their tests so that PR is pure wiring.

## Tests

- **New** `tests/frontend/correspondence_generator/test_generate_correspondences_inline.py`: inline == Dask path for the DetDesc and Image generators, using shape-dependent stubs (a shape not threaded through shows up as a mismatch).
- **New** `tests/test_create_v_corr_idxs_futures.py`: parallel == inline for default / single-pair / single-chunk chunking; empty graph; `valid()` filter honoured.
- Existing suites touching these modules (two-view estimator + cacher, colmap generator, keypoint aggregators, cluster_vggt, cluster-optimizer cacher): 26 passed.
- **Not runnable here:** the MASt3R and Synthetic inline paths (need the mast3r submodule + weights / Tanks & Temples data). Both are behaviour-preserving extractions, syntax- and lint-checked. @akshay-krishnan — a run of `correspondence/mast3r.yaml` through a cluster would close that gap.
- Pre-existing failures on `mapping-vggt`, unrelated to this change (verified on a pristine checkout): `tests/test_gtsfm_runner_olssonloader.py` (requests `loader/olsson_loader`; the config is `olsson`) and `tests/runner/test_frontend_runner.py` (imports `gtsfm.runner.frontend_runner`, which no longer exists).

## Review focus

- `cluster_mvo.py`: futures as task inputs in `create_frontend_graph`; `_run_correspondence_generator(images)` signature.
- `two_view_estimator.py::create_v_corr_idxs_futures`: scatter-as-blob, `pure=False`, `as_completed` merge, chunk sizing (~4 chunks/worker).
- MASt3R / Synthetic: extraction only — no logic change intended.

## Context

Part of the series landing the ECCV'26 SFM-DL *Hierarchical SfM scales feedforward reconstruction* pipeline (research branch `vggt-verified-pipeline`, tag `eccv2026-sfmdl-research`) into `mapping-vggt` as reviewable units. Next up: the global verified view graph.

### DIFF
--- a/gtsfm/frontend/correspondence_generator/colmap_correspondence_generator.py
+++ b/gtsfm/frontend/correspondence_generator/colmap_correspondence_generator.py
@@ -123,6 +123,25 @@ class ColmapCorrespondenceGenerator(CorrespondenceGeneratorBase):
 
         return corr_idxs
 
+    def generate_correspondences_inline(
+        self, images: List[Image], visibility_graph: VisibilityGraph
+    ) -> Tuple[List[Keypoints], Dict[Tuple[int, int], np.ndarray]]:
+        """Read keypoints and matches for the given images from the Colmap DB.
+
+        Note: the matches read from the DB are already geometrically verified.
+
+        Args:
+            images: List of all images.
+            visibility_graph: The visibility graph defining which image pairs to process.
+
+        Returns:
+            List of keypoints, one entry for each input images.
+            Correspondences as indices of keypoints, for pairs of images.
+        """
+        gtsfm_id_to_pycolmap_id, keypoints = self._read_image_ids_and_keypoints(images)
+        corr_idxs = self._read_matches(visibility_graph, gtsfm_id_to_pycolmap_id)
+        return keypoints, corr_idxs
+
     def generate_correspondences(
         self, client: Client, images: List[Future], visibility_graph: VisibilityGraph
     ) -> Tuple[List[Keypoints], Dict[Tuple[int, int], np.ndarray]]:
@@ -137,10 +156,4 @@ class ColmapCorrespondenceGenerator(CorrespondenceGeneratorBase):
             List of keypoints, one entry for each input images.
             Putative correspondence as indices of keypoints, for pairs of images.
         """
-        # Note: we will end up reading verified correspondences from the colmap DB.
-        images_actual = client.gather(images)
-
-        gtsfm_id_to_pycolmap_id, keypoints = self._read_image_ids_and_keypoints(images_actual)
-        corr_idxs = self._read_matches(visibility_graph, gtsfm_id_to_pycolmap_id)
-
-        return keypoints, corr_idxs
+        return self.generate_correspondences_inline(client.gather(images), visibility_graph)
--- a/gtsfm/cluster_optimizer/cluster_mvo.py
+++ b/gtsfm/cluster_optimizer/cluster_mvo.py
@@ -6,11 +6,10 @@ import os
 import time
 from dataclasses import dataclass
 from pathlib import Path
-from typing import Any, Mapping, Optional, cast
+from typing import Any, Mapping, Optional
 
 import numpy as np
 from dask.delayed import Delayed, delayed
-from dask.distributed import Future, worker_client
 from gtsam import Pose3, Similarity3  # type: ignore
 
 import gtsfm.common.types as gtsfm_types
@@ -33,7 +32,7 @@ from gtsfm.multi_view_optimizer import MultiViewOptimizer
 from gtsfm.products.one_view_data import OneViewData
 from gtsfm.products.two_view_result import TwoViewResult
 from gtsfm.products.visibility_graph import AnnotatedGraph, VisibilityGraph
-from gtsfm.two_view_estimator import TwoViewEstimator, create_two_view_estimator_futures
+from gtsfm.two_view_estimator import TwoViewEstimator, create_two_view_results_inline
 from gtsfm.ui.gtsfm_process import UiMetadata
 from gtsfm.utils import transform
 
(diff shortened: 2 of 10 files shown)


gtsfm/frontend/correspondence_generator/colmap_correspondence_generator.py
```diff
@@ -123,6 +123,25 @@ def _read_matches(
 
         return corr_idxs
 
+    def generate_correspondences_inline(
```
akshay-krishnan, line 126: i would avoid more jargon about the dask graph in function names. just call it generate_correspondences()

akshay-krishnan:
@claude why have the functions been made inline? "That tasks-launching-tasks pattern keeps every feature future and every TwoViewResult of the cluster resident on one worker during the nested gather" perhaps we can limit the number of workers? how many threads / workers was used? 

main question: what happens if we are not using metis partitioning, and running on the full viewgraph? this is not nested. is the frontend still run inline? 

@kathirgounder ?

kathirgounder:
@akshay-krishnan Local Claude's Response:

Good questions — taking them in order.

**How many workers/threads?** The failing runs were `--num_workers 1 --threads_per_worker 1` (the
runner default, and the natural setting for a single-GPU node — VGGT serializes on the one GPU
anyway). So "limit the number of workers" was the starting condition, not an available fix.

**Why inline rather than fewer workers?** Two failure classes, and worker count fixes neither:

1. *Coordination fragility (the terminal one).* At num_workers=1 the nested pattern buys zero
   parallelism — there's nobody to fan out to — while paying scheduler round-trips + serialization
   per pair, and parking the parent task in a `worker_client().gather(...)` on the worker for the
   whole stage. On multi-hour scenes (St Peter's: 2,504 images, 48,416 edges) that's exactly what
   kept dying: `CommClosedError` / "Removing worker … lost dependencies" cascades with **no memory
   warnings anywhere in the logs** (node had 2 TB; died identically at 90 GB and 128 GB caps). We
   peeled this over several commits on the research branch — client-side lean gather, then
   worker-side streaming, then full de-nesting — and each layer moved the failure later until runs
   survived. This PR is the distillation.
2. *Gather concentration (the multi-worker hazard).* `gather` has a single destination: all N
   `TwoViewResult`s of a cluster (three per-point reports each — measured 5–8× the size of the
   `v_corr_idxs` we actually consume) land in the parent task's heap at once, while dask still pins
   the distributed copies until release. That concentration is per-cluster-task and independent of
   pool size; fewer workers makes the thread-occupancy side of it worse, not better. (This is the
   class we hit first, as nanny kills at 32/64 GB worker caps.)

Where the parallelism went: **up one level**. Cluster tasks are now peers in a flat graph, so the
scheduler runs many clusters concurrently and can actually see and balance the units it's
scheduling — each bounded by the partitioner (max_cameras=70 → ≤2,415 pairs), so inline per-cluster
is small and fast. Parallelism across clusters instead of nested fan-out within one.

**Main question — no METIS, full view graph:** correct, and deliberately half of a two-PR story: on
this PR alone, a SinglePartitioner run executes the frontend inline in one task. The completion is
#1135 (stacked on this one):

- This PR ships the parallel primitive for exactly that regime: `create_v_corr_idxs_futures` —
  *client-driven* chunked fan-out (shared inputs scattered once, only lean per-chunk `v_corr_idxs`
  gathered, a lost worker recomputes only its own chunks). Fan-out without nesting, fully
  scheduler-visible.
- #1135 wires the full-view-graph frontend through it: the global verification pass runs over the
  entire retrieval graph *before* partitioning, dispatching on pool size (>1 worker → parallel
  chunks across the pool; 1 worker → inline in the main process, where there is no comm surface to
  trip). Per-cluster frontends after that pass are cache hits — or skipped, since clusters reuse
  the global correspondences. In the stacked end state, SinglePartitioner just means "one cluster
  that reuses the (parallel) global result."
- In the window between the two PRs: the practical SinglePartitioner configs (door, gerrard-hall —
  the classic small scenes) are actually *faster* inline, since per-pair task overhead dominated at
  a few hundred pairs. The regime where you'd miss fan-out — large unpartitioned scenes on a big
  pool — is the regime the nested version couldn't survive. If that window is a concern, happy to
  land #1134 + #1135 together.

akshay-krishnan:
@kathirgounder could you make a couple changes? 

1. remove the word _inline_ in most places? not sure that is the right term for this. perhaps _eager_ is. maybe claude can suggest better terms. I would also only use the term when absolutely needed (eg a function with the term _inline dropped already exists). 

2. minimize any code duplication between an inline implementation and a delayed implementation. we want as much  correspondence estimation logic to be shared between them as possible.

akshay-krishnan (approved):

gtsfm/frontend/correspondence_generator/colmap_correspondence_generator.py
akshay-krishnan: this was an old comment, that would collide with an existing function, but i have a newer comment above about this

kathirgounder:
Done in bf0468c6 — both asks addressed.

**Naming** — went one step past `eager`: the repo already has a convention for exactly this distinction — Dask variants carry `_futures` (cf. `create_two_view_estimator_futures`). So the no-client versions now take the **plain names**, and the qualifier lives only on the side that actually takes a `client`:

- `generate_correspondences(images, graph)` (plain) / `generate_correspondences_futures(client, image_futures, graph)` — i.e. your original suggestion, with the collision resolved by renaming the Dask variant onto the convention.
- `create_two_view_results` and `create_v_corr_idxs` (plain) / `create_v_corr_idxs_futures`.
- The word "inline" no longer appears anywhere in the diff (code, docstrings, tests, or the PR title).

**Duplication** — the estimation logic is now single-sourced between the two paths; what remains different is pure orchestration (loop vs. submit/gather):

- DetDesc: module-level `_detect_and_describe` / `_match_features` kernels are the same callables the plain loops run and the futures path `client.submit`s.
- Image: one `_match_image_pair` kernel shared by the plain loop, the futures fan-out, and `generate_correspondences_and_estimate_two_view` (which previously had its own copy of the closure).
- Two-view: the full-result and v_corr-only loops both call a single `_run_2view_for_pair`; the chunked futures variant already used the plain function as its per-chunk body.
- MASt3R (`apply_mast3r` + `_aggregate`), Synthetic (`_prepare_scene` + pair kernel), and Colmap (futures = gather + plain) were already shared.

21 affected tests green after the rename/refactor.

akshay-krishnan (approved):

